PR #13, opened 2016-06-01
Disable output colorizing when running in a non-interactive shell

RandomReaper:
Prevent seeing strange chars in files and non-interactive consoles ```[[33mRuntime[0m]``` -> ```[Runtime]```


Dolu1990:
With Intelij colors works , (even in non interactive mode). But if it doesn't work on eclise, it's globaly bether with your fix.
Thanks  !

PR #47, opened 2016-10-25
Update Sdram.scala

roman3017:
Correcting the capacity calculation when dataWidth is not power of two since SDRAM is word not byte oriented.

The current calculation works for known 8/16/32/64 bit SDRAMs but if there ever were say 24bit one it would fail.

Also, it seems that 4bit SDRAMs are not supported here but I do not have a fix for that.


Dolu1990:
Hooo right, for strange bits width, the following should work :

```scala
def capacity = (BigInt(dataWidth) << wordAddressWidth) / 8
```

Right ?

roman3017:
Yes, looks correct.

PR #48, opened 2016-11-06
Add DES and TripleDES 

Snoopy87:


Dolu1990:
Hmm it look like it's a git or a travis issue ?

PR #75, opened 2017-09-04
FSM boot state should be first state in enum

vvb16:
FSM boot state should be first state in enum for optimal synthesis on devices.
Without this fix initial FSM state was: StateReg and StateNext = first enum element which is not boot state, but state created first in Scala code (it can be even not the Entry state).
This change fix synthesis warnings at least for Spartan 6 FPGAs.
This also fix post-synthesis, post-map and post-PAR simulations.


Dolu1990:
Thanks ^^

PR #76, opened 2017-09-05
Blackbox io ports fix

vvb16:
Keep all io signals for Blackbox - they should not be optimized away
Map unused out ports in VHDL generation to "open".
Probably same fix should be done for Verilog backend


Dolu1990:
Great ^^

PR #108, opened 2018-01-03
removing the need for `.` in `.elsewhen`

### LINKED ISSUE #107: The annoying extra `.` in `.elseWhen` and `.otherwise`
See the following example:
```scala
when(cond1){
  //
}.elsewhen(cond2){
  //
}.otherwise{
  //
}
```
I wish to write the above without the extra `.`
```scala
when(cond1){
  //
} elsewhen(cond2){
  //
} otherwise{
  //
}
```

I investigated the matter on the [scala contributors gitter channel](https://gitter.im/scala/contributors?at=5a4c49db5355812e5745b193) and it turns out this may be a Scalac parser bug.

Follow issue https://github.com/scala/bug/issues/8859 . If resolved, it may solve this issue for Spinal.
I also created a ticked for dotty, which has a similar issue. https://github.com/lampepfl/dotty/issues/3737

soronpo:
Fixes #107


Dolu1990:
Thanks !

som-snytt:
Excellent! I'm not following the issue, sorry, but from the diff you might be amused by http://scalapuzzlers.com/#pzzlr-060 .  I don't remember the solution, though I wrote it.

In any case, glad it worked out for you.

soronpo:
Thanks @som-snytt 
With your help I finally understood what scalac understood from my code, so I then managed to manipulate it to do what I want 😄

PR #112, opened 2018-01-18
lib/sim/Stream: Add ready condition for StreamReadyRandomizer

accuminium:
This is useful to control when the stream is ready and defaults to true.


Dolu1990:
Thanks ^^

PR #140, opened 2018-08-22
Add extra flags to verilator command

dlobato:


dlobato:
@Dolu1990 does it look better?

Dolu1990:
Thanks :D

Issue #145, opened 2018-09-10
RFE: Make switch statement optimized for one-hot encoded arguments


tomverbeure:
Current code:
```scala
        object PcState extends SpinalEnum(binaryOneHot) {
            val Idle           = newElement()
            val WaitReqReady   = newElement()
            val WaitRsp        = newElement()
            val WaitJumpDone   = newElement()
            val WaitStallDone  = newElement()
        }
...
    val instr_final = (pc.cur_state === pc.PcState.WaitStallDone) ? instr_r     | instr
...
        switch(cur_state){
            is(PcState.Idle){
...
```
This generates the following verilog:
```verilog
`define PcState_defaultEncoding_type [4:0]
`define PcState_defaultEncoding_Idle 5'b00001
`define PcState_defaultEncoding_WaitReqReady 5'b00010
`define PcState_defaultEncoding_WaitRsp 5'b00100
`define PcState_defaultEncoding_WaitJumpDone 5'b01000
`define PcState_defaultEncoding_WaitStallDone 5'b10000
...
// GOOD:
  assign instr_final = (((pc_cur_state & `PcState_defaultEncoding_WaitStallDone) != 5'b00000) ? instr_r : instr);
// BAD
    case(pc_cur_state)
      `PcState_defaultEncoding_Idle : begin
```
The regular compare operation is optimized for one-hot encoding, but the switch statement compares the full vector.
Ideally, it should generate something like this:
```verilog
(* parallel case *) 
case(1) // synopsys parallel_case
    ((pc_state & `PcState_defaultEncoding_Idle) != 5'b00000): begin
...
```
The `(* parallel case *)` is supported by some, but not all, synthesis tools. The same is true for `// synopsys parallel_case`. So you need both. It's seriously annoying.

parallel_case usage is usually [highly discouraged](http://www.sunburst-design.com/papers/CummingsSNUG1999Boston_FullParallelCase.pdf), but one-hot encoded case statements are an exception. From the paper:
> Guideline: only use full_case parallel_case to optimize onehot FSM designs.

typingArtist:
I’d suggest to enter the code like the following. It should be functionally equivalent but points out better the intent.

```verilog
(* parallel case *)
case(1) // synopsys parallel_case
    ((pc_state & `PcState_defautEncoding_Idle) == `PcState_defaultEncoding_Idle): begin
```
Also this would allow the code to be used for other *hot encodings as well, e.g. two-hot.

[...]

PR #147, opened 2018-09-13
Change coding style of Verilog to case(1) with parallel_case #145

typingArtist:
This allows the special case of a binaryOneHot encoding to be encoded with best synthesis results

- [ ] compiled
- [ ] tested


typingArtist:
You can see from the CI runs of my commits: when I had this in, the SpinalHDL test cases failed, so I removed it. We definitely need an option for supporting the various tool’s flavors.

typingArtist:
Ah, seems that the attribute `(* parallel case *)` should be replaced by `(* parallel_case *)` (mind the underscore).

Issue #130, opened 2018-07-17
Issue warnings when BusSlaveFactory read multiple signals to the same address


marcoep:
Lets say I have multiple modules all of which I drive from the same bus. For all of them I use the same BusSlaveFactory. There should be a way to detect when reading multiple signals at the same address, as this is most commonly the case when mis-configuring base-addresses, addresses, or bitOffsets.

So for example this spinalHDL code:
```scala
val apbSlaveFactory = new Apb3SlaveFactory(io.apb3, 0)
val thermalCtrlDigitalBridge = thermalCtrlDigital.driveFrom(apbSlaveFactory, 0x00)
val thermalCtrlDigitalBridge_2 = thermalCtrlDigital_2.driveFrom(apbSlaveFactory, 0x04)
```
but  I have made the mistake that one thermalCtrlDigital actually reads from addresses base+0x00 to base+0x07! This then results in the following data model on .printDataModel():
```
...
Address 0x4 :
  R[5:0] io_AIF_TCtrl_ref_current_DAC_o - ThermalControllerDigital: analog interface: Iset [5..0] 
  W[0:0]  - ThermalControllerDigital FSM control bit: start_locking 
  R[0:0]  - ThermalControllerDigital FSM control bit: start_locking 
  W[1:1]  - ThermalControllerDigital FSM control bit: start_ramp_heater 
  R[1:1]  - ThermalControllerDigital FSM control bit: start_ramp_heater 
  W[2:2]  - ThermalControllerDigital FSM control bit: stop_locking 
  R[2:2]  - ThermalControllerDigital FSM control bit: stop_locking 
  R[3:3] io_idle - ThermalControllerDigital FSM control bit: idle 
...
```
and in the following VHDL:
```VHDL
...
      when "00000100" =>
        if apbSlaveFactory_doWrite = '1' then
          zz_13 <= pkg_toStdLogic(true);
        end if;
        io_apb3_PRDATA(5 downto 0) <= zz_30;
        io_apb3_PRDATA(0 downto 0) <= pkg_toStdLogicVector(zz_7);
        io_apb3_PRDATA(1 downto 1) <= pkg_toStdLogicVector(zz_8);
        io_apb3_PRDATA(2 downto 2) <= pkg_toStdLogicVector(zz_9);
        io_apb3_PRDATA(3 downto 3) <= pkg_toStdLogicVector(zz_40);
...
```
[...]

PR #164, opened 2018-12-14
Prohibit BusSlaveFactory to assign reads to the same bit more than once

marcoep:
This small PR fixes issue #130.

An additional check in BusSlaveFactoryDelayed is implemented to assert that on any read command, no bit on the bus read line is assigned multiple times.

There is also a unit test for this case.


Dolu1990:
Nice :D

Dolu1990:
Thanks for the pullrequest :)

PR #165, opened 2018-12-17
Add randBootFixValue parameter in spinalconfig

Snoopy87:


Dolu1990:
Hoo one thing, dev is now using sbt 1.2.7, moved to scala 2.11.12 and changed many little things in the flow. So if you want to try it out you will probably need to do a sbt clean compile

Dolu1990:
In fact i restarted the travis build from the web interface XD

Dolu1990:
hmm it look like you finaly realy need to merge SpinalHDL dev into your repo :)

Dolu1990:
Look good for me :)

Dolu1990:
So now the last thing is to fix the default value of 

                        noRandBoot                     : Boolean = false,
                        randBootFixValue               : Boolean = false,

What should be the default value of randBootFixValue ? Before it was true (always zero) to avoid changing the netlist at each generation.

PR #168, opened 2018-12-19
Syncronous -> Synchronous

tomverbeure:
In ClockDomain.scala, there are 2 public methods that are spelled incorrectly as "...Syncronous". I've renamed them to the correct spelling, but also kept the old calls for backward incompatibility.

I've also fixed the spelling everywhere else.

Tom

### DIFF
--- a/core/src/main/scala/spinal/core/core.scala
+++ b/core/src/main/scala/spinal/core/core.scala
@@ -352,6 +352,10 @@ package object core extends BaseTypeFactory with BaseTypeCast {
   }
 
 
+
+  implicit def BooleanPimped[T <: Data](that : Boolean) = new BooleanPimped(that)
+
+
   /**
     * Assertion
     */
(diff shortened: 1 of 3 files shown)


typingArtist:
The misspelled names should be marked @deprecated. And could you make them one-liners without curly braces and return?

```scala
@deprecated("misspelled method will be removed", "SpinalHDL 1.2.3")
def isSyncronousWith(that: ClockDomain) = isSynchronousWith(that)
@deprecated("misspelled method will be removed", "SpinalHDL 1.2.3")
def setSyncronousWith(that: ClockDomain) = setSynchronousWith(that)
```

And when we are at it: ```setSynchronousWith``` should be given ```: Unit``` as its return type, instead of returning the ```that.synchronizedWith``` ArrayBuffer.

```scala
def setSynchronousWith(that: ClockDomain) : Unit = {
  …
  that.synchronizedWith += this
}
```

PR #179, opened 2019-01-02
SPI changes

dnltz:


Dolu1990:
The issue i have with absolute path is it break ass soon the move the sources. For instance a generated verilog submited into a repository for people to use it in their own environnement will break with an absolut path.

But anyway, i agree that relative path are a paine too, and i often had issue with synthesis/simulation tools to find them :/

tomverbeure:
> The issue i have with absolute path is it break ass soon the move the sources. For instance a generated verilog submited into a repository for people to use it in their own environnement will break with an absolut path.

Yes. For one of my projects, I provide a generated Verilog so that other don't need to install the whole SpinalHDL environment. 

This change would break that.

dnltz:
Mhhh that's true. We could add a shell variable as prefix to the relative path, but this requires to set the variable every time. I'll search for another way.
Since the other patch failed, I will close this pull request and reopen a new one with the fix.

dnltz:
Oh, I force-pushed the origin branch which automatically updates this pull request. Hope you're fine with the remaining patch.

Dolu1990:
So one thing i'm still not exactly sure about, is how to properly name the following things : 
- The elaboration parameters (fifo depth, ...) is that a config ? a parameter ? a generic ?
- The runtime parameters (real wire to configure the baud rate of the uart, ...) is it a config ?  a parameter ? (ok not a generic ^^)

So what would be the apropriate postfix for each of those to differente cases ?

Currently i think Parameter is good for elaboration stuff, and Config is good for runtime stuff. sound good ?

dnltz:
In the reorganizing of the peripherals, I put everything in one config, but I like the idea of runtime vs elaboration. Could be something like this:
```
case class UartParameter(
  busCanWriteFrameConfig: Boolean,
  busCanWriteClockDividerConfig: Boolean,
  dataWidthMax: Int,
  clockDividerWidth: Int,
  preSamplingSize: Int,
  samplingSize: Int,
  postSamplingSize: Int,
  txFifoDepth: Int,
  rxFifoDepth: Int
) {
  require(...)
  def someFunc()
}
object UartParameter {
  def lightweight(): UartParameter = UartParameter(...)
  def default(): UartParameter = UartParameter(...)
  def performance(): UartParameter = UartParameter(...)
}
case class UartConfig(
  baudrate: Int,
  dataLength: Int,
  parity: UartParityType.E,
  stop: UartStopType.E
) {
  require(...)
  def someFunc()
}
object UartConfig {
 default(baudrate: Int) ...
}
```
I actually like these predefined configs (like the CSR configurations) and if you have a special case, you can also define your own configs. That's a easy and safe way for configure IP-Cores!

Dolu1990:
Sure, let's go that way then ^^

PR #182, opened 2019-01-03
RFC: Reorganize peripheral structure

dnltz:
Peripherals are right now in lib.com, lib.bus, lib.graphics or in seperate repos (e.g. VexRiscv). All peripherals should be in the same hierarchy and built by a common pattern. This will make SoCs more reliable and software maintainable.

These two patches show how this could look like:
a) All bus interfaces and public code will be in the first level, e.g. gpio/...
b) Internal code is hidden in a sub directory.

If this RFC looks good, I could start and port all peripherals to this structure.


Dolu1990:
More generaly, we could maybe imagine a way to avoid having to redefine each time the given component for each bus variation ?

dnltz:
> More generaly, we could maybe imagine a way to avoid having to redefine each time the given component for each bus variation ?
That's what I also thought! I'm not sure how to do this but maybe some kind of "BusDomain"?

Dolu1990:
Ok, so, we do it for all peripherals ? It will lightly break some imports in thirds party repository, but that's fine i guess.

Dolu1990:
hmm
I was thinking about defining a abstract component (for instance BusSlaveFactoryUartCtrl), and then, extending for each bus implementation we need it.

```scala
abstract class BusSlaveFactoryUartCtrl(factory : => BusSlaveFactory) extends Component{
   magic here
}

case class Apb3UartCtrl(...) extends BusSlaveFactoryUartCtrl(Apb3BusSlaveFactory(Apb3(8,32)){ empty body }

```

Issue #183, opened 2019-01-04
Deprecate kB, MB, GB, TB and introduce KiB, MiB, GiB, TiB


typingArtist:
We should use and encourage proper binary prefix according to the IEC standard. The old/wrong/JEDEC-only forms should be deprecated.

https://en.wikipedia.org/wiki/Binary_prefix

Edit: Also updated such that in addition to ```Byte``` also ```Bytes``` is supported for the unit.

tomverbeure:
> We should use and encourage proper binary prefix 

I think we should not. 

Insisting on KiB and MiB is like insisting that the new century started in 2001 and not 2000. It may be technically correct, but it’s also annoying. ;-)

typingArtist:
I’d say, insisting on KiB and MiB is like acknowledging that a new century has started at all … :grin:

And the old form is still supported, with deprecation warnings though.

tomverbeure:
What about a reasonable middle ground: Add KiB and MiB, but don't have deprecation warnings? There's really no upside in having those deprecation warnings.

zhutmost:
In most academic papers about computer architecture or hardware, we still use kB instead of KiB much often. The same things also happen in lots of chip datasheets. 
I guess deprecation warnings will confuse users? 

( actually, many standards made by IEEE or other associations are insisted by only a little people. For example, SystemVerilog :D

tomverbeure:
I’ll just leave this here: https://xkcd.com/394/. :-)

Dolu1990:
So, keeping both kB, MB, GB, TB and KiB, MiB, GiB, TiB, without depreciation ?

tomverbeure:
Yes!

typingArtist:
Other projects have already found out about the old form’s ambiguity and have reacted. Why can’t we do as well? I’d like to see the old form go away better soon than late, but @Dolu1990 has already made clear that he’d like to keep the old form around with deprecation warnings.

I think that this is the best we can do, so I vote against un-deprecating the old form.

Dolu1990:
So the thing is, maybe today this sound stupid to deprecate the KB MB stuff, but what about in the long terme ? 
As SpinalHDL is there to reign on the EDA ecosystem for the next 1000 years, it is probably the right move ^.^, or not ?

typingArtist:
[...]

PR #184, opened 2019-01-04
Deprecate for IEC binary prefix

typingArtist:


typingArtist:
This should fix #183 for this repository.

PR #189, opened 2019-01-10
Reorganize peripherals for minimal Zephyr support.

dnltz:
This patch series will reorganize the GPIO and UART controller and introduce a new controller, the MachineTimer. All these controllers are in their name-specific package under 'peripherals', e.g.:
`import spinal.lib.peripheral.uart.{Apb3Uart, UartParameter, UartConfig, Uart}`
All controllers can be configured by two objects; parameters and configs. Parameters will define the hardware environment of a controller and the config how the controller behaves in run-time or start-up.
All controller registers start at address 0x10 so we have four reserved registers for future purposes like storing controller version, type, ...

What are the files for?
Apb3Gpio.scala : Contains a Apb3 BusFactory. (Same for WishboneGpio.scala, ...)
Gpio.scala: Contains the parameter and config object and all public stuff like enums, ...
controller/ : Here are internal, private files like GpioController.scala

I was not able to make the bus ports more generic. Having only one Component for bridging between buses and controller is definitively a good feature for the future.

I added doc.txt files to all controller directory as minimal documentation but I will update the official SpinalHDL doc as soon as possible.


Dolu1990:
Hoo, changing the memory mapping is an issue, as there is already software outthere which is using the currrent mapping.
We have to care about backward compatibility. I have nothing against having alternative memory mapping, but i think the default should remain the  actual one.

The doc.txt issue is it will not be in the scala sources. It is better to have it directly above the related classes via a block comment.

dnltz:
I only changed the offset by 16 to make space for four reserved addresses and not the bit or register inside of the controller. I know that changing the memory mapping is not that easy, but maybe now is the only good time before it's too late. Later version can benefit of theses IP-Core information to handle memory mapping correctly in software. Right now this is not possible. Anyway, all SoC implementations will fail due missing imports from SpinalHDL. So, they need some little adjustment.

The doc.txt is only for me to don't forget the meaning of all bits. We should keep the primary doc on the other repo, but I can also add the content to the controller-file.

dnltz:
I could add a flag to all parameters "emitControllerInformation", which will add a offset of 0x10 or not? If someone has a compatibility problem, they can disable it.

Dolu1990:
There is multiple things crossing :
- Having fit peripherals for small devices, without "overkilled" features
- Some design use a single shared rom to register all peripherals nature and address.

I would say, the best would probably be to over an alternative improved memory map, which is more unified between controllers.
It would have two advantages : 
- Keeping backward compatibility
- Having a better memory map without compromise.

dnltz:
Sorry, didn't get it. What do you mean with "an alternative improved memory map"?

But the peripherals with the same memory map are OK, right?

Dolu1990:
The peripherals with the same memory mapping (or register mapping) is fine.

What i wanted to say, is we can add improved register set / register mapping if the current one isn't good enough. In otherwords, alternative driveFrom implementation :)

dnltz:
sounds good :)

Dolu1990:
Hmm maybe this pull request should be split into multiple sequencial one, there is a lot of changes
Also, we have to discuss about the package naming :)

dnltz:
What are your concerns about the package naming? After I created the MachineTimer controller, I thought 'misc' is not so good as subgroup, but when I'm thinking about this a little bit longer, there will be more controller for internal stuff. Will redo this.

That's definitely not the way to do this, but since I moved or renamed files, GIT has anyway lost the logs about these files and you can threat this as an initial commit. For the next peripherals, I can do this better since I know how they should be ;)

Dolu1990:
For instance, there is spinal.lib.peripheral.uart,  Uart isn't necesserly about peripherals, that's way it was in lib.com.uart before :)

For the git, it would be better to do multiple step by step pull request instead of all in one. It is easier to adjust.

In our PR there is the Storage stuff, it is the Config equivalent ?

dnltz:
yeah, you're right... I'll separate this into smaller PRs for each component.

> In our PR there is the Storage stuff, it is the Config equivalent ?
It is. Since we changed the definition of config this could led to confusions. The storage is, like previously, only to store values in some kind of registers between a bus and controller.

Do you think we should create a second container for all vias between a bus and controller? These case classes make the code readable, access through the '.' is really easy and clean and it's better parameterizable.

Dolu1990:
Hmm actualy Config for the RTL configuration signals and parameters for the elaboration stuff is fine ^^
Storage is kind a bit like picking an unrelated word and applying it to our context, isn't it ?

> Do you think we should create a second container for all vias between a bus and controller? These case classes make the code readable, access through the '.' is really easy and clean and it's better parameterizable.

Can you point an example ? 
there you splited the parameters of the UartController and the parameters of the UartController memory mapping into two case class.
Before, the parameters of the UartController were contained as an instance inside  the UartController memory mapping parameters.
Why changing ?
https://github.com/SpinalHDL/SpinalHDL/pull/189/files#diff-f8cd6ae78c1e96c1990f2b5824ea96c5R35

Dolu1990:
I will write a template for the Uart one, and send it on this channel, then we have something simple to talk about :)

Dolu1990:
I'm reworking the Gpio controller.

In short : 

```scala
  case class Parameter(width : Int,                   //Number of pin
                       var input : Seq[Int] = null,   //List of pin id which can be inputs (null mean all)
                       var output : Seq[Int]  = null, //List of pin id which can be outputs (null mean all)
                       interrupt : Seq[Int]  = Nil,   //List of pin id which can be used as interrupt source
                       readBufferLength : Int = 2)    //Number of syncronisation stages

...

    val interrupt = new Area {
      val enable = new Area{
        val high, low, rise, fall = Bits(p.width bits) //interruptions enables
      }

      io.interrupt := ((enable.high & pin.read)
                     | (enable.low & ~pin.read)
                | (enable.rise & (pin.read& ~last))
                | (enable.fall & (~pin.read& last)))
   }
```

Do you see something else to add/change ? 
Basicaly there is no interrupt pending flag in the controller, as it is in the PLIC gateway.

dnltz:
readBufferLength is currently unused. Leave a TODO?
I personally wouldn't do something like this
```scala
if(p.input == null) p.input = (0 until p.width)
``` 
because it's not good to read (as a C developer :D )
The controller itself looks good for me!

PR #192, opened 2019-01-19
Rise/Fall -> Rose/Fell

tomverbeure:
Resubmit of earlier patch (that might have had integration issues).

Change Rise/Fall -> Rose/Fell and $rise/$fall -> $rose/$fell to mimic the SystemVerilog names. $rise/$fall is essentially syntax error.

Also changed $changed to !$stable because $changed is not a support SystemVerilog name either.

Tom

### DIFF
--- a/core/src/main/scala/spinal/core/Formal.scala
+++ b/core/src/main/scala/spinal/core/Formal.scala
@@ -6,8 +6,8 @@ object Formal {
   def past[T <: Data](that : T, delay : Int) : T = that.formalPast(delay)
   def past[T <: Data](that : T) : T = past(that, 1)
 
-  def rise(that : Bool) : Bool = that.wrapUnaryOperator(new Operator.Formal.Rise)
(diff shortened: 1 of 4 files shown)


tomverbeure:
After the second check-in, 'sbt test' passes (except for the Cocotb test, because I don't have that installed.)

Tom

tomverbeure:
@Dolu1990 What is your opinion on this kind of patch? Do you want rose/fell to be compliant with SystemVerilog or do you want rise/fall (which translated into $rose/$fell during generation) to be consistent with your current .rise and .fall methods?

I would like to have this settled soon so that I don't need to link to my own version of SpinalHDL while writing code that uses it.

Tom

Dolu1990:
I would say, let's got the SystemVerilog way, it is also compatible with the yosys formal stuff right ?

Dolu1990:
So, ready to merge this pull request ^^ ?

tomverbeure:
Go for it!

Yes, that’s the way it’s done with Yosys formal.

Tom

PR #194, opened 2019-01-19
Yosys integration

wifasoi:
### Introduction
This PR will introduce a way to interface/integrate with yosys without leaving scala.
This PR contains (in detail)
* A way to abstract any CLI-commands and keep written log of the output
* Yosys abstraction
  * Define a command, or a sequence of commands
  * Concatenate and combine yosys command
* next-pnr integration (only ice40 for now, but more target can be added easily)
  * This will call the default yosys synthesize script, if a file design is not specified)
* Formal verification (yosys-smtbmc)
  * This will call yosys smt2 model generation script, if a file model file is not given [WIP]
  * Supports different solvers (default yosys abc, because is integrated in yosys)
  * Supports different modes (live,bmc,prove,cover)

Other thins can be added (like formal equivalence checking), but the building block are here. The state of this PR is: more than usable, but still need polish (eg. docstring, remove unused include ecc..)

### How to Use it
The goal was to imitate the use of SpinalConfig and similar commands
```scala
val testFormal = SpinalConfig().includeFormal.generateSystemVerilog(new testCounter(2,10))
val dd = new YosysFlow(testFormal,"tasty").formal(Mode.prove).use(Solver.boolector).step(100).run()

// YosysFlow.formal will return an istance of FormalCommand, so you can have a monline or:
val dd = new YosysFlow(testFormal,workDir="tasty").formal()
dd.mode(Mode.prove)
dd.use(Solver.boolector)
dd.step(100)
dd.run()
```
note: I use the SpinalReport to get the rtls paths, and the top module name

example: https://gist.github.com/wifasoi/309fd0196ba7ce648b6d9832ed67c82e

### Improvements
This PR is not perfect, so **feedback is welcome and encourage**, but i have a personal to-do list:
* Have a class that contain the various output paths of the generated file (like SpinalReport)
* backtracking the line that triggered the formal verification engine (this will help: https://github.com/YosysHQ/SymbiYosys/issues/24#issuecomment-451110632)
* add formal equivalence checking
* integrate area/timing report (I can use icetiming, but I wold prefer to scrape nextpnr output for that)


Dolu1990:
It look good :)

Dolu1990:
There is two general experience feedback i can give :

1)
One thing about the SpinalConfig stuff is that i regrets having it the way it actualy is ^^

SpinalConfig(xxx, xxx ,xxx) is quite borring
I think i prefer 
SpinalConfig.xxxx(yyyy).xxxx(yyyy)
or 
val c = SpinalConfig
c.xxx(yyyy)
c.xxxx(yyyy)

A bit like the SpinalSim stuff :)

What is your preference ?


2)
One good thing about makefile, is incremental compilation / testing ^.^
That's realy one thing which aren't there natively with Scala based testing / unitary tests.
And there as we go to formal stuff, which could take very much time to test things, it can start to be a real issue. (for instance riscv-formal can take easily 4-8 hours)

I don't know how exactly we should aproach this.

wifasoi:
1) I like the spinalsim command style better, I'll try to mimic it as much as possible (transform YosysFlow in a case class)

2) As much I love use gnu make, I don't think is useful in this case because Yosys need to read all the verilog files to synthesise/create a model(Maine I'm wrong on this).. so if you modify something, all the chain need to be executed.. make will do nothing to save time here.

One way to save time is to dump a verilog test bench with the bug the formal solver found, and iterate with that until the bug is solved. For this I can add some basic support for icarus verilog.

GNU make can save time if you want to formally verify the same design with different parameters (eg. Run BMC and then prove). But you can do that manually by use the same model (the implementation now is a bit "flaky" but the idea in the future is to have an object like spinalReport that hold all the path, so you can pass it around to the function than need a model).

If you find a case where use "make" is better than handle it in scala, tell me (or if someone wants to add something to this discussion, please do so). Anyway I'm not totally against use make, so with the right use case I can change mind quite easily :)

Dolu1990:
So, not saying there is no way to workaround it with scala, but the practical case i'm thinking about with RISC-V formal was : 
- I have 40 different formal test to run on VexRiscv, which take hours and hours
- Some will fail, some will pass
- Each test which passed create its PASS file, which tag it as successfull
- I change something in VexRiscv, and i want to rerun only the failing ones.
- with make, only the tests without the PASS file are runned again.

Probably this kind of infrastructure shouldn't be part of the SpinalHDL / SpinalSim / SpinalYosys API, but something specific, case by case ^^

wifasoi:
Ok, I'm ready for round 2
need to create a better class/function for composing a complete makefile (for now `CreateMakefile(op: Makeable*)` works fine for testing)
I moved the core component for this integration on a separate file `makeable.scala`

### How to use it (right now)
The heart of the makable trait is the `prerequisite` list and the `|>` operator:
`val prog = InputFile("test.bin") |> IceProg()`
`prog` will equal a copy of IceProg() with InputFile(Paths.get("test.bin")) in the dependencies list

you can then use `prog.makejob` to generate the make jobstring, ot you can `prog.makefile` and generate all the makejob required to satisfy the prog command/job (`.makefile` will crawl the dependency tree for you)

If the job description is more complex, you can use [CreateMakefile](https://github.com/SpinalHDL/SpinalHDL/pull/194/files#diff-d812d6a4718e55390f2e3529ac30072aR120) and collect and specify all the leaf commands/job, and it will generate a makefile without duplicating jobs (is not intuitive, but it works great, I'll work more on this utility in the future)

```scala
val test = SpinalConfig.includeFormal.generateSystemVerilog(new testCounter(2,10))
val pw = new PrintWriter(new File("Makefile" ))
//YosysFlow.ice40Flow -> InputFile(report) |>(Yosys.load_SystemVerilog(report) + Yosys.synthesize("ice40")).outputFolder(Paths.get(workDir,"syntesys")) |> NextPNR_ice40().outputFolder(Paths.get(workDir,"pnr"))
val ice = YosysFlow.ice40Flow(test)
val ice_1 = InputFile(Paths.get("test.pcf")) |> ice
val ice_2 = form2.outputFolder(Paths.get("programming"))
pw.write(CreateMakefile.makeFolderStructure(ice_2)+"\n\n")
pw.write(CreateMakefile(ice_2))
pw.close()
```
#### Result
```makefile
DIRS=makeWorkplace/syntesys programming/makeWorkplace/pnr programming/makeWorkplace/pnr
$(info $(shell mkdir -p $(DIRS)))


makeWorkplace/syntesys/syntetized.json : testCounter.sv
	yosys -p'read_verilog -sv testCounter.sv; prep -top testCounter; synth_ice40; write_json  makeWorkplace/syntesys/syntetized.json' &> Yosys.log

programming/makeWorkplace/pnr/PASS programming/makeWorkplace/pnr/nextpnr-ice40.asc : makeWorkplace/syntesys/syntetized.json test.pcf
	nextpnr-ice40 --json makeWorkplace/syntesys/syntetized.json --quiet --randomize-seed --no-tmdriv --pcf test.pcf --hx1k --asc programming/makeWorkplace/pnr/nextpnr-ice40.asc --package ct257  && date > programming/makeWorkplace/pnr/PASS
```
### Yosys integration
You can add an input/output file via the `addInputFile/addOutputFile` function
You can add commands via the `addCommand` function.
You can chain together yosys scripts via the `+` operator

The separation between Input Ouput and Command is necessary for the makefile integration, so you can use it like any comand above (for example [here](https://github.com/SpinalHDL/SpinalHDL/pull/194/files#diff-3b267794213ef7556fd311262b7d74b3R11))
### What's need some polish
- I'll like to keep a way to run the command via a .run method, this can be useful if you want integrate test with scaladoc, need to refacotr/modify the trait `Executable`
- A container that keep track of the jobs in the makefile: so i can add commands with the various dependencies, and the container will keep track of all the leafs of the tree and generate the makefile for the user (instead to rely on `.makefile` and `CreateMakefile`
- use af mutable list instead of a normal list in yosys: I'll try to get rid of all mutable list for the immutable counterparts (from old version of the code)

### Conclusion
@Dolu1990 I'll like some feedback on the UX of this PR (I tried to explained above) when you have the time. The main files are `yosys.scala` and `makeable.scala` (the main logic behind this pr)
Unfortunately this PR is **NOT** merge-whorty at this time (IMHO), but the codebase is relatively stable, enough to ask for some review

wifasoi:
Sorry for the wall of text:
TL;DR: not finished, but want some opinion on UX

mzau:
to be able to spawn parallel tests for example i'd like to have something like
```
val test1 = SpinalConfig.includeFormal.generateSystemVerilog(new testCounter(2,10))
val test2 = SpinalConfig.includeFormal.generateSystemVerilog(new testPwm(...))

val ice = test1.success and test2.success |> YosysFlow.ice40Flow
...
```
or 
```
val someflow = (test1, test2, test3).success |> YosysFlow.someFlow
val anotherflow =  (test1, test2, test3).success |> VivadoFlow

val all = (YosysFlow, VivadoFlow).createMakfile  // optionally
```
I'm not so confident in scala to say thats feasible, but intuitive^^

wifasoi:
@mzau thank for your feedback. about:
```scala
val all = (YosysFlow, VivadoFlow).createMakfile  // optionally
```
I liked the idea of have a  tuple/list of flow, and do operation on it, so i implement it.
About the having a bunch of test and trigger, the ground work is already there, i just need to add a way to pass targets (that are not files) to the next command in a sane way

piegamesde:
What needs to be done for this getting merged?

wifasoi:
> What needs to be done for this getting merged?
Two things...
- Fixing some path nonsense
- Have a bit of time (now i have some, so you'll see some new commit)

wifasoi:
There are still some issue (I hate working with Path so much).
The code right now ill work fine, unless you use the "outputFolder" parameter.
So this is still a WIP (I was hoping to finish it at 36C3 but it needs more time to be usable)

piegamesde:
@wifasoi I've tried it out and there is still a lot of work to do – do you prefer a GitHub code review or direct contributions to the pull request?

wifasoi:
@piegamesde both is good.
We can keep here discussion) and for fast code review you can comment here.
If you want partecipate for "taming the beast", you can propose a PR on my fork.
If you want to chat, you can @wifasoi on the spinalHDL gitter channel and I'll try to respond there.

In any case, I'll try to respond as soon as possible.

lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala
```diff
@@ -0,0 +1,297 @@
...
+    modelGen.addCommand("opt", "-keepdc", "-fast")
+    modelGen.addCommand("check")
+    modelGen.addCommand("hierarchy", "-simcheck")
+    modelGen
+  }
+
+  def synthesize(target: String): Yosys = {
+    val command = Yosys()
+    command.addCommand("synth_" + target)
+    command.addOutputFile("syntetized.json", "json")
```
piegamesde: Please don't hardcode the output file type to JSON. In my current setup, I run Yosys with `synth_xilinx -top MyTop -edif output.edif` so I can pipe it directly into Vivado.

lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala
```diff
@@ -0,0 +1,297 @@
...
+import java.io.PrintWriter
+import scala.sys.process._
+import scala.collection._
+import spinal.core._
+
+import org.apache.commons.io.FilenameUtils
+
+trait YosysScript
+
+/** Class that rappresent a yosys command
```
piegamesde: "Class that represents a yosys command"

lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala
```diff
@@ -0,0 +1,297 @@
...
+    command.addOutputFile("syntetized.json", "json")
+    command
+  }
+
+  def formal[T <: Component](report: SpinalReport[T],
+                             mode: String = Mode.bmc,
+                             multiclock: Boolean = false,
+                             memoryMap: Boolean = false,
+                             workDir: String = ".") = {
+    loadSystemVerilog(report) +
```
piegamesde: Formal verification isn't limited to SystemVerilog. Only because the other backends haven't implemented the formal synthesis yet doesn't mean they won't ever have it (see my PR for example).

wifasoi: This function is just a concatenation of various yosys scripts (load + model + export), the user can take the simple building block and change the part that he needs.
For now I'll change the name of the function to "svFormal" so after we can add other formal flow without naming ambiguities.

lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala
wifasoi: I agree, I should not force an output in this command.

piegamesde:
I tried the formal flow (`SymbiFlow.svFormal`) and I get:

```
Exception in thread "main" java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:349)
	at scala.None$.get(Option.scala:347)
	at spinal.lib.eda.symbiflow.FormalCommand.toString(formal.scala:150)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.io.PrintStream.println(PrintStream.java:897)
	at scala.Console$.println(Console.scala:267)
	at scala.Predef$.println(Predef.scala:393)
```

This is because `FormalCommand._smt2` is `Optional`, but the `toString` method simply assumes it will exist. (I have no clue what smt is btw, some more documentation would be appreciated)

wifasoi:
First of all, I thank you to take time and review my code.

Let's get down to business:
Documentation is on the way (I already wrote one page, but didn't push it on my spinaldoc-RTD fork).
For clarity, a bit of background:
This mainly PR implement mainly the "executable" and "makeable" trait.
- "executable" just run a command abstracted via a case class, implements the "run()" comand and take care to log the command output.
- "makeable" will create a makefile out of executable abstracted comand, and make sure that the output of a command goes to the input of another.
This relation is given from the user via the "|>" function:
`command1(parameters) |> command2(parameters)`

The original PR was for the Yosys abstraction layer, but after some discussion there seams that a makefile generator was a nice to have feature.

The yosys class is "executable" and "makeable", so it can be used in the makefile generator or run stand alone in scala.

A clarification on Symbiyosys:
I wanted to create a symbiyosys wrapper/sby file generator at first... but after some time i find out that symbiyosys dosn't do mutch: it just a makefile generator (ans some useful functionality on the side).
I wanted to abstract yosys and yosys-smtbmc anyway, so the functionality similar to symbiyosys was given for free by the "executable" and "makeable" trait.
So i just implemented a way to generate a smt2 model via the Yosys class ([here](https://github.com/wifasoi/SpinalHDL/blob/58240a6f656552c6dd832f1377368dd9d92288d8/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala#L82-L109) the code).
Then i abstracted all the flag of yosys-smtbmc (code [here](https://github.com/wifasoi/SpinalHDL/blob/58240a6f656552c6dd832f1377368dd9d92288d8/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/formal.scala#L25), yes i should change the name).
Finally I just concatenated the commands together via `|>` (code [here](https://github.com/wifasoi/SpinalHDL/blob/58240a6f656552c6dd832f1377368dd9d92288d8/lib/src/main/scala/spinal/lib/eda/symbiflow/workflows.scala#L22)).

The `workflow.scala` will be a collection of ready to go makefiles/workflow.
for example [svFormal](https://github.com/wifasoi/SpinalHDL/blob/58240a6f656552c6dd832f1377368dd9d92288d8/lib/src/main/scala/spinal/lib/eda/symbiflow/workflows.scala#L22) will take a SpinalReport (that is returned when you run `generateSystemVerilog()`) and generate a makefile (than if executed will run the formal verification, in this case) based on the generated file by SpinalHDL. An example can be found [here](https://github.com/wifasoi/SpinalHDL/blob/58240a6f656552c6dd832f1377368dd9d92288d8/tester/src/main/scala/spinal/tester/PlayDev.scala#L1437-L1443)

I hope this helps navigating my code.

---

For the error part,
I should throw an error in this case, because smt2 file is a requirement of the FormalCommand. (so this is a bug)
This should be checked when the command string is generated, but i borked something in the flow (maybe i forgot to add the output command for the formal model to the yosys yesterday)
I'll try to fix it tomorrow (if not even today.. but I'm a bit busy right now).

For now try to check out my similar [code ](https://github.com/wifasoi/SpinalHDL/blob/58240a6f656552c6dd832f1377368dd9d92288d8/tester/src/main/scala/spinal/tester/PlayDev.scala#L1437-L1443) for that use svFormal, and check the if there are some difference.
If you can @piegamesde can you post a snippet of your code here?

piegamesde:
> Documentation is on the way (I already wrote one page, but didn't push it on my spinaldoc-RTD fork).

That's great, I'll try to have a look at it once it's online (you can open pull requests prematurely and mark them as "draft" in GitHub). Without having seen it, I suppose this won't be enough. While readthedocs makes a great tutorial and gives high-level overview over a project, in-code Scala documentation has the advantage that it will show in users' IDE, making them more productive.

> "makeable" will create a makefile out of executable abstracted comand

When you refer to "makefiles", are you talking about the GNU/make tool or the generic concept of a make file?

> i find out that symbiyosys didn't do much: it just a makefile generator (ans some useful functionality on the side).

I know, symbiyosys doesn't do a lot, but it's a really nice front-end to use. As long as smtbmc is as easy to use as sby files (plug in some Verilog and configure the proof mode), I'm fine with any solution.

> I hope this helps navigating my code.

Yes indeed, thanks

> If you can @piegamesde can you post a snippet of your code here?

Sure:

```scala
    val yosys = SymbiFlow.svFormal(report, Mode.cover, false, false, "test123")
              .solver(Solver.yices)
              //.smt2(Paths.get("bla")) // <-- Using this fixes the error
    println(yosys)
```

wifasoi:
> While readthedocs makes a great tutorial and gives high-level overview over a project, in-code Scala documentation has the advantage that it will show in users' IDE, making them more productive.

Well, most of(if not all) function have a scaladoc string already, so that part of the documentation is done. But the code need a bit of overview before someone can understand "how the cogs runs" in the system.

> When you refer to "makefiles", are you talking about the GNU/make tool or the generic concept of a make file?

I meant the "Makefile" GNU/make can understand.

> I know, symbiyosys doesn't do a lot, but it's a really nice front-end to use. As long as smtbmc is as easy to use as sby files (plug in some Verilog and configure the proof mode), I'm fine with any solution.
Yes, I too love symbiyosys: is easy and functional :)
Using smtbmc by itself is not difficult, the more difficult part is to create a model file. In symbiyosys the model command is hardcoded, but make you change some behavior before it runs the model generato yosys script. I emulate the same behavior (or at least I'm trying to).

> ```
    val yosys = SymbiFlow.svFormal(report, Mode.cover, false, false, "test123")
              .solver(Solver.yices)
              //.smt2(Paths.get("bla")) // <-- Using this fixes the error
    println(yosys)
```
### Long story short
If you want print the resulting makefile you should use `println(yosys.makefile)`.

### Short story long (the longer version)
`SymbiFlow.svFormal` will return a `FormalCommand` with `prerequisite` populated with a Yosys class that will generate a smt2 file.
Right now it should trow an error, because the leaf class function "toString" doesn't care about parent classes for calculating input files. If you specify the smt2 file, it doesn't crash because you specify the required file. This is by design, so the user can overload the automatic file association implemented in the "makeable" trait.

The `makefile()` function will craw the dependency tree backward and calculate all the necessary step (and in/out dependecy file) to generate the desired outcome.

### In conclusion
The error is a design one: I should not overload `toString` (funny enough a class that extends "makeable" have the option to override `def makeComand: String = this.toString`, so i should use this instead of relying on `toString`... but i was a bit lazy and didn't change it, but it's a easy fix)

P.S:
if you have multiple "leaf" class, you can bundle it together in a tuple/list, and call `(leaf1,leaf2,leaf3).makefile` to generate the minimal makefile that satisfy the dependency tree of all the leafs in the tuple/List

**Comment by piegamesde** [comment]
What is the simplest way to run the verification without make? I want to make it all from code so that I can make turn my formal proofs into unit tests. Simply running `yosys.makeCommand` isn't enough since it won't build all the dependencies (the smt2 file for example).

**Comment by wifasoi** [comment]
You can do:
(Assuming the snipet of the code above, yosys is a variable, not the class name)
`yosys.run("all")`
And will generate and run the makefile (you need gnu-make installed).
There is an option to change the target name and have a phony one if you like)

Otherwise you need to handle it manually (and so you're responsible to put the file in the correct path) by running the command one by one like:
```
(
  Yosys.loadSystemVerilog(parameters) + 
  Yosys.model(parameters) +
   Hosts.export(parameters)
).run().  //Needed for generating the smt2 file
FormalCommand(parameters).run()
```

**Comment by wifasoi** [comment]
> Also, please make sure that the output directories exist before running commands that will put files into them. (You may assume that the working directory exists, but not for any subdirectories)

I recreate the folder each time (the first line in the makefile is mkdir -p [list of dirs]

Be carefull with the "clean" target though (it does rm -rf [list dir/files](I should change this))
I suggest to not use it right now.

Tell me if it works ok for you out of the box

**Comment by piegamesde** [comment]
I can't find any `run` method.

I played around with the `makeComand` methods and the commands seem to be correct. But while they work from the shell, I can't get Yosys to run from `sys.process.Process`. I found out that it requires the command to be split by arguments in a `Seq` (please implement that, so that `makeCommand` doesn't give me a plain string anymore).

`sys.process.Process.apply(Seq("bash", "-c", yosys.prerequisite(0).makeComand)).!` works for me and will build the stm2 file.

**Comment by piegamesde** [comment]
Also, you need to tell Yosys to set the `FORMAL` macro. (I just lost half an hour on debugging why it was optimizing away all my modules, until I remembered that all the formal code is guarded by `ifdef	FORMAL`)

**Comment by wifasoi** [comment]
`makeComand`, as you discovered, should not be used by the end user (still useful for debugging reason). This is the function that give a string from a abstracted terminal command (and required to implement by a "meakable" class). Is not designed to be a runnable string, but something used by the function implemented in the "makeable" trait.


---

@piegamesde Ok my bad for the run method, I did not implement it for a signle "makeable" class, I implemented only for a `Seq[makeable]` (implemented [here](https://github.com/wifasoi/SpinalHDL/blob/58240a6f656552c6dd832f1377368dd9d92288d8/lib/src/main/scala/spinal/lib/eda/common/makeable.scala#L172), yes is an implicit class)
So, for now, just do:
`List(yosys).run("all")`

I'll see what I can do today, and at least add the formal flag in the import script.

Still thank you for your invaluable testing, without it I wild have not found this little/big bug.
Soo keep breaking this for me :)

**Comment by piegamesde** [comment]
I tried out the command, and it just overwrote my `Makefile` in the project's root directory!? I'm generally not a fan of the fact that Spinal simply dumps its outputs in the root directory by default, and there's why. Have you thought about the situation where there are multiple formal proofs in a program? When their output is in the same folder (because why not?), they will overwrite the other make files every time (same goes for the log files).

I would write the logs of commands to stdout instead of a file by default (yosys failed and I simply got `make: *** [Makefile:15: test123/modelgen/unnamed.smt2] Error 1` which confused me for a few minutes). The API does not seem to expose any methods to disable the log file, I only found `.copy(logFile=None)` which does not feel right. Even that does not work as intended, because it only affects the top level task, not any of its dependencies.

**Comment by wifasoi** [comment]
> I tried out the command, and it just overwrote my Makefile in the project's root directory!?

Sorry about that :(
yep the pain point of this pr is move the makefile in another directory (this is why this branch is not merged yet).

> Have you thought about the situation where there are multiple formal proofs in a program? 

Yes, multiple test are supported, this is why you can specify all the leaf you want in a list/tuple like:
`
val testFormal1 = SymbiFlow.svFormal(report, Mode.bmc, false, false, "bmcTest")
val testFormal2 = SymbiFlow.svFormal(report, Mode.prove, false, false, "proveTest")
List(testFormal1,testFormal2).run("all")`
Othe usefull feature are:
- `<FormalCommand class>.phony("phonyTargetName")` (so you can group test)
- `<FormalCommand class>.pass("path to Pass file")` (so you can interface it with external makefile, will generate a Pass file on sucess)

> I would write the logs of commands to stdout instead of a file by default (yosys failed and I simply got make: *** [Makefile:15: test123/modelgen/unnamed.smt2] Error 1 which confused me for a few minutes)...

Fair, I'll add it to the todo list.

> Regarding my last comment yesterday, you need to add the verilog_defines -DFORMAL; command to yosys: http://www.clifford.at/yosys/cmd_verilog_defines.html

This or just use the flag "-formal" with "load_verilog"

**Comment by wifasoi** [comment]
So for now the todo:
- [ ] usage Documentation
- [ ] save the makefile in a workdir (where all path are relative to) (this only for Seq[makeable])
- [ ] do not rm -rf if possible, or make user configurable
- [ ] log level
- [ ] if not specified, log to stdout
- [ ] make SymbiFlow.svFormal load verilog file with the "-formal" flag

Nice to have:
- [ ] make a way to make a Pass file/target as requirement (multistage testing)

**Comment by piegamesde** [comment]
> Yes, multiple test are supported, this is why you can specify all the leaf you want in a list/tuple like:

This requires me to define all my formal tests in one place. But I usually define only one per unit test and run them independently of each other. Your proposed solution does not work in this case. (I could work around it by defining that common code in a `beforeClass` block, but this will only postpone the problem to the point where I have different formal test suites. I don't think this will scale well to large, formally proven libraries.)

> make a way to make a Pass file/target as requirement (multistage testing)

This is a nice thing to have, many testing libraries offer tests depending on others and I will make great use of it (first do all cover statements, then BMC, then induction proof).

**Comment by wifasoi** [comment]
> This requires me to define all my formal tests in one place. But I usually define only one per unit test and run them independently of each other.

What do you mean?
Are you trying to add the formal verification as one unit in the unit test suite? (with junit I think)
Are you trying to o a single Component formal verification, or multi component?

You'll need to change the "formal code" if you want to verify a multi-component Project (otherwise you risk that the formal verification is not solvable in a timely meaner).

**Comment by piegamesde** [comment]
Yes, I want to do unit tests (with JUnit or scala test) that prove one piece of hardware each. This may be a single component as well as a composition of multiple components. Slow unit test aren't really that much of a problem, since they can be put together in a "slow" test suite that is run less often.

-----

I got the make file somewhat working, but the main target of it has no name:

```
[… other make targets]
 : test123/modelgen/unnamed.smt2
    [… yosys command]
```

Therefore, `make` won't execute that last command: `make: Nothing to be done for 'all'.`.

**Comment by wifasoi** [comment]
mmmm.. 
a easy fix is to add:
`<FormalCommand class>.phony("ponytargetname")`

Explanation:
This will make the target a phony one (so make can execute it) and will be added in the "all" target by my code.

Conclusion:
I should add some sane default in formalCommand for the no-output case (another item in the todo list)

**Comment by piegamesde** [comment]
@wifasoi Any progress on this?

**Comment by wifasoi** [comment]
Status report:
Didn't do much.

It's tax season/upgrade season, so I was, and still I am, a bit swamped.
But good news, I'll have time after the 20 of may (end of my contract with my company), and I'm in pseudo-quarantine (well all Italian citizen are)... so I'll have plenty of time on my hand :)
Expect some more commit by the end of the month.

**Comment by gyn** [comment]
@wifasoi Could I know the status of this feature? I heard formal verification at zipcpu.com, and wish that symbiyosys could just use the produced verilog file with formal sentences directly.

**Comment by wifasoi** [comment]
Status report:
<insert generic "life is hard" excuse here>
But in relaity, I had a surge of freelancer work that is eating my free time. So not much progress (the only thing left is debug path issue, and create some unit test)

Anyway, symbiyosys shuold read the files ok (I personally use symbiyosys so i can abuse the cover statement as a lazy testbench).
Remeber to export from SpinalHdl as system verilog, and use the ".withFormal" flag, and add it as a list of files in the config file. Otherwhise, if you need more help you can tag me in the SpinalHdl gitter chat

**Comment by jiegec** [comment]
I am using yosys for formal verification of SpinalHDL generated code and found this pr. This pr does exactly what I need: do formal verification without leaving Scala. Is there any progress?

**Comment by wifasoi** [comment]
No progress so far,
I'm round robin between a dozen projects, and this year the pool of free hour has shrunken considerably (so no progress).
This pr it's on limbo, but not dead. This is almost complete, the blocking problem is the path issues (fun times, I tell you).

**Comment by jiegec** [comment]
@wifasoi I am reading the source code, hoping to fix the path issues. I have a question: What's the difference and relationship between `target` and `getTarget` in `makeable` trait?

**Comment by wifasoi** [comment]
For now there is no difference. The only one is that getTarget is used for generation of the makefile, so you can return a modified copy (like with changed paths).

**Comment by piegamesde** [comment]
Hi, can you please rebase you branch (and optionally check that none of your changes broke)? It branched off a few years ago, I'm sure SpinalHDL has changed quite a bit since then.

**Comment by wifasoi** [comment]
As this PR will not effect file in the src/lib folder, nor existing one it will rebase cleanly (just tried).
I'm waiting to merge a PR on my PR before I can push my rebased branch.

**Comment by Readon** [comment]
Are there any existing issues preventing it from being widely adopted?

**Comment by wifasoi** [comment]
I'm in path resolving hell :D (need to re-check it).
The PR is still open on my baranch... so maybe I need to work on that first.
after that, there is not else much to do. I'm just lazy :D (and too many project and only 24 Hour for day)

**Comment by Readon** [comment]
> I'm in path resolving hell :D (need to re-check it). after that, there is not else much to do. I'm just lazy :D (too many project and only 24 Hour for day)

ahh，is it necessary to import Makefile when there is a sbt？ have you investegated spinalsby？[spinalsby](https://github.com/craigjb/spinalsby/blob/master/src/main/scala/spinalsby/spinalsby.scala)

**Comment by wifasoi** [comment]
Sorry for the late reposnse.. <insert joke about being a year since i responded here>
This pr implement a way to compose your sby script (class Yosys), a way to abstract binary application (trait executable), and a way to create a makefile with a simple dependency language.

The way i implemented the sby script composition is similar to the one you linked ([My take](https://github.com/SpinalHDL/SpinalHDL/blob/3eff2a428d8a838393e398475dfe051c01a90f34/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala))
You can see I defined some script already [here](https://github.com/SpinalHDL/SpinalHDL/blob/3eff2a428d8a838393e398475dfe051c01a90f34/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/snippets.scala).

I see now that the yosys class is not runnable, So maybe i could make it so (so you don't need to use the makefile approach, if you don't want).

I think this PR is almost ready for prime time... but i need some time to re-test. or I could just drop the makefile part of the pr (the one that give me problems).

**Comment by Readon** [comment]
> Sorry for the late reposnse.. <insert joke about being a year since I responded, here> This pr implement a way to compose your sby script (class Yosys), a way to abstract binary application (trait executable), and a way to create a makefile with a simple dependency language.
> 
> The way i implemented the sby script composition is similar to the one you linked ([My take](https://github.com/SpinalHDL/SpinalHDL/blob/3eff2a428d8a838393e398475dfe051c01a90f34/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala)) You can see I defined some script already [here](https://github.com/SpinalHDL/SpinalHDL/blob/3eff2a428d8a838393e398475dfe051c01a90f34/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/snippets.scala).
> 
> I see now that the yosys class is not runnable, So maybe i could make it so (so you don't need to use the makefile approach, if you don't want).
> 
> I think this PR is almost ready for prime time... but i need some time to re-test. or I could just drop the makefile part of the pr (the one that give me problems).

also a late response, ahhh.

I read the code and trying avoid to use MakeFile method, however what I found is that in SymbiFlow.svFormal function the InputFile is used, where the InputFile implement MakableFile traits. I could not tell if the usage of svFormal only would not take a side effect. 

Can you give an example that do not use Makefile but command directly?

**Comment by Readon** [comment]
> Sorry for the late reposnse.. <insert joke about being a year since I responded, here> This pr implement a way to compose your sby script (class Yosys), a way to abstract binary application (trait executable), and a way to create a makefile with a simple dependency language.
> 
> The way i implemented the sby script composition is similar to the one you linked ([My take](https://github.com/SpinalHDL/SpinalHDL/blob/3eff2a428d8a838393e398475dfe051c01a90f34/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/yosys.scala)) You can see I defined some script already [here](https://github.com/SpinalHDL/SpinalHDL/blob/3eff2a428d8a838393e398475dfe051c01a90f34/lib/src/main/scala/spinal/lib/eda/symbiflow/yosys/snippets.scala).
> 
> I see now that the yosys class is not runnable, So maybe i could make it so (so you don't need to use the makefile approach, if you don't want).
> 
> I think this PR is almost ready for prime time... but i need some time to re-test. or I could just drop the makefile part of the pr (the one that give me problems).

What I found is that the Makefile was tightly coupled with ice tools. When I commented the syntesize related command, Makefile would still generate Makefile searching for ../../test.pcf, which fail the runs.
Here is code I modified in PlayDev.scala.
``` scala
  val form1 = SymbiFlow.svFormal(test,workDir="pesto")
    .log()
    .solver(Solver.z3).prove
    .dumpVCD("pesto/test.vcd")
    .outputFolder("pesto/test1")
    .phony("test5")
    .append(100)
  val form2 = SymbiFlow.svFormal(test,workDir="pesto").phony("test1").solver(Solver.z3).outputFolder("pesto/test2").phony("test").append(100).log()
  val form3 = SymbiFlow.svFormal(test,workDir="pesto").phony("test2").solver(Solver.z3).outputFolder("pesto/test3/1/2/3/").append(100).log()
  val form4 = SymbiFlow.svFormal(test,workDir="pesto").phony("test3").solver(Solver.z3).outputFolder("pesto/test4").append(100).log()
  // val syntetize = SymbiFlow.syntesize(test,"test.pcf")
  //val make = InputFile(test) |> List(form1,form2,form3,form4)
  // println(make.makefile + "\n" + make.bundleTest() + "\n\n" + make.bundle("test2")({case x: Yosys => x}))
  // pw.close()
  // Makefile(form3,form4,syntetize).workDir("test2electricbogaloo/fff").run("all")
  Makefile(form3,form4).workDir("test2electricbogaloo/fff").run("all")

```

wifasoi:
I kwon is too late to respond now :D

yes, I procastoineted and didn't modified my code. As the Yosys-smtbc class doesn't implment a run() function (by using the "runnable" trait) you shuld just:

scala```
val form1 = SymbiFlow.svFormal(test,workDir="pesto")
    .log()
    .solver(Solver.z3).prove
    .dumpVCD("pesto/test.vcd")
    .outputFolder("pesto/test1")
    .phony("test5")
    .append(100)
    .run()
```

This will repalce what sby does (minus multiple solvers and fancy configuration) and add the ability to play with the yosys script (svFormal is a composite of different yosys sniptes I exstracted from the sby codebase)

numero-744:
Hello,

This PR is quite old ^^' what is its current status?

PR #211, opened 2019-04-26
Foldable unit test

wifasoi:
Is still not perfect, bet you can check the result here:
https://travis-ci.com/wifasoi/SpinalHDL/jobs/195742247

There are still some u-test that do not behave, so I'll mark this PR as a WIP.

Explanation: (the magic souce is [here](https://github.com/SpinalHDL/SpinalHDL/compare/dev...wifasoi:foldable-utest?expand=1#diff-e2080510c7aa6a0d1ccc828264aab096))
- This PR will print the travis "magic string" to permit each unit test to have a foldable log.
- Each fold will have a name corresponding to the unit-test class name.
- This behavior will trigger only when the TRAVIS env variable is defined (the value doesn't matter)

### DIFF
--- a/tester/src/test/scala/spinal/tester/scalatest/BusSlaveFactoryDoubleRead.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/BusSlaveFactoryDoubleRead.scala
@@ -40,7 +40,7 @@ object TestTopLevel {
 
 }
 
-class BusSlaveFactoryDoubleReadTester extends FunSuite {
+class BusSlaveFactoryDoubleReadTester extends FunSuite with TravisFold {
 
   test("BusSlaveFactoryDoubleRead") {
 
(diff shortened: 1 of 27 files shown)


Dolu1990:
Ok :D
Let's me know when that's ok

wifasoi:
@typingArtist it makes sense.. I'll try that.
This PR is a bit stale because I'm a bit busy (but i can cut some time this evening for implementing your suggestion)

wifasoi:
Ok, this solution is not perfect.. but this will make travis show all the logs.

numero-744:
Hmm it seems Travis integration stopped a year ago (in favor of GitHub actions?). Maybe this PR can be closed?

numero-744:
Hmm it seems Travis integration stopped a year ago (in favor of GitHub actions?). Maybe this PR can be closed?

wifasoi:
Agree

PR #212, opened 2019-05-15
Pull from Upstream to get all the nifty new stuffle.

madscientist42:


Dolu1990:
ahh ? what was it ?

madscientist42:
Nothing yet.  Was a fubar on my part.  Finally got time to move CRC out of
Crypto to main and was trying to pull from the upstream and screwed up.
Should be providing a real PR there shortly.  Was going to add the missing
CRC-7 support ability after the move so we can add a SD/MMC core for the
growing stuff we have now.

PR #217, opened 2019-05-28
remove output/temp folders from git visibility

typingArtist:


Dolu1990:
Which tool produce those extensions ?

typingArtist:
Sorry for the late response.

They come with scala language servers, here from running the ide-scala plugin in atom.

https://scalameta.org/metals/

PR #220, opened 2019-06-05
Add state machine variant with conditional state transition

### LINKED ISSUE #219: Conditional state transition for statemachine
Is it possible to only transition a StateMachine when a condition is met? Obviously, this is possible with `when()` inside the states, but what if the condition is always the same? Let's say we have the following code:

```scala
val cond : Bool = /* ... */

val fsm = new StateMachine {
  val STATE1 = new State
  val STATE2 = new State
  /* ... */
  STATE1.whenIsActive {
    when(cond) {
      goto(STATE2)
     }
  }
  STATE2.whenIsActive {
    when(cond) {
      goto(STATEN)
     }
  }
}
```

Instead maybe something like this:

```scala
val cond : Bool = /* ... */

val fsm = new StateMachine {
  val STATE1 = new State
  val STATE2 = new State
  /* ... */
  /* Imaginary API */
  onlyTransitionWhen(cond)
  STATE1.whenIsActive {
    goto(STATE2)
  }
  STATE2.whenIsActive {
    goto(STATEN)
  }
}
```

I.e. only latch `stateNext` into `stateReg` when `cond` is `True`. Or would this be bad-practice?

rpls:
So this is the current WIP for the state machine with conditional transition. I opted for an auxiliary constructor to pass the condition to the state machine, as to make it clear that this will change the behaviour of the fsm. The change works by introducing a "candidate" for `stateNext` named `stateNextCand`. Then `stateNext` will only be assigned this candidate, if the transition condition is met. I chose this design to simplify the treatment of the `onEntry`/`onExitTasks` and to avoid a giant `when()` around these.

I made the synthesis of this completely optional, as to preserve the behaviour of the original state machine. I also introduced a `forceGoto` method for the `StateMachineAccessor`, which sets `stateNext` regardless of the condition. This is, for example, used in the boot state, as I think the boot state should always transition to the entry state regardless of the condition.

Closes #219

### DIFF
--- a/lib/src/main/scala/spinal/lib/fsm/Example.scala
+++ b/lib/src/main/scala/spinal/lib/fsm/Example.scala
@@ -26,7 +26,7 @@
 package spinal.lib.fsm
 
 import spinal.core._
-
+import spinal.lib.com.uart._
 
 object StateMachineStyle1 {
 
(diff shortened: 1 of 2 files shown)


rpls:
I verified this change on an Ice40 FPGA, and in the toy example the approach saves a bit of logic resources at a very small cost to frequency.

Dolu1990:
hmm maybe,
Maybe, instead of having an alternate  constructor, we could have a function call to set the condition ?

```scala
new StateMachine{
   setTransitionCondition(xxx) ?
}
````

Because ultimately a state machine could have many more things of that kind. 
Maybe that's just taste and flavor XD

But it look good :D

So do you wanna add the function ? or i merge directly ?

Dolu1990:
@rpls

rpls:
My initial idea with the alternate constructor was to "express" that this is a different type of statemachine (since it behaves differently). But I think your "speaking" API is better. So I took your suggestion into account. I updated and rebased the branch.

PR #222, opened 2019-06-13
RFC: Remove randBoot() for Cadence toolchain

dnltz:
randBoot() generates warnings in Cadence's frontend because initial values without a reset are not allowed. The easiest way is to remove the function calls, but maybe there is a better way to handle this!


Dolu1990:
The reason why those rand boot are there is to allow RTL simulation (avoiding X prop)
You can avoid the generation of the randBoot things into the generated hardware by setting the noRandBoot option in the SpinalConfig.
Is that fine ?

dnltz:
This works perfectly! I should have opened an issue for this, sorry for the unnecessary PR ;)

PR #232, opened 2019-09-05
SInt/UInt fixpoint and other usefull method extensions

jijingg:
##  1. FixPoint API extensions 
the fix-point extension information moved to #233 

## 2. Safe Adder Extensions 
| operation | description          | return               | state |
|-----------|----------------------|----------------------|-------|
| +^        | add with carry(1 bit expand)           | T(max(x,y) + 1 bits) | new   |
| -^        | sub with carry(1 bit expand)           | T(max(x,y) + 1 bits) | new   |
| +\|         | add with saturation without bit expand | T(max(x,y) bits)     | new   |
| -\|         | sub with saturation without bit expand | T(max(x,y) bits)     | new   |

 

## 3. Others

 | operation    | UInt/SInt | description                               | return            | state |
|--------------|-----------|-------------------------------------------|-------------------|-------|
| x.intoSInt     | UInt      | add 1bit 0 for Signbit                    | SInt(w(x)+1 bits) | new   |
| x.expand     | UInt      | expand 1 bit fill with 0                  | UInt(w(x)+1 bits) | new   |
| x.expand     | SInt      | expand 1 bit fill with sign-bit           | SInt(w(x)+1 bits) | new   |
| x.absWithSym | SInt      | symmetric abs -S(-128,8 bit)=U(127,7 bit) | UInt(w(x)-1 bits) | new   |
| x.sign       | SInt      | Sign bit                                  | Bool              | new   |


jijingg:
**floor,round,sround(special round)** 
image
**saturation, sym-saturation, trim** 
image

**Difference between round/sround  show by python code** 

```python
def sround(x):
    "good round performance ,lower complex for HW design"
    return  math.floor(x+0.5)

def round(x):
    "best round performance ,more complex for HW design"
    return  math.ceil(x-0.5) if x<0 else math.floor(x+0.5)

def floor(x):
    "bad performance ,simplest for HW design"
    return math.floor(x)  
```
```matlab
sround(-1.5)    = -1 //There was only a difference on -0.5
 round(-1.5)    = -2

sround(-1.4999) = -1 //same
 round(-1.4999) = -1 //same

sround(-1.5001) = -2 //same
 round(-1.5001) = -2 //same

sround(1.5)     =  2 //same
 round(1.5)     =  2 //same
```

Dolu1990:
Hi,

- x.toSInt vs x.asSInt.  So asSInt is raw convertion, while toSInt is a concervative convertion. I'm just afraid that having both so close to each others can lead to mistakes. maybe something like x.intoSInt ? 
- Probably we should link that issue in the SpinalHDL gitter channel, would that be fine ?
- Do you have much DSP math to do :D ?

jijingg:
Yes, this not finally,  It's necessary  to have more discussion.

Dolu1990:
Look good :D
About naming, fixWithRound, i'm trying to find another name. What not just "round" ?

jijingg:
Yes, the naming issue was resorted, which description in Issue #233

all lower level round api are provided as follows( roundToInf as default round)：

| API         | UInt/SInt | description                | Return      | comments |
|-------------|-----------|----------------------------|-------------|----------|
| floor       | Both      |                            | w(x) bits   |          |
| ceil        | Both      |                            | w(x)+1 bits |          |
| floorToZero | SInt      | equal to floor  in UInt    | w(x) bits   |          |
| ceilToInf   | SInt      | equal to ceil   in UInt    | w(x)+1 bits |          |
| roundUp     | Both      | simple for HW              | w(x)+1 bits |          |
| roundDown   | Both      |                            | w(x)+1 bits |          |
| roundToInf  | SInt      | most Common                | w(x)+1 bits |          |
| roundToZero | SInt      | equal to roundDown in UInt | w(x)   bits |          |
| round       | Both      | SpinalHDL chose roundToInf | w(x)+1 bits |          |

high level Factory function are unified into one  `fixTo(section,roundType)`,
fixWithRound ,fixWithFloor ...  are removed. the code will updated soon .

Dolu1990:
Ok ^^

core/src/main/scala/spinal/core/FixedPoint.scala
```diff
@@ -23,6 +23,17 @@ package spinal.core
 import scala.collection.mutable.ArrayBuffer
 import spinal.core._
 
+trait  RoundType ;
+object Ceil           extends RoundType ;// RoundUp
+object Floor          extends RoundType ;// RoundDown
+object FloorToZero    extends RoundType ;// RoundToZero
+object CeilToInf      extends RoundType ;// RoundToInf
+object RoundUpp       extends RoundType ;// RoundHalfUp    ; RoundUp->RoundUpp Temporary distinguish with roundUp
```
jijingg: @Dolu1990 `RoundUp` conflict to `misc.roundUp`(Scala seems to be case insensitive),  i temporary move to `RoundUpp`, it's not good , Do you have any better suggestions?

Dolu1990: That's one thing i don't like about Scala, enum like stuff often conflict together.
Now i'm always putting enum like stuff into a enclosing object to avoid that hell ^^

```scala

trait RoundType
object RoundType{
  object Floor          extends RoundType ;// RoundDown
  object FloorToZero    extends RoundType ;// RoundToZero
  object CeilToInf      extends RoundType ;// RoundToInf
  object RoundUpp       extends RoundType ;
} 

```

Would be fine ?

Dolu1990:
@jijingg How is it going :D ?

jijingg:
@Dolu1990 Recently, I did some simulation test, and all bug found have fixed ,The regression have been passed.  
Basically, the fixPoint(8 RoundType, Sat) are no serious problem.

Dolu1990:
Thanks !
That's much ^^
So, the single thing before merging, would be about 
https://github.com/SpinalHDL/SpinalHDL/pull/232/files#diff-31098a29a4a9d1b1142d60bc7becaa1eR27

Basicaly, one weakness of scala is it's lack of proper enum support. Using object like this can easily endup into name clash. I made the mistake with some enum in SpinalHDL as HIGH/LOW, and few others XD

What's about having  :

```scala
trait  RoundType ;
object RoundType {
  object Ceil           extends RoundType ;// Wikipedia name: RoundUp
  object Floor          extends RoundType ;// Wikipedia name: RoundDown
  object FloorToZero    extends RoundType ;// Wikipedia name: RoundToZero
  object CeilToInf      extends RoundType ;// Wikipedia name: RoundToInf
  object RoundUpp       extends RoundType ;// Wikipedia name: RoundHalfUp    ; RoundUp->RoundUpp Temporary distinguish with roundUp
  object RoundDown      extends RoundType ;// Wikipedia name: RoundHalfDown
  object RoundToZero    extends RoundType ;// Wikipedia name: RoundHalfToZero
  object RoundToInf     extends RoundType ;// Wikipedia name: RoundHalfToInf
  object RoundToEven    extends RoundType ;// Wikipedia name: RoundHalfToEven
  object RoundToOdd     extends RoundType ;// Wikipedia name: RoundHalfToOdd
}
```

?

jijingg:
Yes ,for long run , this can avoid name clash.
```scala
trait  RoundType ;
object RoundType {
  object Ceil           extends RoundType ;// Wikipedia name: RoundUp
  object Floor          extends RoundType ;// Wikipedia name: RoundDown
  ...
}
```
There a litter change: can we still use all  RoundType name with Upcase ( `Ceil,Floor, FloorToZero ... `), becuase lowacase `ceil,floor, floorToZero ... ` already used as method , is it ok ?

jijingg:
@Dolu1990 Another name issue we discuss before .
- UInt.toSInt ( potential have mening **Convet to**)  
- UInt.asSInt ( potential have mening **Treat as**)  
- UInt.intoSInt (just more literal difference, but no mening of **Convet**)

so i keep UInt.toSInt, is it Ok?

Dolu1990:
> There a litter change: can we still use all RoundType name with Upcase ( Ceil,Floor, FloorToZero ... ), becuase lowacase ceil,floor, floorToZero ... already used as method , is it ok ?

Hmm, what's about having : 

```scala
trait  RoundType ;
object RoundType {
  object CEIL          extends RoundType ;// Wikipedia name: RoundUp
  object FLOOR          extends RoundType ;// Wikipedia name: RoundDown
  ...
}
```

I mean, those are enumeration elements, and in some area, that's quite common to use upper case names for constants / enum elements. Does that sound good ?

> so i keep UInt.toSInt, is it Ok?

The issue i have with   toXXX is it is realy close to many other regular scala transformation function. Ex  :   toInt, toShort, toString
In spinalHDL, the intoSInt / toSInt would be quite special, as it extend the width to specifical keep the information.
I would realy prefer using intoSInt over toSInt to realy mark that special behaviour.
i personnaly often write toSInt / toUInt by mistake instead of asSInt / asUInt, and having a compilation error there is a good thing i think.

jijingg:
@Dolu1990 your comment is resonable. i updated.   
And  i think its almost ready for merging. Do you have any other comment?

Dolu1990:
All good for me ^^
Thanks :D
Let's me know when you want it to be merged

jijingg:
It's ok to be merged now.

Dolu1990:
Thanks :D

PR #235, opened 2019-10-04
Add support for multiple Data arguments for forkSensitive(trigger)

JobNoorman:
`forkSensitive(trigger)` can now be called using `Data` varargs like this: `forkSensitive(data1, data2, ...)`.

This commit also refactors `forkSensitive(trigger)` a bit to avoid code duplication.


Dolu1990:
Thanks ^^

PR #244, opened 2019-12-05
Add support for Nameable members inside Scala containers

JobNoorman:
This patch adds support for storing `Nameable` objects inside Scala containers. The members of the container are given a name in the same way `Vec` gives a name to its members.

I would like to get some feedback, especially on the call to `subObj.asInstanceOf[Nameable].reflectNames()`. This is done to support `Area`s inside containers but I'm not sure if its ok to call this for all `Nameable`s.

### DIFF
--- a/core/src/main/scala/spinal/core/Trait.scala
+++ b/core/src/main/scala/spinal/core/Trait.scala
@@ -491,7 +491,7 @@ trait Nameable extends OwnableRef with ContextUser{
   }
 
   def reflectNames(): Unit = {
-    Misc.reflect(this, (name, obj) => {
+    def reflectName(name: String, obj: Any): Unit = {
       obj match {
         case component: Component =>
           if (component.parent == this.component) {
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hmm i'm not sure it should be added there, but maybe there : 
https://github.com/SpinalHDL/SpinalHDL/blob/52be3be4ba2f243873d37b26b03fc998b3b3edf4/core/src/main/scala/spinal/core/internals/Misc.scala#L168

Just to check and tests, can you provide a all in one example of the exact case, just to reproduce and be sure i'm in sync with the issue ?

JobNoorman:
I'm very confused now since I cannot reproduce the issue anymore. (I created this patch because I thought that `Area`s inside a `Seq` that was created in a call to `Component.rework` were not given a proper name.)

I'll close this now and reopen if I manage to reproduce it again.

Dolu1990:
@JobNoorman Ok, no worries :)
Let's me know how it goes

PR #248, opened 2019-12-23
Various little improvements

piegamesde:
I focused mostly on the Stream/Flow classes, because I use these the most.

- Fixed formatting where I went
- Added and improved some of the documentation
- Added a few helper objects/functions that I missed while using Spinal
- Changed the build system until it worked for me in Eclipse
- Changed default assertion severity to FATAL
- Normal Verilog backend generates formal assertions as well (in an `ifdef FORMAL` block)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,18 +6,17 @@ SpinalHDL is:
 - Much more powerful than VHDL, Verilog and SystemVerilog by its syntaxes and features
 - Much less verbose than VHDL, Verilog and SystemVerilog
 - Not an HLS, nor based on the event driven paradigm
-- Only generating what you asked him in a one to one way (no black-magic, no blackbox)
+- Only generating what you asked it in a one to one way (no black-magic, no blackbox)
 - Not introducing Area/Performance overhead in your design (versus a hand written VHDL/Verilog)
 - Based on the RTL description paradigm, but can go much further
 - Allowing you to use Object Oriented Programming and Functional Programming to elaborate your hardware and verify it
 - Free and can be used in the industry without any license
 
-
-
 ## Links
 - Documentation                  <br> http://spinalhdl.github.io/SpinalDoc/
 - Presentation of the language   <br> http://spinalhdl.github.io/SpinalDoc/presentation/
 - SBT base project               <br> https://github.com/SpinalHDL/SpinalTemplateSbt
+- Gradle base project            <br> https://github.com/SpinalHDL/SpinalTemplateGradle
 - Workshop                       <br> https://github.com/SpinalHDL/SpinalWorkshop
 - Google group                   <br> https://groups.google.com/forum/#!forum/spinalhdl-hardware-description-language
 
@@ -25,7 +24,11 @@ SpinalHDL is:
 
 [![Build Status](https://travis-ci.org/SpinalHDL/SpinalHDL.svg?branch=master)](https://travis-ci.org/SpinalHDL/SpinalHDL)
 
-## SBT (Scala build tool)
+## Get it
+
+SpinalHDL is simply a set of Scala libraries. Include them into your project and you're good to go! If you're unsure about what to do, simply clone one of our example projects (see links above).
+
+### SBT (Scala build tool)
 
 ```scala
 scalaVersion := "2.11.6"
@@ -38,11 +41,25 @@ libraryDependencies ++= Seq(
 
 You can force to pick a specific SpinalHDL version by replacing the 'latest.release'. See https://github.com/SpinalHDL/SpinalTemplateSbt/blob/master/build.sbt
 
-## JAR
+### Gradle
+
+```kotlin
+repositories {
+	mavenCentral()
+}
+
+dependencies {
+	compile group: 'com.github.spinalhdl', name: 'spinalhdl-core_2.12', version: '1.3.6'
+	compile group: 'com.github.spinalhdl', name: 'spinalhdl-lib_2.12', version: '1.3.6'
+}
+```
+
+### JAR
 
     https://oss.sonatype.org/content/groups/public/com/github/spinalhdl/spinalhdl-core_2.11/
     https://oss.sonatype.org/content/groups/public/com/github/spinalhdl/spinalhdl-lib_2.11/
 
+The files are available [on Maven](https://mvnrepository.com/artifact/com.github.spinalhdl) as well.
 
 ## Changes logs
 
@@ -61,5 +78,3 @@ Your obligations (and my wish) are :
 - If you modify the SpinalHDL core (the compiler itself), please, share your improvements.
 
 Also, SpinalHDL is provided "as is", without warranty of any kind.
-
-
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ repositories {
 }
 
 dependencies {
-	compile group: 'com.github.spinalhdl', name: 'spinalhdl-core_2.12', version: '1.3.6'
-	compile group: 'com.github.spinalhdl', name: 'spinalhdl-lib_2.12', version: '1.3.6'
+	compile group: 'com.github.spinalhdl', name: 'spinalhdl-core_2.11', version: '1.3.6'
+	compile group: 'com.github.spinalhdl', name: 'spinalhdl-lib_2.11', version: '1.3.6'
 }
 ```
 
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -494,10 +494,18 @@ class ComponentEmitterVerilog(
                 case `ERROR` => "ERROR"
                 case `FAILURE` => "FAILURE"
               }
-              b ++= s"${tab}if(!$cond) begin\n"
-              b ++= s"""${tab}  $$display("$severity $frontString"$backString);\n"""
-              if (assertStatement.severity == `FAILURE`) b ++= tab + "  $finish;\n"
-              b ++= s"${tab}end\n"
+              b ++= s"${tab}`ifdef FORMAL\n"
+              /* Emit actual assume/assert/cover statements */
+              b ++= s"${tab}  $keyword($cond)\n"
+              
+              b ++= s"${tab}`else\n"
+              /* Emulate them using $display */
+              b ++= s"${tab}  if(!$cond) begin\n"
+              b ++= s"""${tab}    $$display("$severity $frontString"$backString);\n"""
+              if (assertStatement.severity == `FAILURE`) b ++= tab + "    $finish;\n"
+              b ++= s"${tab}  end\n"
+              
+              b ++= s"${tab}`endif\n"
             } else {
               val severity = assertStatement.severity match {
                 case `NOTE` => "$info"
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -990,6 +990,21 @@ class StreamDispatcherSequencial[T <: Data](gen: HardType[T], n: Int) extends Co
   }
 }
 
+/**
+ * This is equivalent to a StreamMux, but with a counter attached to the port selector.
+ */
+// TODOTEST
+object StreamCombinerSequential {
+  def apply[T <: Data](inputs: Seq[Stream[T]]): Stream[T] = {
+    val select = Counter(inputs.length)
+    val stream = StreamMux(select, inputs)
+    when (stream.fire) {
+      select.increment()
+    }
+    stream
+  }
+}
+
 /** Combine a stream and a flow to a new stream. If both input sources fire, the flow will be preferred. */
 object StreamFlowArbiter {
   def apply[T <: Data](inputStream: Stream[T], inputFlow: Flow[T]): Flow[T] = {
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -188,6 +188,34 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
     this.ready := ret.ready && counter.willOverflowIfInc
     (ret, counter)
   }
+  
+  /**
+   * Connect this to a new stream whose payload is n times as wide, but that only fires every n cycles.
+   * It introduces one cycle of latency.
+   */
+  def slowdown(factor: Int): Stream[Vec[T]] = {
+    val next = Stream(Vec(this.payload, factor))
+    val counter = Counter(factor)
+    for (i <- 0 to factor) {
+    	next.payload(i) := RegNextWhen(this.payload, counter === i)
+    }
+    when (counter.willOverflow) {
+      /* All elements are valid, so wait for ready signal to advance */
+      this.ready := next.ready
+      next.valid := True
+      when (next.fire) {
+        counter.increment()
+      }
+    } otherwise {
+      /* Take the elements and save them into a register */
+      this.ready := True
+      next.valid := False
+      when (this.valid) {
+        counter.increment()
+      }
+    }
+    next
+  }
 
 /** Return True when a transaction is present on the bus but the ready signal is low
     */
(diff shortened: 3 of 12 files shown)


project/Version.scala
```diff
@@ -1,5 +1,5 @@
 object SpinalVersion {
-  val compiler = "2.11.12"
+  val compiler = "2.12.0"
```
Dolu1990: Basicaly, from 2.11.x to 2.12.x there is things which break. There is specific SpinalHDL   branch for 2.12.x that i update before each release.
Can you revert that change ?

build.sbt
```diff
@@ -148,7 +148,7 @@ lazy val tester = (project in file("tester"))
     version := SpinalVersion.tester,
     baseDirectory in (Test) := file("./"),
 
-    libraryDependencies += "org.scalatest" % "scalatest_2.11" % "2.2.1",
+    libraryDependencies += "org.scalatest" % "scalatest_2.12" % "3.1.0",
```
Dolu1990: Can't update to 2.12.x, see the review comment about the SpinalVersion.scala file :)
More generaly, i try to avoid updating depedancies if that's not required, as some people use SpinalHDL ofline +  there is not so much value updating something which work

Can you revert those changes ?

README.md
```diff
@@ -38,11 +41,25 @@ libraryDependencies ++= Seq(
...
-## JAR
+### Gradle
+
+```kotlin
+repositories {
+	mavenCentral()
+}
+
+dependencies {
+	compile group: 'com.github.spinalhdl', name: 'spinalhdl-core_2.12', version: '1.3.6'
```
Dolu1990: revert 2.12.x related changes

project/plugin.sbt
```diff
@@ -5,3 +5,5 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0")
 addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.1")
 
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.10")
+
+addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "5.2.4")
```
Dolu1990: I remember that for some time, the sbteclipse-plugin was breaking things. that's why i removed it.
Isn't the scala eclipse support dead / not realy active ?

Dolu1990:
Hi,
Thanks for the pull request :)
So basicaly all seem ok, just the 2.11.x -> 2.12.x isn't possible that way.
There is a SpinalHDL branch specificaly for 2.12.x that i update before each release.

project/plugin.sbt
piegamesde: I'm currently developing Spinal in Eclipse. It's not great, but it works and I didn't encounter anything breaking.

Dolu1990: I had bad experiances with eclipse scala plugin.
Did you tried the intelij comunity edition with the scala plugin ?

piegamesde: @Dolu1990 No, I don't have IntelliJ installed

piegamesde:
@Dolu1990 I left a few TODO comments in the code. Can you please have a look at them?

project/plugin.sbt
Dolu1990: If you have time, try intellij, hard to go back to eclipse after it :)
(triple shift magic)

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -549,9 +581,29 @@ class StreamFork[T <: Data](dataType: HardType[T], portCount: Int) extends Compo
   }
 }
 
+/**
+ * A simpler version of StreamFork. All output streams will always fire together,
+ *  which means that the stream will halt until all output streams are ready. In contrast,
+ *  the normal StreamFork allows that output streams may be ready one at a time, at the cost
+ *  of an additional flip flop (1 bit per output).
+ */
+// TODO this can be integrated into the normal StreamFork by adding a boolean parameter. But I don't fully understand StreamFork. ~piegames
```
Dolu1990: When there is function aguement conflicts in scala Object, i generaly just switch "apply" to a more expressive name, in the given case, could be  :

```scala
object StreamFork{
...
   def vec[T <: Data](source: Stream[T], count: Int): Vec[Stream[T]] 
} 

```

Dolu1990: About StreamFork, it take a stream, and each transaction is replicated on each output streams

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -906,13 +956,19 @@ class StreamCCByToggle[T <: Data](dataType: T, inputClock: ClockDomain, outputCl
   }
 }
 
-object StreamDispatcherSequencial{
+/**
+ * This is equivalent to a StreamDemux, but with a counter attached to the port selector.
+ */
+// TODO is there any reason to have a proper class for this if we simply could reuse StreamDemux instead? ~piegames
```
Dolu1990: That's likely right, could hav eused StreamDemux with some proper logic around it :)

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -906,13 +956,19 @@ class StreamCCByToggle[T <: Data](dataType: T, inputClock: ClockDomain, outputCl
...
+// TODO If there is a StreamDispatcherSequential, there should be a StreamCombinerSequential as well (name open to suggestions) ~piegames
```
Dolu1990: StreamCombinerSequential basicaly already exists as a StreamArbiter with the StreamArbiter.sequentialOrder arbitration :)

lib/src/main/scala/spinal/lib/Stream.scala
piegamesde: Can I change this, then?

lib/src/main/scala/spinal/lib/Stream.scala
piegamesde: I think I completely lost track about which stream modifications exist, what they do and how to name them. All my attempts to order these in some way and make it consistent failed because stream operations are heavily asymmetric (e.g. most operations have an "inverse", but it is not really the *inverse* because it does a few things differently).

lib/src/main/scala/spinal/lib/Stream.scala
Dolu1990: Sure ^^

lib/src/main/scala/spinal/lib/Stream.scala
Dolu1990: Just be care full to not break compatibility with previous releases, also things do not realy need to be 100 % perfect :)

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -494,10 +494,18 @@ class ComponentEmitterVerilog(
                 case `ERROR` => "ERROR"
                 case `FAILURE` => "FAILURE"
               }
-              b ++= s"${tab}if(!$cond) begin\n"
-              b ++= s"""${tab}  $$display("$severity $frontString"$backString);\n"""
-              if (assertStatement.severity == `FAILURE`) b ++= tab + "  $finish;\n"
-              b ++= s"${tab}end\n"
+              b ++= s"${tab}`ifdef FORMAL\n"
```
Dolu1990: Why not using similar things as https://github.com/SpinalHDL/SpinalHDL/blob/fbc7dc94dc14b0fb367d1f326fba14016f0f50c2/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala#L516 ?

piegamesde: Because if I understand it correctly, those formal statements are not part of Verilog, but are a Yosys-specific extension. I can still add the `if` as below if you prefer, but the generated code *must* contain an `ifdef FORMAL` or otherwise it won't be compliant.

piegamesde:
Two thoughts on the consistency of the API:

- `StreamArbiter` is like a `StreamMux` on steorids. This means that we could include the functionality of `StreamMux` and remove that special object (Or even rename `StreamArbiter` into `StreamMux`). The only thing I worry about is that this may result in some additional, but redundant logic that doesn't get optimized away.
- We have a `StreamDemux`, but no enhanced version of it. There is enough symmetry to build a `StreamArbiter` equivalent for demultiplexing, but I don't know if there is actual need for it and if it's worth building nevertheless.

piegamesde:
@Dolu1990 I made some changes and now the test fail. How do I get a trace to what exactly failed? The travis log isn't really helpful, and running `sbt test` throws errors like

```
../common/Makefile.sim:12: /makefiles/Makefile.sim: No such file or directory
make: *** No rule to make target '/makefiles/Makefile.sim'.  Stop.
```

Dolu1990:
> StreamArbiter is like a StreamMux on steorids. This means that we could include the functionality of StreamMux and remove that special object (Or even rename StreamArbiter into StreamMux). The only thing I worry about is that this may result in some additional, but redundant logic that doesn't get optimized away.

Right, also, that's fine if there is two nearly similar implementation, those are little things, no need to have it all merged in one.

> We have a StreamDemux, but no enhanced version of it. There is enough symmetry to build a StreamArbiter equivalent for demultiplexing, but I don't know if there is actual need for it and if it's worth building nevertheless.

That's another aspect of the dev, i try to avoid adding things which have no use/will not be used, as it endup into weakly tested dead code

Dolu1990:
About the travis issue, it seem it is because of the verilator repo, they renamed tags or something like that, i pushed a fix, it should be fine : https://github.com/SpinalHDL/SpinalHDL/commit/0d38c2b24b32561fb28571550a4178f690923de3

piegamesde:
Hm, your fix didn't work for me. I still get that error and a lot more test fail than on Travis:

```
[…]
../common/Makefile.sim:12: /makefiles/Makefile.sim: No such file or directory
make: *** No rule to make target '/makefiles/Makefile.sim'.  Stop.
[info] - cocotbVHDL *** FAILED ***
[info]   SpinalTesterCocotbBase.this.cocotbMustPass was true, and pass was false Simulation fail (SpinalTesterCocotbBase.scala:78)




../common/Makefile.sim:12: /makefiles/Makefile.sim: No such file or directory
make: *** No rule to make target '/makefiles/Makefile.sim'.  Stop.
[info] - cocotbVerilog *** FAILED ***
[info]   SpinalTesterCocotbBase.this.cocotbMustPass was true, and pass was false Simulation fail (SpinalTesterCocotbBase.scala:78)
[info] Run completed in 9 minutes, 27 seconds.
[info] Total number of tests run: 274
[info] Suites: completed 75, aborted 0
[info] Tests: succeeded 200, failed 74, canceled 0, ignored 0, pending 0
[info] *** 74 TESTS FAILED ***
[error] Failed tests:
[error]         spinal.tester.scalatest.RomTesterCocotbBoot2
[error]         spinal.tester.scalatest.Axi4SharedSdramCtrlTesterCocotbBoot
[error]         spinal.tester.scalatest.StreamTesterCocotbBoot
[error]         spinal.tester.scalatest.Axi4CrossbarTesterCocotbBoot
[error]         spinal.tester.scalatest.PllAAssertSDeassertTesterBoot
[…]
```

Dolu1990:
I will check out localy on my computer. Just right now this week i'm realy underwater XD Sorry it will take a bit of time

Dolu1990:
Ahh ? now it pass the travis stuff ?

piegamesde:
Yes it builds again now. The commits where I replaced uses of the `StreamCombinerSequential` class were wrong in some way.

piegamesde:
I currently have function argument conflicts in `StreamJoin`. Which of the methods should I rename, and to what? I'm not sure of how to call this.

piegamesde:
@Dolu1990 The pull request is now done (feature complete) from my perspective. If Travis runs through and you don't request any further changes, this may be merged.

Dolu1990:
Thanks, look good for me :)

Dolu1990:
@piegamesde Ahhh quite a issue here :(

Basicaly, the $finish statement that now appear because assert are set to be failure interract badely with yosys : 

```
  always @ (posedge clkout0) begin
    `ifdef FORMAL
      assert((! ((! io_writeDatas_0_valid) && io_writeDatas_0_ready)))
    `else
      if(!(! ((! io_writeDatas_0_valid) && io_writeDatas_0_ready))) begin
        $display("FAILURE SDRAM write data stream starved !");
        $finish;
      end
    `endif
    if(rspPipeline_output_valid)begin
      rspPipeline_debugData <= rspPipeline_output_payload_fragment_data;
    end
    rspPop_payload_last <= rspPipeline_output_payload_last;
    rspPop_payload_fragment_data <= rspPipeline_output_payload_fragment_data;
    rspPop_payload_fragment_context <= rspPipeline_output_payload_fragment_context;
  end
``` 


Yosys =>  

../../netlist/Ulx3sLinuxUboot.v:3010: ERROR: System task `$finish' outside initial block is unsupported.

I'm not a verilog guru. but is there a clean way to have a statement to exit a sim but not interract with the synthesis ?

piegamesde:
@Dolu1990 The only way I see to resolve this is to put the `$finish` in another `ifdef` block. I think both yosys and most simulators set some define variables when they process the code. The tricky part is more to generalize this to work with different simulators/synthesizers

Dolu1990:
@piegamesde Do you know if other synthesis tool are picky about the $finish ? Or if that's a yosys thing ?

piegamesde:
@Dolu1990 No, in fact, I don't even know other synthesis tools.

Dolu1990:
@piegamesde 
I openned an issue, it seem to be a yosys specific behaviour  :
https://github.com/YosysHQ/yosys/issues/1688

piegamesde:
@Dolu1990 On another note, it seems like I fucked up `Stream.slowdown`. There's a fix on my latest `dev` branch, but I can't really test it at the moment.

Dolu1990:
@piegamesde Ahhh ok, let's me know when it can be merged, there was also a little issue with the streamMux wrapper : 
https://github.com/SpinalHDL/SpinalHDL/commit/384858e3307dedb0f48ec1b4c4d21a93a69668b3

piegamesde:
@Dolu1990 I ran into another issue with `StreamJoin`.

[This](https://github.com/piegamesde/SpinalHDL/blob/0f14fcc31e1099ab7fb106009a605d0d6e7be21a/lib/src/main/scala/spinal/lib/Stream.scala#L793) line does not work, and I can't figure out why and how to fix it. Depending on the exact nature of the issue, `Stream::transmuteWith` might be broken as well.

Dolu1990:
@piegamesde what do you mean exactly by not working ? 

I can see one thing : 
```scala
   val  payloads = List(a,b,c)
   val raw = payloads.reduce(_ ## _) 
   // => abc where a is at the MSB and c at the LSB, which isn't what you want right ?
```

piegamesde:
I have the following helper structure:

```scala
case class Tuple2Bundle[L <: Data, R <: Data](val payloadTypeL: HardType[L], val payloadTypeR: HardType[R]) extends Bundle {
  val _1 = payloadTypeL()
  val _2 = payloadTypeR()
}
```

When I do

```scala
val transaction = new Tuple2Bundle(new Axi4Aw(slaveConf), new Axi4W(slaveConf))
transaction._1 := addressBuffer.payload
transaction._2 := dataBuffer.payload
```

everything works fine. But when I do

```scala
val transaction = new Tuple2Bundle(new Axi4Aw(slaveConf), new Axi4W(slaveConf))
transaction.assignFromBits(addressBuffer.payload ## dataBuffer.payload)
```

it doesn't work anymore. I'm sorry but I haven't further reduced the example yet.

Dolu1990:
@piegamesde That's because the concatenation doesn't work that way : 

addressBuffer.payload ## dataBuffer.payload => 
lsb : dataBuffer
msb : addressBuffer
But as defined in Tuple2Bundle, the lsb should be addressBuffer not dataBuffer

Do you see what i mean ?

piegamesde:
Thank you, I somehow thought elements in a Bundle where MSB to LSB and not the opposite. This leaves me with the issue that I have to join the streams in the "opposite" order that I want them to appear in the Tuple. This leaves me with the following possibilites to fix:

- Accept this unintuitive behaviour in my case and deal with it.
- Simply concatenate the elements of the sequence in the reverse order. But this may break other people's intuition when joining streams that are not aligned with the new bundle type like I do.
- Add the `Tuple2Bundle` to the stdlib and replace the varargs/`Seq` method with some finite arguments ones.

Issue #256, opened 2020-01-13
[Feature] Pretty print of module instantiations


lsteveol:
Hi @Dolu1990 , I have made a local change to the emitSubComponents method so that the module instantiations are printed with aligned spacing. I'm honestly not familar with GitHub outside of cloning projects so I'm trying to figure out how to do a proper pull-request. In meantime, I will post the update I did hear and you can update if desired.

in ComponentEmitterVerilog.scala line 241:
```scala
//Fixing the spacing
      var maxNameLength = 0
      for(data <- child.getOrdredNodeIo) {
        if (emitReferenceNoOverrides(data).toString.length() > maxNameLength) maxNameLength = emitReferenceNoOverrides(data).toString.length()
      }
      var maxNameLengthCon = 0
      for(data <- child.getOrdredNodeIo) {
        val logic = if(openSubIo.contains(data)) "" else emitReference(data, false)
        if (logic.toString.length() > maxNameLengthCon) maxNameLengthCon = logic.toString.length()
      }

      logics ++= s"${child.getName()} ( \n"
      for (data <- child.getOrdredNodeIo) {
        val logic = if(openSubIo.contains(data)) "" else emitReference(data, false)
        //logics ++= s"    .${emitReferenceNoOverrides(data)}($logic),\n"
        logics ++= s"    .%-${maxNameLength}s ( %-${maxNameLengthCon}s ),\n".format(emitReferenceNoOverrides(data), logic) 
      }
      logics.setCharAt(logics.size - 2, ' ')

      logics ++= s"  );"
      logics ++= s"\n"
```

this will cause the module instances to be printed as such:
```verilog
module MyMod2 (
      input  [7:0] io_a,
      input  [7:0] io_b,
      output [7:0] io_c,
      input   tb_somein,
      output  tb_someout,
      inout  tb_myinout,
      input  [2:0] myin);
  wire [7:0] mod_1__c;
  wire  mod_1__tb_someout;
  MyMod mod_1_ ( 
    .a          ( io_a              ),
    .b          ( io_b              ),
    .c          ( mod_1__c          ),
    .tb_somein  ( tb_somein         ),
    .tb_someout ( mod_1__tb_someout ),
    .tb_myinout ( tb_myinout        ),
    .myin       ( myin              ) 
  );
  assign io_c = (io_a + mod_1__c);
  assign tb_someout = mod_1__tb_someout;
  assign tb_myinout = tb_myinout;
endmodule
```

[...]

PR #257, opened 2020-01-13
Update ComponentEmitterVerilog.scala

lsteveol:
Hi @Dolu1990 ,

Here is the pull-request I mentioned in https://github.com/SpinalHDL/SpinalHDL/issues/256

Pretty print for module instantiations.

Let me know if I did something incorrect WRT to the pull-request

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -238,10 +238,22 @@ class ComponentEmitterVerilog(
         }
       }
 
+      //Fixing the spacing
+      var maxNameLength = 0
+      for(data <- child.getOrdredNodeIo) {
+        if (emitReferenceNoOverrides(data).toString.length() > maxNameLength) maxNameLength = emitReferenceNoOverrides(data).toString.length()
+      }
+      var maxNameLengthCon = 0
+      for(data <- child.getOrdredNodeIo) {
+        val logic = if(openSubIo.contains(data)) "" else emitReference(data, false)
+        if (logic.toString.length() > maxNameLengthCon) maxNameLengthCon = logic.toString.length()
+      }
+
       logics ++= s"${child.getName()} ( \n"
       for (data <- child.getOrdredNodeIo) {
         val logic = if(openSubIo.contains(data)) "" else emitReference(data, false)
-        logics ++= s"    .${emitReferenceNoOverrides(data)}($logic),\n"
+        //logics ++= s"    .${emitReferenceNoOverrides(data)}($logic),\n"
+        logics ++= s"    .%-${maxNameLength}s ( %-${maxNameLengthCon}s ),\n".format(emitReferenceNoOverrides(data), logic) 
       }
       logics.setCharAt(logics.size - 2, ' ')
 


jijingg:
i like this.  woud you prefer to add  i/o infos at the tail of line. sometimes it's usefully for debuging 
```
  MyMod mod_1_ ( 
    .a          ( io_a              ), //i
    .b          ( io_b              ), //i
    .c          ( mod_1__c          ), //o
    .tb_somein  ( tb_somein         ), //i
    .tb_someout ( mod_1__tb_someout ), //o
    .tb_myinout ( tb_myinout        ), //io
    .myin       ( myin              )  //i
  );
  assign io_c = (io_a + mod_1__c);
```

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -238,10 +238,22 @@ class ComponentEmitterVerilog(
         }
       }
 
+      //Fixing the spacing
+      var maxNameLength = 0
+      for(data <- child.getOrdredNodeIo) {
+        if (emitReferenceNoOverrides(data).toString.length() > maxNameLength) maxNameLength = emitReferenceNoOverrides(data).toString.length()
```
jijingg: It is recommended to use FP style instead of mutable `var` for short and Stable.
``
val maxNameLength = child.getOrdredNodeIo.map(data => emitReferenceNoOverrides(data).toString.length()).max
``

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -238,10 +238,22 @@ class ComponentEmitterVerilog(
...
+      }
+      var maxNameLengthCon = 0
+      for(data <- child.getOrdredNodeIo) {
```
jijingg:
```
val maxNameLengthCon = child.getOrderedNodeIo
  .filter(!openSubIo.contains(_))
  .map(emitReference(data.false).toString.length())
  .max
```

lsteveol:
Hi @jijingg ,

For the i/o comments, I'm fine with having them or not. I have a script that I wrote several years back that generates a Verilog instance, formatted as I have done in this request and at the end it has
```
), //<input/output/inout> <Width> <type>

type would indicate something like a SV interface, real, etc.
```
This may be a little more verbose than we want. I normally remove these after instantiation, but since this is automatically generated, I don't see a problem with it.

And you have seen I'm not that familar with Scala :). Yes, that does look like a cleaner implementation.

When time allows I would like to also clean up the port declarations, wire declarations, and the assign statements with similar formatting. I know it's auto-generated code, but I do like to debug through the Verilog since I do a bunch of blackboxing.

Thanks for the feedback.

Dolu1990:
> This may be a little more verbose than we want. I normally remove these after instantiation, but since this is automatically generated, I don't see a problem with it.

From the moment the netlist size do not take a big hit, i'm fine ^^

Let's me know when you feel it is ready for review / merge :)

lsteveol:
We can merge this whenever is a good time.

Dolu1990:
Thanks ^^

lsteveol:
Hey @Dolu1990 ,

Just wondering, how long does it take before the patches are updated in the sbt repo?

Dolu1990:
@lsteveol when i do release, it realy depend, i generaly do them when there is a good accumulation, or a big broken things which was fixed.
Additionaly, what you can do, is https://spinalhdl.github.io/SpinalDoc-RTD/SpinalHDL/About%20SpinalHDL/faq.html#how-to-use-an-unreleased-version-of-spinalhdl-but-commited-on-git
Then you can use the unreleased version exactly as if it was released on your computer.

lsteveol:
Ah ok thanks! Somehow I missed this in the docs.

PR #258, opened 2020-01-15
AsyncMemoryBusFactory fix for writeMemWordAligned

reostat:
One line was missing thus preventing BusSlaveFactory.readSyncMemWordAligned() from working


Dolu1990:
Thanks ^^

Issue #255, opened 2020-01-13
Multi-word driveFlow  stays valid forever


distributed:
The `valid` behavior of `BusSlaveFactory.driveFlow` is different for flows that fit within the bus data width and those that don't.

For the former, `valid` is asserted for one clock cycle, with a combinatorial signal.

For larger flows, `valid` is asserted in the next clock cycle and never deasserted.

There is a `regValid` in `driveFlow`:
```
val regValid = Reg(that.valid) init(False)
```

which is set right below

```
onWrite(address + ((wordCount - 1) * wordAddressInc)){ regValid := True }
```

but never reset.

Maybe I misunderstood something, but I believe that `regValid` should stay asserted for only one clock cycle.

Dolu1990:
Hooo right, my bad, clearly a thing to fix

distributed:
No worries. I am currently not using it, but I stumbled across it while reading the code :)

PR #259, opened 2020-01-16
fix drive flow and some surrounding checks for proper recognition of bitOffset #255

typingArtist:


lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
```diff
@@ -489,7 +489,7 @@ trait BusSlaveFactory extends Area{
     */
   def readStreamNonBlocking[T <: Data](that: Stream[T], address: BigInt): Unit = {
 
-    val wordCount = (widthOf(that.payload) - 1 ) / busDataWidth + 1
+    val wordCount = (1 + widthOf(that.payload) - 1 ) / busDataWidth + 1
```
Dolu1990: seem wrong to me. Ex : 
32 bits bus, mapped with a 32 bits word => (1+32-1)/32+1  = 2 words isn't it ?

What's about 

```
((widthOf(that.payload) - 1 ) / busDataWidth).max(0) + 1
```

 ?

typingArtist: But isn’t that what readMultiWord in Line 500/502 will do and require, prepend/append that.valid to that.payload? This adds 1 bit that is not accounted for. Or am I wrong about line 500/502?

typingArtist: Speaking of that … I don’t like that protocol very much. First, it not just introduces another bit of valid information, what shifts the data around and likely requires one more word of address range, as you point out. It also introduces a race if the master thinks that it can read `that.valid` and implicitly write `that.ready` in one go with a full width read operation, but due to the fact that the slave sits behind a width adaption bridge, the two are spaced by, say, one clock cycle, so `that.valid` may return False, but `that.ready` is set to True in the next cycle when actual data may have arrived. So the only way to avoid that is to always read `that.valid` in the smallest possible entitiy, i.e. 8 bits, and then repeating the full width read when True is returned for `that.valid`.

Dolu1990: onWrite(address + ((wordCount - 1) * wordAddressInc)){ regValid := True }

When you write the last word, the valid will fire automaticaly

typingArtist:
@Dolu1990 is there anything missing?

PR #261, opened 2020-01-21
fixPoint improvements

jijingg:
hi @Dolu1990 
i did some improvements on fixPoint， and the simulation regression is done.
**1. fixTo support QFormat**
```scala 
A.fixTo(SQ(10,4))      //need A.tag(SQ(16,8)) first， auto Q tag will add next version
A.fixTo(13 downto 4)
```
and QFormat move spinal.lib to spinal.core

**2.fixPointConfig added**
~~**2. add default fixPointConfig to SpinalConfig**~~ 
```scala
SpinalConfig(Verilog, 
 // fixPointConfig = FixPointConfig(CEIL,true))  //removed

A.fixTo( 13 downto 4)  //equal A.fixTo( 13 downto 4, CEIL, true)
```

**3. wrapper fixTo with Component as defualt.  it's good for readability and easy debugging,** 
``` verilog
module SInt16fixTo13_4_CEIL_sym (
      input  [15:0] din,
      output [9:0] dout);  
......
endmodule 

module MyTop(
....
  SInt16fixTo13_4_CEIL_sym fixTo$1 ( 
    .din(din),
    .dout(fixTo$1_dout) 
  );
...
endmodule    
```
if dont need wrap , just set `SpinalConfig(fixToWithWrap = false) `
**4. dsptool FixSwitch rework with singleton, which used in fixpoint simulation**
```scala
FixOn()
...
FixOff()   //like matlab grid on, grid off 
```
other dsptool feature Compelx , IQ bundle in WIP
 
**5.  some cornes bug fixed** 
- fix Double precision trap on` floor/ceil/==`, instead safe [epsFloor/epsCeil](https://github.com/SpinalHDL/SpinalHDL/pull/261/commits/143550bb3a22453897ee0046b21ecfed53a09dba)
- fixed FixData(-0.01, SQ(8,0)).hex = 00 not 100 
- symmetric erro  on dsptoo.FixData unsigned Type


Dolu1990:
Hi,

Sorry for the delay.

````scala
A.fixTo(13 downto 4)
```` 

This one will select the bit 13 downto 4 from the fixed point ? or it will take the bits from 2^13 downto 2^4 ?

```scala
SpinalConfig(Verilog, 
  fixPointConfig = FixPointConfig(CEIL,true))
```` 

That seem quite dangerous to me when it will come to mix different peoples work.
What's about having something a bit like clockdomain ? where you can say that all the enclosed logic is using the given default fixed point config ?

> SInt16fixTo13_4_CEIL_sym 

There is an alternative way of propagating names and improve debugability. setCompositeName
Did you knew about it ? 

Ex : 
https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/Stream.scala#L245

Basicaly, signal tagged with it, would constructe their name from the name of something else + an post fix.

That would avoid having tonnes and tonnes of component just for little translations.

**Comment by jijingg** [comment]
> This one will select the bit 13 downto 4 from the fixed point ?

yes 13 downto 4,  not 2^13 downto 2^4 

> That seem quite dangerous to me when it will come to mix different peoples work.
What's about having something a bit like clockdomain ? where you can say that all the enclosed logic is using the given default fixed point config ?

good point, I thought about it, too. the implement is very similar to the clockdomain, most of time one project with same fixConfig
image
If a project really does have mix different peoples work,  As you said like ClockingArea
```scala 
ClockingArea(myClockDomain) {
...
}

FixPointConfig(FLOOR, false) {
...
}
```
image

i know setCompositeName, this will tagName on signals . But is it better to wrap all logic under component.
for example 
```scala
class T2 extends Component{
    val a = in SInt(16 bits) 
    val b = a.fixTo(5 downto 0)
    val c = a.fixTo(10 downto 3)
    val d = a.fixTo(10 downto 2)
}
```
if we don't wrap fixTo with Hierarchy(Component), see [fixToWithoutWrap.v](https://gist.github.com/jijingg/a594e63e80b38dd0fb99d91a23122645#file-fixtowithoutwrap-v-L6)
you can found Fixed point logic will be scattered in the whole module, which is very messy.

then we wrap fixTo with Hierarchy, see [fixToWithWrap.v](https://gist.github.com/jijingg/862ca17adbd54ac3434e1a2a179e19a8#file-fixtowithwrap-v-L107), you can found the top module is very clear,and less code lines.

> That would avoid having tonnes and tonnes of component just for little translations.

- first,  i think the fixTo process in Project  is very  limited, not too much. even though , component is clean than messy logic .
- second, i think hierarchy(Component) is good thing almost no extra cost, it's friendly for IC developer and traditional EDA-Flow，when wave debug or ECO. 

 
so i think for some complex and common functions, we should encourage the use of hierarchy wrapper. 
What about your opinion? @Dolu1990

core/src/main/scala/spinal/core/core.scala
```diff
@@ -76,6 +76,8 @@ package object core extends BaseTypeFactory with BaseTypeCast {
 
     def downto(start: Int): Range.Inclusive = Range.inclusive(i, start, -1)
 
+    def ~(start: Int): Range.Inclusive = this.downto(start)
```
jijingg: do you mind add `~` for short of `downto`
```scala
a(8 downto 4)
a(8 ~ 4) 
```

Dolu1990: Issue with ~ is that it is already a integer operator (bitwise not), so that's quite confusing

Dolu1990:
```scala
ClockingArea(myClockDomain) {
...
}

FixPointConfig(FLOOR, false) {
...
}
```

Maybe we should find a general purpose way of handeling such things.
As 

```scala
val miaou = new Area with ClockedArea(x) with FixedPointConfig(y){

}
```

The issue with that way is it isn't so much programable (you can't cancel a scala "with")

An alternative could be 

```scala
val miaou = x(y(newArea{

}))
```

But that's not so pretty 

Else 

```scala
val miaou = new Area{
  set(x)  // x is a clock domain
  set(y)  // y is a FixedPointConfig
  ....
}
```

core/src/main/scala/spinal/core/core.scala
jijingg: ` ~4`(bitwise not)  is a UnaryOperator, `8 ~ 4` is a BinaryOperator.  not perfect， But there seems to be no better choice in `~!@#$%^&*` :joy:
for some scenarios `downto` is too long 
```scala
ret := B(0,6 bits) ## ab(2 downto 0) ## ab(5 downto 3) ## ab(8 downto 6) ## ab(11 downto 9)} 
```
```scala
ret := B(0,6 bits) ## ab(2 ~ 0) ## ab(5 ~ 3) ## ab(8 ~ 6) ## ab(11 ~ 9)
```

jijingg:
a little different to ClockArea， the usage of fixPointConfig is not so often. it more closer to `ClockDomain(RISING,ASYNC)`，  most of time the global `SpinalConfig(fixPointConfig)` is enough, we do not need always change `FixpointConfig`, 

For a team or a company, the default fixed-point method is fixed, There's not much difference in performance，It's just a habit or a choice，(our team use FixPointConfig(ROUNDUP, true) as default).

core/src/main/scala/spinal/core/core.scala
typingArtist: I am also not in favor of adding this shorthand. You can and most probably should split lines once they get too long or just for readability, and `~` won’t change that problem.

Dolu1990:
I realy have to provide that feature : 

```scala
val miaou = new Area{
  set(x)  // x is a clock domain
  set(y)  // y is a FixedPointConfig
  ....
}
```

It could probably be usefull in other cases too. I'm a bit scared of default configs in the SpinalConfig.

Question, isn't CEIL more costly in ressources than FLOOR ?

jijingg:
> I realy have to provide that feature 

good, it's ok
but what you scared of default configs in SpinalConfig?
default configs in SpinalConfig have **lowest Priorty**, it won't do anything bad. When no fixpointconfig is set, It's the last one to take effect

image


you can found the explicit `fixTo(8 downto 5, FLOOR, true)` have highest priority, no matter what fixpointconfig you set up.

> Question, isn't CEIL more costly in ressources than FLOOR ?

- FLOOR have no cost, just discard lower n bit
- CEIL have lower cost  
- ROUNDUP with lower cost and near algorithm performance as ROUNDTOINF
- ROUNDTOINF most cost, and most common, good algorithm  performance
- ROUNDTOENVEN have most cost and complex ,best algorithm performance, but not common

our team favor ROUNDUP, it have lower cost , good timing ,no bad algorithm performance ,  it's highly recommended. 
But I'm not sure whether to set ROUNDUP as the SpinalHDL default , because ROUNDTOINF is more common, What's your opinion? @Dolu1990

core/src/main/scala/spinal/core/core.scala
jijingg: ok, i will remove that

Dolu1990:
Sorry, i forgot about it, i have to provide a scope property feature, ping me when you see no moves from my side ^^ I'm often underwater and forget about things

jijingg:
The new features and improvement  should repeat experimenting on my personal IC project,  what needs to be verified is not only the function, but also the usage logical.  so don't worry，when it turns out to be feasible and stable,  i will pin you harder if need review or merge, ^_^

jijingg:
hi @Dolu1990 I want this PR be merged in 1.4.0 as  a phased improvment, because some fix is  necessary,   
 other improvements and new feature(Complex Data implement) will continue to be updated in the new PR.

Dolu1990:
Ahhh f***, i just see your message now :/
So, do we need a 1.4.1 ?

jijingg:
Hah, I missed the last window. see next version  :)

Dolu1990:
@jijingg I think i have what you need : 

```scala
object PlayScopeProperty extends App{
  object FixedPointProperty extends ScopeProperty[Int]{
    override def default: Int = 42
  }

  class Toplevel extends Component{
    println(FixedPointProperty.get)
    val logic = FixedPointProperty(666) on new Area{
      println(FixedPointProperty.get)
      FixedPointProperty(1234){
        println(FixedPointProperty.get)
      }
      println(FixedPointProperty.get)
    }
    println(FixedPointProperty.get)
  }

  val config = SpinalConfig()
  config.generateVerilog(new Toplevel)
//  42
//  666
//  1234
//  666
//  42
  config.setScopeProperty(FixedPointProperty, 76)
  config.generateVerilog(new Toplevel)
//  76
//  666
//  1234
//  666
//  76
}
``` 

Now i just have to make it working with the "component.rework" API.

Dolu1990:
@jijingg See https://github.com/SpinalHDL/SpinalHDL/blob/514d74900efa8f2ed204d7fdabff3598c29640d3/tester/src/main/scala/spinal/tester/PlayDev.scala#L1497 for the current API, let's me know if you have some idea about improvements

jijingg:
Sorry for miss the message .
Yes, exactly what I want，the scopeProperty way are elegant.  Very good .  
I will update the FixpointConfig Api base on scopeProperty  soon.

```scala 
val area0 = FixedPointProperty(FLOOR, sym = false) on new Area{
....
}
val LowCostFixConfig =  FixedPointProperty(ROUNDTOUP, sym = true) 
val dut = LowCostFixConfig on new Top
```
but i m not sure dose ScopeProperty can set like this ?
```scala 
config.setScopeProperty(FixedPointProperty(ROUNDTOINF, false))
```

jijingg:
hi @Dolu1990 there have a little problem .
i found th ScopeProperty only avliable on Component Scope , Because all them in `GlobalData`
```
def stack = GlobalData.get.scopeProperties.getOrElseUpdate(this.asInstanceOf[ScopeProperty[Any]],new Stack[Any]
```
in my case the FixedPointConfig  not only for HDL ,
```scala
val Pi = in UInt(32 bits)
val Pi16bit = Pi.fixTo(31 downto 16, FLOOR, false)
val Pi12bit = Pi.fixTo(31 downto 20)// roundType, sym from scopeProperty
```
but also used in Algorithm reference platform, something like 
```scala
val pi = FixData(3.1415926, UQ(12,10), ROUNDTOINF, false).value  
val pi = FixData(3.1415926, UQ(12,10)).value  // roundType, sym from scopeProperty
```
so  is it neccessary to put scope Properties to GlobalData? if put theme to another GlobalData which   avliable for all Scala code have other problem ?

Dolu1990:
> config.setScopeProperty(FixedPointProperty(ROUNDTOINF, false))

Fixed, now you can do  :

```scala
  config.setScopeProperty(ComplexPropertyValue(54,65))
```


> i found th ScopeProperty only avliable on Component Scope , Because all them in GlobalData

There is some conflicts here, the issue is mainly that SpinalHDL sandbox the scope properties, that way, if it crash, nothing get corrupted, and everything is cleaned away.
So maybe we move the ScopeProperty away from the SpinalHDL core, but add a sandboxing API in it, so SpinalHDL can use it in a safe way ?

jijingg:
> So maybe we move the ScopeProperty away from the SpinalHDL core, but add a sandboxing API in it, so SpinalHDL can use it in a safe way ?

Yes , it's better

Dolu1990:
@jijingg Fixed, all should be fine now

Dolu1990:
But be carefull, the ScopeProperty are per thread.

jijingg:
@Dolu1990 there are still three question

**1. `push` `pop` should not opened for users？** 
```scala
MyScopeProperty.get
//0 
MyScopePropert(32) on {
   balabala 
  MyScopeProperty.get
// 32
  MyScopePropert.push(44)
}
MyScopeProperty.get
//we got 44 ,not 0 
```
i think it's dangerous and prone to abuse if opend for users , so i add `protected [core] def push`, only can be used in spianl-core, is it oK?
https://github.com/SpinalHDL/SpinalHDL/pull/261/commits/8b6e46773913c990b93a0b1ac3abdeb8e19d65a8#diff-c9166a5d4cebd4536f6b62ffb6394ab1R39

**2. ScoperProperty.default should be updateable？** 

in the old way ScoperProperty default value be fixed. 
You can imagine a scene， 
the spinal FixPointProperty.default is `FixPointConfig(ROUNDTOINF, false)`.
but in my project , i prefer `FixPointConfig(ROUNDUP, true)` as default, and i dount want wrapper 
all my project with 
```scala
FixPointScopeProperty(Mydefault){
  AlgPlateForm()....
  SpinalHDL generator...
  SpinalSim ....
}
```
 if i put `FixPointScopeProperty.setDefault(....)` at the begin  of code should more convenient
```scala
FixPointScopeProperty.setDefault(....)
AlgPlateForm()....
SpinalHDL generator...
SpinalSim ....
```
https://github.com/SpinalHDL/SpinalHDL/pull/261/commits/8b6e46773913c990b93a0b1ac3abdeb8e19d65a8#diff-c9166a5d4cebd4536f6b62ffb6394ab1R46-R48

**3. SpinalConfig.scopepropertys should be Wrapper on generator？** 
i found there only simple push ,https://github.com/SpinalHDL/SpinalHDL/pull/261/files#diff-73def809d8ea6bd3b203f4377fbc60a0L161 , 
 The lifecycle of  SpinalConfig.scopepropertys should only Influence the Spinal generator or simulation ， not other scala code , so SpinalConfig.scopepropertys  should `on` the scope not simple `push`,  right ?

Dolu1990:
> 1. push pop should not opened for users？

Right, shouldn't be used by users, we could imagine a way to make it usable by users, but is it a good thing ?

> 2. ScoperProperty.default should be updateable？

The issue with the proposal, is it modify the state of the thread without restoring it for other usages (ex :  Scala tests reusing the same thread to run sequancialy multiple tests)

To better understand the issue, what is AlgPlateForm() for ? parameters calculation ?

In general what i do, is that i define an alternative SpinalConfig with all my project specific needs, and then i reuse that SpinalConfig for everything in the project.
Would that be good ?

3. SpinalConfig.scopepropertys should be Wrapper on generator？

Sorry, i don't understand, do you have a practical case ?

jijingg:
> 1 To better understand the issue, what is AlgPlateForm() for ? parameters calculation ?

for Example , i develop an FFT module, I need to verify it. I have to have a reference model(algPlateForm). Reference model can be C code, MATLAB, or Python code.  also can be scala code ,now i  write the reference model in scala and compare it with the results dumped by HDL(SpinalHDL) at simulation. 

image     

> 2 SpinalConfig.scopepropertys should be Wrapper on generator？

```scala 
object PlayFixPointProperty2 extends App{
  import spinal.core.RoundType._
  def check(roundType: RoundType, sym: Boolean): Unit = {
    println(s"${FixPointProperty.get}, $roundType, $sym ")
  }

  class MyFFT extends Component {
    check(RoundType.ROUNDUP, true)
    //FixPointConfig(ROUNDTOINF,false), ROUNDTOINF, false   Ok!
    val indata = in SInt(16 bits)
    val outdata = out SInt(16 bits)
    outdata := indata
  }

  check(RoundType.ROUNDTOINF, false)
  //FixPointConfig(ROUNDTOINF,false), ROUNDTOINF, false   Ok!
  val config = SpinalConfig(targetDirectory = "./tmp")
  config.setScopeProperty(FixPointProperty, FixPointConfig(ROUNDUP, true))
  config.generateVerilog(new MyFFT)
  //config.setScopeProperty(FixPointProperty, FixPointConfig(ROUNDUP, true))  need  Pop Out !!!

  check(RoundType.ROUNDTOINF, true)
  //FixPointConfig(ROUNDUP,true), ROUNDTOINF, true   Wrong!
}
```

>3 we could imagine a way to make it usable by users, but is it a good thing ?
>ScoperProperty.default should be updateable

Maybe these two can be combined into one problem, What we need is to change the default value.

PR #267, opened 2020-02-06
changes for output assigns

### LINKED ISSUE #251: [Request]: Remove assigns for direct connections to parent using autoConnect
I wondering if something like this can be adjusted. Similar to https://github.com/SpinalHDL/SpinalHDL/issues/132. If I have a Bundle that I use up a hierarchy, outputs seem to get reassigned.

Here is an example
```scala
class MyBundle extends Bundle{
  val a = in  UInt(8 bits)
  val b = in  UInt(8 bits)
  val c = out UInt(8 bits)
}

class TestBundle extends Bundle{
  val somein  = in  Bool
  val someout = out Bool
}

class MyMod extends Component {
  val io = new MyBundle
  val tb = new TestBundle
  
  io.c := io.a + io.b
  
  tb.someout := False
  
  noIoPrefix()   
}

class MyMod2 extends Component {
  val io = new MyBundle
  val tb = new TestBundle
  
  val mod = new MyMod
  
  mod.io.a := io.a
  mod.io.b := io.b
  
  io.c := io.a + mod.io.c
  
  tb <> mod.tb
  
}
```

This will produce:
```verilog
module MyMod (
      input  [7:0] a,
      input  [7:0] b,
      output [7:0] c,
      input   tb_somein,
      output  tb_someout);
  assign c = (a + b);
  assign tb_someout = 1'b0;
endmodule

module MyMod2 (
      input  [7:0] io_a,
      input  [7:0] io_b,
      output [7:0] io_c,
      input   tb_somein,
      output  tb_someout);
  wire [7:0] mod_1__c;
  wire  mod_1__tb_someout;
  MyMod mod_1_ ( 
    .a(io_a),
    .b(io_b),
    .c(mod_1__c),
    .tb_somein(tb_somein),
    .tb_someout(mod_1__tb_someout) 
  );
  assign io_c = (io_a + mod_1__c);
  assign tb_someout = mod_1__tb_someout;
endmodule
```

It would be ideal if it could do this:
```verilog
module MyMod2 (
      input  [7:0] io_a,
      input  [7:0] io_b,
      output [7:0] io_c,
      input   tb_somein,
      output  tb_someout);
  wire [7:0] mod_1__c;
  MyMod mod_1_ ( 
    .a(io_a),
    .b(io_b),
    .c(mod_1__c),
    .tb_somein(tb_somein),
    .tb_someout(tb_someout ) 
  );
  assign io_c = (io_a + mod_1__c);
endmodule
```

I know this is a byproduct of VHDL, but is there a way to remove this intermediate variable? A reason I'm asking is that in ASIC flows you will often need to set a "dont_touch" on a particular signal. This keeps the synthesis tool from inserting a buffer or other logic. These tools can often handle an assign statement, but I know being able to only have 1 signal name is ideal. I've only seen this on outputs, and not on inputs.

I have dug through the SpinalHDL code some, but I'm pretty fresh with it and Scala in general, so I don't really have much to offer in terms of what I've tried. But I'm willing to try some things out if needed. I'm not sure how much effort this would be.

Thanks!

lsteveol:
This is a *hopeful* solution to https://github.com/SpinalHDL/SpinalHDL/issues/251 that I raised.

ComponentEmitterVerilog.scala
-----------------------------
- Added a check where if an output of a subcomponent was connect to an output of the current component it would set the referencesOverrides() to the component port. This removes the need for the additional assign statement.
- subcomponent outputs that are only used internally to the current component should still produce a wire as expected.
- Added formatting to the portMap in emitEntity. Before adding to the portMaps, it will figure up the spacing.

Spinal.scala
------------
- Added a config parameter called 'removeOutputAssigns'. Defaults to false, keeping the original
  functionality where outputs are set through assigns.

Here is an example of the input Spinal and corresponding output Verilogs
```scala
class MyBB extends BlackBox {
  val thisin  = in Bool
  val thisout = out Bool
}

class MyBottom extends MrvnComponent{
  val somein = in Bool
  val somein2 = in Bool
  val someout = out Bool
  val someout2 = out Bool
  
  someout  := somein & somein2
  someout2 := somein | somein2
  
  val bb = new MyBB
  bb.thisin := somein
}

class MyTop extends MrvnComponent{
  val in1 = in Bool
  val in2 = in Bool
  val in3 = in Bool
  
  val out1 = out Bool
  val out2 = out Bool
  val out3 = out Bool
  val out4 = out Bool

  val bot1 = new MyBottom
  val bot2 = new MyBottom
  
  bot1.somein   := in1
  bot1.somein2  := in2
  out1 := bot1.someout
  
  bot2.somein   := in2
  bot2.somein2  := in3
  out2 := bot2.someout
  
  out3 := bot1.someout2 ^ bot2.someout2
  
  out4 := bot1.someout2
}
```

**Normal/Default - removeOutputAssigns=false**
```verilog
module MyBottom (
      input    somein,
      input    somein2,
      output   someout,
      output   someout2);
  wire  bb_thisout;
  MyBB bb ( 
    .thisin  ( somein     ),
    .thisout ( bb_thisout ) 
  );
  assign someout = (somein && somein2);
  assign someout2 = (somein || somein2);
endmodule


//MyBottom_1_ remplaced by MyBottom

module MyTop (
      input    in1,
      input    in2,
      input    in3,
      output   out1,
      output   out2,
      output   out3,
      output   out4);
  wire  bot1_someout;
  wire  bot1_someout2;
  wire  bot2_someout;
  wire  bot2_someout2;
  MyBottom bot1 ( 
    .somein   ( in1           ),
    .somein2  ( in2           ),
    .someout  ( bot1_someout  ),
    .someout2 ( bot1_someout2 ) 
  );
  MyBottom bot2 ( 
    .somein   ( in2           ),
    .somein2  ( in3           ),
    .someout  ( bot2_someout  ),
    .someout2 ( bot2_someout2 ) 
  );
  assign out1 = bot1_someout;
  assign out2 = bot2_someout;
  assign out3 = (bot1_someout2 ^ bot2_someout2);
  assign out4 = bot1_someout2;
endmodule
```
**removeOutputAssigns=true**
```verilog
module MyBottom (
      input    somein,
      input    somein2,
      output   someout,
      output   someout2);
  wire  bb_thisout;
  MyBB bb ( 
    .thisin  ( somein     ),
    .thisout ( bb_thisout ) 
  );
  assign someout = (somein && somein2);
  assign someout2 = (somein || somein2);
endmodule


//MyBottom_1_ remplaced by MyBottom

module MyTop (
      input    in1,
      input    in2,
      input    in3,
      output   out1,
      output   out2,
      output   out3,
      output   out4);
  wire  bot2_someout2;
  MyBottom bot1 ( 
    .somein   ( in1  ),
    .somein2  ( in2  ),
    .someout  ( out1 ),
    .someout2 ( out4 ) 
  );
  MyBottom bot2 ( 
    .somein   ( in2           ),
    .somein2  ( in3           ),
    .someout  ( out2          ),
    .someout2 ( bot2_someout2 ) 
  );
  assign out3 = (out4 ^ bot2_someout2);
endmodule
```

Fixes #251

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -79,11 +79,25 @@ class ComponentEmitterVerilog(
   }
 
   def emitEntity(): Unit = {
+    var maxSpacing = 1
+    component.getOrdredNodeIo.foreach(baseType => if(emitType(baseType).toString.length() > maxSpacing) maxSpacing = emitType(baseType).toString.length())
+    
     component.getOrdredNodeIo.foreach(baseType =>
       if(outputsToBufferize.contains(baseType) || baseType.isInput)
-        portMaps += s"  ${emitSyntaxAttributes(baseType.instanceAttributes)}${emitDirection(baseType)} ${emitType(baseType)} ${baseType.getName()}${emitCommentAttributes(baseType.instanceAttributes)}"
+        portMaps += s"  %s%s %-${maxSpacing}s %s%s".format(emitSyntaxAttributes(baseType.instanceAttributes),
+                                              emitDirection(baseType),
+                                              emitType(baseType),
+                                              baseType.getName(),
+                                              emitCommentAttributes(baseType.instanceAttributes))
       else
-        portMaps += s"  ${emitSyntaxAttributes(baseType.instanceAttributes)}${emitDirection(baseType)} ${if(signalNeedProcess(baseType) && !outputsToBufferize.contains(baseType)) "reg " else ""}${emitType(baseType)} ${baseType.getName()}${if(outputsToBufferize.contains(baseType)) "" else getBaseTypeSignalInitialisation(baseType)}${emitCommentAttributes(baseType.instanceAttributes)}"
+        portMaps += s"  %s%s %s%-${maxSpacing}s %s%s%s".format(emitSyntaxAttributes(baseType.instanceAttributes),
+                                                 emitDirection(baseType),
+                                                 if(signalNeedProcess(baseType) && !outputsToBufferize.contains(baseType)) "reg " else "",
+                                                 emitType(baseType),
+                                                 baseType.getName(),
+                                                 if(outputsToBufferize.contains(baseType)) "" else getBaseTypeSignalInitialisation(baseType),
+                                                 emitCommentAttributes(baseType.instanceAttributes))
+        
     )
   }
 
@@ -92,6 +106,7 @@ class ComponentEmitterVerilog(
     declarations ++= emitBaseTypeWrap(io, name)
     referencesOverrides(io) = name
   }
+  
 
   def emitArchitecture(): Unit = {
     for(mem <- mems){
@@ -135,12 +150,26 @@ class ComponentEmitterVerilog(
         //        expressionToWrap ++= mux.inputs
       }
     }
-
+    
+    
     component.children.foreach(sub => sub.getAllIo.foreach(io => if(io.isOutput) {
       val name = component.localNamingScope.allocateName(sub.getNameElseThrow + "_" + io.getNameElseThrow)
-      declarations ++= emitExpressionWrap(io, name)
+      //declarations ++= emitExpressionWrap(io, name)
       referencesOverrides(io) = name
+      
+      var addDecl = true
+      if(spinalConfig.removeOutputAssigns){
+        component.getAllIo.foreach(topio => if(topio.isOutput) {
+          if(topio.head.source == io){
+            referencesOverrides(io) = topio.head.target
+            subComponentOutputToNotBufferize += topio
+            addDecl = false
+          } 
+        })
+      }
+      if(addDecl) declarations ++= emitExpressionWrap(io, name)
     }))
+    
 
     //Wrap expression which need it
     cutLongExpressions()
@@ -176,6 +205,7 @@ class ComponentEmitterVerilog(
       if(p.leafStatements.nonEmpty ) {
         p.leafStatements.head match {
           case AssignmentStatement(target: DeclarationStatement, _) if subComponentInputToNotBufferize.contains(target) =>
+          case AssignmentStatement(target: DeclarationStatement, _) if subComponentOutputToNotBufferize.contains(target) => //steven
           case _ => emitAsynchronous(p)
         }
       } else {
--- a/core/src/main/scala/spinal/core/Spinal.scala
+++ b/core/src/main/scala/spinal/core/Spinal.scala
@@ -140,6 +140,7 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         transformationPhases           : ArrayBuffer[Phase] = ArrayBuffer[Phase](),
                         memBlackBoxers                 : ArrayBuffer[Phase] = ArrayBuffer[Phase] (/*new PhaseMemBlackBoxerDefault(blackboxNothing)*/),
                         rtlHeader                      : String = null,
+                        removeOutputAssigns            : Boolean = false,
                         private [core] var _withEnumString : Boolean = true
 ){
 
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitter.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitter.scala
@@ -69,6 +69,8 @@ abstract class ComponentEmitter {
   val outputsToBufferize = mutable.LinkedHashSet[BaseType]() //Check if there is a reference to an output pin (read self outputed signal)
   val subComponentInputToNotBufferize = mutable.HashSet[Any]()
   val openSubIo = mutable.HashSet[BaseType]()
+  
+  val subComponentOutputToNotBufferize = mutable.HashSet[Any]()
 
   def getOrDefault[X,Y](map: java.util.concurrent.ConcurrentHashMap[X,Y], key: X, default: Y) = map.get(key) match {
     case null => default
@@ -100,6 +102,7 @@ abstract class ComponentEmitter {
     else
       null
   }
+
 
   def elaborate() = {
     val asyncStatement = ArrayBuffer[LeafStatement]()
@@ -294,6 +297,7 @@ abstract class ComponentEmitter {
         }
       }
     }
+
 
     //Fill multiplexersPerSelect
     component.dslBody.walkStatements(s => {
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
@@ -109,6 +109,7 @@ class ComponentEmitterVhdl(
     declarations ++= s"  signal $name : ${emitDataType(io)};\n"
     referencesOverrides(io) = name
   }
+  
 
   def emitArchitecture(): Unit = {
     for(mem <- mems){


core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -135,12 +150,26 @@ class ComponentEmitterVerilog(
...
     component.children.foreach(sub => sub.getAllIo.foreach(io => if(io.isOutput) {
       val name = component.localNamingScope.allocateName(sub.getNameElseThrow + "_" + io.getNameElseThrow)
-      declarations ++= emitExpressionWrap(io, name)
+      //declarations ++= emitExpressionWrap(io, name)
       referencesOverrides(io) = name
+      
+      var addDecl = true
+      if(spinalConfig.removeOutputAssigns){
+        component.getAllIo.foreach(topio => if(topio.isOutput) {
+          if(topio.head.source == io){
```
Dolu1990: Which case you try to avoid here ?
if(topio.head.source == io) is weird, it wil likely fail in case of multiple assignement

lsteveol: Hmm, are you saying in the case of multiple assignment on the `topio` in this case? I would need to try that out, maybe I didn't try that in my testing. Let me give that a shot and I can modify if needed.

I basically want to remove the intermediate wire/assign that is produced for subcomponent outputs if they connect to another output port of the component they are instantiated in. If they do not connect to an output port (used internally) then I'm ok with the intermediate wire.

lsteveol: Hey @Dolu1990 ,

Can you give me an example of what you think would break this? I've tried a few different things but haven't been able to break the proposed fix.

I tried to assign the same output multiple times, but I get an ASSIGNMENT OVERLAP compiler error.

Dolu1990: Sorry for the delay.

Basicaly : 

```scala
val a = out Bool()
a := subComponent.x
when(somthing){
  a := True
}
```

In such case, the code will enter the  if(topio.head.source == io){  but the issue is that's not the only assignement.

Dolu1990: To check if a given signal has only one assignement, then : 
```scala
topio.dlcHasOnlyOne
which is equivalent to 
dlcHead == dlcLast && dlcHead != null
```

Dolu1990: But i'm still worrying about the delta cycle / race conditions issues that such update can create.

Do you see why ?

lsteveol: Yes you are correct on the multiple assignment. I tried out the dlcHasOnlyOne check and that seemed to resolve the issue for the few cases I tried.

For the delta cycle/race issue I can see where this could possibly cause some of that to happen. If I can, I'd like to describe the motivation behind removing the assigns. This is from a purely ASIC flow perspective with a heavy mixed signal influence.

When synthesizing with Design Compiler or Genus the tool will often times put in a DRC (design rule check) buffer when an assign is seen. This normally happens when synthesizing with a .lib for an analog component or some other digital PnR block that has been run through PD. For digital logic this is normally ok, as you don't really care about additional buffers provided logical equivalence and timing are preserved. For an analog signal such as a bias, high voltage signal, power/ground, etc. this would obviously be an issue. Since Verilog/VHDL have no way to distinguish digital signals vs. some analog signal that is handled by layout, you have to constrain the design during synthesis to "dont_touch" certain signals.

For example, say you have the following verilog

```verilog
module mod_with_some_analog(
  input  wire some_digital_sig1,
  input  wire some_digital_sig2,
  output wire some_digital_sig3,
  output wire dac, //analog
  input  wire vcc,  //analog
  input  wire vss //analog
);
//...
endmodule
```

`dac`, `vcc`, and `vss` are analog signals in which you would not want the synthesis tool to place a buffer on. To resolve this, you would have the following constraints

```tcl
set_dont_touch [get_pins path_to_mod/dac]
set_dont_touch [get_pins path_to_mod/vcc]
set_dont_touch [get_pins path_to_mod/vss]
```

When an additional assign is put in there is now the *potential* issue that you have two signals to set a dont_touch on. Now I will admit I have not done enough testing to know if the `set_dont_touch` propagates through `assign` statements. It's entirely possible that it does, but I know in the past it has been pushed that we try to remove assign statements for signals which are done in custom layout.

For additional race conditions I'll say that in a traditional ASIC flow we carefully construct the clock trees in such a way that should prevent, or at least significantly reduce the possibility of a delta cycle issue. To put into perspective, in the past 7 years I've only seen it with our analog models where we are doing some rather strange modeling :)

For all of the clocks we use in a design, we actually hand place clock buffers (based on the process/vendor) which are used during synthesis and PnR as a anchor for any design constraints. This is done for a few reasons:
1. Place and Route can modify the respective clock buffers based on speed, power requirements, etc.
2. For design constraints you have a clear point of clock definition. Most synthesis tools will also rename flops in the design. Take the following image
image

Here would be a traditional clock divider where I'm taking the input clock `clk` and creating a divide by 2. One would think to create a generated clock on the `Q` of `FF1`. 
`create_generated_clock -name CLK_DV2 -source [get_ports clk] -edges {1 3 5} [get_pins FF1/Q]`
However during synthesis, `FF1` may be renamed to something like `FF1_reg` or in the case of using multi-bit flops `FF1_SomeOtherReg_MBit`. Sometimes the synthesis tool will print out an updated SDC file with any changes, but this can often cause issues when you merge constraints between IPs.

So we place a clock buffer, that we have set as `dont_touch` so the synthesis/PnR tool will not change it, and we define the clock on that.
`create_generated_clock -name CLK_DV2 -source [get_ports clk] -edges {1 3 5} [get_pins CBUF/Z]`

We now know that this constraint can be used at any stage of the design, pure verilog, Synthesis netlist, PnR netlist for GLS and don't have to worry about constraints referencing a component that is not there.

We also will place a clock mux at the entry point to any IP to allow muxing of a scan clock for DFT purposes. So all flops in a design would be traced back to one conceptual entry point.

That's a long winded answer :) but I also decided that it would be best to have a spinalConfig option to enable/disable this feature, since in the issue you had mentioned the same problem regarding delta cycle issues. 

Are you ok with this feature provided we keep the original methodology (output assigns) as the default?

lsteveol: Also I'll do some more testing and update the pull request. I might need a few days with other things at work.

Dolu1990:
> For additional race conditions I'll say that in a traditional ASIC flow we carefully construct the clock trees in such a way that should prevent ... we carefully construct the clock trees in such a way that should prevent, or at least significantly reduce the possibility of a delta cycle issue.

I was talking about race condition / delta cycle issue only from a digital simulation perspective, it seem you were interpreting it more in a backend / design perspective, or i miss something ?

```
  output wire dac, //analog
  input  wire vcc,  //analog
  input  wire vss //analog
``` 

Then we could imagine setting a specific property on those signal in SpinalHDL to say "That's analog stuff, avoid intermediate signals", it would have the advantage to be explicit and could be reuse for other purposes. I mean, VHDL/Verilog implicit inferation is probably similar to the bilion $$$ null mistake in software XD, some say "lipstick on a pig" :D

> Are you ok with this feature provided we keep the original methodology (output assigns) as the default?

I'm not against as long that's disabled by default (because of the delta cycle potential issues in simulation), but maybe we can do better (explicit analog signal tags for signal in SpinalHDL itself)
Then we can mix digital / analog in a clean and safe way

What is your opinon ?

lsteveol:
Constructing the clock tree in that manner will usually avoid the delta cycle issue in digital simulation as well.

I have been working on a DSL that sits on top of Spinal. Right now I have signals that I declare as a certain type of analog (not the Spinal built in Analog). I mainly do this for creating `set_dont_touch` lists after elaboration.

I will modify the current pull request to resolve the issues you've brought up and give some thought to how we can specify intermingling digital and analog signals. I mainly don't want to bloat Spinal with something that could be handled with an additional DSL and/or some custom DSL as many ASIC companies have their own methodologies  with regards to how they handled mixed signal designs.

numero-744:
Hi Pathfinder @lsteveol, any news about this PR and the linked issue?

numero-744:
This PR is old and has conflicts so I close it, but it is still linked to #251 so that people can read it to fix #251.

PR #270, opened 2020-02-13
Bits method extention

jijingg:
@Dolu1990  i want add some usefull method to Bits, 
like
```scala
val res = BitsA.takeHigh(5)
val res = BitsA.drop(5)
val res = BitsA.splitAt(3)
val res = BitsA.apartBy(List(2,3,5)
```
But what makes me wonder is that these names conflict with the method names of string
like `drop`, `take`, `splitAt`.
when build project , always faild at:
image

image


Dolu1990:
Hmm that's likely because string have implicit convertion to Bits, then

"miaou".splitAt   can also be understand by the compiler as B("miaou").splitAt

Maybe we should drop that implicit convertion in favor of an explicit B"010010"

Dolu1990:
(https://github.com/SpinalHDL/SpinalHDL/blob/dd37f919c873fd31ef5b178bb44b82ea120d1238/core/src/main/scala/spinal/core/core.scala#L233)

jijingg:
ok, I added those method in my local by implicit extension as a workaround. the Bits' slice operations are very frequent. I hope that the later Spinal can support some slice function  natively.

Dolu1990:
@jijingg Normaly

Dolu1990:
Ahh sorry, when i closed the issue, it was to me about fixing the compilation issue that i fixed in : 
https://github.com/SpinalHDL/SpinalHDL/commit/d44a7f0c35756a8097ed2ee88f6ebf20a008af85

I forgot about the addition XD

Dolu1990:
So i'm totaly ok with the functionalty, i'm just thinking if there is no better naming for takeHigh and drop, something a bit more explicit.

Dolu1990:
Maybe 
- dropHigh ?
- takeHigh ?

jijingg:
> Maybe
dropHigh ?
takeHigh ?

sure,  but `drop` little  different to `takeHigh`, `drop(n) = thakeHight(getWidth - n)`
so i keep all 4 `take, takeHigh,  drop,  dropHigh`, is it ok?

next , what about `sliceBy` instead of `apartby` ?  or you have a better name ?

jijingg:
@Dolu1990 is this ok?   should this PR be finished

Dolu1990:
> next , what about sliceBy instead of apartby ? or you have a better name ?

Let's name it sliceAt, as we slice at the position given in the list, i think that make the more sense

Dolu1990:
> Let's name it sliceAt, as we slice at the position given in the list, i think that make the more sense

Forget what i said, i was wrong about the behaviour, you are right, sliceBy is good :)

> so i keep all 4 take, takeHigh, drop, dropHigh, is it ok?

Basicaly, i'm trying to avoid non explicit things. For instance, one issue of the Cat function, is that it doesn't relay say in which direction things are Cat together XD, but in the given case i think that's fine.

But if we want to follow the Scala way of nameing such things, maybe : 
takeRight, dropRight ?

jijingg:
takeRight in scala means take first `List(0,1,2,3,4)` , but in HDL , the last Bit at the Right position 32'h43210 , exact converse， Will it lead to confusion？

jijingg:
takeRight in scala means take last  in List(0,1,2,3,4) , but in HDL , the first Bit at the Right position 32'h43210 , exact converse， Will it lead to confusion？

Dolu1990:
Thanks ^^

PR #283, opened 2020-03-08
little fix

jijingg:
1. `Bits.splitAt(0)` fix 
2. negative constant hexString align instead true as false
```scala 
val a = SInt(6 bits)
a := -3    //assign a = 6'hfd align will lead EDA warning 
``` 
now  `a := -3` got `assign a = 6'h3d` not `assign a = 6'hfd`


jijingg:
@Dolu1990 here is two little fix ,  it‘s ready for merge

Dolu1990:
thanks ^^

PR #286, opened 2020-03-16
adding JtagTap implementation to Lattice ECP5 JTAGG Blackbox

hwengineer:
Hi
I recently added the Lattice ECP5 JTAGG Blackbox to this repos.

With that I made a variant of the JtagTap which uses the JTAGG Blackbox.
It has the same protocol as the existing JtagTap.
I also refactored the "old" code such that both variants now share a common Trait and as such can be used (mostly) the same (the JTAGG has some limitations...)

I had to change the "instructionId" of the existing implementation to an Int and cast it to Bits in the code. I tested it with Symbyflow that this change should generate the same code.
I would say it should be identical. Maybe someone can test the "old" JtagTap implementation?

The JtagTap Implementation of the ECP5 is simulated and tested successfully on a board of mine.


Dolu1990:
Hmm maybe we should thing of a more composite JTAG tap design, so having a universal backend (to manage commands and data) and  a separated front end which could be cleaning implemented for all sorts of FPGA vendor virtual JTAG and regular logic made JTAG ?

That would sort the issue one and for all :)

hwengineer:
Hi
I tried to not break the current implementation just as an addition to the currently used JtagTap.
But sure if you're okay with breaking the current implementation I could work on a more composite design.

Just as a general question:
The JtagTapAccess trait seems to be a good interface for a JtagTap or would you want to change that too?

Dolu1990:
@hwengineer Hi,
sorry, missed the question.

> The JtagTapFunctions trait seems to be a good interface for a JtagTap or would you want to change that too?

I was more thinking about hardware decoupling phy and tap. allowing to expose a tap interface without knowing on what phy you will plug it.

Normaly, i should have some time soon to work on that :)

hwengineer:
@Dolu1990 
I also understood it that way.
I already started with it. It's not done yet but it's the works.

As an outlook:
I decoupled the Phy(Backend) with two Stream(Fragment(Bits(1 bits)) and plan to make a Frontend something like the current access functions.
I think I'm on a good way, but wasted much time with (many..) small syntax errors on the way...

lib/src/main/scala/spinal/lib/com/jtag/lattice/ecp5/JtagTap.scala
```diff
@@ -0,0 +1,176 @@
...
+
+import spinal.core._
+import spinal.lib._
+
+import spinal.lib.blackbox.lattice.ecp5.{JtaggIo, JTAGG, JtaggGeneric}
+import spinal.lib.com.jtag.{JtagTapFunctions, JtagTapShifter, JtagTapFactory}
+
+import scala.collection.immutable.HashMap
+import scala.collection.mutable.ArrayBuffer
+
```
Dolu1990, line 11: Thanks for the comments ^^

lib/src/main/scala/spinal/lib/com/jtag/JtagTapInstructions.scala
```diff
@@ -0,0 +1,145 @@
...
+//
+class JtagTapInstruction(tap: JtagTapAccess, val instructionId: Int) extends Area {
+  def doCapture(): Unit = {}
+  def doShift(): Unit = {}
+  def doUpdate(): Unit = {}
+  def doReset(): Unit = {}
+
+  val instructionHit = tap.getInstruction === B(instructionId, tap.getInstruction.getWidth bits)
+
+  Component.current.addPrePopTask(() => {
```
Dolu1990, line 20: I'm thinking about removing that, and having the callback managed from the jtag tap itself, this would avoid having the JtagTapInstruction having to know the the state of the jtag

Dolu1990:
Ok,

So, i will create a new branch in SpinalHDL to work on this, by first merging your changes in.

Thanks for the pullrequest ^^

Dolu1990:
So basicaly, i think the focus should be on how instruction are managed, as that's the common denominator. What's about getting ride of the JtagTapInstruction, and having a common regular Bundle interface ? 

```scala
case class JtagTapInstructionCtrl() extends Bundle{
  val tdi = Bool()
  val tdo = Bool()
  val enable = Bool()
  val capture = Bool()
  val shift = Bool()
  val update = Bool()
  val reset = Bool()
}


class JtagTapInstructionWrite[T <: Data](data: T, cleanUpdate: Boolean = true, readable: Boolean = true) extends Area{
  val ctrl = JtagTapInstructionCtrl()

  val shifter,store = Reg(Bits(data.getBitsWidth bit))

  when(ctrl.enable) {
    when(ctrl.capture) {
      shifter := store
    }

    when(ctrl.shift) {
      shifter := (ctrl.tdi ## shifter) >> 1
      if (readable) ctrl.tdo := shifter.lsb
    }

    when(ctrl.update) {
      store := shifter
    }
  }

  if (!cleanUpdate)
    data.assignFromBits(shifter)
  else
    data.assignFromBits(store)
}

class JtagTapInstructionRead[T <: Data](data: T) extends Area {
  val ctrl = JtagTapInstructionCtrl()
  val shifter = Reg(Bits(widthOf(data) bits))

  when(ctrl.enable) {
    when(ctrl.capture) {
      shifter := B(data)
    }

    when(ctrl.shift) {
      shifter := (ctrl.tdi ## shifter) >> 1
      ctrl.tdo := shifter.lsb
    }
  }
}

class JtagTapInstructionIdcode[T <: Data](value: Bits, instructionSet: Flow[Bits]) (instructionId: Int) {
  val ctrl = JtagTapInstructionCtrl()
  val shifter = Reg(Bits(32 bit))

  when(ctrl.enable) {
    when(ctrl.shift) {
      shifter := (ctrl.tdi ## shifter) >> 1
      ctrl.tdo := shifter.lsb
    }
  }

  instructionSet.payload := instructionId
  when(ctrl.reset){
    shifter := value
    instructionSet.valid := True
  }
}


class JtagTapInstructionFlowFragmentPush(sink : Flow[Fragment[Bits]], sinkClockDomain : ClockDomain) extends Area{
  val ctrl = JtagTapInstructionCtrl()

  val source = Flow Fragment(Bits(1 bit))
  val valid = RegNext(ctrl.enable && ctrl.shift)
  val data = RegNext(ctrl.tdi)

  source.valid := valid
  source.last := !(ctrl.enable && ctrl.shift)
  source.fragment.lsb := data

  sink << FlowCCByToggle(source, outputClock = sinkClockDomain)
}
```

Dolu1990:
@hwengineer 

A funny one :

How to wrap multiple instructions inside singe tap instruction, using DR_SHIFT header to select which one you want : 

```scala
class JtagInstructionWrapper(headerWidth : Int) extends Area with JtagTapFunctions{
  val ctrl = JtagTapInstructionCtrl()

  val header = Reg(Bits(headerWidth bits))
  val headerNext = (ctrl.tdi ## header) >> 1
  val counter = Reg(UInt(log2Up(headerWidth) bits))
  val done = Reg(Bool())
  val sendCapture = False
  val sendShift   = False
  val sendUpdate  = False


  when(ctrl.enable){
    when(ctrl.capture){
      done := False
      counter := 0
    }
    when(ctrl.shift){
      when(!done) {
        counter := counter + 1
        header := headerNext
        when(counter === headerWidth - 1) {
          done := True
          sendCapture := True
        }
      } otherwise {
        sendShift := True
      }
    }
    when(ctrl.update){
      sendUpdate := True
    }
  }

  ctrl.tdo := False

  def map(userCtrl : JtagTapInstructionCtrl, instructionId : Int): Unit ={
    val hit = header === instructionId
    userCtrl.tdi     := ctrl.tdi
    userCtrl.enable  := True
    userCtrl.capture := headerNext === instructionId && sendCapture
    userCtrl.shift   := hit && sendShift
    userCtrl.update  := hit && sendUpdate
    userCtrl.reset   := ctrl.reset
    when(hit) { ctrl.tdo := userCtrl.tdo }
  }

  // implement traits of JtagTapFunctions
  override def idcode(value: Bits)(instructionId: Int) = ???
  override def read[T <: Data](data: T, light : Boolean = false)(instructionId: Int) = {
    val area = new JtagTapInstructionRead(data, light = light)
    map(area.ctrl, instructionId)
    area
  }
  override def write[T <: Data](data: T, cleanUpdate: Boolean = true, readable: Boolean = true)(instructionId: Int) = {
    val area = new JtagTapInstructionWrite(data, cleanUpdate, readable)
    map(area.ctrl, instructionId)
    area
  }
  override def flowFragmentPush[T <: Data](sink : Flow[Fragment[Bits]], sinkClockDomain : ClockDomain)(instructionId: Int) = {
    val area = new JtagTapInstructionFlowFragmentPush(sink, sinkClockDomain)
    map(area.ctrl, instructionId)
    area
  }
}
```

Dolu1990:
JtagInstructionWrapper usage : 
```scala
  val jtag = ClockDomain(io.jtag.tck)(new Area{
    val tap = new JtagTap(io.jtag, 4)
    val idcodeArea = tap.idcode(B"x10001FFF")(1)
    val wrapper = tap.instructionWrapper(headerWidth = 4)(instructionId = 2)
    val writeArea = wrapper.flowFragmentPush(system.cmd,JtagBridge.this.clockDomain)(6)
    val readArea = wrapper.read(system.rsp.addTag(crossClockDomain), light = true)(7)
  })
```

hwengineer:
Hi @Dolu1990 

When we want to have a backend for different FPGA vendor jtag cells, we need another approach.
I can only speak for the lattice ecp5 jtag cell, but let me explain:

Jtagg

That picture is from the documentation of the hard-ip.
The stared signals TMS TCK TDI TDO are mentioned in the picture but are not accessible by the fpga respectively by our synthesized code.
actually the pins are not in the the netlist and the place-and-route will assert en error, that it not found the signals.
By the way, you see the explanation which signal does what in 
[debug.scala](https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/blackbox/lattice/ecp5/debug.scala)

thus if we want to reuse as much code as possible, the interface with the tdo, tdi, etc. will not be feasible

So the jtag cell does decode the low level signals, implements the idcode and bypass etc. registers but reroutes the TDI and TDO signals when the idcode register is loaded with the right id.
But...
Where do we know which idcode is loaded in the jtag cell?
Well we have to observe these signals JCEx / JRTIx
The JTAGG hard-ip declares two idcodes (0x32 / 0x38) as usercodes thus 1 / 2

and this is a pain because we only know for sure which idcode is used when the (internal hard-ip) jtag enters Capture_DR state.
The JCEx will be high if we're in Capture_DR or Shift_DR and JRTIx will be high if where in IDLE 

In an allowed transition, from Update_IR to Select_DR we have no indication, that the last Update_IR changed the idcodes and thus changed between 0x32 and 0x38.
So we can't rely on the last idle state and thus JRTIx.
In the end we just have to observe JCEx to determine the current idcode

Jtagg_timing

The picture shows the internal signals connected to an output pin. And measured via a scope.
The JTDI Signal is the samples TDI pin with one clock cycle delay, so the JTDO and JTDI streams have one clock cycle difference to each other.
The embedded jtag cell also makes the half cycle shift of the JTDO signal.

You see, the embedded cell makes already much decoding of the TMS and Jtag FSM.
I can only assume that other embedded jtag cells will make similar approaches, like shifting the tdo and a sampling delay of the tdi signal.
and of course the decoding of the instruction register

hwengineer:
I would propose the following approach:
for nomenclature: I called the FPGA vendor the "backend" and the generic access the "frontend"

Because the tdo and the tdi timings of the vendor jtag cells can be with a clock delay. I would implement a Flow respectively a Stream interface to connect them.
Also the embedded cells will most likely decode the idcode and only signalling it.

I would split the backend and the frontend this way:
The backend has to decode the idcode and generate the stream / flow of the tdi / tdo
and they implement it as a Fragment(Bits(1 bits). The idcode is a Bits() and is only valid for the time when the serialOutput stream is valid

```scala
// JtagSlaveBackendIo Interface
//
// SlaveCtrl <- SlaveBackend 
// Slave     <- Master 
case class JtagSlaveBackendIo(instrIdWidth: Int) extends Bundle 
                                            with IMasterSlave{
  val instrId = Bits(instrIdWidth bits)
  val serialIn  = Flow(Fragment(Bits(1 bits)))
  val serialOut = Stream(Fragment(Bits(1 bits)))
}
```
the fronted implements and arbiter for the two stream and access methods to attach to it.

As I stated I work on that, but I will take some time...

Dolu1990:
@hwengineer 

> thus if we want to reuse as much code as possible, the interface with the tdo, tdi, etc. will not be feasible
So the jtag cell does decode the low level signals, implements the idcode and bypass etc. registers but reroutes the TDI and TDO signals when the idcode register is loaded with the right id.
But...
Where do we know which idcode is loaded in the jtag cell?
Well we have to observe these signals JCEx / JRTIx
The JTAGG hard-ip declares two idcodes (0x32 / 0x38) as usercodes thus 1 / 2

Right, that was one mistake of the old JtagTapInstruction api, is that it was made to be quite much integrated into the tap, with quite some inner view that aren't always provided by hard tap to custom instructions.


Because the tdo and the tdi timings of the vendor jtag cells can be with a clock delay.

Hoo damned, realy ? do you know one vendor which expect your custom instruction to have one cylce delay between tdi and tdo ? I mean, that would be ennoying.


> Because the tdo and the tdi timings of the vendor jtag cells can be with a clock delay.I would implement a Flow respectively a Stream interface to connect them.

The issue of that is that anyway, the user should implement the tap with zero latency between the controlles and tdo, because that the only way it can be connected to a vendor phy which assume zero latency.


What is the issue of : 

```scala
case class JtagTapInstructionCtrl() extends Bundle{
  val tdi = Bool()
  val tdo = Bool()
  val enable = Bool()
  val capture = Bool()
  val shift = Bool()
  val update = Bool()
  val reset = Bool()
}
```

see https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/com/jtag/JtagTapInstructions.scala 

?

You might event be able to reuse all of them in the ECP hard tap isn't it ?

hwengineer:
@Dolu1990 

I also looked briefly at the Altera Max 10 Jtag:
[Intel MAX 10 JTAG BST Overview : JTAG Instructions](https://www.intel.com/content/www/us/en/programmable/documentation/sss1396244916395.html#sss1395223016816)
Their jtag cell also decodes the idcode for itself and signals it via a USER0 / USER1 signal.

---

Back to the tdi  timing issue.

I appended you the timing diagram i made for the analysis of the ECP5 as a hole.
There you see the difference from the "normal" jtag signals to the output of the JTAGG cell.
[timing.pdf](https://github.com/SpinalHDL/SpinalHDL/files/4500050/timing.pdf)

The JTDI signal lag's one cycle behind.
This is due to the fact, that the hard-ip the signal already samples with an internal Flipflop.
Now we could use the output of the JTDI signal directly for driving combinatorial logic, because its already sampled it can be used the same as a register output. 
then we would have the same timing as you got it with the gpio implementation
The JTDI output must then be the MSB of the shifter I guess.
That might work

---

> What is the issue of :
```
case class JtagTapInstructionCtrl() extends Bundle{
  val tdi = Bool()
  val tdo = Bool()
  val enable = Bool()
  val capture = Bool()
  val shift = Bool()
  val update = Bool()
  val reset = Bool()
}
```
Hmm... I think i see your point now.
The idcode is in the enable value... sorry I didn't recognized that yesterday.

I will try this out with my implementation!

Dolu1990:
> I appended you the timing diagram i made for the analysis of the ECP5 as a hole.
There you see the difference from the "normal" jtag signals to the output of the JTAGG cell.

So the second diagram is from JTAGG ? and the TDI is the one getting out of the IP to the user ? So the IP is providing the TDI with one extra cycle latency more than the states of the jtag (CE, SHIFT, UPDATE) ?
That's realy weird, or i missunderstood the test setup ?

By " In the next example, JTDI is connected to JTDO directly. (daisy chaining)" you mean you connect JTAGG.TDO and JTAGG.TDI together combinatorialy ?

> Hmm... I think i see your point now.
The idcode is in the enable value... sorry I didn't recognized that yesterday.
I will try this out with my implementation!

No worries ^^, but the JTAGG seem realy weird to me XD
Hmm about the branches
Actualy, everything is back into SpinalHDL dev, including your changes, but if we want to go the JtagTapInstructionCtrl way, we will need to update your ECP changes to use it :)

hwengineer:
Hi @Dolu1990 

So... I'm nearly done with it. Just some testing (without the JtagInstructionWrapper)

A question: What is the purpose of the JtagInstructionWrapper exactly?
I didn't saw it used somewhere in the repository.

>So the second diagram is from JTAGG ? and the TDI is the one getting out of the IP to the user ? >So the IP is providing the TDI with one extra cycle latency more than the states of the jtag (CE, >SHIFT, UPDATE) ?
>That's realy weird, or i missunderstood the test setup ?

short answer : Yes it shows the JTAGG internal signals
The first image is the transmission measured at the physical lanes on the Jtag interface with an oscilloscope.
You see two words (transactions) 0xA5 / 0x5A send after each other, but placed below each other for visualization.
The second image are all FPGA internal signals routed to FPGA pins and measured with an Oscilloscope (except for JTDO, which is measured at the physical interface...)

>By " In the next example, JTDI is connected to JTDO directly. (daisy chaining)" you mean you >connect JTAGG.TDO and JTAGG.TDI together combinatorialy ?

Yes, the JTAGG.JTDO connected internally with the JTAGG.JTDI. And measured TDO at the pins, (the text is a bit misleading...)

I don't think its that weird, its just the sampled TDI signal from the input register stage.
You so or so have to sample it.
Maybe other vendors will do it the same...

Dolu1990:
> So... I'm nearly done with it. Just some testing

 ok ^^
So did you toke the changes from the dev branch ?

> A question: What is the purpose of the JtagInstructionWrapper exactly?
I didn't saw it used somewhere in the repository.

Basicaly, it allow to encapsulate multiple "virtual" JTAG instruction into one single "real" JTAG instruction. basicaly, i'm using it to allow having the two instruction used by the VexRiscv JTAG inside a single FPGA hardware tape user instruction.
It work by adding a header to the DR-shift, used to redirect the DR operations to the right "virtual" instruction.

> Yes it shows the JTAGG internal signals

The thing which is realy weird to me, is the control signals as SHIFT have one cycle less latency than TDI on the second chart. They both come from the JTAGG right ?

What is the the first wave ? i mean, what is TDI1 ? do you have access to it as this in the user FPGA design ?

Seem like my proposed interface would be confused by the JTAGG, as TDI value only come one cycle after SHIFT, instead than at the same time.

> Maybe other vendors will do it the same...

The only other vendor i tried do not do as the ECP5, but provide the control signals (SHIFT, ...) and TDI with the same phase, all combinatorialy.

Ahhhh
paine in the ass :)

hwengineer:
Hi @Dolu1990 

I took the current dev branch and updated my repo accordingly.

I made several changed.
First I used a sane default for the ecp5 Jtagg Blackbox.
Then I added the changes to the JtagTap.
Indeed the new interface reduced the complexity of the code!

i reworked the JTagTapFunctions. 
I think it makes sense to make an own implementation to this quirky JtagTap.
Most of the hard stuff I exported to a own Shifter area.
The functions are mostly looking the same for the rest of the code.

I didn't implement an equivalent for the Wrapper for the moment.
Maybe at another day when I can test it a little bit better.

---

>The thing which is realy weird to me, is the control signals as SHIFT have one cycle less latency 
> than TDI on the second chart. They both come from the JTAGG right ?

Yes both are from the JTAGG. Yeah, like some information is combinatorial logic directly and then boom a registered JTDI...
Not very consistent.
But nevertheless my implementation know works transparent to your's

---

I talked to a colleague which did some Usercode Jtag stuff in the company and he also stated that he only saw this type of jtag usercode handling at Lattice...

Dolu1990:
Thanks :)
All look good to me, so, good for you to merge ?

hwengineer:
@Dolu1990 
Yes

Dolu1990:
Thanks ^^

PR #287, opened 2020-03-19
Various little fixups

piegamesde:
- Re-written `Stream::slowdown` as it's broken
- Re-design of `StreamJoin` to make it easier to use
- Added `TupleBundle1` to `TupleBundle22`

### DIFF
--- a/lib/src/main/scala/spinal/lib/Fragment.scala
+++ b/lib/src/main/scala/spinal/lib/Fragment.scala
@@ -56,6 +56,36 @@ class StreamFragmentPimped[T <: Data](pimped: Stream[Fragment[T]]) {
 
   def toStreamOfFragment : Stream[T] = pimped.translateWith(pimped.fragment)
 
+/**
+ * Stepwise reduction of all fragments into a single value. With this, you can calculate
+ * the (check)sum of every n elements or parity bits.
+ * @param accumulator A combinatorial function that accumulates the current (param 3) and the previous
+ * 	(param 2) value into the next one (param 1)
+ * @example {{{ outStream = inStream.reduce[SInt]((a, b, c) => { a := b + c }) }}}
+ */
+  def reduce[U <: Data](identity: U, accumulator: (U, U, T) => Unit): Stream[U] = {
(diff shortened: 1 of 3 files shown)


piegamesde:
Ping @Dolu1990

Dolu1990:
@piegamesde Hooo right, i'm on it now :)

lib/src/main/scala/spinal/lib/Fragment.scala
```diff
@@ -56,6 +56,36 @@ class StreamFragmentPimped[T <: Data](pimped: Stream[Fragment[T]]) {
 
   def toStreamOfFragment : Stream[T] = pimped.translateWith(pimped.fragment)
 
+/**
+ * Stepwise reduction of all fragments into a single value. With this, you can calculate
+ * the (check)sum of every n elements or parity bits.
+ * @param accumulator A combinatorial function that accumulates the current (param 3) and the previous
+ * 	(param 2) value into the next one (param 1)
+ * @example {{{ outStream = inStream.reduce[SInt]((a, b, c) => { a := b + c }) }}}
```
Dolu1990, line 64: Why not using something similar to the scala collection library ? 

(b, c) => { b + c }   instead of (a, b, c) => { a := b + c } ?

lib/src/main/scala/spinal/lib/TupleBundle.scala
```diff
@@ -0,0 +1,875 @@
...
+  val _15 = payloadType15()
+  val _16 = payloadType16()
+  val _17 = payloadType17()
+  val _18 = payloadType18()
+  val _19 = payloadType19()
+  val _20 = payloadType20()
+  val _21 = payloadType21()
+}
+
+case class TupleBundle22[
```
Dolu1990, line 805: <3

lib/src/main/scala/spinal/lib/Fragment.scala
piegamesde: I just copied the pattern from `Stream::translateInto` (among others). If we change it, we should change it everywhere to remain consistent. Also, I don't really know how to do it the other way.

PR #297, opened 2020-04-14
QSysify emitters for AXI4, AXI4 Lite and Interrupt Sender

c-thaler:
Configs for AXI4 and AXI4 had to be extended:
max number of pending reads/writes and reorder FIFO depth

I also added a check for AXI4 Lite data width. The spec only allows 32 and 64 bit.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
@@ -55,10 +55,10 @@ case class Axi4Config(addressWidth : Int,
                       wUserWidth   : Int = -1,
                       bUserWidth   : Int = -1,
                       
-                      readIssuingCapability     : Int = 1,
-                      writeIssuingCapability    : Int = 1,
-                      combinedIssuingCapability : Int = 1,
-                      readDataReorderingDepth   : Int = 1) {
+                      readIssuingCapability     : Int = -1,
+                      writeIssuingCapability    : Int = -1,
+                      combinedIssuingCapability : Int = -1,
+                      readDataReorderingDepth   : Int = -1) {
 
 
   def useArUser = arUserWidth >= 0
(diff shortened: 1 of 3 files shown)


lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
```diff
@@ -49,11 +49,16 @@ case class Axi4Config(addressWidth : Int,
...
-                      arUserWidth   : Int = -1,
-                      awUserWidth   : Int = -1,
+                      arUserWidth  : Int = -1,
+                      awUserWidth  : Int = -1,
                       rUserWidth   : Int = -1,
                       wUserWidth   : Int = -1,
-                      bUserWidth   : Int = -1) {
+                      bUserWidth   : Int = -1,
+                      
+                      readIssuingCapability     : Int = 1,
```
Dolu1990, line 58: We might want to put it to Int.MaxValue or -1 by default ? to say "unspecified" by default ?

Dolu1990:
Sorry for the delay :)
see my above review comment about AXI.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
c-thaler: Yep, makes sense. We should use -1 by default and prevent the QSysifyEmitter from emitting the corresponding `set_interface_property` in this case.
I think that the `*Capability` parameters are optional, so Quartus will use some "safe" default. I'm not sure of this, I will check it ASAP.

c-thaler: Quartus Platform Designer defaults to 1 when `*Capability` is not set in the TCL script. I will proceed as proposed in my former comment.

c-thaler:
The QSys parameters are now disabled by default.

Dolu1990:
ok :)
So all good for merge ?

c-thaler:
Yes, please :)

Dolu1990:
Thanks ^^

PR #299, opened 2020-04-15
Various Bugfixes and improvements to VPI/GHDL backend

bellaz89:
- Now the VPI handles are stored internally in SimVpi to avoid getting invalid values in re-run simulations
- Thread locking on VpiBackend creation and .instanciate
- VpiBackend.instanciate waits the simulation to be ready before returning
- Elaboration and run step was splitted in GhdlBackend. Now elaboration is executed in  .analyze
- Added lazy compilation/analyze step to reduce the number of compilations to be done
- Added  spinal.sim.TestGhdl15 to test Parallel execution of GhdlBackend .. passing
- A crashing simulation doesn't hang the JVM anymore

Now spinal.tester.code.PlaySimGhdl3 runs fine :)

### DIFF
--- a/sim/src/main/java/spinal/sim/vpi/JNISharedMemIfaceJNI.java
+++ b/sim/src/main/java/spinal/sim/vpi/JNISharedMemIfaceJNI.java
@@ -39,5 +39,6 @@ public class JNISharedMemIfaceJNI {
   public final static native void SharedMemIface_eval(long jarg1, SharedMemIface jarg1_);
   public final static native void SharedMemIface_randomize(long jarg1, SharedMemIface jarg1_, long jarg2);
   public final static native void SharedMemIface_close(long jarg1, SharedMemIface jarg1_);
+  public final static native void SharedMemIface_check_ready(long jarg1, SharedMemIface jarg1_);
   public final static native boolean SharedMemIface_is_closed(long jarg1, SharedMemIface jarg1_);
 }
--- a/sim/src/main/java/spinal/sim/vpi/JNISharedMemIfaceJNI.java
+++ b/sim/src/main/java/spinal/sim/vpi/JNISharedMemIfaceJNI.java
@@ -40,5 +40,6 @@ public class JNISharedMemIfaceJNI {
   public final static native void SharedMemIface_randomize(long jarg1, SharedMemIface jarg1_, long jarg2);
   public final static native void SharedMemIface_close(long jarg1, SharedMemIface jarg1_);
   public final static native void SharedMemIface_check_ready(long jarg1, SharedMemIface jarg1_);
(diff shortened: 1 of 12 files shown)


bellaz89:
The CI fails to compile the SharedMemIface.cpp . Maybe boost/ghdl is missing? on my computer it runs fine until NoBB..

wifasoi:
HI,
Travis is built from source:
https://github.com/SpinalHDL/SpinalHDL/blob/310b041274db181981813b8e1f07f9d776fc537d/travis.sh#L120-L128
There are some flag missing for having the required library compiled?

bellaz89:
@wifasoi 
No they shouldn't. But boost-interprocess has to be installed

```
sudo apt install -y libboost-all-dev
```
This install the required boost dependencies on my machine (Debian 11)

wifasoi:
ok, later I'll try to add it as a dependency and check if it break things (and make a PR on your PR). For now you can ignore travis :)

bellaz89:
Thanks :D !

umarcor:
> Added spinal.sim.TestGhdl15 to test Parallel execution of GhdlBackend .. passing

Is the purpose of this to test whether you can build multiple simulations models/binaries/libraries in paralell or whether you can load an execute the in parallel? In the later case, is it functionally equivalent to mixing examples `dlopen` and `shghdl` from https://ghdl.github.io/ghdl-cosim/vhpidirect/examples/shared.html#shared with `pthreads`?

bellaz89:
> > Added spinal.sim.TestGhdl15 to test Parallel execution of GhdlBackend .. passing
> 
> Is the purpose of this to test whether you can build multiple simulations models/binaries/libraries in paralell or whether you can load an execute the in parallel? In the later case, is it functionally equivalent to mixing examples `dlopen` and `shghdl` from https://ghdl.github.io/ghdl-cosim/vhpidirect/examples/shared.html#shared with `pthreads`?

Hi!

It is the first case. SpinalSim testsuite can be executed in parallel, so it has to be assured that in every simulation the acquision of shared resources is done correctly. Each simulation is loaded in a separate process that communicates with the JVM through shared memory. So no dlopen/pthread (at least not directly...)

Dolu1990:
Thanks ^^

It fix the issue of re-running the same compiled test :)
Now, i have to add GHDL to all the exisiting SpinalSim tests ^^

wifasoi:
hei, @Dolu1990 that was too fast :P
I'll rebase my PR for CI fix in this repo, and not at @bellaz89's.
Now I'm testing the spinal/dev container, before push the commit to the "docker" repo.
The new docker dev container should be build before this branch is merge to master (otherwise the CI will complain a lot)

Issue #146, opened 2018-09-11
VHDL support in SpinalSim using GHDL as a backend


eine:
While reading https://spinalhdl.github.io/SpinalDoc/spinal/sim/introduction/, I wondered whether any effort has been done in order to replicate the workflow with VHDL sources. [ghdl/ghdl](https://github.com/ghdl/ghdl) can be used to compile VHDL files to C (shared) objects, just as [Verilator](https://www.veripool.org/wiki/verilator) is used with Verilog sources.

I am actually seeking to generate and compile a VHDL version of [Murax SoC](https://github.com/SpinalHDL/VexRiscv#murax-soc), to then simulate the execution of some of the example programs.

Dolu1990:
Actualy, there is this interface in the spinal.sim sources : 
https://github.com/SpinalHDL/SpinalHDL/blob/master/sim/src/main/scala/spinal/sim/SimRaw.scala#L15
which as to be implemented for a given target simulator
For Verilator it's implemented as following : 
https://github.com/SpinalHDL/SpinalHDL/blob/master/sim/src/main/scala/spinal/sim/SimVerilator.scala
+
https://github.com/SpinalHDL/SpinalHDL/blob/master/sim/src/main/scala/spinal/sim/VerilatorBackend.scala

So in the case of verilator it wasn't a crazy amount of work. Then i have no idea how the ghdl shared object interface appear :)

Dolu1990:
the def eval() function is called by spinalHDL to ask the simulatior to do delta cycles until everything is stabilized.
maybe that's a tricky thing ?

1138-4EB:
Hi @Dolu1990! Thanks for the references.

I think that implementing something as `eval()` with GHDL might be possible with either the [VPI](https://ghdl.readthedocs.io/en/latest/using/InvokingGHDL.html#vpi-build-commands) or the [VHPI](https://ghdl.readthedocs.io/en/latest/references/ImplementationOfVHDL.html#interfacing-to-other-languages) interfaces. I have gathered some examples in [1138-4EB/hwd-ide/tree/vpi-ex-ghdl/examples](https://github.com/1138-4EB/hwd-ide/tree/vpi-ex-ghdl/examples), but I have not thoroughly learnt it yet.

I was looking for something 'easier' to start with. I tried the following for now:

[...]

PR #304, opened 2020-04-20
SimGHDL Fixes/improvements

bellaz89:
In this pull request the following changes were done:

- Improved spinlock performances when many simulations are run at the same time. This was accomplished adding a yield directive that pass the thread control to the os if the lock is spinning for too long. A cache line optimization was added too based on https://probablydance.com/2019/12/30/measuring-mutexes-spinlocks-and-how-bad-the-linux-scheduler-really-is/ . Since I am not an expert I am unsure if the way I did that is the best one. Still I can get >30% performance improvement on TestGHDL16 vs TestGHDL17
- Fixed a memory leak on vpi_iterate and vpi_scan.
- Added windows support. Please note that since ghdl-mcode doesn't work on 64bit Windows platform, the code was tested only on ghdl-llvm.

Please note that https://github.com/ghdl/ghdl/issues/1233 fixed a memory leak on GHDL. Therefore some tests may show a memory leak if an older version of GHDL is used

This PR is part of issue #146

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,8 @@ project/plugins/project/
 .settings
 .bloop
 .metals
-
+*.swp
+*.swo
 .idea
 out
 
@@ -46,7 +47,9 @@ log.txt
 *.out
 *.o
 *.bin
+*.exe
 *.so
+*.dll
 *.vpi
 *.vpl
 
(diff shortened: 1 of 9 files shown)


Dolu1990:
thanks ^^

PR #307, opened 2020-04-22
Corrected bug in SimVpi. Sign Extension on signed values was missing

bellaz89:
Corrected bug in SimVpi. Sign Extension on signed values was missing.
 Now if the datatype is signed and the signal has a width which is smaller than an Int or a Long (depending from the method called) a sign extension is applied when using getInt and getLong


Dolu1990:
Thanks :D

PR #311, opened 2020-04-28
merged code paths for Delay

charliehorse55:
removes the old `DelayWithInit` function and merge it into the main `Delay`. Seems to be currently breaking VexRiscv :(.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Utils.scala
+++ b/lib/src/main/scala/spinal/lib/Utils.scala
@@ -780,7 +780,7 @@ class TraversableOncePimped[T <: Data](pimped: Seq[T]) {
 
 
 object Delay {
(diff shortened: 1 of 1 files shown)


Dolu1990:
Don't remove the DelayWithInit as it will break backward compatibility :)

Dolu1990:
Hi ^^

All good with VexRiscv  ?

charliehorse55:
Yes. the problem with Vexriscv was that I didn't recompile it. Changing the number of optional arguments changes the ABI under the hood, that was causing the problem. If you have optional arguments, it's the caller's job to insert a function call with the extra arguments filled out with their default values. 

So, even though the code in VexRiscv didn't change, it needed to be recompiled.

charliehorse55:
One more change - I moved the rename call to *before* the onEachReg function. I saw someone on gitter asking how to rename all their delay registers. Now you can.

Dolu1990:
Thanks ^^

PR #315, opened 2020-05-07
Make blackboxing of single port and true dual port RAM respect the readUnderWrite policy

sxtyzhangzk:
The current blackboxing process set the `readUnderWrite` attribute of `Ram_1wrs` and `Ram_2wrs` to `dontCare` regardless of what is specified in the `mem.readWriteSync` statement. 

This patch makes them respect the `readUnderWrite` policy of the corresponding `MemReadWrite` ports.


Dolu1990:
Hoo right, Thanks :)
This will be released very soon in 1.4.1

PR #320, opened 2020-05-15
Snoopy

Snoopy87:
- Rework AhbInterconnect (Decode + arbitrer) 
- Add AHB tests  for arbitrer/decoder/interconnect

- Add PhaseInitReg (Init all register uninitiated )

- You can add comments of a component in the rtf file generated


.gitignore
```diff
@@ -67,6 +67,10 @@ log.txt
 simWorkspace/
 tmp/
 null/
+
+#mac
+*.DS_Store
```
typingArtist, line 72: I suggest to make that part of the global gitignore file. See various hints on Google or just here https://stackoverflow.com/questions/7335420/global-git-ignore

core/src/main/scala/spinal/core/Component.scala
```diff
@@ -324,4 +324,11 @@ abstract class Component extends NameableByComponent with ContextUser with Scala
     }
     null
   }
+
+  /**
+    * User can add comment in the rtl file generated on the top of the component
+    */
+  val rtlComments = new LinkedHashSet[String]()
+  def addComment(comment: String) = rtlComments += comment
```
typingArtist, line 332: `def addComment(comment: String): **Unit**`

core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
```diff
@@ -96,8 +99,10 @@ class PhaseVerilog(pc: PhaseContext, report: SpinalReport[_]) extends PhaseMisc
       () => componentBuilderVerilog.result
     } else {
       emitedComponentRef.put(component, oldComponent)
+      val comments = "\n" + component.rtlComments.map(_.split("\n")).flatten.map(comment => f"$commentSymbol $comment").mkString("\n") + "\n"
+      val str      =  s"$commentSymbol ${component.definitionName} replaced by ${oldComponent.definitionName}\n"
       component.definitionName = oldComponent.definitionName
-      () => s"\n//${component.definitionName} replaced by ${oldComponent.definitionName}\n"
```
typingArtist, line 100: Are you fixing a hidden bug here? Seems that previously, `component.definitionName` was overwritten before use. Now the string is built before overwriting which makes sense. Or, if this wasn’t a bug, perhaps it is one now?

lib/src/main/scala/spinal/lib/bus/amba3/ahblite/AhbLite3Arbiter.scala
```diff
@@ -53,42 +72,71 @@ case class AhbLite3Arbiter(ahbLite3Config: AhbLite3Config, inputsCount: Int, rou
...
+    /** Backup address phase */
+    val addressPhaseData  = Vec(Reg(AhbLite3AddrPhase(ahbLite3Config)), io.inputs.length)
+    val addressPhaseValid = Vec(RegInit(False), io.inputs.length)
+
+    for((input, index) <- io.inputs.zipWithIndex){
+      when(input.HSEL & input.HTRANS === 2 & input.HREADYOUT){
+        addressPhaseData(index).assignFromBus(io.inputs(index))
+        addressPhaseValid(index) := True
+      }
+    }
```
typingArtist, line 86: for consistency with previous widths, use `inputsCount`. Zip the three parts together for simplicity.
```scala
val addressPhaseData = Vec(Reg(AhbLite3AddrPhase(ahbLite3Config)), inputsCount)
val addressPhaseValid = Vec(RegInit(False), inputsCount)

for ((input, data, valid) <- (io.inputs, addressPhaseData, addressPhaseValid).zipped) {
  when(input.HSEL & input.HTRANS === 2 & input.HREADYOUT) {
    data.assignFromBus(input)
    valid := True
  }
}
```

core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
Snoopy87: Good question.. I have no idea.. it is weird because in the PhaseVerilog and PhaseVhdl it is not the same (https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/internals/PhaseVhdl.scala#L81, https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/internals/PhaseVerilog.scala#L99). @Dolu1990 which version is correct ?

.gitignore
Snoopy87: I have no opinion about that part. What do you think @Dolu1990 ?

core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
Dolu1990: Hooo nice catch XD

Yes that was a bug ^^

jijingg:
> Add PhaseInitReg (Init all register uninitiated )
hi @Snoopy87 i have question ,dose this will automatic init all register uninitiated? or just a manually option

.gitignore
Snoopy87: I have configured the global gitignore file 👍

Snoopy87:
> > Add PhaseInitReg (Init all register uninitiated )
> 
> hi @Snoopy87 i have question ,dose this will automatic init all register uninitiated? or just a manually option

It is a manually option. In the SpinalConfig you can add transfomration phases
`
  SpinalConfig(
    mode = VHDL,
    transformationPhases  = ArrayBuffer(new PhaseInitReg())
  ).generate(new `TopLevel)`
`

jijingg:
@Snoopy87,   manually  option is ok .  if you can add report log of uninitiated register it 'will be great .  
because for some team, they prefer to remove reset on data-path for purpose of saving area(there is no functional risk), but on control path must with reset logic, if spinal give report log of uninitiated, the designer can find out which control path without uninitiated.

Snoopy87:
> @Snoopy87, manually option is ok . if you can add report log of uninitiated register it 'will be great .
> because for some team, they prefer to remove reset on data-path for purpose of saving area(there is no functional risk), but on control path must with reset logic, if spinal give report log of uninitiated, the designer can find out which control path without uninitiated.

I have added messages in the PhaseInitReg

lib/src/main/scala/spinal/lib/bus/amba3/ahblite/AhbLite3Arbiter.scala
```diff
@@ -53,42 +72,72 @@ case class AhbLite3Arbiter(ahbLite3Config: AhbLite3Config, inputsCount: Int, rou
...
       maskLocked := maskRouted
       locked     := True
+      addressPhaseValid(requestIndex) := False
 
-      when(io.output.HREADY){ //fire
-        when(io.output.last && !io.output.HMASTLOCK) { //End of burst and no lock
-          locked := False
-        }
+      when(io.output.HREADY & io.output.last && !io.output.HMASTLOCK) { // End of burst and no lock
+        locked := False
```
typingArtist, line 98: I very much dislike this code, both in its old and in its new form. [AMBA 3 AHB-Lite specification,](https://static.docs.arm.com/ihi0033/a/IHI0033A_new_eula.pdf) Section 3.5.2 states that

> Although masters are not permitted to terminate a burst request early, slaves must be designed to work correctly if the burst is not completed.

This means that decoding the `last` information is quite worthless and leading to wrong conclusions in unlocking from a burst. IMHO, the only condition for unlocking is checking for `HMASTLOCK`, and whether `HTRANS` is either `IDLE` or `NONSEQ`. I am aware of the implications on timing, but relying on `last` is improper.

lib/src/main/scala/spinal/lib/bus/amba3/ahblite/AhbLite3.scala
```diff
@@ -138,7 +138,7 @@ case class AhbLite3(config: AhbLite3Config) extends Bundle with IMasterSlave {
   //return true when the current transaction is the last one of the current burst
   def last(): Bool = {
     val beatCounter        = Reg(UInt(4 bits)) init(0)
-    val isLast             = Vec(U"0000", U"0011", U"0111", U"1111")(U(HBURST >> 1)) === beatCounter || (HREADY && ERROR)
+    val isLast             = Vec(U"0000", U"0011", U"0111", U"1111")(U(HBURST >> 1)) === beatCounter || ERROR
```
typingArtist, line 141: `isLast` doesn’t indicate properly a burst of type `INCR`, i.e. a burst of undefined length. In its current form, it takes a burst of type `INCR` just the same like the type `SINGLE` which is quite the opposite of `INCR`.

However, see my other note on using `last()` in the AhbLite3Arbiter for why I see this method useless in the first place.

PR #321, opened 2020-05-17
Axi4/lite simulation tools

c-thaler:
Two little helpers for simulation of Axi4Lite and Axi4 busses.

AxiLite4Driver is a simple driver for read/write access on register interfaces on Axi4Lite.

AxiMemorySim is a memory model for Axi4 buses. Its memory uses the whole bus address range.
Internally, it uses a sparse memory model based on pages. Pages are created on write access only.
Data can be loaded from a file to the memory and saved from memory to file.
It can also simulate bus load by pausing writes/reads with a given probability.

I used this in my current project to provide image data to a image processing core and write the processed data back to a file. There is much room for improvement but it is already a nice little tool.


Dolu1990:
Thanks :)

PR #322, opened 2020-05-20
Added backend support for direct memory read/write

bellaz89:
This PR modifies the Vpi and Verilator simulation backends to allow direct memory read and write access.
The SimRaw interface is extended with the following methods:

```scala
  def getIntMem(signal : Signal, index : Long) : Int
  def getLongMem(signal : Signal, index : Long) : Long
  def setLongMem(signal : Signal, value : Long, index : Long)
  def getBigIntMem(signal : Signal, index : Long) : BigInt
  def setBigIntMem(signal : Signal, value : BigInt, index : Long)
```
to read from an arbitrary memory location. Since no bound check is done in the backend, such a task should be performed by the frontend to avoid segmentation fault if an invalid index value is passed to the methods. Currently GHDL is not supported due to its own limitations in the VPI API (see https://github.com/ghdl/ghdl/issues/237). Multiple memory RW tests/examples are present in TestIVerilog.scala and TestVerilator.scala in sim/src/test/spinal/sim/ .
To make a verilog memory register public in Verilator-generated code, it has to be tagged with `/* verilator public */`


Dolu1990:
All good ^^ ?

bellaz89:
I think so :D

PR #324, opened 2020-05-24
Wave tracing off for sub component

jijingg:
hi @Dolu1990 
now we can closeOff the sub component wave dumping which we don't care.
Usage:
```scala 
SimConfig.compile {
  val dut = new GnssTB(cfg)
  ...
  dut.uAE_Top.uCorrelater.tracingOff()
  dut
}
```
Another good news is that the latest **Verilator4.0.34** have greatly improved on simulation speed.
and also fix FstWave with slowest speed issue . the current sim speed of FST is even faster than VCD. you cant try that.

here is simple compare report for my project case
 
 verilator4.034 | waveSize | simTime | comment      
--------------------------|----------|---------|--------------
 VcdWave                  | 12G      | 463.4s  |  with larger Size
 FstWave                  | 1.1G     | 370.1s  | even fast than VCD
 FstWave(tracingOff sub)   | 12M      | 76.6s   | tracing Off uCorrelatoer

*cpu: Intel-i5*

### DIFF
--- a/core/src/main/scala/spinal/core/Attribute.scala
+++ b/core/src/main/scala/spinal/core/Attribute.scala
@@ -63,5 +63,13 @@ object Verilator{
   object public extends AttributeFlag("verilator public", COMMENT_ATTRIBUTE){
     override def isLanguageReady(language: Language) : Boolean = language == Language.VERILOG || language == Language.SYSTEM_VERILOG
   }
+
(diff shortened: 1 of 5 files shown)


Dolu1990:
Great :) 
Finaly Fst is usable ^^
Thanks !

PR #332, opened 2020-06-20
Fix typo in Phy.scala

roman3017:


roman3017:
This PR is replaced with #334

PR #338, opened 2020-07-25
Regif

jijingg:
@Dolu1990 Some new issue have been fixed. 
@c-thaler  i prefer your previous regif_visitor work should be merged now ，the other improvement should be done step by step, is it ok?


c-thaler:
Well, it's ok. But I'm still working on a kind of merge with the Bus Slave Factory (see [MMSlaveFactory](https://github.com/c-thaler/SpinalHDL/commits/slave_factory))

Dolu1990:
Hooo that's a big PR that i completly forgot about :(

What is the status ?

jijingg:
@Dolu1990 Actually, there are several  feature and branches over there，i use regif branch publishLocal for my local use.  
branchs: 
- binarySystem  (BigInt/Long/Int to HexList/DecList/BinList  Mutual conversion)
- ram  (a memWrapper Solution)
- bits2Bundle  (Bits and Bundle/Vec Mutual conversion )
- regif   (regif some update)

i will add those branch to PR one by one later, there have still some discussion and document should be added .

Dolu1990:
@jijingg 
Ok thanks :)

andreasWallner:
@jijingg I have been using/playing around with the regif implementation for a bit. First off: I think it's a really nice thing and definitely a move in a good direction.

But I have two major gripes with it:

First off the register fields themselves are not typed. In my code (and maybe I'm doing something wrong) I have to use `.asBits()` and `.assignFromBits()` for pretty much every register. For single bits it can also be annoying because when trying to assign a `Bool` have index into the `Bits`, but I can't do so directly as that would bind the to implicit parameter for the name.
My code looked mostly like this then (typed from memory, please excuse syntax errors ;-):
```
val ctrl = factory.newReg(doc="Control...")
val prescaler = ctrl.field(block.io.some_uint.getBitWidth() bits, RW)
block.io.some_uint.assignFromBits(prescaler)
block.io.run := ctrl.field(1 bit, RW)(SymbolName("run"))(0)
val status = factory.newReg(doc="Status...")
status.field(1 bit)(SymbolName("busy"), R) := block.io.run.asBits()
```
In my tries (https://github.com/andreasWallner/SpinalHDL/blob/typed_regif/lib/src/main/scala/spinal/lib/bus/regif/RegInst.scala) I modified the interface such that you pass the type (like the BusSlaveFactory does). Apart from making the code simpler it has to added advantage that the underlying type is then known to the generators (I story the HardType of the given field type). This is especially interesting when generting files for e.g. a debugger or header files for C++ where the type information can be used.

Would you be open to a change on that side? I'd be more than willing to clean up my POC.

The second one is that to me there is currently a gap with providing information about special values / bit meanings as part of the data model, but for that I'll just open a PR once things have settled. I've already implemented that in my POC, but since it would in any way be a backwards incompatible change there is no need to discuss upfront.

Dolu1990:
> whether busif exists independently so that I can close this PR

Is is ok for me :)

jijingg:
@Dolu1990 I finally made up my mind to finish this PR, XD， It's really a little difficult to merge regif into BusSlaveFactor, and some subsequent improvements for regif may not be able to be promoted， as I asked before regif exists independently, I'm glad to get your yes reply.
now it's ready for merge!
the document also have been updated here https://github.com/SpinalHDL/SpinalDoc-RTD/pull/73

Dolu1990:
thanks ^^

PR #340, opened 2020-08-19
Comment rtl entity

Snoopy87:
Comments can be added to the rtl entity


sebastien-riou:
Thanks for this, it is a huge plus for projects which involve some kind of source level audit (source meaning verilog since auditors most likely know only this language). It would be great to get it merged.

Dolu1990:
I completly losed track of this issue XD

So, here the way comments are saved is in their own list. But what's about, instead, we save them as SpinalTag ? that way they could be registred into every SpinalTag ready things (BaseType + Component) ?

Dolu1990:
I pushed two feature on dev : 

```scala
myComponent.addComment("Rawrr") which add comment where it is instanciated
myComponent.definition.addComment("Wuff") which add the comment above the definition of the component
```

wswslzp:
> I pushed two feature on dev :
> 
> ```scala
> myComponent.addComment("Rawrr") which add comment where it is instanciated
> myComponent.definition.addComment("Wuff") which add the comment above the definition of the component
> ```

Are these two features available now in 1.7.0? I try it on the SpinalHDL 1.7 but it doesn't work.

Dolu1990:
Just tried in the SpinalTemplateSbt project : 


```scala
object PlayComment extends App{
  SpinalVerilog(new Component {
    val sub, sub2 = new Component {
      definition.addComment("I'm the definition")
    }

    sub.addComment("I'm sub")
    sub2.addComment("i'm sub2")

    setDefinitionName("miaou")
  })
}
```

It generate

```scala
// Generator : SpinalHDL v1.7.1    git head : 00cf989b959731bd244eca1c43961a3a30c59dff
// Component : miaou
// Git hash  : 00cf989b959731bd244eca1c43961a3a30c59dff

`timescale 1ns/1ps

module miaou (
);


  //I'm sub
  unamed sub (
  );
  //i'm sub2
  unamed sub2 (
  );

endmodule

//unamed replaced by unamed

//I'm the definition
module unamed (
);



endmodule

````

Which is what i'm expecting. what is your config ?

**Comment by wswslzp** [comment]
I add some lines in your example, the comment on both the definition and the instantiation of the module can be generated but the comment on the data inside cannot. 
```scala
object GenBlackBoxTest extends App{
  SpinalVerilog(new Component {

    val sub, sub2 = new Component {
      definition.addComment("I'm the definition")
    }
    val a = out(RegNext(in Bits(3 bit) setName("ia")))
    a.addTag(new CommentTag("this is a"))

    sub.addComment("I'm sub")
    sub2.addComment("i'm sub2")

    setDefinitionName("miaou")
  })

}
```

Dolu1990:
Ahhh yes right, for data it isn't implemented.

Should be easy to add.

PR #341, opened 2020-08-31
Add a blackbox PLLE2_BASE for xilinx series 7 FPGA

DeanoC:
Exposes the PLLE2_BASE PPL to SpinalHDL. Small example at the top of the black box. Has all the parameters (except 1 which Xilinx advise not to use as the tooling will...)


Dolu1990:
thanks ^^

PR #344, opened 2020-09-17
oneFilePerComponent update

jijingg:
@Dolu1990 , here is little two fix.
- 1: oneFilePerComponent Enumdefine generate every .v file will lead EDA-tool `redefine` warning, fixed 
image

- 2: generate filelist


Dolu1990:
Cool thanks :)

Issue #343, opened 2020-09-13
UFix fails on simple test. Any Idea?


yenchih:
I wrote a simple class to test the limit of UFIx and the compiler gave me " assertion failed: Literal 3.75 is to big to be assigned" (test on SpinalHDL v1.4.1 and SpinalHDL v1.4.0). However, the assigned value is the legal maximum value of the defined UFix variable.
``` Scala
class Top extends Component{
    val A = UFix(2 exp, -2 exp)  
    A := 3.75
}
```
No idea why this simple test fails.

philipaconrad:
I was baffled by this, but I tried it out in the Scala console, and traced the error back to `FixedPoint.scala` in the core library.

The issue seems to be caused by a bug in the `UFix` maximum value calculation logic. It looks like a copy/paste bug that went unnoticed during development, as you can see if you look at the `maxValue` methods for `SFix` and `UFix` side-by-side:

 - [SFix definition of `maxValue` on `master`](https://github.com/SpinalHDL/SpinalHDL/blob/3c91edd524fbc70de9e4361101a128f62ffbab38/core/src/main/scala/spinal/core/FixedPoint.scala#L389):
    ```scala
    def maxValue: BigDecimal = Math.pow(2.0, maxExp) * (1.0 - BigDecimal(1.0 / math.pow(2.0, bitCount - 1)))
    ```
 - [UFix definition of `maxValue` on `master`](https://github.com/SpinalHDL/SpinalHDL/blob/3c91edd524fbc70de9e4361101a128f62ffbab38/core/src/main/scala/spinal/core/FixedPoint.scala#L510):
    ```scala
    def maxValue: BigDecimal = Math.pow(2.0, maxExp) * (1.0 - 1.0 / math.pow(2.0, bitCount - 1))
    ```

If we plug in the `peak` and `resolution` constants from your `UFix` instance, we'll get the following:
```scala
scala> Math.pow(2.0, 2.0) * (1.0 - 1.0 / Math.pow(2.0, 4 - 1))
res3: Double = 3.5
```
Which is *not* what is the [docs on SFix/UFix](https://spinalhdl.github.io/SpinalDoc-RTD/SpinalHDL/Data%20types/Fix.html) suggest on the maximum value should be for the `UFix` type.

The docs page says that we should have a max value of `2^peak - 2^resolution`, which *should* be `3.75`, as shown below:

```scala
scala> 4 - (Math.pow(2.0, -2.0))
[...]

PR #348, opened 2020-09-21
Fix UFix maxValue bug from issue #343.

philipaconrad:
This commit fixes a copy/paste bug in the `maxValue` calculation for the
`UFix` type in `FixedPoint.scala`.

The original `maxValue` calculation was copied from `SFix`, and resulted
in spurious assertion failures for users who were trying to assign the
maximum possible value to their `UFix` instances.

The issue was fixed by removing the `- 1` in the `UFix` `maxValue`
calculation that was inherited from the `SFix` `maxValue` calculation
to account for the sign bit.


Dolu1990:
Thanks :)

PR #357, opened 2020-11-26
readStreamNonBlocking: use payload width instead of stream width

distributed:
I verified by looking at the generated output which looks sensible.


Dolu1990:
cool thanks ^^

PR #363, opened 2021-01-09
Little fix

jijingg:
hi @Dolu1990 here is two little fix

1: max bit width 4095 -> 4096
2: regif accesstype **WS** type bug fixed


Dolu1990:
Thanks ^^

PR #364, opened 2021-01-12
Initial pmod support

dotcypress:
Hi there,

This PR contains initial support for [PMODs](https://reference.digilentinc.com/reference/pmod/specification).
A new abstract bundle(PMODBundle) used to describe any PMOD-compatible  interface. This give us  more readable code and ability to port programs between different development boards very easily.

Also a few simple component implementations added for demo purposes (DIP switch PMOD, IceBreaker SnapOff PMOD, Seven segment display PMOD).

See more examples here: https://github.com/dotcypress/icebreaker-fpga-playground/tree/master/src/main/scala/ice/pmods


Dolu1990:
Hi,

Cool thanks ^^ 

- can you move the package into spinal.lib.com.pmod ?
- Should we have all pin defined as bool in the Bundle, or try instead to have them in a single Bits ? (decoupling of the physical layer from the logical layer

dotcypress:
>can you move the package into spinal.lib.com.pmod?

Yep, moved to spinal.lib.com.pmod


>Should we have all pin defined as bool in the Bundle, or try instead to have them in a single Bits ? (decoupling of the physical layer from the logical layer

Oh, didn't think about decoupling on tat level before, thanks!

dotcypress:
@Dolu1990 what do you think about adding other PMOD controllers to the repo? 
Maybe separated library is a better way?

Dolu1990:
> @Dolu1990 what do you think about adding other PMOD controllers to the repo in additional PRs?
Maybe separated library is a better way?

hmm, if you want one possibility is to have a separated repo for now, but which implement the spinal.lib.com.pmod package. And then when it is ready/tested/matured then it can be merged into the SpinalHDL repo (assimulation way XD)

The most important thing is to have the right fundation, as it is always a pain to update afterward :)

dotcypress:
@Dolu1990 gotcha!

What about PMOD interface itself, is it good to merge or you have any comments?

lib/src/main/scala/spinal/lib/com/pmod/PMOD.scala
```diff
@@ -0,0 +1,23 @@
...
+abstract class PMODBundle extends Bundle {
+  val pin1 = Bits(1 bit)
+  val pin2 = Bits(1 bit)
+  val pin3 = Bits(1 bit)
+  val pin4 = Bits(1 bit)
+  val pin7 = Bits(1 bit)
+  val pin8 = Bits(1 bit)
+  val pin9 = Bits(1 bit)
+  val pin10 = Bits(1 bit)
+
```
Dolu1990, line 14: What about instead having Bits(8 bits) ?

lib/src/main/scala/spinal/lib/com/pmod/PMOD.scala
```diff
@@ -0,0 +1,23 @@
...
+  def build()
```
Dolu1990, line 15: What is the build function ? is it to set the direction right ?
What's about using the IMasterSlave interface ?

dotcypress: Make sense. Let me try to do this

lib/src/main/scala/spinal/lib/com/pmod/PMOD.scala
dotcypress: Do we can set direction for each bit  in Bits? We don't know which pins gonna be input and which ping gonna be output here.

lib/src/main/scala/spinal/lib/com/pmod/PMOD.scala
dotcypress: @Dolu1990 done! 
It looks cleaner with IMasterSlave

lib/src/main/scala/spinal/lib/com/pmod/PMOD.scala
Dolu1990: Hoooooooo
That's right, that's a issue :/ Only solution then is to use tristate pins. or as you did having separated signal for each pin.
So i would say let's go you way.

Dolu1990:
Over all, what is the advantage of having a shared PMODBundle over having specific implementation for each usages ?

dotcypress:
@Dolu1990 the main advantage is having easy prototyping with PMOD devices.
For example, I have a [constraint file for IceBreaker](https://github.com/dotcypress/icebreaker-fpga-playground/blob/main/src/main/resources/constraints.pcf) with all pins named regarding PMODBundle naming convention:

```
...
set_io -nowarn io_pmod1b_pin1      43
set_io -nowarn io_pmod1b_pin2      38
set_io -nowarn io_pmod1b_pin3      34
...
```

Now all top modules gonna be really minimalistic and easy to read:

```scala
case class LedMatrix8x8() extends Component {
  val io = new Bundle {
    val pmod2 = master(LedMatrix())
  }

  val matrix = new LedMatrixCtrl()
  io.pmod2 <> matrix.io.pins
}
```

And if you decide to replug your PMOD device in another connector, you must change only 2 lines of code:
```scala
case class LedMatrix8x8() extends Component {
  val io = new Bundle {
    val pmod1 = master(LedMatrix())
  }

  val matrix = new LedMatrixCtrl()
  io.pmod1 <> matrix.io.pins
}
```

Of course, we can achieve that by defining a bundle for each PMOD without a base PMODBundle class, but I think this is a kind of standard connector interface, and having a base PMOD bundle in Spinal lib will be convenient for novices in HDL(like myself 😄 ).

PR #367, opened 2021-01-20
Add support for multi word memory access in BusSlaveFactory

jiegec:
Before this commit, the functions `readSyncMemWordAligned` and `writeMemWordAligned` only support aligned word access. This pr implements two new functions that support multi word memory access (elements of Mem span across several words).


Dolu1990:
thanks ^^

PR #369, opened 2021-01-22
Add workaround for oshi not working under Apple M1

jiegec:
See https://github.com/oshi/oshi/issues/1462 for oshi issue on Apple M1.

By the way, I am curious about why using oshi instead of `Runtime.getRuntime().availableProcessors()`?

### DIFF
--- a/lib/src/main/scala/spinal/lib/eda/bench/Bench.scala
+++ b/lib/src/main/scala/spinal/lib/eda/bench/Bench.scala
@@ -1,9 +1,9 @@
 package spinal.lib.eda.bench
 import spinal.core._
-import java.util.concurrent.ForkJoinPool
-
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks :)

> By the way, I am curious about why using oshi instead of Runtime.getRuntime().availableProcessors()?

Basicaly, sometime it return crap, like 1, seems it depend the system load.

dbwiddis:
FYI: https://groups.google.com/g/jna-users/c/UwUa5qWbIco

jiegec:
> FYI: https://groups.google.com/g/jna-users/c/UwUa5qWbIco

Yes, after the release of new jna and oshi, I will submit a new pr.

PR #371, opened 2021-01-24
WIP: Fix multicycle read: don't overflow counter

jiegec:
In the original implementation, when `cycles = 2`, the counter will wrap around every 2 cycles:

```
valid 010101010101
ready 000000000000
```

However, this can violate the requirement in AXI:

```
The slave can assert the RVALID signal only when it drives valid read data. When asserted, RVALID must remain asserted until the rising clock edge after the master asserts RREADY. Even if a slave has only one source of read data, it must assert the RVALID signal only in response to a request for data.
``` 

If `RREADY` stays low for some cycles, `RVALID` will deassert before the master asserts `RREADY`.

This problem is found when formal verifying an `AXILite4` slave created with `AxiLite4SlaveFactory`.

With this fix, the valid signal would be:

```
valid 01110
ready 0001*
```

But there is a small concern whether this fix might break someone's logic.

EDIT:  ~~I found a problem with this fix. DO NOT MERGE.~~ FIXED.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
+++ b/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
@@ -555,15 +555,22 @@ trait BusSlaveFactory extends Area{
 
   def multiCycleRead(address: AddressMapping, cycles: BigInt): Unit = {
     val counter = Counter(cycles)
+    val reading = False
     onReadPrimitive(
       address       = address,
       haltSensitive = false,
       documentation = null
     ){
-      counter.increment()
+      reading := True
       when(!counter.willOverflowIfInc){
         readHalt()
+        counter.increment()
       }
+    }
+
+    // clear counter before next read
+    when(!reading) {
+      counter.clear()
     }
   }
 


Dolu1990:
hmm the fix seems bugy too, as if two concecutive reads come, the counter isn't flushed, right ?

Dolu1990:
@jiegec 

> We need a extra class to mark the stream as Irrevocable(comes from chisel) for AXI (see rocket chip). Maybe we can call it StrictStream? StrictStream means the Stream has the property above.

I understand the concept, and i agree that Irrevocable and revocable stream are two necessessary things in term of hardware design.
But then about the writting of it, i'm not sure that the Irrevocable typing is more good than harm.
So far, my opinon about it is that it isn't that good as a solution, especialy that the behaviour of some stream logic / stream bridges, even if itself behave as irrevocable, will mutate into revocable depending its inputs. So to me, the proper handeling of that Stream behaviour goes behond what staticaly typing can offer.

> Add formal properties to spinal lib after pr #194 is merged.

Do we realy need formal for this ? I mean a few regular asserts with a few Reg would be enough isn't it ?

PR #372, opened 2021-01-25
fix when some SizeMappings are completely overlap.

pdaxian:
val mappings = Seq(
      SizeMapping(0x0000, 1 KiB),
      SizeMapping(0x0000, 1 KiB),
      SizeMapping(0x0000, 1 KiB)
    )
    println(verifyOverlapping(mappings))

it should be true not false.


Dolu1990:
Thanks ^^

PR #373, opened 2021-01-25
Move doCmd to misc and cleanup duplicate code in spinalhdl.eda.bench

jiegec:
Move duplicated `doCmd` function to Misc.scala and cleanup some unused code.

PR #374, opened 2021-01-26
Add support for existing Quartus project

volatile-static:
The former `QuartusFlow` always creates a new project. So I add a new function to compile and program a configured project. Also has an example usage for Windows ^_^


volatile-static:
conflicts with #373

Dolu1990:
Hoo, #373 no merged :)

PR #375, opened 2021-01-29
Add support for existing Quartus project

volatile-static:
The former `QuartusFlow` always creates a new project. So I add a new class to compile and program a configured project. 
A sample usage in Windows:
```scala
val prj = new QuartusProject(
      quartusPath = "F:/intelFPGA_lite/20.1/quartus/bin64/",
      workspacePath = "G:/"
)
prj.compile()
prj.program()  // automatically find Chain Description File of the project 
```
And print report as well:
```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Area: 2,981 LUT 1,883 FF 
FMax: 115.370000 MHz
```
By the way, the private method `search` will recursive find project files in `workspacePath`.

### DIFF
--- a/lib/src/main/scala/spinal/lib/eda/altera/QuartusFlow.scala
+++ b/lib/src/main/scala/spinal/lib/eda/altera/QuartusFlow.scala
@@ -33,21 +33,20 @@ object QuartusFlow {
     return fMax*1e6
   }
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks ^^

PR #376, opened 2021-01-29
Apb3businterface bugfix

pdaxian:
Apb3BusInterface bugfix
if use readSync, the PRDATA and PSLVERR were delayed one beat, but the PREADY didn't halt. so the master recivered the wrong data.
 
register the PWDATA/askWrite/askRead to improve timing closure.
if read sync , one read cost 3 beats, else 2 beats.

and add a  testbench.


Dolu1990:
Miaou ?

PR #377, opened 2021-01-31
RegIf Apb3BusInterface readsync bugfix, improve timing closure.

pdaxian:
Apb3BusInterface bugfix
if use readSync, the PRDATA and PSLVERR were delayed one beat, but the PREADY didn't halt. so the master recivered a wrong data.

register the PWDATA/askWrite/askRead to improve timing closure.
if read sync , one read cost 3 beats, else 2 beats.

and add a testbench.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/Apb3BusInterface.scala
+++ b/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/Apb3BusInterface.scala
@@ -1,8 +1,14 @@
 package spinal.lib.bus.regif
 
 import spinal.core._
-import spinal.lib.bus.amba3.apb.Apb3
+import spinal.lib.bus.amba3.apb.{Apb3, Apb3Config}
 import spinal.lib.bus.misc.SizeMapping
+import spinal.lib.slave
+import spinal.core.sim._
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

> register the PWDATA/askWrite/askRead to improve timing closure.

This can be an option, but shouldn't be the default, as it consume area. Also, most pipelinings can generaly be inserted outside the slave logic in a pipelining bridge to keep things as simple and modular as possible.

> val askWrite: Bool = RegNext(bus.PSEL(selId) && !bus.PENABLE && bus.PWRITE)

I'm afraid it will create a bug in case writeHalt is used, same for askRead.
Basicaly it will lose the askWrite/askRead condition if the user halt the bus.

pdaxian:
Thanks,I learned a little more.
I am working on FPGA and never think about the register area.

You are right ,the askWrite/askRead will be missed when writeHalt/readHalt.  
But the askWrite/askRead isn't used in the RegIf package.
The RegBase class use `doRead` as read enable, which is combined from PREADY is True. It's a problem when use readSync. 
I think it should use askRead instead of doRead as read enable, but I haven't build test bench for other bus interface.

I'll close the PR,and ask jijing to fix it. ^ ^

Dolu1990:
No worries :)

PR #378, opened 2021-02-02
Add support for multiple rtl file benchmark

jiegec:
Extend the original `Rtl` trait for multiple rtl files.

 I only have Vivado in my environment..


Dolu1990:
thanks ^^

PR #379, opened 2021-02-04
fix 3 files

volatile-static:
1. fix a syntax error in `readme.md` sbt example. only add a `,` 😂 
2. last week I added a class `QuartusProject` but I find that user cannot access it. So I add an object.
and then we can use in this way:
```scala
QuartusProject(...).compile().program()
```
3. the `UartCtrlUsageExample` maybe too old to use. I add an object for simple usage and update the example.
I think it's more friendly to starters and very convenient to debug FPGA. For example, when you need to read data from single RX pin:
```scala
val uartCtrl: UartCtrl = UartCtrl(
    config = UartCtrlInitConfig(
      baudrate = 921600,
      dataLength = 7,  // 8 bits
      parity = UartParityType.NONE,
      stop = UartStopType.ONE
    ),
    readonly = true
  )
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,7 @@ scalaVersion := "2.11.12"
 
 libraryDependencies ++= Seq(
   "com.github.spinalhdl" % "spinalhdl-core_2.11" % "latest.release",
-  "com.github.spinalhdl" % "spinalhdl-lib_2.11" % "latest.release"
-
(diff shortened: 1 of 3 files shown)


volatile-static:
And I forgot `object QuartusProject`.
User cannot access it in release 1.4.3, sorry.

Dolu1990:
Thanks ^^

PR #381, opened 2021-02-21
Fix gitHash when building as a subproject

kleinai:
This is a small change that lets git find the current hash when building as a subproject with a different working directory.

For example I use this in my project's `build.sbt` when working from sources:
```
lazy val spinalHdlIdslPlugin = ProjectRef(file("./ext/SpinalHDL"), "idslplugin")
lazy val spinalHdlCore = ProjectRef(file("./ext/SpinalHDL"), "core")
lazy val spinalHdlLib = ProjectRef(file("./ext/SpinalHDL"), "lib")

dependsOn(spinalHdlCore,spinalHdlLib,spinalHdlIdslPlugin)

scalacOptions += (artifactPath in(spinalHdlIdslPlugin, Compile, packageBin)).map { file =>
  s"-Xplugin:${file.getAbsolutePath}"
}.value
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -133,6 +133,7 @@ lazy val core = (project in file("core"))
     version := SpinalVersion.core,
     sourceGenerators in Compile += Def.task {
       val dir = (sourceManaged in Compile).value
+      dir.mkdirs()
       val file = dir / "Info.scala"
       IO.write(file, """package spinal.core
                        |object Info {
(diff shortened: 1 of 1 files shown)


Dolu1990:
travis isn't happy :(

Dolu1990:
Magnificent ^^

PR #382, opened 2021-02-22
Add inc and dec for UInt.

volatile-static:
An elegant way to describe `a:=a+b`.
Like Pascal.
- `a` must be `Reg`

### DIFF
--- a/core/src/main/scala/spinal/core/UInt.scala
+++ b/core/src/main/scala/spinal/core/UInt.scala
@@ -64,7 +64,7 @@ class UInt extends BitVector with Num[UInt] with MinMaxProvider with DataPrimiti
 
   /**
     * Concatenation between two UInt
-    * @example{{{ val myUInt = uInt1 @@ uInt2 }}}
+    * @example {{{ val myUInt = uInt1 @@ uInt2 }}}
     * @param that an UInt to append
     * @return a new UInt of width (w(this) + w(right))
     */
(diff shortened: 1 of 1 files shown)


Dolu1990:
Why not a :+=  :-= instead ?

volatile-static:
I think `:+=` looks strange...
Is this ok?
```scala
def :+=(that: UInt): Unit = {
      assert(isReg)
      this := this + that
}
```
Yet how about `:+^=` and `:+|=` ?

> `:=+` looks sexy 😂  hahaha...

volatile-static:
Don't `a.inc` or `a.incCarry(b)` or `b.dec(2)` more reasonable?

volatile-static:
```scala
val ui: UInt = UInt(8 bits) setAsReg() init 0
    when(True) {
      ui inc()
    } otherwise  {
      ui dec 2
    }
```
This code generated:
```verilog
always @ (posedge clk or posedge reset) begin
    if (reset) begin
      ui <= 8'h0;
    end else begin
      if(1'b1)begin
        ui <= (ui + 8'h01);
      end else begin
        ui <= (ui - 8'h02);
      end
    end
```
`inc()`like`setAsReg()` while `dec 2` like `init 0`
As far as I am concerned, `inc` is better than `:+=`. What do you think? @Dolu1990

Dolu1990:
> Yet how about :+^= and :+|= ?

So, those two do not apply to an increment operation, as anyway the number of bits has to be the same than the left operande.

I would realy prefer : 

```scala
val ui: UInt = Reg(UInt(8 bits)) init(0)
    when(True) {
      ui :+= 1
    } otherwise  {
      ui :-= 2
    }
```

The problem with inc / dec keyword is that they do not disociate themself from others identifiers, while :+=  and :-=  clearly show themself as operators :)

volatile-static:
OK. I agree with you.

PR #383, opened 2021-02-27
Pass simulator flags to iverilog and ghdl

jiegec:
Currently, user can pass simulator flags via `addSimulatorFlag` function, but only iverilator backend receives the flags https://github.com/SpinalHDL/SpinalHDL/blob/adf552d8f500e7419fff395b7049228e4bc5de26/sim/src/main/scala/spinal/sim/VerilatorBackend.scala#L459.

This pr passes `simulatorFlags` to iverilog and ghdl backend as well.


Dolu1990:
Nice thanks ^^

PR #387, opened 2021-03-04
Add structs and records

kleinai:
Add SystemVerilog Struct and VHDL Record under a unified SpinalStruct datatype.

Current state is a hack to get some limited record support in VHDL. The plan is to integrate changes deeper in the statement and expression logic.

This is not meant to advocate using records and structs over bundles but to provide support for blackboxes which do use records or structs.

### DIFF
--- a/core/src/main/scala/spinal/core/Data.scala
+++ b/core/src/main/scala/spinal/core/Data.scala
@@ -255,6 +255,7 @@ trait Data extends ContextUser with NameableByComponent with Assignable with Spi
 
   private[core] var dir: IODirection = null
   private[core] def isIo = dir != null
+  private[core] def isSuffix = parent != null && parent.isInstanceOf[Suffixable]
 
   var parent: Data = null
   def getRootParent: Data = if(parent == null) this else parent.getRootParent
--- a/core/src/main/scala/spinal/core/Struct.scala
+++ b/core/src/main/scala/spinal/core/Struct.scala
@@ -1,0 +1,178 @@
+package spinal.core
+
+import spinal.core.internals.TypeStruct
+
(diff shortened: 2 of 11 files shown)


Dolu1990:
Hooo that's quite a big one ^^

Do you have any example how to use it ?

kleinai:
The idea is to have structs match bundles as much as possible. Although my work doesn't extend Bundle presently. I'm still learning how Spinal works under the hood.
```
case class MyStruct extends SpinalStruct("some_existing_struct") {
   val a = Bool()
   val c = Bits(8 bit)
}
```
An optional type name can be provided if for instance the struct is already in an external library, or you just want a different name (like maybe snake case).

Dolu1990:
> This is not meant to advocate using records and structs over bundles but to provide support for blackboxes which do use records or structs.

I think that's right, don't bother trying to cover any other usage than from blackbox, as else it will become quite difficulte and hellish XD

kleinai:
Took a few tried to roll back the complex commits correctly...

> I think that's right, don't bother trying to cover any other usage than from blackbox, as else it will become quite difficulte and hellish XD

I realized a true structs implementation is not going to be easy.

Dolu1990:
> I realized a true structs implementation is not going to be easy.

Right, also, there is so many thing that you can't do with Struct / records in VHDL and verilog, that's just not possible. ^^

> Took a few tried to roll back the complex commits correctly...

To roll back to blackbox support only ?

kleinai:
> To roll back to blackbox support only ?

Yes, I started delving far too deep.

Dolu1990:
Thanks ^^

PR #391, opened 2021-03-23
Wrap JDK includes in quotes to allow paths with spaces

JobNoorman:
When compiling a design using Verilator, JDK include paths are not wrapped in quotes causing problems when the JDK is installed in a path with spaces (which might be the case, for example, when using IntelliJ on macOS). This PR fixes this.


Dolu1990:
Thanks <3

PR #393, opened 2021-03-24
verilog blackbox generic parameter BigInt support

jijingg:
@Dolu1990 verilog could suport bigInt like  with format as `65'h1ffff_ffff_ffff_ffff`, so add BigInt support 
```scala
  val generic = new Generic{
    val PHV_1B_NUM = c.phvc.cB1Num
    val TEST  = BigInt("ffff1234ff",16)
  }
```

```verilog
  mybb #(
    .PHV_1B_NUM(64),
   ...
    .TEST(40'hffff1234ff) 
  )  uutmybb (   
```


Dolu1990:
Thanks ^^

PR #395, opened 2021-03-27
add altera virtual-jtag elements

michg:
Hello Charles,
this pull request adds the equivalent of the xilinx BSCANE2 element for the virtual jtag element from altera.
You can have a look at https://github.com/michg/pyocdriscv32 at the saxon target.
Regards,
Michael


Dolu1990:
Thanks <3

Issue #394, opened 2021-03-25
Axi4Decoder: WriteOnly: missing a conditional check to avoid deref NULL pointer


lastweek:
Hi @Dolu1990,

I'm using the `Axi4Crossbar` and found when the slave address mapping covers the whole space, there will be a NULL pointer error in `Axi4WriteOnlyDecoder`.

Using whole address space is enforced due to this line: https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Decoder.scala#L29

The error line is:
https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Decoder.scala#L106
```scala
io.input.writeData.ready := ((pendingSels & io.outputs.map(_.writeData.ready).asBits).orR || (pendingError && errorSlave.io.axi.writeData.ready)) && allowData
```

It should have a `if (decodingErrorPossible)` like this line does: https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Decoder.scala#L33

Or am I missing something? Looking forward to your reply

---

My code
```scala
  val axiCrossbar = Axi4CrossbarFactory()
  axiCrossbar.addSlaves (
    parsers(0).io.axi   -> (0, 0x4000),
    parsers(1).io.axi   -> (0x4000, 0xC000)
  )
  axiCrossbar.addConnections (
    io.axi   -> List(parsers(0).io.axi, parsers(1).io.axi)
  )
  axiCrossbar.build()
```

The error msg:
```
ERROR !                                             
A null pointer access has been detected in the JVM.                                                      
This could happen when in your SpinalHDL description, you access an signal which is only defined further.
For instance :                                      
  val result = Bool                                 
  result := a ^ b  //a and b can't be accessed there because they are only defined one line below (Software rule of execution order)
  val a,b = Bool                                    
                                                                                                
[info]   java.lang.NullPointerException:            
[info]   at spinal.lib.bus.amba4.axi.Axi4WriteOnlyDecoder.<init>(Axi4Decoder.scala:97)                   
[info]   at spinal.lib.bus.amba4.axi.Axi4CrossbarFactory$$anonfun$2$$anon$2.<init>(Axi4Crossbar.scala:142)                               
[...]

PR #396, opened 2021-03-29
Axi4Decoder: fix null pointer in Axi4WriteOnlyDecoder

lastweek:
In Axi4WriteOnlyDecoder, when the aggregated slave address mapping
is smaller than what the address bits can cover, the errorSlave
variable will be set, otherwise it will be NULL.

The original code does not check whether this errorSlave is NULL
or not. Therefore, in some cases, the code will not be able to compile.


lastweek:
Originally mentioned in https://github.com/SpinalHDL/SpinalHDL/issues/394

Dolu1990:
Thanks :)

PR #401, opened 2021-04-07
Fix signal names when wrapping suffixed sub inputs

kleinai:
Small bug fix, I missed wrapped suffixed (structs) subcomponent inputs during my initial development.


Dolu1990:
thanks ^^

PR #415, opened 2021-05-05
one line bug fix for AxiMemorySim

pwang7:
The bug is that when AxiMemorySim handing read channel, AXI slave should wait for the RREADY signal from AXI master before sending next data.


Dolu1990:
Thanks ^^

PR #423, opened 2021-05-21
Allow enum.toEnum to work in simulation with custom encodings

C-Elegans:
When using a SpinalEnum type with a custom encoding, such as

```object TransactionSize extends SpinalEnum{
  val BYTE       = newElement()
  val HALFWORD   = newElement()
  val WORD       = newElement()
  val DOUBLEWORD = newElement()
  val QUADWORD   = newElement()
  defaultEncoding = SpinalEnumEncoding("myEncoding")(
    BYTE -> 1,
    HALFWORD -> 2,
    WORD -> 4,
    DOUBLEWORD -> 8,
    QUADWORD -> 16)
}```

it's impossible to access the enum value using .toEnum in simulation as it gives a not implemented error.

This PR has each enum type store a mapping between encoded enum values and enum elements, so that the enum element can be easily retrieved during simulation.

### DIFF
--- a/core/src/main/scala/spinal/core/Enum.scala
+++ b/core/src/main/scala/spinal/core/Enum.scala
@@ -82,6 +82,7 @@ class SpinalEnum(var defaultEncoding: SpinalEnumEncoding = native) extends Namea
 
   /** Contains all elements of the enumeration */
   @dontName val elements = ArrayBuffer[SpinalEnumElement[this.type]]()
+  @dontName var encodingMap = Map[BigInt, SpinalEnumElement[this.type]]()
 
   def apply() = craft()
   def apply(encoding: SpinalEnumEncoding) = craft(encoding)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/src/main/scala/spinal/core/Enum.scala [hunk 1]
```diff
@@ -82,14 +82,21 @@ class SpinalEnum(var defaultEncoding: SpinalEnumEncoding = native) extends Namea
 
   /** Contains all elements of the enumeration */
   @dontName val elements = ArrayBuffer[SpinalEnumElement[this.type]]()
+  @dontName var encodingMap = Map[BigInt, SpinalEnumElement[this.type]]()
```
Dolu1990, line 85: This is problematic, basicaly the SpinalEnum class is a kind of template which has a recerence to a default encoding.

Only encodings themself know the mapping between a given element and its hardware values, not the SpinalEnum itself.

I will give a look now

Dolu1990:
Fixed ^^

Tested with : 

```scala
object PlayBugEnumSpinalSim extends App{
  import spinal.core.sim._
  val MyEnum = new SpinalEnum{
    val A, B, C, D = newElement()
    defaultEncoding = SpinalEnumEncoding("miaou")(
      A -> 1,
      B -> 2,
      C -> 4,
      D -> 8
    )
  }

  SimConfig.compile(new Component {
    val input = in Bits(4 bits)
    val output = out(MyEnum)
    output.assignFromBits(input)
  }).doSim{dut =>
    dut.input #= 2
    sleep(1)
    println(dut.output.toEnum.getName())
  }

  SimConfig.compile(new Component {
    val input = in(MyEnum)
    val output = out(Bits(4 bits))
    output := input.asBits
  }).doSim{dut =>
    dut.input #= MyEnum.C
    sleep(1)
    println(dut.output.toInt)
  }
}
``` 

Thanks anyway for the PR ^^

PR #424, opened 2021-05-22
Wrap Axi4Crossbar in a component to declutter HDL output

kleinai:
I don't know how useful this might be but it does make the generated HDL far cleaner, especially with one component per file switched on.

This probably isn't the most efficient way to do this by using an anonymous class but was easier to write.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Crossbar.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Crossbar.scala
@@ -121,157 +121,218 @@ case class Axi4CrossbarFactory(/*decoderToArbiterConnection : (Axi4Bus, Axi4Bus)
         onThat.setCompositeName(bus.component,bus.getName() + "_" + name)
     }
 
-    val decoders = for(master <- masters) yield master match {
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

Sorry for the delay, i losed the track of issues :(

Did you tested it on some SoC ? (ex : Briey) ?

About all the port.setName("axi_m_ro_%d".format(id))

What's about using the name of the original master port ?

kleinai:
I haven't run this against any of the tests yet. This was more of a request for comment.

Not sure what I was thinking with the port renaming. Seems to work fine if I copy the port names.

I think I'll remove the anonymous `Component` and do something similar to `StreamArbiterFactory`.

Dolu1990:
> This was more of a request for comment.

Hoo, so, can you describe the issue, i mean, more precisely ?
Which aspects of the current implementation need declutter ? (generated hardware / scala description)

kleinai:
I feel a bit that "heavy" functions such as the AXI crossbar shouldn't be pushed directly into the current scope. I get that areas can help a little but even then the HDL output can get pretty unwieldly. I do understand that human readable HDL isn't exactly a main goal of the project but even a little bit of componentization could help, especially with more traditional HDL analysis.

PR #428, opened 2021-05-25
Add bounds checking to sim memory accesses

C-Elegans:
Attempting to access an out of bounds address in a SpinalHDL Mem object causes a segmentation fault which is rather hard to debug. This adds bounds checking to Mem accesses from simulation to make it easier to debug this.


Dolu1990:
Thanks ^^

PR #433, opened 2021-06-01
Add handling of atrributes with integer argument

nbstrike:
Some attributes require an integer argument value (eg Vivado CASCADE_HEIGHT; Quartus max_fan), allow SpinalHDL to handle these attributes.


Dolu1990:
Thanks ^^

PR #437, opened 2021-06-04
Allow Verilog emitter to render MaskedLitterals as case items into a pure case statement

nbstrike:
This change allow the Verilog emitter to render:
```scala
switch(cond1 ## cond2 ## cond3)
  is(M"1--") { }
}
```
more idiomatically, with casez & ? for don't care bits:
```
casez(...)
  3'b1??: ...
endcase
```

As opposed to the chain of if/else that it would currently generate

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -627,7 +627,18 @@ class ComponentEmitterVerilog(
             lastWhen = treeStatement
             statementIndex = emitLeafStatements(statements,statementIndex, scopePtr, assignmentKind,b, tab + "  ")
           case switchStatement : SwitchStatement =>
-            val isPure = switchStatement.elements.foldLeft(true)((carry, element) => carry && element.keys.forall(_.isInstanceOf[Literal]))
+            def checkPure(o : Any): Boolean = o match {
+              case l: Literal => true
+              case k: SwitchStatementKeyBool => k.key != null
+              case _ => false
+            }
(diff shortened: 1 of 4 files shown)


typingArtist (commented):
Nice change!

nbstrike:
Fixed both items in update.

typingArtist (approved):
LGTM

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -627,17 +627,17 @@ class ComponentEmitterVerilog(
...
             }
-            def checkMaskedLiteral(o : Any): Boolean = o match {
+            def checkMaskedLiteral(o : Expression): Boolean = o match {
               case k: SwitchStatementKeyBool => k.key != null
               case _ => false
             }
-            val hasMaskedLiterals = switchStatement.elements.exists(_.keys.exists((k)=>checkMaskedLiteral(k)))
-            val isPure = switchStatement.elements.forall(_.keys.forall(checkPure(_)))
+            val hasMaskedLiterals = switchStatement.elements.exists(_.keys.exists(checkMaskedLiteral))
+            val isPure = switchStatement.elements.forall(_.keys.forall(checkPure))
```
typingArtist, line 641: For beauty we could do

```scala
val allExpressions = switchStatement.elements.flatMap(_.keys)
val hasMaskedLiterals = allExpressions.exists(checkMaskedLiteral)
val isPure = allExpressions.forall(checkPure)
```

Dolu1990:
Thanks  ^^

PR #440, opened 2021-06-07
Specifying the file path when generating Qsys .tcl

volatile-static:
```scala
QSysify(SpinalVerilog(new TopLevel).toplevel, "../")
```
The Quartus Platform Designer requires that ".tcl" and ".v" in the same path.


volatile-static:
These changes work well on my Quartus projects.
But I don't know why `The Travis CI build failed`.

Dolu1990:
Thanks ^^

Basicaly, the travis stuff always fail because of docker changing their free services amounts.

@wifasoi is currently migrating things away from docker.

PR #402, opened 2021-04-10
Add writeByteEnable to BusSlaveFactory and support unaligned access in read/writeMultiWord for AXI4

jiegec:
Currently, BusSlaveFactory does not support masked write. This pr adds preliminary support to it. I am working on a Axi4BusSlaveFactory.

By the way, the code style in the repo is pretty inconsistent... I wish they are described and enforced by scalafmt.conf.

PR #441, opened 2021-06-08
Fix AvalonMMSlaveFactory when no byteEnable signal exists.

nbstrike:
#402 causes a null pointer for an AvalonMM bus that is configured without byteEnable (signal is of course missing) -- this will correct that problem.


Dolu1990:
That's funny, i just had a kind of similar issue yesterday.

I made the BusSlaveFactory happy with def writeByteEnable(): Bits = null 

So my guess is that we can close this PR, as with the very recent SpinalHDL:dev upstream it should be fine already ?

Dolu1990:
(basicaly now, writeByteEnable is implemented by default as a null, and the generation is happy with it)

nbstrike:
Thanks, I just tested that and can confirm that I don't experience the problem when SpinalHDL is built from ed3aac7c4ed7c50ca6de9b533668dc9bd6940ec1

PR #444, opened 2021-06-14
added some documentation for the PLIC mapper, and some documentation string

allexoll:
as i said on gitter, here is some doc for the PLIC mappter.

I'm adding the doc string because i'd like at some point to add a SVD generator, similar to the BSP generator. having some doc for that will help autogenerate the description.

remarks welcome, this is just a proposal for this documentation, i'm fine changing it as you see fit


Dolu1990:
Thanks :D

Dolu1990:
RGB XD

PR #449, opened 2021-06-17
Fixed sbt clean publishLocal issues on clean dev branch in centos

kartikp4892:


jiegec:
It `-c` correct?

```
usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]
```

kartikp4892:
Ohh `-c` is different than `-C`.. Not sure why on my Centos 7 `git -C` is not recognized. My git version is `1.8.3.1`

kartikp4892:
But weird thing is `-c` works at my end. Git or Spinal doesn't complain about `-c` at all.

andreasWallner:
Checked this on my machines (Windows & Ubuntu  21): both are running git 2.30, using the uppercase version. I'd guess it's related to the git version that you are running.

Dolu1990:
So, -C is correct, but not -c right ?

kartikp4892:
Yes, After upgrading to the git version it started working. So this was git version issues. I will close the request now.

PR #456, opened 2021-06-22
Cross scala versions build

jiegec:
This pr uses sbt cross building feature: https://www.scala-sbt.org/1.x/docs/Cross-Build.html.


Dolu1990:
<3 <3

PR #457, opened 2021-06-27
added more useful error message

charliehorse55:
Makes it more obvious what you did wrong if you don't include mode in spinalConfig. I wasn't sure how SpinalHDL handles errors, so I just made an exception.


Dolu1990:
Thanks ^^

PR #459, opened 2021-06-30
Use Gpio readBufferLength

andreasWallner:
The readBufferLength parameter was unused but present up to now.


Dolu1990:
Thanks ^^

PR #462, opened 2021-07-05
add toggleWhen to Bool

typingArtist:


Dolu1990:
The issue with that design, is that the assignement override the previous value, even when there is no toggle.

What's about :

```scala
when(cond){
   this := !this
}
```

Dolu1990:
Thanks ^^

PR #464, opened 2021-07-06
Improve PulseCCByToggle, Unify StreamCCByToggle and StreamCCByToggleWithoutBuffer

typingArtist:
makes use of new APIs ```clock on new Area {}``` and ```toggleWhen()``` and ```edge()```.

I don’t think that the added ```RegNext(pushArea.data)``` for the ```stream.payload``` in case of ```withInputWait = true``` warrants another pipeline stage on ```popArea.target```’s ```BufferCC```, as it was there in the old ```StreamCCByToggleWithoutBuffer``` implementation. As the correlated control signal ```popArea.target``` is at least one ```outputClock``` cycle slower than the always registered data in ```stream.payload```, it will for sure not be losing the payload update. At least it arrives in the same clock cycle as the data.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -202,7 +202,7 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
   }
 
   def ccToggleWithoutBuffer(pushClock: ClockDomain, popClock: ClockDomain): Stream[T] = {
-    val cc = new StreamCCByToggleWithoutBuffer(payloadType, pushClock, popClock).setCompositeName(this,"ccToggle", true)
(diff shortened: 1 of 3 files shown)


lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -1166,92 +1166,54 @@ object StreamCCByToggle {
...
     stream.valid := (target =/= hit)
-    stream.payload := pushArea.data
-    stream.payload.addTag(crossClockDomain)
 
-    when(stream.fire) {
-      hit := !hit
+    if (!withInputWait) {
+      stream.payload := pushArea.data
+    } else {
+      stream.payload := RegNext(pushArea.data)
```
Dolu1990, line 1209: My guess is that both could be  stream.payload := pushArea.data, and that the bufferCC(xx, 3) of the previous implementation was superflux (could have been 2)

Else all good :D

PR #477, opened 2021-07-20
add cache for verilator binaries

### LINKED ISSUE #475: Feature request: Reuse compiled binaries
I'm working on a large project where the verilator compilation needs up to 30 minutes. Often I only make changes to the test - leaving the main source code unchanged. SpinalHDL always recompiles everything though it isn't needed (I guess).

It would be useful if SpinalHDL would automatically reuse the compiled binaries. Alternatively it would be helpful I there were a way to target a directory containing the compiled binaries which are then used. Then one could run the test normally in the first place and then set the option targeting the directory containing the binaries.

jonnykl:
Creates a cache for the binaries generated by verilator. The hash of the verilator options and the source file contents will be saved and compared before verilator gets invoked. If the hashes equal, the binaries are copied from the cache and verilator compilation will be skipped.

This works for me for some components. SpinalHDL does not always generate the same RTL given the same scala source code. In this case the cached is not used (because the hash always changes). A possible fix (besides fixing SpinalHDL to generate the same RTL given the same scala source) would be to use the scala source code for the hash generation, but this would be more complicated (how to get the corresponding scala source filenames?) and would maybe introduce some problems (for some reason SpinalHDL should generate not the same RTL, then the cached binaries will be used though verilator should be invoked).

Tested on Linux (Ubuntu). Should also be tested on Windows and Mac.

Fix #475

### DIFF
--- a/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
@@ -506,7 +506,7 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
       case false => null
     }
 
-    // skip verilator compilation if nothing changed (hashed equal)
+    // skip verilator compilation if nothing changed (hashes equal)
     if (hash == previousHash) {
       val workspaceDir = new File(s"${workspacePath}/${workspaceName}")
       val backupDir = new File(s"${cachePath}/${workspaceName}")
--- a/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
@@ -509,11 +509,11 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
     // skip verilator compilation if nothing changed (hashes equal)
     if (hash == previousHash) {
       val workspaceDir = new File(s"${workspacePath}/${workspaceName}")
-      val backupDir = new File(s"${cachePath}/${workspaceName}")
+      val cacheDir = new File(s"${cachePath}/${workspaceName}")
 
-      if (backupDir.exists()) {
+      if (cacheDir.exists()) {
         FileUtils.deleteQuietly(workspaceDir)
-        FileUtils.moveDirectory(backupDir, workspaceDir)
+        FileUtils.moveDirectory(cacheDir, workspaceDir)
 
         println("[info] Verilator options and sources unchanged, using cached binaries")
         return
--- a/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
@@ -7,6 +7,7 @@ import net.openhft.affinity.impl.VanillaCpuLayout
 import org.apache.commons.io.FileUtils
 import java.security.MessageDigest
 
+import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.util.Random
@@ -35,6 +36,10 @@ class VerilatorBackendConfig{
 }
 
 
+object VerilatorBackend {
+  private val cachePathLockMap = mutable.HashMap[String, Object]()
+}
+
 class VerilatorBackend(val config: VerilatorBackendConfig) extends Backend {
   import Backend._
 
@@ -45,13 +50,30 @@ class VerilatorBackend(val config: VerilatorBackendConfig) extends Backend {
   val wrapperCppName = s"V${config.toplevelName}__spinalWrapper.cpp"
   val wrapperCppPath = new File(s"${workspacePath}/${workspaceName}/$wrapperCppName").getAbsolutePath
 
+  def cacheSynchronized(function: => Unit): Unit = {
+    if (cacheEnabled) {
+      var lock: Object = null
+      VerilatorBackend.cachePathLockMap.synchronized {
+        lock = VerilatorBackend.cachePathLockMap.getOrElseUpdate(cachePath, new Object())
+      }
+
+      lock.synchronized {
+        function
+      }
+    } else {
+      function
+    }
+  }
+
   def prepareCache(): Unit = {
     if (!cacheEnabled) {
       return
     }
 
-    (new File(cachePath)).mkdirs()
-    assert((new File(cachePath)).isDirectory(), s"${cachePath} could not be created or is not a directory")
+    cacheSynchronized {
+      (new File(cachePath)).mkdirs()
+      assert((new File(cachePath)).isDirectory(), s"${cachePath} could not be created or is not a directory")
+    }
   }
 
   def clean(): Unit = {
@@ -456,6 +478,8 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
 
 
     val verilatorBinFilename = if(isWindows) "verilator_bin.exe" else "verilator"
+
+    // when changing the verilator script, the hash generation (below) must also be updated
     val verilatorScript = s""" set -e ;
        | ${verilatorBinFilename}
        | ${flags.map("-CFLAGS " + _).mkString(" ")}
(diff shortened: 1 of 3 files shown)


Dolu1990:
Hi,

Cool :D

> This works for me for some components. SpinalHDL does not always generate the same RTL given the same scala source code. In this case the cached is not used (because the hash always changes)

This happen when :
- By mistake in SpinalHDL i used a set / map data structed which is underdered. If you have the issue, please let's me know, and i can try to track down where in SpinalHDL itself the mistake was done
- In the user code some HashMap/Set are used

Easy workaround is to use the LinkedHashSet and LinkedSet.

I'm a bit in a rush, i will check soon :)

jonnykl:
> * In the user code some HashMap/Set are used

Yep, fixed but still having the problem.

> * By mistake in SpinalHDL i used a set / map data structed which is underdered. If you have the issue, please let's me know, and i can try to track down where in SpinalHDL itself the mistake was done

The following scala code does not always generate the same verilog code:

```scala
import spinal.core._
import spinal.lib._
import spinal.lib.fsm._


case class Fsm() extends StateMachine {
  val state: State = new State with EntryPoint {}
}


case class Foo() extends Component {
  val fsmA = Fsm()
  val fsmB = Fsm()

  val fsm = new StateMachine {
    val stateA: State = new StateFsm(fsmA) with EntryPoint {}
    val stateB: State = new StateFsm(fsmB) {}
  }
}

object Foo {
    SpinalConfig(targetDirectory="rtl-gen").generateVerilog(Foo())
  }
}
```

For example these are the first non-empty and non-comment lines (there are more different lines):

Version 1:
```
`define fsmB_enumDefinition_binary_sequential_type [0:0]
`define fsmB_enumDefinition_binary_sequential_fsmB_BOOT 1'b0
`define fsmB_enumDefinition_binary_sequential_fsmB_state 1'b1

`define fsmA_enumDefinition_binary_sequential_type [0:0]
`define fsmA_enumDefinition_binary_sequential_fsmA_BOOT 1'b0
`define fsmA_enumDefinition_binary_sequential_fsmA_state 1'b1
```

Version 2:
```
`define fsmA_enumDefinition_binary_sequential_type [0:0]
`define fsmA_enumDefinition_binary_sequential_fsmA_BOOT 1'b0
`define fsmA_enumDefinition_binary_sequential_fsmA_state 1'b1

`define fsmB_enumDefinition_binary_sequential_type [0:0]
`define fsmB_enumDefinition_binary_sequential_fsmB_BOOT 1'b0
`define fsmB_enumDefinition_binary_sequential_fsmB_state 1'b1
```

The defines are generated here: https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/internals/PhaseVerilog.scala#L147

And `enums` is defined as `LinkedHashMap[SpinalEnum, LinkedHashSet[SpinalEnumEncoding]]` here: https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/internals/Phase.scala#L46 

So using `LinkedHashMap`/`LinkedHashSet` is not enough. I did not analyze the problem further but the enums must have been filled in in a different order. Maybe from multiple concurrent threads?

Dolu1990:
<3
Thanks, fixed with the above commit, that was a Set not being a LinkedSet in the StateMachine builder

jonnykl:
I should have opened a new issue. Could you reopen this PR?

Dolu1990:
Hooooo my bad XD, github closed the issue automaticaly when i pushed with the #tag in the commit

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
```diff
@@ -460,15 +483,59 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
...
+      }
+
+      // skip verilator compilation if nothing changed (hashes equal)
+      if (hash == previousHash) {
+        val workspaceDir = new File(s"${workspacePath}/${workspaceName}")
+        val cacheDir = new File(s"${cachePath}/${workspaceName}")
+
+        if (cacheDir.exists()) {
+          FileUtils.deleteQuietly(workspaceDir)
+          FileUtils.moveDirectory(cacheDir, workspaceDir)
```
Dolu1990, line 527: moveDirectory, should it be copyDirectory ?

jonnykl:
> <3
> Thanks, fixed with the above commit, that was a Set not being a LinkedSet in the StateMachine builder

Tested in my project, now the new cache is working properly :)

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
```diff
@@ -548,10 +615,27 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
     }
   }
 
+  def copyOutputToCache(): Unit = {
+    if (!cacheEnabled) {
```
Dolu1990, line 619: We may also skip the cache update if things were already cached ?

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
```diff
@@ -548,10 +615,27 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
     }
   }
 
+  def copyOutputToCache(): Unit = {
```
Dolu1990, line 618: copyOutputToCache should be executed atomicaly (global mutex), in case of multiple compilation update the cache at the same time ?

Dolu1990:
> Tested in my project, now the new cache is working properly :)

Great ^^

Those Map and Set can be so easily forget about XD

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
jonnykl: Yes you're right

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
jonnykl: moved to the `compileVerilator` method so the cache will only be updated when verilator was run

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
```diff
@@ -460,15 +483,59 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
...
+      // calculate hash of verilator version+options and source file contents
+
+      val md = MessageDigest.getInstance("SHA-1")
+      md.update(verilatorScript.getBytes())
+
+      val verilatorVersionProcess = Process(Seq(verilatorBinFilename, "--version"), new File(workspacePath))
+      val verilatorVersion = verilatorVersionProcess.lineStream.mkString("\n")    // blocks and throws an exception if exit status != 0
+      md.update(verilatorVersion.getBytes())
+
+      verilatorInputFiles.foreach { filename =>
```
jonnykl, line 501: `verilatorInputFiles` is defined as the input source files for verilator. So this includes all `*.v`, `*.sv` and `*.h` files but not `*.mem` or `*.bin` files. Im not sure if that's really correct?

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
jonnykl: Dou you mean the cache could be updated at the same time when running the compiled scala program two times concurrently?

In case you mean to just run `simCfg.compile()` more than one time, the passed directory `cacheDir` is also renamed. So in this case there should be no conflicts.

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
```diff
@@ -491,6 +558,16 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
...
     FileUtils.copyFile(new File(s"${workspacePath}/${workspaceName}/V${config.toplevelName}${if(isWindows) ".exe" else ""}") , new File(s"${workspacePath}/${workspaceName}/${workspaceName}_$uniqueId.${if(isWindows) "dll" else (if(isMac) "dylib" else "so")}"))
+
+    if (cacheEnabled) {
+      // update cache
+
+      val workspaceDir = new File(s"${workspacePath}/${workspaceName}")
+      val workspaceCacheDir = new File(s"${cachePath}/${workspaceName}")
+
+      FileUtils.deleteQuietly(workspaceCacheDir)
+      FileUtils.copyDirectory(workspaceDir, workspaceCacheDir)
```
jonnykl, line 569: If only the `*.so`/`*.dylib`/`*.dll` file is needed (?), I could copy that file instead of the whole directory to save disk space. Then I could also save multiple versions of this file without creating a huge cache (think of multiple simulation configs, always switching between them).

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
Dolu1990: > Dou you mean the cache could be updated at the same time when running the compiled scala program two times concurrently?

Yes, right, basicaly, a mutex defined in a scala object could be locked / unlocked when checking / copying / write the cache.

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
Dolu1990: verilatorInputFiles => file that are translated by verilator
config.rtlSourcesPaths.filter(s => s.endsWith(".bin") || s.endsWith(".mem"))    =>  files which are used as ressources.
config.rtlSourcesPaths => everytiong

Maybe you should use  config.rtlSourcesPaths instead of verilatorInputFiles ?

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
Dolu1990: I'm not exactly sure, but i would say : 
*.so/*.dylib/*.dll is enough. You don't need to include the bin / mem, because they are embeded into the verilator model.

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
jonnykl: done

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
jonnykl: Just tested it. You only have to copy the `*.so`/`*.dylib`/`*.dll`. I also realized that the shared library contains the `uniqueId` which must match also match the assigned `uniqueId` when using the cached binary. So I now changed the code to only copy the `__ALL.a`, `*.mk` and `*.h` files and also always (re)run make but skip the verilator compilation when the cache is used. So the wrapper will always be compiled with the assigned `uniqueId`.

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
jonnykl: I don't understand why that is needed. When running the program multiple times concurrently they would run verilator in the same directory (please correct me if I'm wrong). So this would fail anyway.

Can you give an example which shows where the current implementation (of the cache) would fail but should succeed?

Dolu1990: > When running the program multiple times concurrently they would run verilator in the same director

It won't, basicaly there is a protection against that, and if a natural workspace name is Foo, then the second time that workspace name is asked, it will get Foo_1 then Foo_2.

Basicaly, that was put in place to allow aggressive multithreading when running regressions in parallel.

> Can you give an example which shows where the current implementation (of the cache) would fail but should succeed?

So my guess is that so far, two compilations of the same test could consider that the cache is empty at the same time (which is fine for me):

https://github.com/SpinalHDL/SpinalHDL/blob/6f3763730c20379395ef935b49e9e76a32631e27/sim/src/main/scala/spinal/sim/VerilatorBackend.scala#L526

The issue would come later, when both will try to update the same cache entry at the same time ?

So to avoid that, just having a mutex to protect the cache read / copy / write operations should be enough i guess.

jonnykl: > > When running the program multiple times concurrently they would run verilator in the same director
> 
> It won't, basicaly there is a protection against that, and if a natural workspace name is Foo, then the second time that workspace name is asked, it will get Foo_1 then Foo_2.

That protection should also apply to the current implementation of the cache. At least when calling `.compile()` multiple times there will also be created separate directories for `Foo_1`, `Foo_2`, ... in the cache. So two threads will never try to write/read to/from the same files.

But now I realize that this behaviour is not wanted because verilator compilation is done `n` times in the first run (when running `n` parallel test). To fix this I have to use a single cache (directory) for a module and then I can / have to protect the cache using a mutex. Right?

Just for clarification: This only applies when running mulitple tests in a single process? Or does the protection (which renames the workspaces) also work when running multiple tests but each in its own process?

jonnykl: Done, hope I understood you correctly

Dolu1990: Ahhhhh, i were assuming the cache was global, that was the missunderstanding

> To fix this I have to use a single cache (directory) for a module and then I can / have to protect the cache using a mutex. Right?

I would say instead, we could have a single folder containing all caches. Each entry in that cache would have the hash as folder name. That way, nothing is related to specific modules. The key would be enough to ensure we pick the proper cache entry. This would easily support having the same module with different parameters being cached.

> Just for clarification: This only applies when running mulitple tests in a single process? Or does the protection (which renames the workspaces) also work when running multiple tests but each in its own process?

This only applies when running mulitple tests in a single process. If you run two JVM, this won't work.

jonnykl: > I would say instead, we could have a single folder containing all caches. Each entry in that cache would have the hash as folder name. That way, nothing is related to specific modules. The key would be enough to ensure we pick the proper cache entry. This would easily support having the same module with different parameters being cached.

Ok, but then we need a way to automatically delete unused cache entries. Otherwise the cache can get really large in a short time. Maybe specify a maximum number of cache entries and delete the directory with the oldest modification time when a cache entry has to be replaced. Or any other idea?

Dolu1990: Right, To me we can keep about 100 in cache, and when there is more, it can either randomly delete one, either pick the one with the oldest date as you proposed ^^
As you want :)

jonnykl: done

Dolu1990:
Hi ^^ 

What is the status ? Ready for merging ?

Dolu1990:
Ahh seems there is a compilation issue : 

```scala
Error:  /home/runner/work/SpinalHDL/SpinalHDL/sim/src/main/scala/spinal/sim/VerilatorBackend.scala:584:29: value sortInPlaceWith is not a member of Array[java.io.File]
Error:              cacheEntriesArr.sortInPlaceWith(_.lastModified() < _.lastModified())
```

jonnykl:
Now all tests pass (except one that also fails without the cache) for all three scala versions. Tested on Linux (Ubuntu), should also be tested on Windows and Mac (I don't have Windows and also no Mac so someone else could run the test? (`sbt 'testOnly spinal.tester.scalatest.VerilatorCacheTester'`))

PR #479, opened 2021-07-23
Widen the address bus so the masters don’t interfere

typingArtist:
This is in preparation of bringing in https://github.com/SpinalHDL/CocotbLib/pull/11


Dolu1990:
should i merge already ?

typingArtist:
Yes, please.

PR #480, opened 2021-07-23
Mem: add models for readUnderWrite modes, add noChange mode

nbstrike:
Adds the ability to generate different intra-port memory read-under-write policies for a read/write memory port in Verilog (readFirst, writeFirst & noChange) with the existing dontCare behavior left as it was.

For memories without a byte lane enable (mask) this infers the expected modes for Xilinx Vivado, however Xilinx blockrams do not have distinct write-enable and mask ports. Vivado does not usually synthesis the expected structures but instead generates a SDP structure for the writeFirst and noChange cases (I'd hoped that it could generate an and gate with the writeEnable and mask signals, then use a single port). I wonder if there's a better way to tackle this?

Given Xilinx/Vivado's limitations I could omit the writeEnable signal (or pass a bit vector there instead of at the mask) -- what do you think? Based on documentation it would seem that Intel FPGAs do have distinct WE and mask signals (Lattice too?), whereas Microsemi seems to be more similar to Xilinx.

Is there a reasonable way to use SpinalHDL expressions to generate the RTL logic, rather than hard-coded strings of Verilog here? That would of course make the change portable to VHDL as well (or any other backend that could be implemented).

The additional of the noChange mode is helpful for Xilinx UltraRAM, if there are to be dual R/W ports on such a memory then it is not possible to read and write from both ports simultaneously. While it is possible to insert readSync and write primitives and arrange the signals for this condition Vivado synthesis is very flaky about inferring the expected memories (due I guess to having 2 processes accessing the memory, instead of the 1 that it expects), the noChange mode alleviates this..

### DIFF
--- a/core/src/main/scala/spinal/core/Mem.scala
+++ b/core/src/main/scala/spinal/core/Mem.scala
@@ -47,6 +47,9 @@ object readFirst extends ReadUnderWritePolicy {
   override def readUnderWriteString: String = "readFirst"
 }
 
+object noChange extends ReadUnderWritePolicy {
+  override def readUnderWriteString: String = "noChange"
+}
 
 object auto extends  MemTechnologyKind{
   override def technologyKind: String = "auto"
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -1153,19 +1153,53 @@ end
           emitWrite(b, memWrite.mem,  if (memWrite.writeEnable != null) emitExpression(memWrite.writeEnable) else null.asInstanceOf[String], memWrite.address, memWrite.data, memWrite.mask, memWrite.mem.getMemSymbolCount(), memWrite.mem.getMemSymbolWidth(), tab)
         }, null, tmpBuilder, memWrite.clockDomain, false)
       case memReadWrite: MemReadWrite  =>
-        if(memReadWrite.readUnderWrite != dontCare) SpinalError(s"memReadWrite can only be emited as dontCare into Verilog $memReadWrite")
         if(memReadWrite.aspectRatio != 1) SpinalError(s"Verilog backend can't emit ${memReadWrite.mem} because of its mixed width ports")
-        emitClockedProcess((tab, b) => {
-          val symbolCount = memReadWrite.mem.getMemSymbolCount()
-          b ++= s"${tab}if(${emitExpression(memReadWrite.chipSelect)}) begin\n"
-          emitRead(b, memReadWrite.mem, memReadWrite.address, memReadWrite, tab + "  ")
-          b ++= s"${tab}end\n"
-        }, null, tmpBuilder, memReadWrite.clockDomain, false)
 
-        emitClockedProcess((tab, b) => {
-          val symbolCount = memReadWrite.mem.getMemSymbolCount()
-          emitWrite(b, memReadWrite.mem,s"${emitExpression(memReadWrite.chipSelect)} && ${emitExpression(memReadWrite.writeEnable)} ", memReadWrite.address, memReadWrite.data, memReadWrite.mask, memReadWrite.mem.getMemSymbolCount(), memReadWrite.mem.getMemSymbolWidth(),tab)
-        }, null, tmpBuilder, memReadWrite.clockDomain, false)
+        memReadWrite.readUnderWrite match {
+          case mode: dontCare.type =>
+            emitClockedProcess((tab, b) => {
+              val symbolCount = memReadWrite.mem.getMemSymbolCount()
+              b ++= s"${tab}if(${emitExpression(memReadWrite.chipSelect)}) begin\n"
+              emitRead(b, memReadWrite.mem, memReadWrite.address, memReadWrite, tab + "  ")
+              b ++= s"${tab}end\n"
+            }, null, tmpBuilder, memReadWrite.clockDomain, false)
+
+            emitClockedProcess((tab, b) => {
+              val symbolCount = memReadWrite.mem.getMemSymbolCount()
+              emitWrite(b, memReadWrite.mem,s"${emitExpression(memReadWrite.chipSelect)} && ${emitExpression(memReadWrite.writeEnable)} ", memReadWrite.address, memReadWrite.data, memReadWrite.mask, memReadWrite.mem.getMemSymbolCount(), memReadWrite.mem.getMemSymbolWidth(),tab)
+            }, null, tmpBuilder, memReadWrite.clockDomain, false)
+          case mode: noChange.type =>
+            emitClockedProcess((tab, b) => {
+              val symbolCount = memReadWrite.mem.getMemSymbolCount()
+              b ++= s"${tab}if(${emitExpression(memReadWrite.chipSelect)}) begin\n"
+              emitWrite(b, memReadWrite.mem,  if (memReadWrite.writeEnable != null) emitExpression(memReadWrite.writeEnable) else null.asInstanceOf[String], memReadWrite.address, memReadWrite.data, memReadWrite.mask, memReadWrite.mem.getMemSymbolCount(), memReadWrite.mem.getMemSymbolWidth(), tab + "  ")
+              b ++= s"${tab}  if(~|${emitExpression(memReadWrite.writeEnable)}) begin\n"
+              emitRead(b, memReadWrite.mem, memReadWrite.address, memReadWrite, tab + "    ")
+              b ++= s"${tab}  end\n"
+              b ++= s"${tab}end\n"
+            }, null, tmpBuilder, memReadWrite.clockDomain, false)
+          case mode: writeFirst.type =>
+            emitClockedProcess((tab, b) => {
+              val symbolCount = memReadWrite.mem.getMemSymbolCount()
+              b ++= s"${tab}if(${emitExpression(memReadWrite.chipSelect)}) begin\n"
+              b ++= s"${tab}  if(${emitExpression(memReadWrite.writeEnable)}) begin\n"
+              emitWrite(b, memReadWrite.mem, null, memReadWrite.address, memReadWrite.data, memReadWrite.mask, memReadWrite.mem.getMemSymbolCount(), memReadWrite.mem.getMemSymbolWidth(), tab + "    ")
+              b ++= s"${tab}    ${emitExpression(memReadWrite)} <= ${emitExpression(memReadWrite.data)};\n"
+              b ++= s"${tab}  end else begin\n"
+              emitRead(b, memReadWrite.mem, memReadWrite.address, memReadWrite, tab + "    ")
+              b ++= s"${tab}  end\n"
+              b ++= s"${tab}end\n"
+            }, null, tmpBuilder, memReadWrite.clockDomain, false)
+          case mode: readFirst.type =>
+            emitClockedProcess((tab, b) => {
+              val symbolCount = memReadWrite.mem.getMemSymbolCount()
+              b ++= s"${tab}if(${emitExpression(memReadWrite.chipSelect)}) begin\n"
+              emitRead(b, memReadWrite.mem, memReadWrite.address, memReadWrite, tab + "  ")
+              emitWrite(b, memReadWrite.mem,  if (memReadWrite.writeEnable != null) emitExpression(memReadWrite.writeEnable) else null.asInstanceOf[String], memReadWrite.address, memReadWrite.data, memReadWrite.mask, memReadWrite.mem.getMemSymbolCount(), memReadWrite.mem.getMemSymbolWidth(), tab + "  ")
+              b ++= s"${tab}end\n"
+            }, null, tmpBuilder, memReadWrite.clockDomain, false)
+          case _ => SpinalError(s"memReadWrite can only be emited as readFirst, writeFirst, noChange or dontCare into Verilog $memReadWrite")
+        }
 
       case memReadSync: MemReadSync   =>
         if(memReadSync.aspectRatio != 1) SpinalError(s"Verilog backend can't emit ${memReadSync.mem} because of its mixed width ports")
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -1127,17 +1127,12 @@ end
         if(memBitsMaskKind == SINGLE_RAM || symbolCount == 1)
           b ++= s"$tab${emitExpression(target)} <= ${emitReference(mem, false)}[${emitExpression(address)}];\n"
         else{
-          val symboleReadDataNames = for(i <- 0 until symbolCount) yield {
-            val symboleReadDataName = component.localNamingScope.allocateName(anonymSignalPrefix + "_" + mem.getName() + "symbol_read")
-            declarations ++= s"  reg [${mem.getMemSymbolWidth()-1}:0] $symboleReadDataName;\n"
-            b ++= s"$tab$symboleReadDataName <= ${emitReference(mem,false)}_symbol$i[${emitExpression(address)}];\n"
-            symboleReadDataName
+          val symWidth = mem.getMemSymbolWidth()
+          for (i <- 0 until symbolCount) {
+            val upLim = symWidth * (i + 1) - 1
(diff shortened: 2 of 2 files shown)


Dolu1990:
Hi ^^

> For memories without a byte lane enable (mask) this infers the expected modes for Xilinx Vivado, however Xilinx blockrams do not have distinct write-enable and mask ports. Vivado does not usually synthesis the expected structures but instead generates a SDP structure for the writeFirst and noChange cases (I'd hoped that it could generate an and gate with the writeEnable and mask signals, then use a single port). I wonder if there's a better way to tackle this?

So to be honnest, i kind of surrender when it is about trying to infer this kind of behaviour across all vendor tools, and instead eventualy rely on blackboxes.

> Is there a reasonable way to use SpinalHDL expressions to generate the RTL logic, rather than hard-coded strings of Verilog here?

What do you mean ? having the possiblity to specify raw verilog string from SpinalHDL ?

What is exactly the noChange mode ?

nbstrike:
> 
> 
> Hi ^^
> 
> > For memories without a byte lane enable (mask) this infers the expected modes for Xilinx Vivado, however Xilinx blockrams do not have distinct write-enable and mask ports. Vivado does not usually synthesis the expected structures but instead generates a SDP structure for the writeFirst and noChange cases (I'd hoped that it could generate an and gate with the writeEnable and mask signals, then use a single port). I wonder if there's a better way to tackle this?
> 
> So to be honnest, i kind of surrender when it is about trying to infer this kind of behaviour across all vendor tools, and instead eventualy rely on blackboxes.

I'm not sure that there's a great answer for this problem, unfortunately Xilinx have this slightly different model even via their memory IP generator (where as the Intel/Altera FPGAs appear to map directly to this model); perhaps if I set the write line to null then the RAM model should be able to use the mask directly to signal writing?

> > Is there a reasonable way to use SpinalHDL expressions to generate the RTL logic, rather than hard-coded strings of Verilog here?
> 
> What do you mean ? having the possiblity to specify raw verilog string from SpinalHDL ?

Not quite, I mean can the logic of the ram model be 'native' SpinalHDL -- ie (for a non-blackboxed memory), can some SpinalHDL logic be substituted in, eg for a writeFirst port you might substitute:
```scala
when (chipEnable) {
  when (writeEnble) {
    array(address) := writeData // this would implement a lower level operation accessing a 2D array
    target := writeData
  } else {
    target := array(address) // access a 2D array
  }
}
```
The advantage being that neither the Verilog nor the VHDL emitters would need to have this logic coded into them (and presumably other variations, such as splitting the memory into symbols to handle the write mask would also be abstracted). What I'm a little unclear on is how this might be achieved -- perhaps via another Phase?
 
> What is exactly the noChange mode ?

In the noChange mode readout is conditional on write not being active (the last read value is preserved); this is helpful for Xilinx's UltraRAMs in which "each port can only execute either a write or a read operation in one cycle" (Xilinx UG573, pg92).

Dolu1990:
> write line to null

what is "line" ?

> Not quite, I mean can the logic of the ram model be 'native' SpinalHDL -- ie (for a non-blackboxed memory), can some SpinalHDL logic be substituted in, eg for a writeFirst port you might substitute:

Ahhhh, the issue of that is that we would kind of go the same hellish direction than the vendors with VHDL / Verilog. So far, i have made on purpose that things are not infered but are explicit (instanciate memory ports via functions). Also this would make the blackboxification harder.

> noChange

Seems good to me :D

i have seen in the code, it emit it with a 

if(write) doWrite
if(!write) doRead

Would the xilinx inferation be fine with : 

if(write) doWrite
else doRead

I guess it may be reconized better by synthesis tools as a single ported ram ?

nbstrike:
> 
> 
> > write line to null
> 
> what is "line" ?

I meant the `write` argument (/signal) of `Mem.readWriteSync()` -- sorry for the confusion.

> > Not quite, I mean can the logic of the ram model be 'native' SpinalHDL -- ie (for a non-blackboxed memory), can some SpinalHDL logic be substituted in, eg for a writeFirst port you might substitute:
> 
> Ahhhh, the issue of that is that we would kind of go the same hellish direction than the vendors with VHDL / Verilog. So far, i have made on purpose that things are not infered but are explicit (instanciate memory ports via functions). Also this would make the blackboxification harder.

I was wondering more if it would be feasible to use the `rework` function to insert a memory model written in Scala (and then expect the backend to translate it appropriately), I think something similar is done to insert black-boxed memories (I have to admit I'm a little unsure of SpinalHDL's architecture around here). I agree that inference would be a nightmare!

> > noChange
> 
> Seems good to me :D
> 
> i have seen in the code, it emit it with a
> 
> if(write) doWrite
> if(!write) doRead
> 
> Would the xilinx inferation be fine with :
> 
> if(write) doWrite
> else doRead
> 
> I guess it may be reconized better by synthesis tools as a single ported ram ?

Good point, I'll update that

Dolu1990:
> perhaps if I set the write line to null then the RAM model should be able to use the mask directly to signal writing

Ahhh so, no write signal, but still the enable signal ?


> I was wondering more if it would be feasible to use the rework function to insert a memory model written in Scala (and then expect the backend to translate it appropriately)

Ahhh, i would say, it isn't a good direction, would be better to fully capture the behaviour in a very explicit way (explicit port instanciation with enough parameters to specify its full behaviour)

nbstrike:
> 
> 
> > perhaps if I set the write line to null then the RAM model should be able to use the mask directly to signal writing
> 
> Ahhh so, no write signal, but still the enable signal ?

Yes, exactly.

> > I was wondering more if it would be feasible to use the rework function to insert a memory model written in Scala (and then expect the backend to translate it appropriately)
> 
> Ahhh, i would say, it isn't a good direction, would be better to fully capture the behaviour in a very explicit way (explicit port instanciation with enough parameters to specify its full behaviour)

i.e. the code as it is now? or perhaps you mean having an explicit Memory Component that could be parametrized with a port specification?

nbstrike:
Updated code uses `else` for noChange mode, adjusts `emitRead` as writeFirst mode was broken for the multi-symbol case.

Dolu1990:
> perhaps if I set the write line to null then the RAM model should be able to use the mask directly to signal writing
Ahhh so, no write signal, but still the enable signal ?
Yes, exactly.

Hmm this seems like a tricky combination to me

Just for me, to do a recap, to be sure of the context :
We want to implement single ported ram using the noChange mode, right ?

I will look around what is the general behaviour of the read data port on such ram :). Will come back

> I was wondering more if it would be feasible to use the rework function to insert a memory model written in Scala (and then expect the backend to translate it appropriately)
..
Ahhh, i would say, it isn't a good direction, would be better to fully capture the behaviour in a very explicit way (explicit port instanciation with enough parameters to specify its full behaviour)
..
i.e. the code as it is now? or perhaps you mean having an explicit Memory Component that could be parametrized with a port specification?

I mean, "insert a memory model written in Scala" seems tricky. What do you mean by memory model written in scala ? can you give a dummy example (code) ?

nbstrike:
> 
> 
> > perhaps if I set the write line to null then the RAM model should be able to use the mask directly to signal writing
> > Ahhh so, no write signal, but still the enable signal ?
> > Yes, exactly.
> 
> Hmm this seems like a tricky combination to me

An alternative could be allowing the write signal to accept a bit vector, implying a mask (while retaining the possibility of independent write + enable signals). 

> Just for me, to do a recap, to be sure of the context :
> We want to implement single ported ram using the noChange mode, right ?

Yes (I also have some use cases for writeFirst).

> I will look around what is the general behaviour of the read data port on such ram :). Will come back
> 
> > I was wondering more if it would be feasible to use the rework function to insert a memory model written in Scala (and then expect the backend to translate it appropriately)
> > ..
> > Ahhh, i would say, it isn't a good direction, would be better to fully capture the behaviour in a very explicit way (explicit port instanciation with enough parameters to specify its full behaviour)
> > ..
> > i.e. the code as it is now? or perhaps you mean having an explicit Memory Component that could be parametrized with a port specification?
> 
> I mean, "insert a memory model written in Scala" seems tricky. What do you mean by memory model written in scala ? can you give a dummy example (code) ?

So could there be a component something along these lines (example, not extracted from actual code!):
```scala
class Memory(nReadPorts : Int, nRwPorts : Int, nWritePorts : Int, readLatency : Int, depth : Int, ...) extends Component {
   val io = new Bundle {
      val rwPorts = if(nRwPorts>0) Vec(slave RwPort(...), nRwPorts) else null // perhaps this is split into separate signals; I assume different clock domains per port would make it through
      ....
   }

   val mem = BitArray2D(wordType(), depth) // BitArray2D being a lower level abstraction of a 2d register array; for multi-symbol memories this could be generated as a vector

   io.rwPorts.foreach { case port =>
     // example of writeFirst RW port
     val rdata = Reg(wordType())
     port.rdata := rdata

     when (port.enable) {
       when (port.write) {
         mem(port.address) := port.wdata
         rdata := port.wdata
       } otherwise {
         rdata := mem(port.address)
       }
     }
   }
}
```

Then from a Phase this could be substituted for Mem similarly to a blackbox memory similarly to now. The advantage I'd hope for would be to have a single abstracted model for the memories in scala, instead of duplicating it into the VHDL & Verilog emitters (hopefully this might then make it easier to deal with oddness like this write signal vs. enable signals question).

In anycase if this made sense then I'd want to tackle something along these lines as a separate PR.

Dolu1990:
>>> perhaps if I set the write line to null then the RAM model should be able to use the mask directly to signal writing
Ahhh so, no write signal, but still the enable signal ?
Yes, exactly.
>>Hmm this seems like a tricky combination to me
>An alternative could be allowing the write signal to accept a bit vector, implying a mask (while retaining the possibility of independent write + enable signals).

Hmm what is the issue of single ported ram with  enable + write [+ mask] as controles ?

> Yes (I also have some use cases for writeFirst).

About the readUnderWrite policy, initialy, it was intended to specify for that read port, if the read should happen before / after / don't care any other writes from other memory ports.

Here, adding a "no change"  mode in that specific paramter seems wrong. i guess we should keep :
- dontCare / writeFirst / readFirst
- Add a Boolean paramter, to specify if durring a write, the read should also occure (default true)

> So could there be a component something along these lines (example, not extracted from actual code!)

I'm not for it so far, i mean, for instance, how do you describe a "dontCare" mode with it ?

nbstrike:
> > > > perhaps if I set the write line to null then the RAM model should be able to use the mask directly to signal writing
> > > > Ahhh so, no write signal, but still the enable signal ?
> > > > Yes, exactly.
> > > > Hmm this seems like a tricky combination to me
> > > > An alternative could be allowing the write signal to accept a bit vector, implying a mask (while retaining the possibility of independent write + enable signals).
> 
> Hmm what is the issue of single ported ram with enable + write [+ mask] as controles ?

For me it is finding a way to match Xilinx's BRAM implementation which combines the function of the write + mask signals.

> > Yes (I also have some use cases for writeFirst).
> 
> About the readUnderWrite policy, initialy, it was intended to specify for that read port, if the read should happen before / after / don't care any other writes from other memory ports.
> 
> Here, adding a "no change" mode in that specific paramter seems wrong. i guess we should keep :
> 
>     * dontCare / writeFirst / readFirst
> 
>     * Add a Boolean paramter, to specify if durring a write, the read should also occure (default true)

I'm not sure what this parameter does in practice, in the current implementations the emitters fix the mode to either dontCare for sync read or writeFirst for async read. Clearly it is not possible to assert anything useful when crossing clock domains, and reviewing FPGA vendor documentation inter-port relationships are not always well defined (Intel suggest that they would be implemented with external logic) and I wonder how multiple ports should be specified to infer the expected modes?

For read-during-write behaviour on a single port it should be possible to choose between readFirst, writeFirst or noChange modes (and dontCare) i.e. not just a Boolean -- these modes are usually documented by the FPGA vendors.

> > So could there be a component something along these lines (example, not extracted from actual code!)
> 
> I'm not for it so far, i mean, for instance, how do you describe a "dontCare" mode with it ?

I'll leave that for now, I may play around with implementing something then it would be clearer

Dolu1990:
> For me it is finding a way to match Xilinx's BRAM implementation which combines the function of the write + mask signals.

Hmm, ok then ^^ So what about, providing a "True" (not a null) to the Mem.readWriteSync.write would emit the Verilog in a way Xilinx like it ?

> inter-port relationships are not always well defined

So at least for simple dual port is is defined not so badly, but i agree in general ^^
But !
Using some SpinalHDL compilation phase, the required logic / fixes could be added to a Mem to impement it properly.
That's one of the idea, allowing to capture "precisely" the memory you want, and later one in some compilation phase, make it compatible with the target programaticaly.

> Intel suggest that they would be implemented with external logic)

Right, i guess that the safe practice, avoid the case by design / external logic.

> For read-during-write behaviour on a single port it should be possible to choose between readFirst, writeFirst or noChange modes (and dontCare) i.e. not just a Boolean -- these modes are usually documented by the FPGA vendors.

I'm realy fine with it, but, so far, i would say the dontCare shouldn't be part of the enumeration, but a specific boolean parameter in the Mem.readWriteSync argument list. To allow being more precise.

Dolu1990:
@nbstrike Hi,

Github consider it merged, but that was a automatic action when i manualy merged your changes ^^.
I do no consider this as a closed subject XD

So, there is the current status : 

- I added in readWriteSync a arguement duringWrite : DuringWritePolicy  which can be  : dontCare, doRead, dontRead
This specify the behaviour of the port when a write is happening.

- For the xilinx UltraRAM, basicaly, the fact that they don't have a write signal but use the mask directly is one of their specificity, but is also a subset of what SpinalHDL can specify, so the idea here is that if they can't be infered  when using a write signal, then to blackbox them using the automatic blackboxing option, and manualy write the verilog that would make them happy.

- About read during write behaviour. Depending the vendor, this will create issues. It is in general a real mess to properly support across vendors. So basicaly here, i guess the best is to stay with SpinalHDL only supporting VHDL/Verilog emition of "dontCare" mode. And if the user realy need more precise behaviour as readFirst writeFirst, he can specify it in the Mem ports, but will have to provide the blackbox for it that make the vendor happy. I'm not against having a repo of verilog/vhdl files implementing the memory blackboxes for the different vendors ^^. But supporting it in the generation backend is just a tricky riscky thing.  

- readDuringWrite is realy to specify for the given read port, how other write ports are seen.

So the idea is to give enough tools to document enough how the user want a Mem to behave, not to match what a synthesis tool can reconise / infer. This create a gap, but this gap can be filled with blackboxes or additional SpinalHDL phase to implement surounding ram logic.

Does that seems good ?

nbstrike:
Thanks, that seems reasonable.

- For the `duringWrite` argument, I would think that the default `dontCare` should probably correspond to `doRead` because this looks like the existing behaviour where as `dontRead` would seem to be a change (arguable the default should simply be `doRead`);
- The `readFirst` and `writeFirst` modes are both useful -- I used code from https://www.doulos.com/httpswwwdouloscomknowhow/fpga/fpga-technotes/ tech note 1 "HDL Synthesis Inference of FPGA Memories" which claims to "work across a wide variety of synthesis tools and FPGA families" (I haven't verified that claim). For instance the writeFirst mode could be used to implement the read port of a common-clock FIFO and resolve a write collision by writing the data through the read port if there is a collision, allowing a lower latency for the fifo; similarly readFirst allows swapping sematics to be implemented.

A difficultly with the blackbox approach can be making sure that I chose an available combination of ports (eg the FIFO case I mentioned would want 1w + 1rws ports).

Dolu1990:
> For the duringWrite argument, I would think that the default dontCare should probably correspond to doRead because this looks like the existing behaviour where as dontRead would seem to be a change (arguable the default should simply be doRead);

with the goal  to stay as close than the previous behaviour ? or to help as much as possible to infer things ?

> HDL Synthesis Inference of FPGA Memorie

Hmm, ok then, i will merge them too ^^

> A difficultly with the blackbox approach can be making sure that I chose an available combination of ports (eg the FIFO case I mentioned would want 1w + 1rws ports).

Right, there is a limited subset of support

nbstrike:
> > For the duringWrite argument, I would think that the default dontCare should probably correspond to doRead because this looks like the existing behaviour where as dontRead would seem to be a change (arguable the default should simply be doRead);
> 
> with the goal to stay as close than the previous behaviour ? or to help as much as possible to infer things ?

Yes, retain the previous behaviour -- also that behaviour matches most vendor samples of a single read/write port describe a behaviour with a simultaneous read + write so this would seem to be the least surprising option.

Dolu1990:
Fixed ^^

PR #484, opened 2021-07-28
enumdefine.v issue fix oneFilePerComponent

jijingg:
@Dolu1990 a little bug fix

here will not generate enumdefine.v if there no enum exists on mode of oneFilePerComponent


Dolu1990:
thanks ^^

PR #490, opened 2021-08-11
Added some size check to Apb3Decoder

allexoll:
there was an issue on gitter with someone using a 23bit mapping connected to a 22bit Apb3, which did not raise any error. this fixes it.


Dolu1990:
Thanks :)

PR #494, opened 2021-08-19
fix memory sim double valid when interrupted more than once

Pyromuffin:
memory sim rvalid goes high twice for the same data if the memory sim is interrupted twice in a row. this change prevents that error. I didnt make changes to the altera behavior part of the code because I don't really know how that works.


Dolu1990:
@Pyromuffin Thanks !

@c-thaler This is for you ^^ All seems good ?

c-thaler:
Looks good! I hope I find the time to look into the Altera implementation soon.

@Pyromuffin Thanks for the patch!

PR #500, opened 2021-08-26
Formal: add enhanced past method

piegamesde:
I had a lot of code like

```scala
cover(inStream.fire
  && past(inStream.fire, 1)
  && past(inStream.fire, 2)
  && past(inStream.fire, 3)
  && past(inStream.fire, 4)
  && past(inStream.fire, 5))
```

and I was looking for a solution that reduces the boilerplate. In the end I found something that generalizes quite well to other use cases. The core concept is that we give in a sequence of time points and get out the history of the value. The desired functionality can then be achieved by reducing with an appropriate operator. By the way I find it quite useful even outside of formal verification (for example, writing `past(signal)` instead of `RegNext(signal)`, so maybe it would fit better somewhere else…

This is almost completely untested, but let me know what you think.

### DIFF
--- a/core/src/main/scala/spinal/core/Formal.scala
+++ b/core/src/main/scala/spinal/core/Formal.scala
@@ -16,6 +16,48 @@ object Formal {
   }
   def past[T <: Data](that : T) : T = past(that, 1)
 
+  /** past implementation optimizied for multiple delays
+   *
+   * Instead of giving a delay in cycles, you can provide a sequence of delays and will
+   * get a Vec containing the history of your value at the given time points.
+   *
+   * Examples:
+   *
+   * A signal should have been held for the last four cycles: `past(signal, 1 to 4).reduce(_ && _)` (start
+   * the range at 0 if the signal should still be high in the current cycle.
+   *
+   * A signal should have been raised in the last four cycles: `past(signal, 1 to 4).reduce(_ || _)`.
+   *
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hoo nice ^^
Hmm this make me think to the History feature in the spinal.lib have you seen it ?
https://github.com/SpinalHDL/SpinalHDL/blob/36e2be8451eeeccd687fc4c2363094d0e038e84e/lib/src/main/scala/spinal/lib/Utils.scala#L788

Your implementation seems even more general, so even better XD

what's about updating the History utility to implement at its root your idea ? (having a list of the different delays you want)

piegamesde:
Great, something like `History` was exactly what I was looking for. However, I'm not sure that I fully understand what the `when` and `init` input do in detail.

Dolu1990:
@piegamesde Sorry, i forgot that issue XD

So, the History "init" is to give a reset value to the registers of the chaine. The "when" arguement allow to specify when the registers need to sample their input.

Dolu1990:
@piegamesde Many improvements are on the way in SpinalHDL dev, including a backend which suport symbiyosys. See the formal doc : 
https://spinalhdl.github.io/SpinalDoc-RTD/dev/SpinalHDL/Formal%20verification/index.html#external-assertions

Dolu1990:
No worries ^^
I just told you, as a notification. i think all is good now :)

PR #506, opened 2021-08-31
Add option to enable logging of ghdl/iverlog backend

jiegec:
Enable logging can help debugging when the backend fails.


Dolu1990:
Thanks ^^

Issue #319, opened 2020-05-14
Can I use localparameter in FSM/Enum instead of define?


lifoo:
Question1：
Why is the state of FSM and Emun defined with define? This introduces hidden risks, for example:
There are 3 verilog files, a.v, b.v and c.v. The synthesizer reads the files in sequence, a.v is my handwritten code with verilog, there is a define DEFINE_USE_IN_A_C, b.v is a verilog file generated by SpinalHDL, there is a state machine or Enum, it happens to redefine DEFINE_USE_IN_A_C, but value is not equal as in a.v, c.v is also handwritten code,did not define DEFINE_USE_IN_A_C, but used DEFINE_USE_IN_A_C, but DEFINE_USE_IN_A_C has been redefined in b.v, resulting in cv circuit error.
I think it would be better use localparameter to solve the problem in verilog.

Question2：
`ifndef SYNTHESIS
...
`endif
When using FSM or Enum, can the above codes be generated selectively? Otherwise, I need to remind back-end engineers to define SYNTHESIS to continue their work.

Dolu1990:
Hi,
> Question1：

Hoo, so in Verilog, define definitions can cross from one .v to another ?
or you mean when you include a verilog file into another ?
(sorry, i was raised as a VHDL guy, don't know so much about Verilog :))

> Question2：

Sorry i'm not sure to understand the exact case you ask. You mean that you want to have some generated code to be in ifndef SYNTHESIS ... endif ? or you ask about some ifndef SYNTHESIS ... endif which are generated by SpinalHDL itself already ?

lifoo:
Hi Dolu:
Question1:
Verilog's macro definition is global from the definition point, and remains valid for subsequent files (this is different from C language) until another macro definition changes the value of this macro definition, or use undef to cancel this Macro definition. Since the macro definition remains valid for the files that are subsequently read, it is usually required to compile the files in a certain order.
The most typical problem with using macro definitions is that another file's macro definition uses the same name. When encountering this kind of problem, it may take a lot of cost to debug.
Question2:
When using StateMachine or Enum, the following verilog code will be generated:
`ifndef SYNTHESIS
  always @(*) begin
    case(fsm_stateReg)
      `fsm_enumDefinition_binary_sequential_boot : fsm_stateReg_string = "boot      ";
      `fsm_enumDefinition_binary_sequential_fsm_stateA : fsm_stateReg_string = "fsm_stateA";
[...]

PR #508, opened 2021-09-05
support localparam for verilog

dxsm:
As #319 , I add localparam support for verilog.
```scala
class top extends Component{
    object mainState extends SpinalEnum() {
    val Idle, Start, Data, Parity, Stop = newElement()
    setLocal()   //default is global define
//    rawElementName()  //default prefix with object name
  }

  val curr_st = mainState()
  curr_st := mainState.Idle

  val next_st = mainState()
  next_st := mainState.Idle

}

object main {
    SpinalVerilog(new top())
//    SpinalConfig().withoutEnumString().generateVerilog(new top())
  }
}
```
use `setLocal()`, you can generate localparam instead of define for enum

```verilog
// Generator : SpinalHDL v1.6.1    git head : 524dbb3d7d3aae8680c579a4592db4c57588aeb8
// Component : top
// Git hash  : 524dbb3d7d3aae8680c579a4592db4c57588aeb8


module top (
);
  localparam mainState_Idle = 3'd0;
  localparam mainState_Start = 3'd1;
  localparam mainState_Data = 3'd2;
  localparam mainState_Parity = 3'd3;
  localparam mainState_Stop = 3'd4;

  wire       [2:0]    curr_st;
  wire       [2:0]    next_st;

  assign curr_st = mainState_Idle;
  assign next_st = mainState_Idle;

endmodule
```

use `rawElementName()`, you can remove the prefix name of object name
```scala
class top extends Component{
    object mainState extends SpinalEnum() {
    val Idle, Start, Data, Parity, Stop = newElement()
    setLocal()   //default is global define
    rawElementName()  //default prefix with object name
  }

  val curr_st = mainState()
  curr_st := mainState.Idle

  val next_st = mainState()
  next_st := mainState.Idle

}

object main {
    SpinalVerilog(new top())
//    SpinalConfig().withoutEnumString().generateVerilog(new top())
  }
}
```
will generate
```verilog
// Generator : SpinalHDL v1.6.1    git head : 524dbb3d7d3aae8680c579a4592db4c57588aeb8
// Component : top
// Git hash  : c3cb5a1e22f552ac73dee6d9d519ab5e88db8499


module top (
);
  localparam Idle = 3'd0;
  localparam Start = 3'd1;
  localparam Data = 3'd2;
  localparam Parity = 3'd3;
  localparam Stop = 3'd4;

  wire       [2:0]    curr_st;
  wire       [2:0]    next_st;

  assign curr_st = Idle;
  assign next_st = Idle;

endmodule
```

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -52,6 +52,7 @@ class ComponentEmitterVerilog(
   val portMaps     = ArrayBuffer[String]()
   val definitionAttributes  = new StringBuilder()
   val declarations = new StringBuilder()
+  val localparams = new StringBuilder()
   val logics       = new StringBuilder()
   def getTrace() = new ComponentEmitterTrace(definitionAttributes :: declarations :: logics :: Nil, portMaps)
 
@@ -60,6 +61,7 @@ class ComponentEmitterVerilog(
     s"""
       |${definitionAttributes}module ${component.definitionName} (
       |${ports}
+      |${localparams}
       |${declarations}
       |${logics}
       |endmodule
@@ -206,6 +208,7 @@ class ComponentEmitterVerilog(
     emitAnalogs()
     emitMuxes()
     emitEnumDebugLogic()
+    emitEnumParams()
 
     processes.foreach(p => {
       if(p.leafStatements.nonEmpty ) {
@@ -486,6 +489,22 @@ class ComponentEmitterVerilog(
       }
       logics ++= "  `endif\n\n"
     }
+  }
+
+  def emitEnumParams(): Unit = {
+    for((e,encoding) <- localEnums) {
+      for(element <- e.elements) {
+        localparams ++= s"  localparam ${emitEnumLiteral(element, encoding,"")} = ${idToBits(element, encoding)};\n"
+      }
+    }
+  }
+
+  def idToBits[T <: SpinalEnum](enum: SpinalEnumElement[T], encoding: SpinalEnumEncoding): String = {
+    //      val str    = encoding.getValue(enum).toString(2)
+    val str    = encoding.getValue(enum).toString(10)
+    val length = encoding.getWidth(enum.spinalEnum)
+    //      length.toString + "'b" + ("0" * (length - str.length)) + str
+    length.toString + "'d" + str
   }
 
   def emitAsynchronousAsAsign(process: AsyncProcess) = process.leafStatements.size == 1 && process.leafStatements.head.parentScope == process.nameableTargets.head.rootScopeStatement
@@ -926,6 +945,8 @@ class ComponentEmitterVerilog(
 
   var memBitsMaskKind: MemBitsMaskKind = MULTIPLE_RAM
   val enumDebugStringList = ArrayBuffer[(SpinalEnumCraft[_ <: SpinalEnum], String, Int)]()
+  val localEnums          = mutable.LinkedHashSet[(SpinalEnum, SpinalEnumEncoding)]()
+
   def emitSignals(): Unit = {
     val enumDebugStringBuilder = new StringBuilder()
     component.dslBody.walkDeclarations {
@@ -936,13 +957,23 @@ class ComponentEmitterVerilog(
         if(spinalConfig._withEnumString) {
           signal match {
             case signal: SpinalEnumCraft[_] => {
-              val name = component.localNamingScope.allocateName(emitReference(signal, false) + "_string")
-              val stringWidth = signal.spinalEnum.elements.map(_.getNameElseThrow.length).max
-              enumDebugStringBuilder ++= s"  reg [${stringWidth * 8 - 1}:0] $name;\n"
-              enumDebugStringList += Tuple3(signal , name, stringWidth)
+	      if(signal.spinalEnum.isGlobalEnable) {
+                val name = component.localNamingScope.allocateName(emitReference(signal, false) + "_string")
+                val stringWidth = signal.spinalEnum.elements.map(_.getNameElseThrow.length).max
+                enumDebugStringBuilder ++= s"  reg [${stringWidth * 8 - 1}:0] $name;\n"
+                enumDebugStringList += Tuple3(signal , name, stringWidth)
+	      }
             }
             case _ =>
           }
+        }
+        signal match {
+          case signal: SpinalEnumCraft[_] => {
+            if(!signal.spinalEnum.isGlobalEnable) {
+              localEnums.add((signal.spinalEnum, signal.encoding))
+            }
+          }
+          case _ =>
         }
       case mem: Mem[_] =>
     }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -488,6 +491,22 @@ class ComponentEmitterVerilog(
...
+    for((e,encoding) <- localEnums) {
+      for(element <- e.elements) {
+        localparams ++= s"  localparam ${emitEnumLiteral(element, encoding,"")} = ${idToBits(element, encoding)};\n"
+      }
+    }
+  }
+
+  def idToBits[T <: SpinalEnum](enum: SpinalEnumElement[T], encoding: SpinalEnumEncoding): String = {
+    //      val str    = encoding.getValue(enum).toString(2)
+    val str    = encoding.getValue(enum).toString(10)
```
Dolu1990, line 504: This may overflow the Verilog 32 bits integer for enumeration with onehot encoding and more than ?32? elements i guess ?

core/src/main/scala/spinal/core/internals/VerilogBase.scala
```diff
@@ -106,17 +106,31 @@ trait VerilogBase extends VhdlVerilogBase{
   }
 
   def emitEnumLiteral[T <: SpinalEnum](enum: SpinalEnumElement[T], encoding: SpinalEnumEncoding, prefix: String = "`"): String = {
-    prefix + globalPrefix + enum.spinalEnum.getName() + "_" + encoding.getName() + "_" + enum.getName()
+//    prefix + enum.spinalEnum.getName() + "_" + encoding.getName() + "_" + enum.getName()
```
Dolu1990, line 109: Here, and also in the other functions of this modified file, globalPrefix seems to be dropped.

what globalPrefix does, is it allow to add some prefix in the front of everything, to avoid name conflicts if you generate multiple SpinalHDL netlist and use them together in a simulation / synthesis.

Would it be possible to bring that feature back ?

Dolu1990:
Nice, thanks :D

Seems good to me, just need to fix two things. I added the comments about them just above ^^

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
dxsm: This function is completely referenced here
https://github.com/SpinalHDL/SpinalHDL/blob/524dbb3d7d3aae8680c579a4592db4c57588aeb8/core/src/main/scala/spinal/core/internals/PhaseVerilog.scala#L170

Is that may overflow too?

core/src/main/scala/spinal/core/internals/VerilogBase.scala
dxsm: yeah, I fixed the dropped.

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
Dolu1990: Hoo my bad, i had a reflex from VHDL where integer are 32 bits only, i guess a verilog literal dxxxxxx is fine. 
I have seen the toString(10) and got triggered XDXD

jijingg:
@dxsm good job
 @Dolu1990 could the default config set to localparameter, because enum value as a macro  is really uncommon in Verilog 
```scala
 var  isPrefixEnable = true
 var  isGlobalEnable = true
 ```

**Comment by Dolu1990** [comment]
@jijingg I guess you right, i pushed a few fixes, i pushed a few fixes and now the default is  
 var  isPrefixEnable = true
 var  isGlobalEnable = false
 
 And the encoding of the enum is only added to the name if it isn't the default encoding.
 Note, the enumString is still enabled by default (usefull for gtkwave for instance)

**Comment by dxsm** [comment]
@Dolu1990 If so, I think you should add an additional `setGlobal` function here if I want to set a specific Enum to global.
https://github.com/SpinalHDL/SpinalHDL/blob/3677620be63b5823279e1781c49a2ac657b4c0e0/core/src/main/scala/spinal/core/Enum.scala#L113

**Comment by Dolu1990** [comment]
@dxsm done, thanks ^^

**Comment by jiegec** [comment]
Hi, I have encountered a problem when the enum is used in more than one modules:

```
module A();
assign xxx = Enum_Value;
endmodule

module B();
local param Enum_Value = 2'b00;
assign yyy = Enum_Value;
```

**Comment by Dolu1990** [comment]
@jiegec  Hi,

Do you use very last SpinalHDL ? 

I just tested 

```scala
new Component {
      val enum = new SpinalEnum{
        val A, B, C = newElement()
      }

      val x = enum.A()
      val sub = new Component {
        val y = enum.B()
        setDefinitionName("Rawrrr")
      }
      setDefinitionName("Miaou")
    }
    
    ```
    
    And that seems fine to me.
    
    Can you provide a broken example ?

**Comment by jiegec** [comment]
@Dolu1990 Yes:

// Generator : SpinalHDL v1.6.1    git head : 3100c81b37a04715d05d9b9873c3df07a0786a9b

I am trying to recreate the problem in a minimal example.

**Comment by jiegec** [comment]
I can reproduce the problem with the following code:

```scala
object MyPlay {
  object TestEnum extends SpinalEnum{
    val A, B, C = newElement()
  }

  case class InnerBlackBox() extends BlackBox {
    val io = new Bundle {
      val i = in(Bits(2 bits))
      val o = out(Bits(2 bits))
    }
  }

  case class Inner() extends Component {
    val inner = InnerBlackBox()
    inner.io.i := TestEnum.A.asBits
  }

  case class TopLevel() extends Component {
    val i = in(TestEnum)
    val o = out(Bits(2 bits))

    val inner = new Inner()
    o := 0
  }

    SpinalVerilog(new TopLevel)
  }
}
```

It produces:

```verilog
// Generator : SpinalHDL v1.6.1    git head : 3100c81b37a04715d05d9b9873c3df07a0786a9b
// Component : TopLevel
// Git hash  : 3100c81b37a04715d05d9b9873c3df07a0786a9b



module TopLevel (
  input      [1:0]    i,
  output     [1:0]    o
);
  localparam TestEnum_A = 2'd0;
  localparam TestEnum_B = 2'd1;
  localparam TestEnum_C = 2'd2;

  `ifndef SYNTHESIS
  reg [7:0] i_string;
  `endif


  Inner inner_1 (
  );
  `ifndef SYNTHESIS
  always @(*) begin
    case(i)
      TestEnum_A : i_string = "A";
      TestEnum_B : i_string = "B";
      TestEnum_C : i_string = "C";
      default : i_string = "?";
    endcase
  end
  `endif

  assign o = 2'b00;

endmodule

module Inner (
);

  wire       [1:0]    inner_1_io_i;
  wire       [1:0]    inner_1_io_o;

  InnerBlackBox inner_1 (
    .io_i    (inner_1_io_i[1:0]  ), //i
    .io_o    (inner_1_io_o[1:0]  )  //o
  );
  assign inner_1_io_i = TestEnum_A;

endmodule
```

Dolu1990:
Thanks ^^

Dolu1990:
(fixed)

PR #513, opened 2021-09-21
avoid using functions for bit assignment

dxsm:
For instance:
```scala
class top extends Component{
  val myBits = B(8 bits, (7 downto 5) -> B"101", 4 -> true, 3 -> True, default -> false)
}

object main {
    SpinalVerilog(new top())
  }
}
```
will generate
```verilog
// Generator : SpinalHDL v1.6.1    git head : 524dbb3d7d3aae8680c579a4592db4c57588aeb8
// Component : top
// Git hash  : 58e077ba1f6e9688b293766733b3b090d3310cc8

module top (
);

  reg        [7:0]    myBits;
  function [7:0] zz_myBits(input dummy);
    begin
      zz_myBits = 8'h0;
      zz_myBits[7 : 5] = 3'b101;
      zz_myBits[4] = 1'b1;
      zz_myBits[3] = 1'b1;
    end
  endfunction
  wire [7:0] _zz_1;

  assign _zz_1 = zz_myBits(1'b0);
  always @(*) myBits = _zz_1;

endmodule
```
For function is rarely used in verilog. I optimized this part, generated as follow:
```verilog
// Generator : SpinalHDL v1.6.1    git head : 524dbb3d7d3aae8680c579a4592db4c57588aeb8
// Component : top
// Git hash  : 58e077ba1f6e9688b293766733b3b090d3310cc8

module top (
);

  reg        [7:0]    myBits;

  always @ (*) begin
    myBits = 8'h0;
    myBits[7 : 5] = 3'b101;
    myBits[4] = 1'b1;
    myBits[3] = 1'b1;
  end


endmodule
```
As you see, itermediate variable and function definition are all eliminated.

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -532,28 +532,30 @@ class ComponentEmitterVerilog(
           //assert(process.nameableTargets.size == 1)
           for(node <- process.nameableTargets) node match {
             case node: BaseType =>
-              val funcName = "zz_" + emitReference(node, false).replaceAllLiterally(".", "__")
-              declarations ++= s"  function ${emitType(node)} $funcName(input dummy);\n"
-//              declarations ++= s"    reg ${emitType(node)} ${emitReference(node, false)};\n"
-              declarations ++= s"    begin\n"
+//              val funcName = "zz_" + emitReference(node, false).replaceAllLiterally(".", "__")
+//              declarations ++= s"  function ${emitType(node)} $funcName(input dummy);\n"
 
+//              declarations ++= s"    begin\n"
+
+              logics ++= s"  always @ (*) begin\n"
               val statements = ArrayBuffer[LeafStatement]()
               node.foreachStatements(s => statements += s.asInstanceOf[LeafStatement])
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^

So, first, the previous implementation (function based) was made to be 100% sure things would work as i was't sure how tools would behave on VHDL process / verilog always  which had no sensitivity inputs.

Are we sure that alwas blocks without trigger are properly applied ? 
I mean, i don't know, this is more about cooking receip than science ^^

- I guess synthesis tools are fine
- Not sure about even driven simulators (regular simulation)
- Not sure how formal tools will handel it too

If it is fine, then sure, this can be merged :)

dxsm:
I think you are right. I'll try to use concatenate assignments to avoid function in verilog as below: 
```verilog
module top (
);

  wire        [7:0]    myBits;

  assign myBits = {3'b101, 1'b1, 1'b1, 3'b0};

 //always @ (*) begin
 //  myBits = 8'h0;
 //   myBits[7 : 5] = 3'b101;
 //  myBits[4] = 1'b1;
 //  myBits[3] = 1'b1;
 // end

endmodule
```

Dolu1990:
> I think you are right. 

I'm absolutly not sure i'm right. i just don't realy know, so toke the "safe" way :)

Issue #445, opened 2021-06-15
Support In-line Verilog


volatile-static:
Is Spinal going to support black boxes with inline Verilog like Chisel?
```scala
import chisel3._
import chisel3.util.HasBlackBoxInline
class BlackBoxRealAdd extends BlackBox with HasBlackBoxInline {
  val io = IO(new Bundle {
    val in1 = Input(UInt(64.W))
    val in2 = Input(UInt(64.W))
    val out = Output(UInt(64.W))
  })
  setInline("BlackBoxRealAdd.v",
    """module BlackBoxRealAdd(
      |    input  [15:0] in1,
      |    input  [15:0] in2,
      |    output [15:0] out
      |);
      |always @* begin
      |  out <= $realtobits($bitstoreal(in1) + $bitstoreal(in2));
      |end
      |endmodule
    """.stripMargin)
}
```

Dolu1990:
Hi

It could be done, doesn't seem a big deal to implement. but more generaly, what's about inline verilog ? Allowing to insert a "blackboxExpression" into the AST, which will emit a user raw code into the netlist ?

I mean, what are the usages of black boxes with inline verilog ?

As far as i can see, emiting very specific chunk of verilog code, right ?

volatile-static:
Hi
I haven't given the question much thought. How do people use this in Chisel? Maybe someone will need it in the future 😄

volatile-static:
Imagine you're working on a project in pure SpinalHDL and all RTL files are in the gitignore. However, there is a little logic hard to be described in Spinal.
The point is that, can Verilog briefer than Spinal describing such a logic? 
For example, I have a switch matrix scanner circuit.
```verilog
always @(posedge clock, posedge reset)
	if (reset)
		key_down = 0;
	else begin: key_decode
		key_down = key_down;
		case (row_cnt)
			1: begin
				key_down[1] = col_sync[0];
				key_down[16:2] = col_sync[16:2];
			end
			2: begin
			  	key_down[37:17] = col_sync[20:0];
			end
			3: begin
			  	key_down[58:38] = col_sync[20:0];
			end
			4: begin
			  	key_down[70:59] = col_sync[11:0];
				key_down[71] = col_sync[13];
				key_down[74:72] = col_sync[19:17];
			end
			5: begin
			  	key_down[85:75] = col_sync[10:0];
				key_down[86] = col_sync[13];
				key_down[87] = col_sync[15];
[...]

PR #517, opened 2021-10-07
support inline rtl for BlackBox

dxsm:
#445, add black boxes with inline Verilog feature support  like Chisel . If you are not against inline verilog, I think it is good for pure scala project.
```scala
class adder(width: Int) extends BlackBox {
  addGeneric("WIDTH", width)

  val io = new Bundle {
    val ain = in UInt(width bits)
    val bin = in UInt(width bits)
    val add_out = out UInt(width bits)
  }

  setInline("adder.v",
    """module adder #(
      |    parameter WIDTH = 16
      |) (
      |    input      [WIDTH-1:0] ain    ,
      |    input      [WIDTH-1:0] bin    ,
      |    output reg [WIDTH-1:0] add_out
      |);
      |
      |always @(*) begin
      |  add_out = ain + bin;
      |end
      |
      |endmodule
    """.stripMargin)
}

class top extends Component {
  val io = new Bundle {
    val ain = in UInt(16 bits)
    val bin = in UInt(16 bits)
    val add_out = out UInt(16 bits)
  }
  var adder = new adder(16)
  io.ain <> adder.io.ain
  io.bin <> adder.io.bin
  io.add_out <> adder.io.add_out

}

object main {
    SpinalVerilog(new top())
  }
}
```
blackbox `adder.v` will be generated
```verilog
module adder #(
    parameter WIDTH = 16
) (
    input      [WIDTH-1:0] ain    ,
    input      [WIDTH-1:0] bin    ,
    output reg [WIDTH-1:0] add_out
);

always @(*) begin
  add_out = ain + bin;
end

endmodule
```

### DIFF
--- a/core/src/main/scala/spinal/core/Area.scala
+++ b/core/src/main/scala/spinal/core/Area.scala
@@ -200,5 +200,9 @@ class ResetArea(reset: Bool, cumulative: Boolean) extends Area with PostInitCall
 
 
 trait AreaObject extends Area{
-  setWeakName(this.getClass.getSimpleName.replace("$",""))
+  setName(this.getClass.getSimpleName.replace("$",""))
+}
+
+trait AreaRoot extends Area{
+  setName("")
 }
--- a/core/src/main/scala/spinal/core/Area.scala
+++ b/core/src/main/scala/spinal/core/Area.scala
@@ -134,7 +134,7 @@ class ClockingArea(val clockDomain: ClockDomain) extends Area with PostInitCallb
 
 
 /**
-  * Clock Area with a specila clock enable
+  * Clock Area with a special clock enable
   */
 class ClockEnableArea(clockEnable: Bool) extends Area with PostInitCallback {
 
@@ -198,3 +198,7 @@ class ResetArea(reset: Bool, cumulative: Boolean) extends Area with PostInitCall
   }
 }
 
+
+trait AreaObject extends Area{
+  setWeakName(this.getClass.getSimpleName.replace("$",""))
+}
--- a/core/src/main/scala/spinal/core/BaseType.scala
+++ b/core/src/main/scala/spinal/core/BaseType.scala
@@ -44,6 +44,7 @@ object BaseType{
   final val isTypeNodeMask = 2
   final val isVitalMask    = 4
   final val isAnalogMask   = 8
+  final val isFrozen       = 16
 }
 
 /**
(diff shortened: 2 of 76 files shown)


Dolu1990:
Hi ^^

I'm not against it. But for a Blackbox.setInline feature, i would say we could have more.

let's me explaine : 
- Your actual implementation would be more like a separated API. More like a InlineVerilog(....), but not related to the BlackBox class
- For a complet inlined blackbox API i would say : 

```scala
object Adder extends BlackBoxImplementation{
  override def content () : String =   
    """module adder #(
      |    parameter WIDTH = 16
      |) (
      |    input      [WIDTH-1:0] ain    ,
      |    input      [WIDTH-1:0] bin    ,
      |    output reg [WIDTH-1:0] add_out
      |);
      |
      |always @(*) begin
      |  add_out = ain + bin;
      |end
      |
      |endmodule
    """.stripMargin)
}

class Adder(width: Int) extends BlackBox {
  addGeneric("WIDTH", width)

  val io = new Bundle {
    val ain = in UInt(width bits)
    val bin = in UInt(width bits)
    val add_out = out UInt(width bits)
  }

  setInline(BlackBoxImplementation)
}

```

This way, we clearly define one unique BlackBoxImplementation which can be reconnized by multiple Adder instances.
Then the SpinalHDL backend can build a set of all BlackBoxImplementation required, and add them either directly into the single generated file, wither into one file for eachones.

Would that make sens ?

Dolu1990:
Ahhh to be clear, the idea of BlackBoxImplementation, was to use the object itself as a unique key identifying the blackbox (not relying on the string itself).

There is also the possibility to have instead of def content, a def getVerilog    def getVhdl

Dolu1990:
Hi ^^

dxsm:
> What is the status of the PR ?

It has been implemented and you can merge it if you like the usage:
```scala
class adder(width: Int) extends BlackBox {
  addGeneric("WIDTH", width)

  val io = new Bundle {
    val ain = in UInt(width bits)
    val bin = in UInt(width bits)
    val add_out = out UInt(width bits)
  }
  noIoPrefix()

  setInline("adder.v",
    """module adder #(
      |    parameter WIDTH = 16
      |) (
      |    input      [WIDTH-1:0] ain    ,
      |    input      [WIDTH-1:0] bin    ,
      |    output reg [WIDTH-1:0] add_out
      |);
      |
      |always @(*) begin
      |  add_out = ain + bin;
      |end
      |
      |endmodule
    """.stripMargin)
}

class top extends Module {
  val io = new Bundle {
    val ain = in UInt(16 bits)
    val bin = in UInt(16 bits)
    val add_out = out UInt(16 bits)
  }
  noIoPrefix()
  var adder0 = new adder(16)
  io.ain <> adder0.io.ain
  io.bin <> adder0.io.bin
  io.add_out <> adder0.io.add_out
}
showRtl(new top())
```

Dolu1990:
I'm currently looking to merge it and release 1.6.1

Why creating a new file for the BB, instead of just genereting everything in the main file ?

PR #519, opened 2021-10-13
avoid using functions for bit assignment

dxsm:
For #513 , I change the implementation method. now
```scala
class top extends Component{
  val myBits = B(8 bits, (7 downto 5) -> B"101", 4 -> true, 3 -> True, default -> false)
}

object main {
    SpinalVerilog(new top())
  }
}
```
will generate
```verilog
module top (
);

  reg        [7:0]    myBits;
  wire       [7:0]    myBits_const;

  assign myBits_const = 8'h0;
  always @(*) begin
      myBits = myBits_const;
      myBits[7 : 5] = 3'b101;
      myBits[4] = 1'b1;
      myBits[3] = 1'b1;
  end


endmodule
```
I think this is a safe way without using function in verilog.


Dolu1990:
Ahhhhh ticky ticky ^^

Thanks :)

PR #523, opened 2021-10-28
Update InOutWrapper.scala to handle opendrain IO

azaparov:
IO  that is generated by I2C module is not handled by any case currently. This is one possibly a fix to allow inout ports for i2c on a hardware.


Dolu1990:
Ok thanks :)

I fixed it by using an alternative to the isMasterInterface
It wasn't trigering because the generator was using the toIo function which work at a flattened level.

PR #527, opened 2021-11-04
Adding readAndWrite to trait and documenting trait

allexoll:
added the readAndWrite to the trait, and implementing it for ECP too. added some documentation too


Dolu1990:
Thanks ^^

PR #528, opened 2021-11-06
make terminate unused channels easier.

Readon:
this commit is to fit the gap while connecting Axi4ReadOnly bus to Axi4.
the setIdle function is added by the way, to terminate it.

two situations are covered.
if the master bus is in type Axi4, then user should use toReadOnly function.
if the master bus is in type Axi4ReadOnly, then user should use toAxi4 function.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -461,6 +461,13 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
     val last = counter.willOverflowIfInc
     return addFragmentLast(last)
   }
+  
+  def setIdle(): this.type = {
+    this.valid.assignDontCare()
+    this.ready.assignDontCare()
+    this.payload.assignDontCare()
+    this
+  }
 
   override def getTypeString = getClass.getSimpleName + "[" + this.payload.getClass.getSimpleName + "]"
 }
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -463,9 +463,13 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
   }
   
   def setIdle(): this.type = {
-    this.valid.assignDontCare()
-    this.ready.assignDontCare()
+    this.valid := False
     this.payload.assignDontCare()
+    this
+  }
+  
+  def setBlocked(): this.type = {
+    this.ready := False
     this
   }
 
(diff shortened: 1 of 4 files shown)


lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -461,6 +461,13 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
     val last = counter.willOverflowIfInc
     return addFragmentLast(last)
   }
+  
+  def setIdle(): this.type = {
+    this.valid.assignDontCare()
```
Dolu1990, line 466: Shouldn't it be this.valid := False ?

Dolu1990: Else all good i think ^^

Readon: The problem is that stream can be a mater as well as slave. In each case, the signal to be assigned should be different.
valid is only assigned to false when the stream is a sink. otherwise, only ready should be assigned.
I am not that familiar with the utility system of SpinalHDL, so I just tried assignDontCare and it works.^^

Readon: Do you have any suggestion on it? 
actually i wonder that how can i identify the stream is a source or sink? 
if there are functions can help to get that infomation, things can be reformed to
```
if(this.isMaster) this.valid := False else this.ready := False
```

Dolu1990: Hmmm, i would say, setIdle should always be from a master perspective, so always this.valid := False

as from a slave perspective it would be a myStream.setNotReady() or something in that kind ?

Readon: > Hmmm, i would say, setIdle should always be from a master perspective, so always this.valid := False
> 
> as from a slave perspective it would be a myStream.setNotReady() or something in that kind ?

How about setBlocked which stands for the stream is blocked by slave.
NotReady seems could not represent the stream is always unready.

Dolu1990: Good for me ^^

Readon:
Sorry, I don't know why it will push to this PR very time i push to my repo.

Dolu1990:
I guess that's github which automaticaly take everything from the branch ^^
For each PR you need a separated branch :)

let's me know when things are ready to merge

Readon:
> I guess that's github which automaticaly take everything from the branch ^^ For each PR you need a separated branch :)
> 
> let's me know when things are ready to merge

ok, can you revert the last commit? I think the PRs before that works, after my test manually.

Dolu1990:
Thanks ^^

PR #531, opened 2021-11-15
add extender to extend the input transfer into several outputs.

Readon:
There are several cases that extend one input transaction into several is required.
For example, some instructions can be executed in multiple times.
The stream extender is designed to fit those needs easily, especially for full pipelined usage.
It's worth noting that the final output trasaction number should be count + 1.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1491,3 +1491,62 @@ object StreamFifoMultiChannelBench extends App{
 
   Bench(rtls, targets)
 }
+
+object StreamExtender {
+    def apply[T <: Data](input: Stream[T], count: UInt)(
+        driver: (UInt, T) => T = (_: UInt, p: T) => p
+    ): Stream[T] = {
+        val c = new StreamExtender(input.payloadType, input.payloadType, count.getBitsWidth, driver)
+        c.io.input << input
+        c.io.count := count
+        c.io.output
+    }
+
+    def apply[T <: Data, T2 <: Data](input: Stream[T], output: Stream[T2], count: UInt)(
+        driver: (UInt, T) => T2
+    ): StreamExtender[T, T2] = {
+        val c = new StreamExtender(input.payloadType, output.payloadType, count.getBitsWidth, driver)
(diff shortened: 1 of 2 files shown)


Readon:
is there any difference between the requirements of SpinalHDL lib and design simulation? I have test the extender in our own repository, however, when I tried to move it into the lib repo, it fails with errors listed below.
```
 java.lang.AssertionError: assertion failed: Compilation of SharedMemIface.cpp failed
at scala.Predef$.assert(Predef.scala:170)
at spinal.sim.VpiBackend.delayed_compilation$lzycompute(VpiBackend.scala:88)
at spinal.sim.VpiBackend.delayed_compilation(VpiBackend.scala:76)
at spinal.sim.VpiBackend.instanciate_(VpiBackend.scala:133)
at spinal.sim.VpiBackend.instanciate(VpiBackend.scala:156)
at spinal.sim.VpiBackend.instanciate(VpiBackend.scala:168)
at spinal.sim.SimVpi.<init>(SimVpi.scala:15)
```

Dolu1990:
Hi,

> java.lang.AssertionError: assertion failed: Compilation of SharedMemIface.cpp failed

that's is weird. there should be no difference with spinal.lib
To have a better idea of the issue, you can enable verbose output in the SimConfig by : SimConfig.withLogging.doSim ...

Dolu1990:
About the PR itself, what is the difference with https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/Stream.scala#L1232 ? 

Mostly that with this implementation you can specify in hardware the number of splits via io.count right ?

Dolu1990:
Hmm, what's about StreamSerializerDynamic as a name ? Or something it this kind ? StreamExtender may be too generic ?

Readon:
> StreamExtender may be too generic ?

Serializer might be confused with a parallel to serial component, although it could be one application of this. :)

Readon:
> Hmm, what's about StreamSerializerDynamic as a name ? Or something it this kind ? StreamExtender may be too generic ?

How about StreamTransactionExtender ?

Dolu1990:
Good ^^

Thanks !

PR #533, opened 2021-11-20
Add support for write byte enable in BusSlaveFactory.writeMemWordAligned

jiegec:


lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
```diff
@@ -620,14 +620,17 @@ trait BusSlaveFactory extends Area{
                                      bitOffset     : Int = 0) : Mem[T] = {
     val mapping    = SizeMapping(addressOffset,mem.wordCount << log2Up(busDataWidth / 8))
     val memAddress = writeAddress(mapping) >> log2Up(busDataWidth / 8)
-    val port       = mem.writePort
+    val port       = mem.writePortWithMask
```
Dolu1990, line 623: The issue i have is that it enfore writePortWithMask for cases which doesn't require it (writeByteEnable == null)
If possible, when possible, it should keep the mem.writePort to ease the memory inferation.

jiegec: Done.

Dolu1990:
Cool thanks ^^

PR #545, opened 2021-12-12
Axi4 on chip ram

Readon:
legacy on chip ram has too much combination code to debug. this is a stream based one that fix bugs when tester is a free run master agent.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
@@ -150,7 +150,7 @@ case class Axi4SharedOnChipRamPort(config: Axi4Config) extends ImplicitArea[Axi4
 
             readStream.translateFrom(readAddrStream) { (to, from) =>
                 if (axi.config.useId) to.id := from.id
-                to.last := from.last
+                if (axi.config.useLast) to.last := from.last
                 to.data := 0
                 to.setOKAY()
                 if (axi.config.useRUser) to.user := from.user
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
@@ -92,7 +92,8 @@ case class Axi4SharedOnChipRamPort(config: Axi4Config) extends Bundle {
             if (axi.config.useWUser) to.user := from._1.user
             to.setOKAY()
         }
-        writeStream.ready := True
+        val writeReady = RegInit(False).riseWhen(True)
+        writeStream.ready := writeReady
 
         val bStream      = cloneOf(writeStream)
         val writeLast    = writeCombined.fire && writeCombined._1.last
@@ -143,22 +144,32 @@ case class Axi4SharedOnChipRamPort(config: Axi4Config) extends Bundle {
         axi.arw.ready.noBackendCombMerge //Verilator perf
         axi
     }
+
+    def withSoftResetOf(ram: Mem[Bits], softReset: Bool): Axi4Shared = {
+        val clock = ClockDomain.current.copy(softReset = softReset)
+        val area = new ClockingArea(clock) {
+
+            val axi = attach(ram)
+        }
+        area.axi
+    }
 }
 
 object Axi4SharedOnChipRamMultiPort {
-   def apply(portCount : Int, dataWidth : Int,byteCount : BigInt,idWidth : Int) = {
-       val axiConfig = Axi4SharedOnChipRam.getAxiConfig(dataWidth,byteCount,idWidth)
-       val wordCount = byteCount / axiConfig.bytePerWord
-       new Axi4SharedOnChipRamMultiPort(axiConfig, wordCount, portCount)
-   }
-   def apply(config: Axi4Config, wordCount: BigInt, portCount: Int) = new Axi4SharedOnChipRamMultiPort(config, wordCount, portCount)
+    def apply(portCount: Int, dataWidth: Int, byteCount: BigInt, idWidth: Int) = {
+        val axiConfig = Axi4SharedOnChipRam.getAxiConfig(dataWidth, byteCount, idWidth)
+        val wordCount = byteCount / axiConfig.bytePerWord
+        new Axi4SharedOnChipRamMultiPort(axiConfig, wordCount, portCount)
+    }
+    def apply(config: Axi4Config, wordCount: BigInt, portCount: Int) =
+        new Axi4SharedOnChipRamMultiPort(config, wordCount, portCount)
 }
 
 class Axi4SharedOnChipRamMultiPort(config: Axi4Config, wordCount: BigInt, portCount: Int) extends Component {
     val io = new Bundle {
         val axis = Vec(slave(Axi4Shared(config)), portCount)
     }
-    
+
     if (config.useLock || config.useRegion || config.useCache || config.useProt || config.useQos)
         SpinalWarning("Axi4SharedOnChipRamMultiPort might not support Axi4 Lock, Region, Cahce, Prot and Qos featrues!")
 
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
@@ -102,9 +102,9 @@ case class Axi4SharedOnChipRamPort(config: Axi4Config) extends ImplicitArea[Axi4
             )
 
             val writeCombined = StreamJoin(writeAddrStream, writeDataStream)
-            val addr     = arw.addr(wordRange)
+            val address       = arw.addr(wordRange).resize(ram.addressWidth bits)
             ram.write(
-                address = addr.resized,
+                address = address,
                 data = writeCombined._2.data,
                 enable = writeCombined.fire,
                 mask = writeCombined._2.strb
@@ -133,7 +133,7 @@ case class Axi4SharedOnChipRamPort(config: Axi4Config) extends ImplicitArea[Axi4
 
             val readData = ram
                 .readSync(
-                    address = addr.resized
+                    address = address
                 )
                 .resized
             val savedReadData = Reg(cloneOf(readData))
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
@@ -64,84 +64,87 @@ case class Axi4SharedOnChipRamPort(config: Axi4Config) extends Bundle {
     def attach(ram: Mem[Bits]): Axi4Shared = {
         val wordRange       = config.wordRange
         val axi             = Axi4Shared(config)
-        val arw             = axi.arw.unburstify
-        val writeDataStream = axi.writeData
+        
+        val logic = new Composite(axi, "logic"){
+            val arw             = axi.arw.unburstify
+            val writeDataStream = axi.writeData
 
-        val readStream      = Stream(Axi4R(axi.config))
-        val readAddrStream  = cloneOf(arw)
-        val writeStream     = Stream(Axi4B(axi.config))
-        val writeAddrStream = cloneOf(arw)
+            val readStream      = Stream(Axi4R(axi.config))
+            val readAddrStream  = cloneOf(arw)
+            val writeStream     = Stream(Axi4B(axi.config))
+            val writeAddrStream = cloneOf(arw)
 
-        Vec(readAddrStream, writeAddrStream) <> StreamDemux(
-            arw,
-            U(arw.write),
-            2
-        )
-
-        val writeCombined = StreamJoin(writeAddrStream, writeDataStream)
-        val writeAddr     = writeCombined._1.addr(wordRange)
-        ram.write(
-            address = writeAddr.resized,
-            data = writeCombined._2.data,
-            enable = writeCombined.fire,
-            mask = writeCombined._2.strb
-        )
-
-        writeStream.translateFrom(writeCombined) { (to, from) =>
-            if (axi.config.useId) to.id := from._1.id
-            if (axi.config.useWUser) to.user := from._1.user
-            to.setOKAY()
-        }
-        val writeReady = RegInit(False).riseWhen(True)
-        writeStream.ready := writeReady
-
-        val bStream      = cloneOf(writeStream)
-        val writeLast    = writeCombined.fire && writeCombined._1.last
-        val writePayload = RegNextWhen(writeStream.payload, writeLast)
-        val writeDone    = RegInit(False)
-        when(writeLast) {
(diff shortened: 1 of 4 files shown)


lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
```diff
@@ -60,60 +60,123 @@ case class Axi4SharedOnChipRam(dataWidth : Int, byteCount : BigInt, idWidth : In
 }
 
 
-//WARNING, do not support backpresure
-case class Axi4SharedOnChipRamMultiPort(portCount : Int, dataWidth : Int,byteCount : BigInt,idWidth : Int) extends Component{
-  val axiConfig = Axi4SharedOnChipRam.getAxiConfig(dataWidth,byteCount,idWidth)
+case class Axi4SharedOnChipRamPort(config: Axi4Config) extends Bundle {
```
Dolu1990, line 63: Why extends Bundle ?, i mean that class has only functions right ?

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
```diff
@@ -60,60 +60,123 @@ case class Axi4SharedOnChipRam(dataWidth : Int, byteCount : BigInt, idWidth : In
...
+    def attach(ram: Mem[Bits]): Axi4Shared = {
```
Dolu1990, line 64: One issue is that this function will generate much logic without naming any of its content.

What's about 
```scala
def attach(ram: Mem[Bits]): Axi4Shared = {
   val axi = Axi4Shared(config)
   val logic = new Composite(axi, "logic"){
      
      ...
   }
   axi
}
```

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
```diff
@@ -60,60 +60,123 @@ case class Axi4SharedOnChipRam(dataWidth : Int, byteCount : BigInt, idWidth : In
...
+            writeDone := False
+        }
+        bStream.valid := writeDone
+        bStream.payload := writePayload
+        axi.writeRsp << bStream
+
+        val readAddr = readAddrStream.addr(wordRange)
+        val readData = ram
+            .readSync(
+                address = readAddr.resized
```
Dolu1990, line 114: I'm a bit worrying about the inferation, basicaly, the behind Axi4SharedOnChipRamPort, was to use a true-dual-port ram to infer a two port axi ram.

Here, the address ued for the readSync and the write aren't the same signal, this may confuse synthesis ?

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: IMO, the port is a bundle of wires which have specific functions.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: OK, I am not very familiar with the name policy. However, attach a name is really great, while debugging.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: Does that means readSync and write should use the same address signal to help synthesizer correctly infer a true-dual-port?

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: I have tried this Composite in attach. However, I could not find any difference after adding that.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: I tried this in Vivado, there is no trouble to it's synthesizer which generate a correct netlist. A real dual block ram has been initiated.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: I suppose that it is because the usage of Axi4SharedOnChipRamPort, where a vector but a standalone signal is assigned.
And I also tried to create new area, unfortunately it does not helps.
```
val ports = Vec(Axi4SharedOnChipRamPort(config), portCount)
```

Readon: I have tried if an individual assignment is applies as below, the name system works.
```
       val axis1 = ports(1).attach(ram)
```
But it would fail when
```
for (i <- 0 until portCount) {
        val axi = ports(i).attach(ram)
        io.axis(i) >> axi
}
```
So I suppose that the problem is the compatibility between name system and container iteration.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Dolu1990: Issue is, vivado may be a bit less stupid than other synthesis tools, but in general, it is a realy pain in the ass.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Dolu1990: very important to attache names ^^ 

```scala
val miaou = for (i <- 0 until portCount) yield new Area{
        val axi = ports(i).attach(ram)
        io.axis(i) >> axi
}
````

File: lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala (hunk 3 above)
Comment (Readon) [reply lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala:114]: ahh，I will use a standalone addr to do that

File: lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala (hunk 1 above)
Comment (Dolu1990) [reply lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala:63]: the Axi4SharedOnChipRamPort class is a collection of function which use the same config i would say, so, i would say, they can be moved into a scala object, with the added config parameter as arguement (trying to avoid this pattern where you have to create a class just to hide passing one parameter later on)

Comment (Readon) [reply lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala:63]: do you mean use an object pattern, as below
```
object Axi4SharedOnChipRamPort{
   def apply(config: Axi4Config):
        .....
}

class Axi4SharedOnChipRamPort extends xxxx{
   ..
}
```
The problem is that if I extends a Component class the io is required to connect ram, which seems make it too complicated.

Comment (Dolu1990) [reply lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala:63]: no, i mean : 

```scala
object Axi4SharedOnChipRamPort{
   def attach(config: Axi4Config, ram: Mem[Bits]): Axi4Shared = { ... }
}
```

Readon: I hide the attach function with a direct object initialization.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: The lazy initialization Handle strategy has been adopted to create the port logic while attaching. And it's still accessable through object property so that the name can be kept.
It seems that calling a function will not let the reference hold the name of logic.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
```diff
@@ -60,60 +61,137 @@ case class Axi4SharedOnChipRam(dataWidth : Int, byteCount : BigInt, idWidth : In
...
+            val wordRange       = config.wordRange
+            val arw             = axi.arw.unburstify
+            val writeDataStream = axi.writeData
+
+            val readStream      = Stream(Axi4R(axi.config))
+            val readAddrStream  = cloneOf(arw)
+            val writeStream     = Stream(Axi4B(axi.config))
+            val writeAddrStream = cloneOf(arw)
+
+            Vec(readAddrStream, writeAddrStream) <> StreamDemux(
```
Dolu1990, line 98: Why forking the stream on the address side ? instead of forking it at the response side ? 
Was the bug of the original implementation related to the unburstify ?

Readon:
the fork is intend to avoid comb logic with !write or write. the principle of this version is too deal with one feature once. so that the response side logic should be independent for better understanding.
I did not investigate the reason cause bugs of old version due to its complex comb logic.&nbsp;





---Original---
From: ***@***.***&gt;
Date: Sun, Dec 26, 2021 17:18 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Axi4 on chip ram (PR #545)




 
@Dolu1990 commented on this pull request.
 
 
In lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala:
 &gt; +    override def implicitValue: Axi4Shared = this.axi + +    val ram = Handle[Mem[Bits]] + +    val logic = Handle { +        new Area { +            val wordRange       = config.wordRange +            val arw             = axi.arw.unburstify +            val writeDataStream = axi.writeData + +            val readStream      = Stream(Axi4R(axi.config)) +            val readAddrStream  = cloneOf(arw) +            val writeStream     = Stream(Axi4B(axi.config)) +            val writeAddrStream = cloneOf(arw) + +            Vec(readAddrStream, writeAddrStream) <&gt; StreamDemux(  
Why forking the stream on the address side ? instead of forking it at the response side ?
 Was the bug of the original implementation related to the unburstify ?
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
```diff
@@ -60,60 +61,137 @@ case class Axi4SharedOnChipRam(dataWidth : Int, byteCount : BigInt, idWidth : In
...
+            bStream.valid := writeDone
+            bStream.payload := writePayload
+            axi.writeRsp << bStream
+
+            val readData = ram
+                .readSync(
+                    address = addr.resized
+                )
+                .resized
+            val savedReadData = Reg(cloneOf(readData))
```
Dolu1990, line 139: Why adding this buffer here ? Would it be possible to avoid it ?

Readon:
it's because that new readAddr could gen data in next clock it issued，but address changing might happen any time. so it's necessary to cache the data in buffer to avoid the affect that caused by address changing before transaction is fired。



---Original---
From: ***@***.***&gt;
Date: Wed, Jan 5, 2022 18:11 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Axi4 on chip ram (PR #545)




 
@Dolu1990 commented on this pull request.
 
 
In lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala:
 &gt; +            val writeDone    = RegInit(False) +            when(writeLast) { +                writeDone := True +            } elsewhen (bStream.fire) { +                writeDone := False +            } +            bStream.valid := writeDone +            bStream.payload := writePayload +            axi.writeRsp << bStream + +            val readData = ram +                .readSync( +                    address = addr.resized +                ) +                .resized +            val savedReadData = Reg(cloneOf(readData))  
Why adding this buffer here ? Would it be possible to avoid it ?
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4SharedOnChipRam.scala
Readon: The key point is that when input stream is fired the address would changed, so that the data is updated one cycle after that. but the output might not fire right on the next cycle, so that the cache reg is necessary.

Dolu1990:
hmmm, i think it was one of the trade off of the previous implementation, not handeling backpresure to avoid this buffer, while having a easy to infer RW ram port.

Ultimately, it is possible to not have that buffer if the  ram RW port only update its readed value after a  arw.fire

But my guess is that for systems without backpresure, ready always driven high, the synthesis tool will optimize that buffer away. so no producing overhead, which is good.

So one very last thing before the merge, the address used on the readSync and write, should realy be the same signal.
Actualy, you used the resized keyword, which return a new signal with resizeing capabilities.
That's why from a SpinalHDL point of view both port will use different address signal, which will mess up the memory port RW detection.
You realy need one single signal to drive those two address :)

Readon:
> hmmm, i think it was one of the trade off of the previous implementation, not handeling backpresure to avoid this buffer, while having a easy to infer RW ram port.
> 
> Ultimately, it is possible to not have that buffer if the ram RW port only update its readed value after a arw.fire
> 
> But my guess is that for systems without backpresure, ready always driven high, the synthesis tool will optimize that buffer away. so no producing overhead, which is good.
> 
> So one very last thing before the merge, the address used on the readSync and write, should realy be the same signal. Actualy, you used the resized keyword, which return a new signal with resizeing capabilities. That's why from a SpinalHDL point of view both port will use different address signal, which will mess up the memory port RW detection. You realy need one single signal to drive those two address :)

I have moved the resize to definition to generate a unified address for one port.

Readon:
> hmmm, i think it was one of the trade off of the previous implementation, not handeling backpresure to avoid this buffer, while having a easy to infer RW ram port.
> 
> Ultimately, it is possible to not have that buffer if the ram RW port only update its readed value after a arw.fire
> 
> But my guess is that for systems without backpresure, ready always driven high, the synthesis tool will optimize that buffer away. so no producing overhead, which is good.
> 
> So one very last thing before the merge, the address used on the readSync and write, should realy be the same signal. Actualy, you used the resized keyword, which return a new signal with resizeing capabilities. That's why from a SpinalHDL point of view both port will use different address signal, which will mess up the memory port RW detection. You realy need one single signal to drive those two address :)

I did not pay attention to the resource usage of this buffer previously. 
If it is important, then adding a explicit parameter is better than implicit. ^^
However, for a standard Axi4 bus, the buffer is necessary.

Dolu1990:
> If it is important, then adding a explicit parameter is better than implicit. ^^

That's fine, if it become a issue we can fix it.

Thanks ^^

PR #547, opened 2021-12-15
[feature] Component add stub method for empty component use

jijingg:
@Dolu1990  as we talk in gitter . i add a stub method for component to get an empty Component

i have try to remove useless statement in component to instead `SpinalConfig(removePruned = true)` but faild . 

could you help me to add that

### DIFF
--- a/core/src/main/scala/spinal/core/Component.scala
+++ b/core/src/main/scala/spinal/core/Component.scala
@@ -340,4 +340,23 @@ abstract class Component extends NameableByComponent with ContextUser with Scala
     ctx.restore()
     ret
   }
+
+  /**
+    * Empty Component, remove logic in component and assign zero on output port as stub
+    * @example {{{ val dut = (new MyComponent).stub }}}
+    */
+  def stub: this.type = this.rework{
+    this.getAllIo.foreach{ p =>
+      if (p.isOutput | p.isInOut) {
+        p.removeAssignments()
+        p match {
+          case x: Bool      => x := False
+          case x: BitVector => x.clearAll()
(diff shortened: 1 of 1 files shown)


Dolu1990:
Ahhh it was more tricky than what i was thinking.
So, clearning the children list of component isn't enough, it is also required to remove all connection to it.

Here is a version which reaaaaaly cleanup everything. It has to be done in two phase, (remove then patch) as when we iterate on stuff with the foreachStatements, it is difended to remove another node than the current one.

```scala
    def stub: this.type = this.rework{
      //First remove all we don't want
      this.children.clear()
      this.dslBody.foreachStatements{
        case bt : BaseType if !bt.isDirectionLess =>
        case s => s.removeStatement()
      }

      //Then patch things
      this.dslBody.foreachStatements{
        case bt : BaseType if bt.isOutput | bt.isInOut => {
          bt := bt.getZero
        }
        case s =>
      }
      this
    }
 ```
 
 let's me know if you have any question about it ^^

jijingg:
> It has to be done in two phase, (remove then patch) as when we iterate on stuff with the foreachStatements, it is defended to remove another node than the current one or to add nodes.

here need iterate foreachStatement two times , I understand why caused me to fail all the time. 
now it work perfect

Basically, you have completed all the implementations of this PR , I just recorded it , haha ..   @Dolu1990

Dolu1990:
No worries ^^
Just one thing, adding the brackets to def stub()
As this function has a side effect it is better for clarity.

jijingg:
> Just one thing, adding the brackets to def stub()

done !

Dolu1990:
thanks ^^

PR #550, opened 2021-12-17
Unify fixed-point into a single class

kleinai:
This is my first pass at creating a single fixed-point class. This is a work in progress but I'd appreciate feedback.
Rationale for this change:
- Simply data storage types
- Mixed signed and unsigned operations

### DIFF
--- a/core/src/main/scala/spinal/core/BaseType.scala
+++ b/core/src/main/scala/spinal/core/BaseType.scala
@@ -31,13 +31,13 @@ trait TypeFactory{
 /**
   * Base type factory
   */
-trait BaseTypeFactory extends BoolFactory with BitsFactory with UIntFactory with SIntFactory with VecFactory with SFixFactory with UFixFactory
+trait BaseTypeFactory extends BoolFactory with BitsFactory with UIntFactory with SIntFactory with VecFactory
 
 
 /**
   * Base type Cast
   */
-trait BaseTypeCast extends SFixCast with UFixCast
+trait BaseTypeCast
 
 object BaseType{
   final val isRegMask      = 1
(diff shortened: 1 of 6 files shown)


Dolu1990:
Hi ^^
So, personnaly, there is some things i wasn't realy happy of UFix SFix, mainly that they are separated API. But also, for it works on bit count, not on value ranges. I mean let's imagine: 

```scala
val a = Number(signed(8 exp, 10 bits))
val b = Number(max = 3, min = 0)
val result = a + b //Number(max = 255.75+3, min = -256)
```

basicaly, every operation will keep the exact result range of things. That way we can chain stuff with exactly what we need as a result.

So a kind of completly different API

For you, what were the incentives to unify fixed-point ?

kleinai:
I'm interested in making it easier to define generic math algorithms (FIR, IIR, FFT, etc) in a way that could mix type inputs. I thought it might be easier to make fixed-point into a single class that can represent all aspects and functions of fixed-point numbers. In this way we can leverage scala's functional programming style.

I tried to adhere to the pre-existing Q-format type while providing most of the functions defined by the number interface.
The backing type would be bits up until sign-ness needs to be resolved into native HDL types to allow for synthesis optimizations.

Does that all seem to make sense or would you prefer something different? At the very least I'd like to dissolve SFix and UFix.

Dolu1990:
It make sense, my main concerns are : 
- Not breaking the existing API for the already existing user codes
- Trying to patch something flawed (the actual implementation), while the ultimate solution may lay somewhere else (new universal number class which hold the min/max value + resolution)

> The backing type would be bits up until sign-ness needs to be resolved into native HDL types to allow for synthesis optimizations.

Yes ^^

PR #555, opened 2021-12-20
patches to adopt Axi4 options in a wide range.

Readon:
The configuration with useLen = false works only when the last signal is not sensitive. It cannot run a variable-length transaction determined by the last signal.


Dolu1990:
Thanks ^^

PR #558, opened 2021-12-25
fix globalPrefix bypass by setDefinitionName issue fix

### LINKED ISSUE #557: globalPrefix not available when  setDefinitionName is called
hi, @Dolu1990  `SpinalConfig(globalPrefix = "xx")` not available when the module use `setDefinitionName`
is there any way give the `goblaPrefix`  with high priority add prefix to the Component Name Whether Component has setDefinitionName or not.

```scala
object paly25 extends App{
  class SubModule extends Component {
    this.setDefinitionName("MyModule")
    val a = in Bits(8 bit)
    val b = out Bits(8 bit)
    b := a
  }
  class  Top extends Component {
    val io = new Bundle {
      val a = in Bits(8 bit)
      val b = out Bits(8 bit)
    }
    val dut = new SubModule
    dut.a  := io.a
    io.b := dut.b
  }
  SpinalConfig(globalPrefix = "xxx").generateVerilog(new Top)
}
```
The expectation is  `xxMyModule`  but get `MyModule `

jijingg:
@Dolu1990  I think `globalPrefix` should add pre to any Component except Blackbox. If you have no opinion, I add a PR to fix #557.

### DIFF
--- a/core/src/main/scala/spinal/core/internals/Phase.scala
+++ b/core/src/main/scala/spinal/core/internals/Phase.scala
@@ -924,13 +924,15 @@ class PhaseNameNodesByReflection(pc: PhaseContext) extends PhaseMisc{
     }
     for (c <- sortedComponents) {
       if(c != topLevel) {
-        if (c.definitionName == null) {
-          val pre = c match {
+        val pre = c match {
             case t: BlackBox => ""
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks ^^

PR #562, opened 2021-12-28
Fix Axi4 signal overview

7FM:
This fixes the header of the Axi4 signal overview table (read data and address channels were swapped).


Dolu1990:
Thanks :)

PR #566, opened 2021-12-29
Unify fixed-point into a single class - Rewrite

kleinai:
Rewrite of #550 but leaving the original SFix and UFix as they are.
This still needs functional simulations and unit tests before I'm confident with it.

### DIFF
--- a/core/src/main/scala/spinal/core/FixedPoint.scala
+++ b/core/src/main/scala/spinal/core/FixedPoint.scala
@@ -136,6 +136,298 @@ object getFixSym{
 //}
 
 
+/**
+ * Fixed-point number representation.
+ *
+ * Underlying bits are:
+ * Fix = N.(M-N)
+ * Where:
+ * N = intWidth
+ * M = bitWidth
+ *
+ * @param intWidth Integer part bit width
+ * @param bitWidth Total number bit width
+ * @param signed   Interpret number as signed if true
+ */
+class Fix(val intWidth: Int, val bitWidth: Int, val signed: Boolean) extends MultiData with Num[Fix] {
+  override def elements: ArrayBuffer[(String, Data)] = ???
+
+  def signWidth = if (signed) 1 else 0
+  /** Total binary width without sign */
+  def numericWidth = bitWidth - signWidth
+  /** Whole binary width without sign bit */
+  def wholeWidth = intWidth - signWidth
+  /** Fractional binary width */
+  def fracWidth = bitWidth - intWidth
+
+  val raw: Bits = Bits(bitWidth bit)
+
+  override def tag(q: QFormat): Fix = ???
+
+  /** Aligns the decimal point of two fixed-point numbers */
+  protected def align(left: Fix, right: Fix): (Bits, Bits) = {
+    val shift = left.fracWidth - right.fracWidth
+    if (shift > 0)
+      (left.raw, right.raw << shift)
+    if (shift < 0)
+      (left.raw << -shift, right.raw)
+    (left.raw, right.raw)
+  }
+
+  /** Addition */
+  override def +(right: Fix): Fix = {
+    val (_left, _right) = align(this, right)
+    val res = new Fix(this.intWidth + right.intWidth, this.bitWidth + right.bitWidth, this.signed | right.signed)
+    res.raw := (_left.asUInt + _right.asUInt).asBits
+    res
+  }
+
+  /** Safe Addition with 1 bit expand */
+  override def +^(right: Fix): Fix = this.expand(1) + right.expand(1)
+
+  /** Safe Addition with saturation */
+  override def +|(right: Fix): Fix = {
+    val expandedAdd = this +^ right
+    val realigned = expandedAdd >> (expandedAdd.fracWidth - this.fracWidth)
+    realigned.sat(this.bitWidth)
+  }
+
+  /** Subtraction */
+  override def -(right: Fix): Fix = {
+    val (_left, _right) = align(this, right)
+    val res = new Fix(this.intWidth + right.intWidth, this.bitWidth + right.bitWidth, this.signed | right.signed)
+    res.raw := ((this.signed, right.signed) match {
+      case (false, false) => _left.asUInt - _right.asUInt
+      case (false, true) => _left.asUInt.intoSInt - _right.asSInt
+      case (true, false) => _left.asSInt - _right.asUInt.intoSInt
+      case (true, true) => _left.asSInt - _right.asSInt
+    }).asBits.resize(res.bitWidth)
+    res
+  }
+
+  /** Safe Subtraction with 1 bit expand */
+  override def -^(right: Fix): Fix = this.expand(1) - right.expand(1)
+
+  /** Safe Subtraction with saturation */
+  override def -|(right: Fix): Fix = {
+    val expandedAdd = this -^ right
+    val realigned = expandedAdd >> (expandedAdd.fracWidth - this.fracWidth)
+    realigned.sat(this.bitWidth)
+  }
+
+  /** Multiplication */
+  override def *(right: Fix): Fix = {
+    val (_left, _right) = align(this, right)
+    // A + B + 1 if result is signed
+    val res = new Fix(this.intWidth + right.intWidth + (this.signWidth | right.signWidth),
+                      this.fracWidth + right.fracWidth + (this.signWidth | right.signWidth),
+                      this.signed | right.signed)
+    res.raw := (_left.asUInt * _right.asUInt).asBits
+    if (res.signed)
+      res.raw(res.raw.high) := _left.high ^ _right.high
+    res
+  }
+
+  /** Division */
+  override def /(right: Fix): Fix = ???
+
+  /** Modulo */
+  override def %(right: Fix): Fix = ???
+
+  /** Is less than right */
+  override def <(right: Fix): Bool = {
+    (this.signed, right.signed) match {
+      case (false, false) => this.raw.asUInt < this.raw.asUInt
+      case (false, true) => this.raw.asUInt.intoSInt < this.raw.asSInt
+      case (true, false) => this.raw.asSInt < this.raw.asUInt.intoSInt
+      case (true, true) => this.raw.asSInt < this.raw.asSInt
+    }
+  }
+
+  /** Is equal or less than right */
+  override def <=(right: Fix): Bool = {
+    (this.signed, right.signed) match {
+      case (false, false) => this.raw.asUInt <= this.raw.asUInt
+      case (false, true) => this.raw.asUInt.intoSInt <= this.raw.asSInt
+      case (true, false) => this.raw.asSInt <= this.raw.asUInt.intoSInt
+      case (true, true) => this.raw.asSInt <= this.raw.asSInt
+    }
+  }
+
+  /** Is greater than right */
+  override def >(right: Fix): Bool = {
+    (this.signed, right.signed) match {
+      case (false, false) => this.raw.asUInt > this.raw.asUInt
+      case (false, true) => this.raw.asUInt.intoSInt > this.raw.asSInt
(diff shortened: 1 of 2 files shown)


Dolu1990:
What's about having the Fix class keeping a accurate track of its value range ? Instead of only knowing  the number of bits, it would know up to which values it can go. This would allow operators to only add bits when required.

For instance : 

```scala
val value = Fix(peak = 8 exp, resolution = -2 exp, signed = false)
val tmp = value + 1 //Result => peak = 9 exp
val result = tmp + 1 //Result => peak = 9 exp  (not adding bit here, as 9 bits is enough)
```

More over, this would may the  Fix(peak = 8 exp, resolution = -2 exp, signed = false) definition a abstraction over the true nature of how the range is stored in the Fix class.

kleinai:
I'll need a little more explanation of the `peak` and `resolution` parameters, and how those factor into underlying bit width.
Currently I'm working on finishing the bit-centric `Fix` re-implementation. We could rename this class and make `Fix` a smart wrapper.

Dolu1990:
> I'll need a little more explanation of the peak and resolution parameters, and how those factor into underlying bit width.

So, i'm not sure of the exact implementation, but the idea is mostly that for each given hardware instances of that number type, it keep track of the min max practicaly possible.

So likely, internaly it would need to store (in scala, rempacing maxExp: Int, bitCount: Int) :
- resolution :  Int     for instance -3  would mean 0.125
- max : BigInt    for instance 56 would mean up to  56*2^resolution
- min  : BigInt    for instance 56 would mean down to -56*2^resolution
I think with that kind of internal storage it should be fine.

As for the hardware storage, a single Bits, as you did.

Does it make sense ?

kleinai:
Unit tests are finished. It checked out locally so I'll see what Travis thinks of it. It only works in VHDL because Verilog and VHDL modulo outputs differ and VHDL is closer to Python.
Next up is adding the value range tracking.

Dolu1990:
I would say, don't use hardware modulo and divide at all on synthtisable expression, is there now way to workaround it ?

Dolu1990:
What's about the : 

> So likely, internaly it would need to store (in scala, rempacing maxExp: Int, bitCount: Int) :

resolution : Int for instance -3 would mean 0.125
max : BigInt for instance 56 would mean up to 56*2^resolution
min : BigInt for instance 56 would mean down to -56*2^resolution
I think with that kind of internal storage it should be fine.

?

kleinai:
I spent some time trying to implement a division function but it's proving to be difficult to debug quickly.
I've also been thinking about the value tracking. The arithmetic functions will automatically resize for the appropriate output representation. I'll need to rework the bit expand add `+^` and subtract `-^` to avoid adding additional unnecessary bits.
I'll add some values which return the decimal representable ranges and granularity.
Also useful, I've been thinking it should be possible to extend the Fix class to use device specific functions, those should handle bit widths as optimally as possible.

kleinai:
> Division is always a special thing, "most of the time" multi cycle. So better to no have it in a divide / module operator.

I figured that out quick. N number of N-bit subtractions will never close timing.

> If we start to put into the scala type system device related information, then we lose parametrability.

Good point. Things could get messy very quickly.

>let's me know if you want to do a call about it ^^

Is your github email up-to-date?

core/src/main/scala/spinal/core/FixedPoint.scala
```diff
@@ -883,6 +886,236 @@ class Fix(val intWidth: Int, val bitWidth: Int, val signed: Boolean) extends Mul
...
+    val maxValue = BigDecimal(2).pow(f.wholeWidth)-BigDecimal(2).pow(-f.fracWidth)
+    val minvalue = if (f.signed) -BigDecimal(2).pow(f.wholeWidth) else BigDecimal(0)
+    val ret = new AutoFix(maxValue, minvalue, -f.fracWidth exp)
+    ret.raw := f.raw
+    ret
+  }
+
+}
+
+class AutoFix(val maxValue: BigDecimal, val minValue: BigDecimal, val exp: ExpNumber) extends MultiData {
```
Dolu1990, line 901: Hi ^^

I would say, use BigInt instead of BigDecimal, as BigInt is just the raw storage of a number, while BigDecimal as some "funky" stuff attached.

core/src/main/scala/spinal/core/FixedPoint.scala
```diff
@@ -135,8 +138,1437 @@ object getFixSym{
...
+      case (true, false) => _left.asSInt % _right.asUInt.intoSInt
+      case (true, true) => _left.asSInt % _right.asSInt
+    }).resize(res.bitWidth).asBits
+    res
+  }
+
+  /** Is less than right */
+  override def <(right: Fix): Bool = {
+    (this.signed, right.signed) match {
+      case (false, false) => this.raw.asUInt < this.raw.asUInt
```
Dolu1990, line 353: this.raw.asUInt < this.raw.asUInt :D ?

core/src/main/scala/spinal/core/FixedPoint.scala
```diff
@@ -135,8 +138,1437 @@ object getFixSym{
...
+  private def trimOrExpand(b: Bits, w: Int): Bits = {
+    if (b.getWidth > w) {
+      b.takeLow(w)
+    } else {
+      b
+    }
+  }
+
+  def +(right: AutoFix): AutoFix = {
+    val (lMax, lMin, rMax, rMin) = alignRanges(this, right)
```
Dolu1990, line 979: <3

core/src/main/scala/spinal/core/FixedPoint.scala
```diff
@@ -135,8 +138,1437 @@ object getFixSym{
...
+  raw.setRefOwner(this)
+  raw.setPartialName("", weak = true)
+
+  override def elements: ArrayBuffer[(String, Data)] = {
+    ArrayBuffer("" -> raw)
+  }
+
+  // These are used for reference tracking
+  private var parents: Seq[AutoFix] = Nil
+  private val children: mutable.Set[AutoFix] = mutable.Set()
```
Dolu1990, line 943: For which purpose is this children ?

core/src/main/scala/spinal/core/FixedPoint.scala
kleinai: Copy and paste error, oops

core/src/main/scala/spinal/core/FixedPoint.scala
kleinai: The thought was that in the future it might be used to graph out a chain of operations for analysis. I can remove it. It has no purpose at the moment.

Dolu1990: Ahhh, no worries, we can keep it ^^

Dolu1990:
So overall how things where going ?
Is there things you are happy / not happy with ?

kleinai:
With the tests in place I believe it's functionally ready. This covers adding, subtracting, multiplying with the bit growth and value range calculations in place. This also doesn't break anything for existing SFix or UFix code.
I don't think I'll need to add anything else at the moment. If I find things I'll bundle them up and make another PR.

kleinai:
I take that back. There's still a bug in the multiplication width calculations...

kleinai:
Threw in an extra fix I think might be safer

Dolu1990:
Look nice ^^

Few idea, what's about : 
- renaming AutoFix into AFix 
- Adding a few construction altenatives : 

```scala
    AFix.u(12 bits)            //Q12.0
    AFix.u(8 exp, 12 bits)     //Q8.4
    AFix.u(8 exp, -4 exp)      //Q8.4
    AFix.u(255  , -4 exp)      //Q8.4

    AFix.s(12 bits)             //Q11.0 + sign bit
    AFix.s(8 exp, 12 bits)      //Q8.4  + sign bit
    AFix.s(8 exp, -4 exp)       //Q8.4  + sign bit
    AFix.s(255  , -256, -4 exp) //Q8.4
 ```

Dolu1990:
Hmm one thing is how we want to handle assignements. For instance :

- Q8.8 := Q12.12   Should that trigger a error, because we lose information ?
- Q8.8 := Q4.4   Should that trigger a error because we add padding ?

If i remember well, the UFix SFix were using a .truncated tag (Q8.8 := Q12.12.truncated) to allow losing information, while padding was alowed by default.

What is you opinon on this ?

kleinai:
I think it should be hard to lose information without being explicit. For example to reduce bits you will need to use saturation or rounding functions.

Padding should be unnecessary as that's handled by the AFix class mechanisms automagically but since that's the case there's no reason to disallow it. A warning at most.

Padding could be use case for the operation graphing stubs in the future. Detecting when a number is over-padded.

Dolu1990:
> I think it should be hard to lose information without being explicit. For example to reduce bits you will need to use saturation or rounding functions.

Right

> Padding should be unnecessary as that's handled by the AFix class mechanisms automagically but since that's the case there's no reason to disallow it. A warning at most.

Ok ^^, maybe the warning as a not enabled by default option ?

> Padding could be use case for the operation graphing stubs in the future. Detecting when a number is over-padded.

Sorry, i'm not sure to understand ? Do you have a example ?

Can i push stuff directly into your branch ?

kleinai:
> > Padding should be unnecessary as that's handled by the AFix class mechanisms automagically but since that's the case there's no reason to disallow it. A warning at most.
> 
> Ok ^^, maybe the warning as a not enabled by default option ?

Sounds like the path I'd choose. Warnings would primarily be for auditing.

> > Padding could be use case for the operation graphing stubs in the future. Detecting when a number is over-padded.
> 
> Sorry, i'm not sure to understand ? Do you have a example ?

No example but this falls under what I imagined an analyzer tool could do.

> Can i push stuff directly into your branch ?

Sure!

Dolu1990:
Hi ^^

Would you have time one of those days for a call ?

Dolu1990:
(just pushed https://github.com/kleinai/SpinalHDL/commit/a3e92d366c0ec561f1df5371b3fc1f3c35f7e7c6 on a separated branch for now)

kleinai:
I feel this is just about done and ready for review.

Do we want to keep around the original fixed-point code rewrite or only keep the new auto fixed-point code? I'd personally prefer people only use the new code.

Dolu1990:
> Do we want to keep around the original fixed-point code rewrite or only keep the new auto fixed-point code? 

I would say, let's keep the orginal fixed-point thing as it was, to avoid breaking anything. and only keep the new auto fixed point code :)

Dolu1990:
I will take some times in the incoming days to play around the syntax ^^
Just right now i have too many in flight things :/

Dolu1990:
> Is the class argument change expected to change anything logically? I assume this is a refactor and a retyping of exp.

Right, just refractoring, now behaviour changes ^^

> Some of those functions covered older usages I thought I might need.

I did a few mistakes in the original SFix UFix, we don't realy need to stick to it ^^

kleinai:
> @kleinai Hi ^^ I added the skeleton for the assignement adjustement API, see the following as a example: [10c7ae3#diff-7fb6876d56b3e0039538609441e87aa9193b098e60d10d9907ba51647bfbcfc2R52](https://github.com/SpinalHDL/SpinalHDL/commit/10c7ae336f552b4381c9121fef5669921f5e4cfe#diff-7fb6876d56b3e0039538609441e87aa9193b098e60d10d9907ba51647bfbcfc2R52)
> 
> The actual behaviour/handeling of it isn't implemented yet.
> 
> Does that seems good to you ?
> 
> Also i added the AFix UQ/SQ

Yes this seems good.

I played round with adding rounding to specific decimal points (shift, round, shift back). It's more flexible to allow a programmer to assign a value with rounding to avoid manually counting bits.

Dolu1990:
Rawrrrr

wswslzp:
Will there be a document for AFix?

kleinai:
> Will there be a document for AFix?

There isn't at the moment but I will put one together.

PR #570, opened 2022-01-03
Component stub clock/reset removed bug fix

jijingg:
@Dolu1990 , Component stub method clock/reset accident removed bug fix

### DIFF
--- a/core/src/main/scala/spinal/core/Component.scala
+++ b/core/src/main/scala/spinal/core/Component.scala
@@ -346,6 +346,9 @@ abstract class Component extends NameableByComponent with ContextUser with Scala
     * @example {{{ val dut = (new MyComponent).stub() }}}
     */
   def stub(): this.type = this.rework{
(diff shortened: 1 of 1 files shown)


Dolu1990:
hmmm using a phase thing there seems to me quite a bit dangerous.

What about instead doing something like ClockDomain.current.readClockWire, same for reset/clockEnable,softReset (if the signal existe)

jijingg:
> hmmm using a phase thing there seems to me quite a bit dangerous.
> 
> What about instead doing something like ClockDomain.current.readClockWire, same for reset/clockEnable,softReset (if the signal existe)

I was thinking about it.  If there are multiple clocks, even submodules use `Clockdomain.external`, in those case, I don't know how to walk all clockdomain out.

Dolu1990:
I isolated the behaviour of the phase in some API, you can use it that way PhasePullClockDomains.recursive(this)

jijingg:
@Dolu1990  done!

PR #577, opened 2022-01-10
[RFC] Add support for multiple CS pin for SDRAM controller

t123yh:
This patch adds support for multiple CS pin, useful if anyone wants 128MB or 256MB with SDRAM. It's tested on a board with 4 IS42S16320F SDRAM chips, forming 32bit x 2CS structure.

Note that it currently only works with the sdr controller, not xdr controller. I'm not sure how to modify xdr controller, so any help is appreciated.


Dolu1990:
I guess 
https://github.com/SpinalHDL/SpinalHDL/runs/4786705600?check_suite_focus=true#step:10:167
regs.CSn(0)   := io.ctrl.phases(0).CSn    should fix the compilation error

PR #582, opened 2022-01-15
Bundle(a, b, c, d)  := Bundle(x, y , z)

jijingg:
implement of Syntax for free Bundle to Bundle connection

```scala

Bundle(a, b, c, d)  := B(0, 32 bit)
Bundle(a, b, c, d)  := Bundle(e, f, g)
Bundle(a, b, c, d)  := Vec(....)
Bundle(a, b, c, d)  := io.data
Bundle(a, b, c, d)  := U(3211, 32 bit)
```


Dolu1990:
Hi ^^

Hmm, what are the differences with the existing : 

```scala
 (a, b, c, d) := B(1832, 32 bit)
```
?

(it isn't a well known feature XD. From https://github.com/SpinalHDL/SpinalHDL/blob/bf604af246f561dfd00bade6a1ba14d7e988efdd/core/src/main/scala/spinal/core/core.scala#L463)

Dolu1990:
So the existing feature need Bits  :   (a, b, c, d) := U(3211, 32 bit).asBits

jijingg:
OK, my mistake , I never noticed that  existed.  

good , the tuples seems more elegant  👍

Dolu1990:
No worries, it is also a quite hidden feature.
One question is how you document such thing in a way people can find it XD ??

jijingg:
I'm sure someone in the gitter asked you more than once. Class method  is easy to found in source-code , those independency Object Method not easy to find them .

First, The most reliable source is [SpinalDoc](https://spinalhdl.github.io/SpinalDoc-RTD/master/index.html#), so the best way is still recode that, Of course, this is a great workload, which is difficult to complete by one person， you can encourage the questioner to update him in [SpinalDoc](https://spinalhdl.github.io/SpinalDoc-RTD/master/index.html#)  when he gets the answer

Second, maintain a FAQ database, but I'm not sure where should be located. Maybe it's git wiki, maybe it's a chapter in last of Spinal-Doc, Maybe you have a better idea. 

the FAQ-list can remove duplicated question.
otherwise these questions will be asked again and again, and then you need to answer them in the gitter again and again.

Dolu1990:
>  you can encourage the questioner to update him in SpinalDoc when he gets the answer

This is a good idea  i think ^^

> FAQ

What's about the cheat sheets  instead ?

jijingg:
> What's about the cheat sheets instead ?

good idea!  pdf  or html ?  where we can update it

Dolu1990:
They are : 

https://github.com/SpinalHDL/SpinalDoc/blob/master/cheatsheet/cheatSheet_symbolic.odt
https://github.com/SpinalHDL/SpinalDoc/blob/master/cheatsheet/cheatSheet_core.odt
https://github.com/SpinalHDL/SpinalDoc/blob/master/cheatsheet/cheatSheet_lib.odt

The online doc pointing to the samefolder's pdf

So, if you open them, you will possibly have arrangement / layout missmatch, i did those file long ago with open office if i remember well.

PR #586, opened 2022-01-18
add options support to Axi4 unbursitfy.

Readon:
if size, len and burst is not available on bus than use default value described in specification.


Dolu1990:
Thanks ^^

PR #589, opened 2022-01-19
api renaming,  bigInt --> toBigInt

jijingg:
Api naming adjust. At present, there is not much influence， just give a better name .

Keep in sync with documents :  https://github.com/SpinalHDL/SpinalDoc-RTD/pull/71


Dolu1990:
<3

PR #590, opened 2022-01-20
core: Units: Hertz-/TimeNumber: Rework decompose

dnltz:
Use match in TimeNumber's decompose method and add a new decompose
method to HertzNumber.

Signed-off-by: Daniel Schultz <d.schultz@phytec.de>


Dolu1990:
Thanks ^^

PR #623, opened 2022-02-27
add pipelined function to support timing optimization.

Readon:
- [ ] Did you add some test code?
           Based on stream pipeline which seems stable enough. This would only increase the resource usage but logic.
- [ ] Did you add appropriate documentation in docs/src?
           TBD
- [ ] Did you state the API impact?
           An API borrow from BMB pipelined. Where should I state the API?
- [ ] Did you specify the code generation impact?
- [ ] Did you request a desired merge strategy?
- [ ] Did you add text to be included in the Release

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -4,6 +4,28 @@ import spinal.core._
 import spinal.lib.eda.bench.{AlteraStdTargets, Bench, Rtl, XilinxStdTargets}
 import scala.collection.Seq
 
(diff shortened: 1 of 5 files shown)


Readon:
The problem now is that if I add the StreamPipe version pipelined function to Bmb bus, then all arguments of the function could not be set at the same time. Because there are already exist a pipelined function with full default arguments.

Dolu1990:
Hooooo f*** XD

Yes, i forgot that scala do not like multiple function with the same name and default arguement :/

So don't bother :)

Thanks for the PR !

PR #630, opened 2022-03-01
Correct MaskMapping hit() function

### LINKED ISSUE #628: MaskMapping Bug?
https://github.com/SpinalHDL/SpinalHDL/blob/2ab339a36cfd2d3f8b902e1254dfc65d895e7a42/lib/src/main/scala/spinal/lib/bus/misc/Misc.scala#L50-L55

Is `(address & base) === mask` correct here? Shouldn't the inverted mask be applied then compared to the base address?

Furthermore, `removeOffset` will not work because inverting `mask` directly will result in a negative number (I tried it myself).

This works as a possible fix:
```
override def hit(address: UInt): Bool = (address & ~U(mask).resize(address.getWidth)) === base
```

Although I'm sure there's a more elegant solution.

kleinai:
Fixes #628

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/misc/Misc.scala
+++ b/lib/src/main/scala/spinal/lib/bus/misc/Misc.scala
@@ -47,9 +47,18 @@ case class SingleMapping(address : BigInt) extends AddressMapping{
 }
 
 
+/**
+ * Creates an address mapping using a bit mask.
+ *
+ * MaskMapping(0x0000, 0x7FFF) => matches 0x0000-0x8000
+ * MaskMapping(0x80, 0xF) => matches 0x80 - 0x8F
+ *
+ * @param base Address offset to use. Must be outside of the mask
+ * @param mask Bit mask of bits to accept in the mapping
+ */
 case class MaskMapping(base : BigInt,mask : BigInt) extends AddressMapping{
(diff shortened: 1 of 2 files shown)


Dolu1990:
Hoo i think there is a mixup here.

```scala
override def hit(address: UInt): Bool = (address & mask) === base
override def removeOffset(address: UInt) = address & ~U(mask, widthOf(address) bits)
```

isn't it ?

kleinai:
Would it not be easier to specify the bits to mask off? Those that should be within the mapping.
`MaskMapping(0x400, 0x0FF)` is seems more clear when dealing with unknown address widths.

Dolu1990:
> Would it not be easier to specify the bits to mask off?

Ahhh, that's right ^^, so :


```scala
  override def hit(address: UInt): Bool = (address & ~U(mask, widthOf(address) bits)) === base
  override def removeOffset(address: UInt) = address & ~U(mask, widthOf(address) bits)
```

For the memory mapping of the usb device controller, this should be good : 

```scala
val memoryMapping = MaskMapping(0x0000, 0x7FFF)
```

Dolu1990:
remove offset => address & ~U(mask, widthOf(address) bits) 

(~) :D

Dolu1990:
hmm, now that i'm thinking about it, remove offset may be the reverse of what i was thinking

Dolu1990:
You are right, XD

Thanks !

PR #632, opened 2022-03-06
rename stream function '~~' to map

pwang7:


Dolu1990:
Hooo map is a good idea, Thanks :)

PR #637, opened 2022-03-09
Fix assertion in wishbone `>>` operator

Jeff-Ciesielski:
According to docs, `master >> slave` should allow for slaves with address widths
<= master, however this assertion fires in that particular case.


Jeff-Ciesielski:
Haven't been able to fully test this as I'm fighting with VexRiscV integration.

Jeff-Ciesielski:
Tested with a working wishbone interconnect + Vex core.

Dolu1990:
Thanks ^^

PR #642, opened 2022-03-11
Try to split huge Stream file into several parts.

Readon:
Questions exists:
1. EventFactory seems off topic of Stream file. It just use the Stream. where to put it is better?
2. EventEmitter has a totally non-related function. The place to put it is a problem.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1267,6 +1267,15 @@ class StreamCCByToggle[T <: Data](dataType: HardType[T],
   }
 }
 
+/**
+ * Enumeration to present order of slices.
+ */
+sealed trait SlicesOrder
+/** Slice with lower bits process first */
+object LOWER_FIRST extends SlicesOrder
+/** Slice with higher bits process first */
+object HIGHER_FIRST extends SlicesOrder
+
 object StreamWidthAdapter {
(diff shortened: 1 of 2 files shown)


Readon:
then the utility would contains about 1.5k lines.



---Original---
From: ***@***.***&gt;
Date: Fri, Mar 11, 2022 20:45 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Try to split huge Stream file into severalparts. (PR #642)




 
hmm
 What's about having Stream.scala with the Stream itself, and SteamUtilities.scala for all the rest?
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

Readon:
The splitting principle is to put utilities with 1 input and N outputs together, and N inputs and 1 output together. The FIFO related classes could be standalone facility.

Dolu1990:
More generaly, the question is what is the right partitioning of the source code.

Maybe it depend the workflow that people are using ?

personnaly, (with intellij) the current partitioning isn't a issue as the IDE can navigate me through the sources fast. (shift + shift =>  class name => HIT)

The goal of the partionning into more smaller files is to help visibility when exploring a unknown code base ?

If yes, maybe the scaladoc is the best way to processed ?
http://spinalhdl.github.io/SpinalHDL/#package

PR #645, opened 2022-03-14
add driver to flow for simulation.

Readon:
The implementation is heavily borrowed the code from StreamDriver, which seems work fine.


Dolu1990:
Thanks ^^

PR #656, opened 2022-03-20
add order parameter for StreamWidthAdapter with corresponding test.

Readon:
fix the problem that when padding = true the width miss match reported.
the reason is that resized function could not get correct bits width from assignFromBits function.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1267,6 +1267,15 @@ class StreamCCByToggle[T <: Data](dataType: HardType[T],
   }
 }
 
+/**
+ * Enumeration to present order of slices.
+ */
+sealed trait SlicesOrder
+/** Slice with lower bits process first */
+object LOWER_FIRST extends SlicesOrder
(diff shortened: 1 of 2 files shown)


Readon:
- [x] Did you add some test code?
 - [x] Did you add appropriate documentation in docs/src?
Wait the merge of this patch.
 - [x] Did you state the API impact?
 - [ ] Did you specify the code generation impact?
     consistent with old style.
 - [ ] Did you request a desired merge strategy?
     no 
 - [ ] Did you add text to be included in the Release
     What does this means?

Dolu1990:
hmm weird, do you have an example of configuration where it was failing ?

Readon:
&nbsp;you can test it with the test case with padding in scalatest dir.



---Original---
From: ***@***.***&gt;
Date: Fri, Mar 25, 2022 18:05 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] add order parameter for StreamWidthAdapterwith corresponding test. (PR #656)




 
hmm weird, do you have an example of configuration where it was failing ?

Dolu1990:
Thanks ^^

PR #661, opened 2022-03-23
Add basic Axi4 Stream support

kleinai:
Starter code for true Axi4 Streams. Goes beyond what functionality Bsb had because Axi4 Stream may use both strobe and keep bits. Although no support for eliminating null bytes from streams was planned as part of this PR.


Readon:
Good work!
In my opinion, something can be considered.
It's best to keep naming style consistent with other Axi4 related bus.
1. Such as tdata => data, tstrb => strb, and so on.
2. There are bytesPerWord for Axi4 which have the same meaning as dataBytes of Axi4StreamConfig. It's derived from the dataWidth in the Axi4Config.

Could you please give an example on what is the endianness in a stream-like bus? 
Normally, it's coupled with address.

Some other suggests is as below:
1. the config is a good mechanism to support custom setting, so why not check if the signal is used by evaluating config.useId by a direct stream.tid != null?

How about add some test?

One more concern is that if the implicit classes like RichStream, RichStreamFragment would have naming conflict in future?
That is out of my knowledge base.^^

kleinai:
Updated naming. Couldn't get `keep` as that's already used :(

> Could you please give an example on what is the endianness in a stream-like bus?
> Normally, it's coupled with address.

Because Axi4 streams only support whole byte width buses any data less than a byte will need to be padded. The endianness setting chooses where the padding bits are. I think I've only seen this with numerical data but since any Spinal data type could be converted to a byte stream it should be supported generally.

> How about add some test?

There's not much functionally going on. It's a fancy bundle type. I might get around to contributing some stream infrastructure in the future.
My driving factor here was to provide an easy conversion to an Axi stream like interface for Vivado.

> One more concern is that if the implicit classes like RichStream, RichStreamFragment would have naming conflict in future?

I don't think these will be a problem. If it is it can be renamed when there's a conflict. It beats names like `StreamPimper` IMO.

Readon:
> Updated naming. Couldn't get `keep` as that's already used :(
> 
> > Could you please give an example on what is the endianness in a stream-like bus?
> > Normally, it's coupled with address.
> 
> Because Axi4 streams only support whole byte width buses any data less than a byte will need to be padded. The endianness setting chooses where the padding bits are. I think I've only seen this with numerical data but since any Spinal data type could be converted to a byte stream it should be supported generally.

For the definition of the padding bits, I suggest use HIGHxxxx and LOWxxxx. 
However, it really needs a consistent clear enum for that.
I think the definition of SlicesOrder in #656 could not cover this also. 
Do you have any idea on this? @Dolu1990 

> 
> > How about add some test?
> 
> There's not much functionally going on. It's a fancy bundle type. I might get around to contributing some stream infrastructure in the future. My driving factor here was to provide an easy conversion to an Axi stream like interface for Vivado.

The test is not for the design now, but more important is for the changes would be made by the core and related libs in future.
The spinal lib itself should keep the logic consistency in a long run, then stable tests are non-trivial.

> 
> > One more concern is that if the implicit classes like RichStream, RichStreamFragment would have naming conflict in future?
> 
> I don't think these will be a problem. If it is it can be renamed when there's a conflict. It beats names like `StreamPimper` IMO.
XD, it is. How about name them with their purpose, like AxiS2StreamPimper, where xxxxPimper are all of the implicit classes.

Readon:
It seems that the tkeep is used. 
IMO, the best practice is not to mix two naming convention txxxx with direct signal name xxxx. 

So, how about direct use keep as the signal name?

Dolu1990:
> For the definition of the padding bits, I suggest use HIGHxxxx and LOWxxxx.
However, it really needs a consistent clear enum for that.
I think the definition of SlicesOrder in https://github.com/SpinalHDL/SpinalHDL/pull/656 could not cover this also.
Do you have any idea on this? @Dolu1990

So i think the endianness in this PR isn't realy about endianness. I would the "endianness" feature isn't realy a thing which should be there, as if it is there we could argue about so many more alternative way to adapte Stream to AxiStream.

So i would say, the best would be to drop the endianness stuff from that PR and instead having the user providing a landa function (if they need).

Ex : 

myAxiStream.toStreamBits().map(_.as(RGB(8,8,8))
myAxiStream.toStreamBits().map((_ >> 8).as(RGB(8,8,8))

?

else i would say the current implementation is more about "truncateLsb, truncateMsb" ?

kleinai:
I removed the endian code. It does make more sense to let the programmer deal with that then for Spinal to support it itself.
Also removed the implicit classes as I don't see the need to always have the Axi4Stream conversion available. This makes the imports a little nicer too.

The problem with `keep` is that it's already used in the `Data` base type. We either leave it as-is or rename everything to match `t*`. I don't have a preference either way.

Dolu1990:
I just removed the Data.keep, as it was redondant with dontSimplifyIt, mostly a artefact of a time where SpinalHDL wasn't preserving named signals by default,  and because keep is a precious keyword

Feel free to go the keep way ^^

Readon:
that's good. I has wondered about why keep was used.:)



---Original---
From: ***@***.***&gt;
Date: Wed, Mar 30, 2022 22:20 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Add basic Axi4 Stream support (PR #661)




 
I just removed the Data.keep, as it was redondant with dontSimplifyIt, mostly a artefact of a time where SpinalHDL wasn't preserving named signals by default,  and because keep is a precious keyword
 
Feel free to go the keep way ^^

PR #662, opened 2022-03-25
Axi4SlaveFactory: writeHaltRequest stall on last burst txn

kleinai:
While I was working with adding Axi4 write backpressure to a design I noticed that when using single non-burst transactions (always asserting last) the `writeHaltRequest` isn't considered. This should be taken into account even on the last beat of a burst in case some FIFO can't accept the data.


Dolu1990:
Thanks :D

PR #671, opened 2022-03-31
discard issue catching from oshi on M1 Mac

allexoll:
The issue [#1462](https://github.com/oshi/oshi/issues/1462) was fixed so this catch is no longer necessary if we bump the dependency version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,8 +25,8 @@ val defaultSettings = Defaults.coreDefaultSettings ++ xerial.sbt.Sonatype.sonaty
 
   libraryDependencies += "org.scala-lang" % "scala-library" % scalaVersion.value,
 
-  dependencyOverrides += "net.java.dev.jna" % "jna" % "4.2.2",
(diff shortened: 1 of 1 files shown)


allexoll:
yes, i need to check what's up with that, I wonder if there was an API change between 3.4.0 and 6.1.5, i'll see if I have time to do that later today

dbwiddis:
> yes, i need to check what's up with that, I wonder if there was an API change between 3.4.0 and 6.1.5, i'll see if I have time to do that later today

You need to update JNA to the version matching OSHI.  6.1.5 used 5.10.0; 6.1.6 uses 5.11.0.
>   dependencyOverrides += "net.java.dev.jna" % "jna" % "4.2.2",
>  dependencyOverrides += "net.java.dev.jna" % "jna-platform" % "4.2.2",

allexoll:
ah thanks @dbwiddis  that fixed the issue!

@Dolu1990 ping since this has been fixed?

Dolu1990:
Thanks ^^

PR #673, opened 2022-04-03
Add a `check` method to Stream.

losfair:
This PR adds a `check` method to `Stream` that enforces [Stream semantics](https://spinalhdl.github.io/SpinalDoc-RTD/dev/SpinalHDL/Libraries/stream.html#semantics) using assertions.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -505,6 +505,24 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
   }.next
 
   override def getTypeString = getClass.getSimpleName + "[" + this.payload.getClass.getSimpleName + "]"
+
+  def check(payloadInvariance: Boolean = false): this.type = {
+    val rValid = RegInit(False) 
+    val rData = RegNextWhen(this.payload, this.valid && !rValid)
+
+    rValid setWhen(this.valid && !rValid) clearWhen(this.fire)
+
+    val stack = Thread.currentThread().getStackTrace().mkString
+    assert(!(!this.valid && rValid), "Stream transaction disappeared: " + stack)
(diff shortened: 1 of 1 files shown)


lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -505,6 +505,24 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
...
   override def getTypeString = getClass.getSimpleName + "[" + this.payload.getClass.getSimpleName + "]"
+
+  def check(payloadInvariance: Boolean = false): this.type = {
+    val rValid = RegInit(False) 
+    val rData = RegNextWhen(this.payload, this.valid && !rValid)
+
+    rValid setWhen(this.valid && !rValid) clearWhen(this.fire)
+
+    val stack = Thread.currentThread().getStackTrace().mkString
+    assert(!(!this.valid && rValid), "Stream transaction disappeared: " + stack)
```
kleinai, line 514: Maybe use `ScalaLocated.long` to be consistent with other code?

Dolu1990:
Nice :D

One thing, about the "\\n". Normaly in dev there is already a change which remove the need of doing it, you can just do "\n".

Can you check on your side ?

Readon:
Could you please clarify what to be checked?
To me, it leads to misunderstanding when only function name "check" is provided.

Readon:
Could you please clarify what to be checked?
To me, it leads to misunderstanding when only function name "check" is provided.

Dolu1990:
Ahh, this make sense, maybe instead of "check", we should be more explicit, as : "withAsserts(xxx)" ?

losfair:
Renamed the method to `withAsserts` and added comments explaining its function.

Dolu1990:
Thanks :D

PR #675, opened 2022-04-05
Try to implement a comparison on UInt with wrapping nature.

Readon:
The bit width selected should guarantee the two UInt's distance would never exceed one quarter of the whole range.


Dolu1990:
Thanks ^^

PR #678, opened 2022-04-06
Memory: multiple ports with writeFirst should be permited.

nbstrike:
This allows logic to be generated that will allow address collisions to be resolved between a write & read port (eg to reduce latency of a FIFO).

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -1312,7 +1312,6 @@ end
                   emitWrite(b, memReadWrite.mem,s"${emitExpression(memReadWrite.chipSelect)} && ${emitExpression(memReadWrite.writeEnable)} ", memReadWrite.address, memReadWrite.data, memReadWrite.mask, memReadWrite.mem.getMemSymbolCount(), memReadWrite.mem.getMemSymbolWidth(),tab)
                 }, null, tmpBuilder, memReadWrite.clockDomain, false)
               case `writeFirst` =>
-                assert(mem.cldCount == 1)
                 emitClockedProcess((tab, b) => {
                   val symbolCount = memReadWrite.mem.getMemSymbolCount()
                   b ++= s"${tab}if(${emitExpression(memReadWrite.chipSelect)}) begin\n"
@@ -1325,7 +1324,6 @@ end
                   b ++= s"${tab}end\n"
                 }, null, tmpBuilder, memReadWrite.clockDomain, false)
               case `readFirst` =>
-                assert(mem.cldCount == 1)
                 emitClockedProcess((tab, b) => {
                   val symbolCount = memReadWrite.mem.getMemSymbolCount()
                   b ++= s"${tab}if(${emitExpression(memReadWrite.chipSelect)}) begin\n"


nbstrike:
The current implementation of `writeFirst` is reasonable, and matches vendor samples (eg https://www.xilinx.com/content/dam/xilinx/support/documents/sw_manuals/xilinx2021_2/ug901-vivado-synthesis.pdf page 122) -- the intent being to bypass the new data to the read side of the port. Given the current implementation that is emitted Xilinx's Vivado infers a port with WRITE_FIRST mode selected.

I don't believe that it's possible to infer ordering of port operations with HDL (except when instantiating vendor primitives directly, where the vendor may have some rules for the block) -- as such I don't believe that a user of SpinalHDL should expect any particular ordering between ports either.

However, being able to create multiple ports is useful as it becomes possible to resolve address collisions between port (with additional logic), eg:
```scala
val mem = Mem(dataType(), depth)
val collision = (rAddr === wAddr)
mem.write(wAddr, wData, wEn && (!collision || !rEn))
val rData = mem.readWriteSync(rAddr, wData, rEn, wEn && collision, duringWrite=writeFirst)
```
ie when writing at different addresses 2 ports are used, however if the addresses collide then use only the reading port and forward the data so that the reader will see the new data -- used for instance in a FIFO this would mean that there is no need to delay to the next delta cycle to implement a read to obtain that data (the current StreamFifo implementation would thus be able to reduce it's latency by a cycle when empty -- although that would require writeFirst to be implement for VHDL too).

nbstrike:
To push on this a bit more and address what is perhaps your concern, the semantics of writeFirst as implemented are correct, a read initiated in the same cycle as a write will result in the new data.

In addition to the non-blocking assignment to the memory, there is also a parallel (non-blocking) assignment to the output register when simultaneously reading & writing (ie a bypass path is created).

The main value of writeFirst is that it can be used to resolve address collisions between multiple ports, for which reason being restricted to a single port reduces the usefulness.

Dolu1990:
@nbstrike sorry, i losed track of the issue.
I agree.
So, maybe we could add a argument to the readWriteSync. Something like "I do not care about the ordering related ot the other ports".
That way we still keep a strong capture of what the user wish for, allowing blockboxification of a ram which need ordering across ports.

So, what's about having : 
                    readUnderWriteLocal: ReadUnderWritePolicy = dontCare,  // ordering for the given port 
                    readUnderWrite: ReadUnderWritePolicy = dontCare, // ordering against all other write ports
                    
That was readUnderWriteLocal = writeFirst with ReadUnderWritePolicy = dontCare  would be inferable

Would that make sense ?

Dolu1990:
> I don't believe it's worth adding an extra parameter as neither RTL nor SpinalHDL have a way to clearly identify inter-port dependencies

So, i would say, it depend. In NaxRiscv, i used to add new SpinalHDL elaboration to translate uninferable Mem into inferable ones. For instance, when a Mem has multiple write ports, the phase is translating it multiple simple dual ports ones.

Sure phase could easily use specific FPGA target IP / add bypass logic.
See : 
https://github.com/SpinalHDL/NaxRiscv/blob/969378560222ef310913addc8dfe4bda78834948/src/main/scala/naxriscv/Gen.scala#L346

So overall, the goal is to capture as precisly the behaviour that the user want, and handle it later on, depending the target, on some added SpinalHDL phases.

> It is clearly not possible to specify any useful relationship between ports when they are independently clocked

Right

> and even with a common clock it is not clear that any vendor can infer any deterministic inter-port relationship

Then bypass logic could be added by SpinalHDL via a custom phase.

> Because inter-port relationships seem to be very vendor specific perhaps it would be best for users to blackbox those memories to guarantee the behaviour they expect? 

Yes, that's kind of the actual goal. allowing as much as possible safe memory inferation, but then, rely on blackboxing / custom phases

so overall, adding that readDuringWriteLocal parameter would allow the user to specify  intra port stuff, (something inferable).

while readDuringWrite will stay the memory global behaviour, hardly inferable, which would normaly require blackboxing or something in that kind.

PR #679, opened 2022-04-06
AvalonMMSlaveFactory: allow byteEnable to be used

nbstrike:


lib/src/main/scala/spinal/lib/bus/avalon/AvalonMMSlaveFactory.scala
```diff
@@ -7,12 +7,13 @@ import scala.collection.Seq
 
 object AvalonMMSlaveFactory{
   def getAvalonConfig(addressWidth: Int,
-                      dataWidth: Int) = {
+                      dataWidth: Int,
+                      useByteEnable: Boolean) = {
```
Dolu1990, line 11: I think this should have default value false, to be backward compatible

nbstrike: fixed with additional commit.

Dolu1990:
Thanks ^^

PR #681, opened 2022-04-07
shuffle function for Vec.

### LINKED ISSUE #677: How about add a regroup function for BitVector?
sometimes, a regroup on bits is required, such as D[31:0] ==> {D[7:0], D[15:8], D[23:16], D[31:24]}.
how about add a regroup function for BitVector, so that the usage can be as below.
``` scala
     val a = Bits(32 bits)
     a.regroup(8 bits) { (i, total) =>
         total - 1 - i
     }
     // the implementation could be 
     def regroup(sliceWidth bits)(indexMap: (Int, Int) => Int) {
          val inVec = this.subdivideIn(sliceWidth)
          val outVec = cloneOf(inVec)
          for(i <- inVec.indexes) {
              outVec(indexMap(i)) := inVec(i)
          }
         val out = cloneOf(this)
         out.asBits := outVec.asBits   //this might require correction.
     }
```

Readon:
fix #677

### DIFF
--- a/core/src/main/scala/spinal/core/Vec.scala
+++ b/core/src/main/scala/spinal/core/Vec.scala
@@ -279,22 +279,6 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult
     elementsCache
   }
 
-  def shuffle(indexMapping: (Int) => Int): Vec[T] = {
-    val out = cloneOf(this)
-    for(i <- range){
-      out(indexMapping(i)) := vec(i)
-    }
-    out
-  }
-
-  def shuffleWithSize(indexMapping: (Int, Int) => Int): Vec[T] = {
-    val out = cloneOf(this)
-    for(i <- range){
-      out(indexMapping(length, i)) := vec(i)
-    }
-    out
-  }
-
   override def clone: this.type = new Vec[T](dataType, vec.map(cloneOf(_))).asInstanceOf[this.type]
 
   override def toString() = s"${getDisplayName()} : Vec of $length elements"
--- a/core/src/main/scala/spinal/core/Vec.scala
+++ b/core/src/main/scala/spinal/core/Vec.scala
@@ -279,11 +279,12 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult
     elementsCache
   }
 
-  def shuffle(indexMapping: (Int, Int) => Int) {
+  def shuffle(indexMapping: (Int, Int) => Int): Vec[T] = {
     val out = cloneOf(this)
     for(i <- range){
-      out(indexMapping(length, i)) := access(i)
+      out(indexMapping(length, i)) := vec(i)
     }
+    out
   }
 
   override def clone: this.type = new Vec[T](dataType, vec.map(cloneOf(_))).asInstanceOf[this.type]
--- a/core/src/main/scala/spinal/core/Vec.scala
+++ b/core/src/main/scala/spinal/core/Vec.scala
@@ -279,6 +279,13 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult
     elementsCache
   }
 
+  def shuffle(indexMapping: (Int, Int) => Int) {
+    val out = cloneOf(this)
+    for(i <- range){
+      out(indexMapping(length, i)) := access(i)
+    }
+  }
+
   override def clone: this.type = new Vec[T](dataType, vec.map(cloneOf(_))).asInstanceOf[this.type]
 
   override def toString() = s"${getDisplayName()} : Vec of $length elements"
--- a/core/src/main/scala/spinal/core/Vec.scala
+++ b/core/src/main/scala/spinal/core/Vec.scala
@@ -279,7 +279,15 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult
     elementsCache
   }
 
-  def shuffle(indexMapping: (Int, Int) => Int): Vec[T] = {
+  def shuffle(indexMapping: (Int) => Int): Vec[T] = {
+    val out = cloneOf(this)
+    for(i <- range){
+      out(indexMapping(i)) := vec(i)
+    }
+    out
+  }
+
+  def shuffleWithSize(indexMapping: (Int, Int) => Int): Vec[T] = {
     val out = cloneOf(this)
     for(i <- range){
       out(indexMapping(length, i)) := vec(i)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/spinal/core/Vec.scala
```diff
@@ -279,6 +279,14 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult
     elementsCache
   }
 
+  def shuffle(indexMapping: (Int, Int) => Int): Vec[T] = {
```
wswslzp, line 282: Why the `indexMapping` function takes two integer params?

Readon: one for total length, one for current index.

Readon: I think the shuffle in tester would explain more.

wswslzp: The name of the `indexMapping` function implies mapping one index to another index. Two integer params may confuse users who doesn't chain `shuffle` after `subdividenIn`, I think. 
Say your test case, `total` can be `data.getWidth/8` and not a necessary param.

wswslzp: An alternative implemenation of your test example, 
```scala
  outData := data.subdivideIn(8 bits).shuffle { i =>
    val total = data.getWidth / 8
    total - 1 -i
  }
```

Readon:
the problem is that subdivideIn would accept noninteger times dividing.



---Original---
From: ***@***.***&gt;
Date: Fri, Apr 8, 2022 13:01 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] shuffle function for Vec. (PR #681)




 
@wswslzp commented on this pull request.
 
 
In core/src/main/scala/spinal/core/Vec.scala:
 &gt; @@ -279,6 +279,14 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult      elementsCache    }   +  def shuffle(indexMapping: (Int, Int) =&gt; Int): Vec[T] = {  
An alternative implemenation of your test example,
   outData := data.subdivideIn(8 bits).shuffle { i =&gt;     val total = data.getWidth / 8     total - 1 -i   }

wswslzp:
> the problem is that subdivideIn would accept noninteger times dividing.
> […](#)
> ---Original--- From: ***@***.***&gt; Date: Fri, Apr 8, 2022 13:01 PM To: ***@***.***&gt;; Cc: ***@***.******@***.***&gt;; Subject: Re: [SpinalHDL/SpinalHDL] shuffle function for Vec. (PR #681) @wswslzp commented on this pull request. In core/src/main/scala/spinal/core/Vec.scala: &gt; @@ -279,6 +279,14 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult elementsCache } + def shuffle(indexMapping: (Int, Int) =&gt; Int): Vec[T] = { An alternative implemenation of your test example, outData := data.subdivideIn(8 bits).shuffle { i =&gt; val total = data.getWidth / 8 total - 1 -i } — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you authored the thread.Message ID: ***@***.***&gt;

I don't understand how your problem affects the `shuffle` implementation. 
1. As far as I know, the `subdivideIn` function cannot accept non-integer parameters. 
2. Still, even `subdivideIn` function takes non-integer parameters, `shuffle` function should not depends on `subdivideIn` function.

core/src/main/scala/spinal/core/Vec.scala
Dolu1990: i agree that the second intparameter seems unatural. (i understand it can be usefull, but then i would say it may be proposed into a second cuntion, ex shuffleWithSize)

Readon:
> > the problem is that subdivideIn would accept noninteger times dividing.
> > […](#)
> > ---Original--- From: _**@**_._**> Date: Fri, Apr 8, 2022 13:01 PM To: _**@**_.**_>; Cc: _**@**_._**_**@**_.**_>; Subject: Re: [SpinalHDL/SpinalHDL] shuffle function for Vec. (PR #681) @wswslzp commented on this pull request. In core/src/main/scala/spinal/core/Vec.scala: > @@ -279,6 +279,14 @@ class Vec[T <: Data](val dataType: HardType[T], val vec: Vector[T]) extends Mult elementsCache } + def shuffle(indexMapping: (Int, Int) => Int): Vec[T] = { An alternative implemenation of your test example, outData := data.subdivideIn(8 bits).shuffle { i => val total = data.getWidth / 8 total - 1 -i } — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you authored the thread.Message ID: _**@**_.***>
> 
> I don't understand how your problem affects the `shuffle` implementation.
> 
> 1. As far as I know, the `subdivideIn` function cannot accept non-integer parameters.
> 2. Still, even `subdivideIn` function takes non-integer parameters, `shuffle` function should not depends on `subdivideIn` function.

1. I think you misunderstood what I mean "divide integer times".
as the code below

subdivedeIn accept non restrict mode which allow non-integer times division.

2. the lambda function relying on external parameter is not a good practice, which also take side effect.
3. IMO, Dolu's solution is a good solution. Other suggestions are also welcome.:)

Dolu1990:
@Readon One thing, is if you add the two function into https://github.com/SpinalHDL/SpinalHDL/blob/51406f98d0bad890cfaad9d708e388cc0bed92ca/lib/src/main/scala/spinal/lib/Utils.scala#L927 to work on abstract Seq, it would work on all scala collections, including Vec ^^

It is triggered by https://github.com/SpinalHDL/SpinalHDL/blob/d01b62b0d5ab50ad5b2baace613b18ff18ed9d32/lib/src/main/scala/spinal/lib.scala#L41

So, the only requirement is to import spinal.lib._

I think moving it there would make sense, then all is good IMO

Readon:
> @Readon One thing, is if you add the two function into
> 
> https://github.com/SpinalHDL/SpinalHDL/blob/51406f98d0bad890cfaad9d708e388cc0bed92ca/lib/src/main/scala/spinal/lib/Utils.scala#L927
> 
> to work on abstract Seq, it would work on all scala collections, including Vec ^^
> It is triggered by
> 
> https://github.com/SpinalHDL/SpinalHDL/blob/d01b62b0d5ab50ad5b2baace613b18ff18ed9d32/lib/src/main/scala/spinal/lib.scala#L41
> 
> So, the only requirement is to import spinal.lib._
> 
> I think moving it there would make sense, then all is good IMO

The problem is that what to be returned.
If the returned value is ArrayBuffer by
``` scala
def shuffle(indexMapping: (Int) => Int): ArrayBuffer[T] = {
    val out = new ArrayBuffer[T]()
    for((v, i) <- pimped.zipWithIndex){
      out(indexMapping(i)) = v
    }
    out
}
```
then asBits could not be called directly.
If the returned value is Vec than it's a spinal specified component, which would be still constrained to Vec.

Dolu1990:
Yes right ^^
Great, thanks !

PR #682, opened 2022-04-08
Minor axi4 updates

dnltz:
- Add check for valid data width
- Fix channel indentation


kleinai:
Isn't WID only used in AXI 3?

kleinai:
If you do require AXI3-like WID support maybe a conversation component should be considered?

dnltz:
@kleinai  - I'm fine, thanks. I wrote an IP with an AXI4 port and was curious this signal is missing and did not read this small hint inside the table :smile: as more as I work with it, I understand why people don't like it ...

PR #684, opened 2022-04-11
Add map and translate payload shortcuts to Flow

kleinai:
Stream already has these. They can be directly applied to Flow.


Dolu1990:
Right, nice ^^

PR #685, opened 2022-04-11
RFC: Axi4S: Make AxiS type a subclass of stream for overrides

kleinai:
I was working on building an Axi stream component and realized the width resizing methods weren't actually applying.
After some time I found I couldn't get an implicit class to override the connect methods in `Stream`. Using a subclass seems to be the only pattern that works.

I'm just wondering what the thoughts on this change are. This doesn't appear to break using any of the existing `Stream[T]` functions like forks and joins.

I'll also have to add unit tests to make sure the Axi stream connect functions are enforced going forward. I missed this in the original tests.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala
@@ -41,6 +41,56 @@ object Axi4Stream {
     def isLast = if (this.last != null) this.last else False
 
     override def clone: Axi4StreamBundle = Axi4StreamBundle(config)
+
+    override def bundleAssign(that: Bundle)(f: (Data, Data) => Unit): Unit = {
+      that match {
+        case that : Axi4StreamBundle => {
+          assert(that.config.dataWidth <= this.config.dataWidth, s"Axi4Stream $that directly drives stream $this with smaller data width! (${that.config.dataWidth} > ${this.config.dataWidth})")
+          if (that.config.useId)
+            assert(that.config.idWidth <= this.config.idWidth, s"Axi4Stream $that directly drives stream $this with smaller ID width! (${that.config.idWidth} > ${this.config.idWidth})")
+          if (that.config.useDest)
+            assert(that.config.destWidth <= this.config.destWidth, s"Axi4Stream $that directly drives stream $this with smaller destination width! (${that.config.destWidth} > ${this.config.destWidth})")
+          if (that.config.useUser)
+            assert(that.config.userWidth <= this.config.userWidth, s"Axi4Stream $that directly drives stream $this with smaller user width! (${that.config.userWidth} > ${this.config.userWidth})")
+
+          this.data := that.data.resized
+          Axi4StreamBundlePriv.driveWeak(that,this,that.id,this.id, () => U(this.id.bitsRange -> false), allowResize = true, allowDrop = false)
+          Axi4StreamBundlePriv.driveWeak(that,this,that.strb,this.strb, () => B(this.strb.bitsRange -> true), allowResize = true, allowDrop = false)
+          Axi4StreamBundlePriv.driveWeak(that,this,that.keep,this.keep, () => B(this.keep.bitsRange -> true), allowResize = true, allowDrop = false)
+          Axi4StreamBundlePriv.driveWeak(that,this,that.last,this.last, () => False, allowResize = false,allowDrop = false)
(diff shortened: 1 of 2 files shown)


Dolu1990:
In general, i prefer composition over inheritence, i'm not that hot for extending the Stream class itself XD
If i remember well it was breaking some part of the SpinalHDL API, where it would make Stream[X] not compatible anymore with MyExtendedStream ?

About the issue in itself, with resizing method weren't applying ? I'm not sure to understand the context, Could you provide some code example of where the issue appeared ?

Thanks ^^

kleinai:
Understood. I didn't like doing it either but I just couldn't get the implicit class to work right.

An example case:
```
class AxisExample() extends Component {
  val io = new Bundle {
    val a = slave(Axi4Stream(Axi4StreamConfig(dataWidth = 4, useStrb = false)))
    val b = master(Axi4Stream(Axi4StreamConfig(dataWidth = 4, useStrb = true)))
  }

  io.b << io.a
}
```
Results in:
```
Bundle assignment is not complete. toplevel/io_b_payload : Axi4StreamBundle need 'strb' but toplevel/io_a_payload : Axi4StreamBundle doesn't provide it.
    spinal.lib.Stream.connectFrom(Stream.scala:294)
    spinal.lib.Stream.$less$less(Stream.scala:97)
    [snip]
    spinal.sim.JvmThread.run(SimManager.scala:51)
```


Now that I've thought about it, it might be better to provide a function to explicitly upsize the stream into a different config.

Dolu1990:
> Now that I've thought about it, it might be better to provide a function to explicitly upsize the stream into a different config.

That would work.

Another way we could do it, is to have a special trait that a Bundle could extends. Which would define a function to smoothly connect two of those bundle.

Then in the Stream.<< operator, we would check if the payload implement that trait, if yes we use that smooth connection function, else we use the regular := ( payload)

kleinai:
> Another way we could do it, is to have a special trait that a Bundle could extends. Which would define a function to smoothly connect two of those bundle.
> 
> Then in the Stream.<< operator, we would check if the payload implement that trait, if yes we use that smooth connection function, else we use the regular := ( payload)

I like this idea. It's cleaner than having to call a function before connecting. Although personally I'd prefer to keep base stream and flow as generic as possible. Maybe it could be worked into the bundle class itself?
I'll play around with some ideas.

Dolu1990:
> Maybe it could be worked into the bundle class itself?

Ahhh having the myBundle := anotherBundle being relaxed by a specific implementation of the user into the myBundle class ?

That could be a thing yes. there is the assignFromImpl function, could be made more frendly to override via a cleaner separated function.

Would that be good ?

kleinai:
Yes that sounds perfect! I was originally going to override assignFromImpl but it's private.

Dolu1990:
Got the following : 

```scala
  def bundleAssign(that : Bundle)(body : (Data, Data) => Unit): Unit ={
    for ((name, element) <- elements) {
      val other = that.find(name)
      if (other == null) {
        LocatedPendingError(s"Bundle assignment is not complete. $this need '$name' but $that doesn't provide it.")
      }
      else {
        body(element, other)
      }
    }
  }

  private[core] override def assignFromImpl(that: AnyRef, target: AnyRef, kind: AnyRef): Unit = {
    that match {
      case that: Bundle =>
        if (!this.getClass.isAssignableFrom(that.getClass)) SpinalError("Bundles must have the same final class to" +
          " be assigned. Either use assignByName or assignSomeByName at \n" + ScalaLocated.long)
        bundleAssign(that)((to, from) => to.compositAssignFrom(from,to,kind))
      case _ => throw new Exception("Undefined assignment")
    }
  }
  
```

I'm not totaly in love with it, but, why not ? 

So the user would have to override bundleAssign, and call the body function on each couple of elements. The reasons why it is "convoluted" is to presreve the compositeAssignFrom.

Should i push it ?

kleinai:
I also don't like this implementation. Doesn't this omit what element is being assigned? The Axi4Stream bundle has several elements which are the same data type. How would the function know what is being connected?

Dolu1990:
> How would the function know what is being connected?

```scala
that match {
   case that : Axi4Stream => { ... }
}
```

kleinai:
Ah maybe I misunderstood. If you don't think this will negatively impact existing code I'd be happy to give it a try later!

Dolu1990:
Pushed ^

kleinai:
New code works as expected! I added a keep signal to the tester to make sure it'll use the new assignment.

If this passes tests it will be ready to merge in.

Dolu1990:
thanks ^^

PR #692, opened 2022-04-20
initially add FormalConfig which call symbiyosys with yices as default solver.

Readon:


Readon:
The code style for formal verification can be borrowed from [doFormal](https://github.com/jiegec/EspinalLib/blob/ffd34f7bd7db9f2662aadf594135d78d7d85b953/espinallib/src/espinallib/zipcpu/equiv/SyncFifoEquiv.scala) and [operator](https://github.com/tomverbeure/cxxrtl_eval/blob/3cb9b955113b22138bd4da183c2b13baf709d83c/spinal/src/main/scala/cc/Apb3CC.scala) where the code for formal can be.
``` scala
doFormal { (outerReset, pastValid) =>
            assume(io.reset_ === !initstate())

            assume(rose(src.PENABLE)    |-> stable(src.PSEL))
            assume(rose(src.PENABLE)    |-> stable(src.PADDR))
            assume(rose(src.PENABLE)    |-> stable(src.PWRITE))
            assume(rose(src.PENABLE)    |-> stable(src.PWDATA))

            assume(!stable(src.PSEL)    |=> (fell(src.PENABLE) || !src.PENABLE))
            assume(!stable(src.PADDR)   |=> (fell(src.PENABLE) || !src.PENABLE))
            assume(!stable(src.PWRITE)  |=> (fell(src.PENABLE) || !src.PENABLE))
            assume(!stable(src.PWDATA)  |=> (fell(src.PENABLE) || !src.PENABLE))
}
```
This style is very similar to SVA.

Dolu1990:
Nice <3

Seems to work well for me.

@jiegec  
about the [doFormal](https://github.com/jiegec/EspinalLib/blob/ffd34f7bd7db9f2662aadf594135d78d7d85b953/espinallib/src/espinallib/zipcpu/equiv/SyncFifoEquiv.scala), i can see it does quite a few reset handeling, why is it necessary ? I do not know much about formal
(vs having a global assumeInitial(ClockDomain.current.isResetActive))

Readon:
> Nice <3
> 
> Seems to work well for me.
> 
> @jiegec about the [doFormal](https://github.com/jiegec/EspinalLib/blob/ffd34f7bd7db9f2662aadf594135d78d7d85b953/espinallib/src/espinallib/zipcpu/equiv/SyncFifoEquiv.scala), i can see it does quite a few reset handeling, why is it necessary ? I do not know much about formal (vs having a global assumeInitial(ClockDomain.current.isResetActive))

As I understand, what he did is to implement the function that ZipCPU's tutorial slides described, where a past valid signal is used to demonstrate when the past can be used.

jiegec:
> As I understand, what he did is to implement the function that ZipCPU's tutorial slides described, where a past valid signal is used to demonstrate when the past can be used.

Yes, the logic was taken from ZipCPU. ChiselTest uses firrtl transform(SafePastSignalsPass) to handle this logic instead, but it can be hard to port to SpinalHDL.

Readon:
> > As I understand, what he did is to implement the function that ZipCPU's tutorial slides described, where a past valid signal is used to demonstrate when the past can be used.
> 
> Yes, the logic was taken from ZipCPU. ChiselTest uses firrtl transform(SafePastSignalsPass) to handle this logic instead, but it can be hard to port to SpinalHDL.

according to yosys states "$initstate evaluates to 1 in the initial state and to 0 otherwise."
That seems **identical** to the definition of f_past_valid from ZipCPU.
``` verilog
reg f_past_valid;

initial f_past_valid = 1'b0;
always @(posedge clk)
         f_past_valid <= 1'b1;
```
So the only necessary task should be to check on where "past" is located? I have concluded some points in the [discussion](https://github.com/SpinalHDL/SpinalHDL/discussions/689#discussioncomment-2617103).

Readon:
The problem exist is that if some verification using the same name then it would be put in the same directory.

Readon:
> The problem exist is that if some verification using the same name then it would be put in the same directory.

The FormalWorkspace would do the rename tasks while all formal verifications are running in the one run/command/process.

core/src/main/scala/spinal/core/Formal.scala
```diff
@@ -7,18 +7,19 @@ import spinal.core.internals.Operator
 object Formal {
 
   //that.formalPast(delay) replacement
-  def past[T <: Data](that : T, delay : Int)(moreAction: (Int, T) => Unit) : T = {
+  def past[T <: Data](that : T, delay : Int, action: (Int, T) => Unit = null) : T = {
```
Dolu1990, line 10: I'm not realy for it, see the bottom of https://github.com/SpinalHDL/SpinalHDL/discussions/689#discussioncomment-2628742

Dolu1990: i mean, having action to past.

Dolu1990:
I'm currently loocking at adding the yosys toolchain into the github action env

Dolu1990:
I think it would be good to have to PR merged before adding to much side stuff. Once merged, we can have separated github issues to track the addition of additional API.

I'm working on getting github action working.

Readon:
> I think it would be good to have to PR merged before adding to much side stuff. Once merged, we can have separated github issues to track the addition of additional API.
> 
> I'm working on getting github action working.

should I split&resubmit them?

Dolu1990:
that's fine, we can keep them there, and rework them afterward, now, i'm waiting to see if the regression pass, then i will merge.

core/src/main/scala/spinal/core/Formal.scala
Readon: I just revert them.

Dolu1990:
Thanks :D

PR #696, opened 2022-04-25
Don't generate VHDL port list when there are no ports defined

kleinai:
Components without any signals in the IO bundle still generate a port list but without elements. The VHDL spec doesn't allow an empty list so we shouldn't emit one if it's empty.


Dolu1990:
Thanks

PR #697, opened 2022-04-26
Add Axi4StreamWidthAdapter

kleinai:
Adds an Axi4 Stream width adapter. This processes an incoming sparse stream of width one width into another sparse stream of a second width.

I tried to pipeline this as best as possible. There's a lot of data movement into and out of the buffer which happens on the same cycle to keep throughput high. But I haven't actually benchmarked it. The sim waveforms look to sustain constant reads/writes.

This should conform to the packing, downsizing, and upsizing portion of the AMBA 4 spec.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala
@@ -71,7 +71,7 @@ object Axi4Stream {
               sourceSlicesPad.zip(sinkSlices).foreach(pair => {
                 val (sourceSlice, sinkSlice) = pair
                 if (sourceSlice != null) {
-                  sourceSlice := sinkSlice.resized
+                  sinkSlice := sourceSlice.resized
                 } else
                   sinkSlice := B(sinkSlice.bitsRange -> false)
               })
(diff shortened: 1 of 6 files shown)


Dolu1990:
>  This processes an incoming sparse stream of width one width into another sparse stream of a second width.

So, it does not compress/collaps the stream data right ?

what manipulations are done on the bytes ?

kleinai:
It does compress or more technically pack the stream. Bytes flagged with TSTRB=0 can be freely removed from the data stream. This removes those and provides a dense stream. When a TLAST beat is seen the rest of the data is flushed out with strobes to indicate the last partial transaction if needed.

I had planned on using this to repack serdes data where I know the order of the data transmitted but know which bytes of the data are valid. In this way I can merge the streams as a large sparse stream then repack the data for handling.

I'm also thinking I'll need a normal width adapter too. This one is logic heavy on wide streams.

kleinai:
Maybe I'll add a few more components before this is ready. Merge useful infrastructure in one PR.

kleinai:
This is ready for review. There's two options, a full sparse optional repacking width converter and a simple shift-like width converter.

The integrity of byte-based fields is preserved. Metadata like DEST and ID are a little tricky and so aren't yet checked by the tests. I don't plan on using these features myself.

PR #700, opened 2022-04-27
lib: bus: amba4: Axi4ToApb3: Ignore writes without strb

dnltz:
When downsizing an AXI4 channel to 32 bits, it adds transmission without
strobe asserted to the 32 bit channel. The Apb3 bridge does not check
whether the strobe signals are set or not and forwards wrong
transmissions.

Ignore a write transmission when all strobe signals are deasserted and
simply acknowledge the transmission without forwarding.

Signed-off-by: Daniel Schultz <dnltz@aesc-silicon.de>


Dolu1990:
I think there is a issue, io.axi.sharedCmd.write should be used instead of io.axi.writeData.valid as condition. Because you may have a read access followed by a write access, in which the data channel is already there.

I will push a fix.

PR #707, opened 2022-05-01
Reduce the load of CI by moving verification of controllable fifo to tester as an App.

Readon:
Further more, fix the usage of anyseq in Trigger.ANY. 
The anyseq variable can be any one of the options, but just one, so it would lead "anyseq(Bool())" True or False depending on random values it choosen.
So if only below line is called then it would pass, while it fails when it is running from the for loop.
``` scala
shouldFail(fifoTest(Error.LOCK_PUSH, Trigger.ANY))
```
All of this is due to that anyseq is only used as trivial data but key controlling signal.
A proper macro is allseq, where the load for computing with allseq and allconst is too high.

### DIFF
--- a/tester/src/main/scala/spinal/tester/code/FormalFifo.scala
+++ b/tester/src/main/scala/spinal/tester/code/FormalFifo.scala
@@ -1,0 +1,184 @@
+package spinal.tester.code
+
+import spinal.core.formal._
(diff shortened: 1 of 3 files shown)


Dolu1990:
Thanks ^^

PR #708, opened 2022-05-01
Fix wishbone arbitration

Jeff-Ciesielski:
I believe that I've found a bug in the Wishbone Arbiter, specifically the round robin arbitration when used in a dual master Interconnect with shared memories (like you might find in a Vex SOC).

A snippet from my configuration: 
```
 // Wishbone Interconnect
  val wbIntercon = WishboneInterconFactory()

  wbIntercon.addSlaves(
    bootRom.io.wishbone -> SizeMapping(0x80000000L >> 2, (bootRomSize / 1024) kB),
    io.imem_wishbone -> SizeMapping(0x60000000L >> 2, 64 kB),
    io.dmem_wishbone -> SizeMapping(0x40000000L >> 2, 64 kB),
    <snip>
  )

  wbIntercon.addMasters(
    core.iBus -> List(bootRom.io.wishbone, io.imem_wishbone, io.dmem_wishbone),
    core.dBus -> List(
      bootRom.io.wishbone,
      io.imem_wishbone,
      io.dmem_wishbone,
     <snip>
    )
  )
```
(Note that all memories are available to both the instruction and data bus)

With the current version, I noticed that my firmware would fail to correctly execute, and that I was getting invalid instruction traps. After significant digging through waveforms, what I came to notice was that this condition would arise after a period with NO grants from the bootRom arbiter, followed by BOTH the instruction and data memory busses both issuing reads to the same memory on the same clock cycle.

With a minor adjustment to the arbiter test bench, this shows up pretty clearly (though it's not being tested for specifically)

image

In my situation:
* The instruction and data bus both issue at the same cycle
* Data address would make it through to the bootRom along with CYC/STB
* ACK would be intercepted by the fetch unit
* Invalid instruction trap is thrown

I believe that the issue is the fact that the selector (via the round robin algorithm) is allowed to drop to zero rather than a one-hot encoding (which is explicitly called out in the comments as being a requirement), while the mask lock is only set when there a grant active.  The implication of this being that while the MuxOH selector will default to the first entry in the mux, the round-robin arbitration may grant to a different slave than was initially selected on the first clock cycle, leading to invalid response routing and the issue at hand.  By simply ensuring that the selector is in a known state at all times, this issue is avoided, and responses are routed accordingly.

image

I've also included an option for fair arbitration as well as well.  To demonstrate: Vex will immediately issue additional instruction fetches without deasserting CYC in between, which will prevent the arbitration from advancing to the next requestor (in effect giving 'burst' privileges to the granted master, and potentially indefinitely starving the bus)

This *is* expected behavior based on the comments (...implement a blocking round robin algorithm...), however I posit this could give unexpected/confusing results to end users if (for example) something like a DMA engine were included in their design, which could cause long bursts of inactivity on the instruction fetch unit.  I've put this functionality behind a 'fairArbitration' flag plumbed through the arbiter and interconnect constructors which defaults to the older behavior and is purely opt in.

image

(I'd actually argue that the only time 'burst'/blocking behavior should be allowed is if the 'use lock'  feature were turned on, but I think breaking current users is generally not great)

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/wishbone/WishboneArbiter.scala
+++ b/lib/src/main/scala/spinal/lib/bus/wishbone/WishboneArbiter.scala
@@ -46,19 +46,20 @@ class WishboneArbiter(config : WishboneConfig, inputCount : Int) extends Compone
     Wishbone.driveWeak(io.output.TGD_MISO, in.TGD_MISO, null, false, false)
   }
 
-  //Implement a blocking round robin algorithm.
+  //Implement a round robin algorithm.
   //The channel will remain selected is either LOCK or CYC are equals to true
   //and pass to the next channel after the master clears CYC and LOCK
   //this output a One Hot encoded vector/bit array
   val requests =  if(config.useLOCK)  Vec(io.inputs.map(func => func.CYC && !func.LOCK))
-                  else                Vec(io.inputs.map(_.CYC))
+                  else                Vec(io.inputs.map(func => func.CYC))
 
   val maskLock : Vec[Bool] = RegInit(B(1,inputCount bits).asBools)
   val roundRobin : Vec[Bool] = OHMasking.roundRobin(requests, maskLock)
+  val selector: Vec[Bool] = RegNext(roundRobin.orR ? roundRobin | maskLock,
+                                    B(1, inputCount bits).asBools)
 
-  val selector : Vec[Bool] = RegNext(roundRobin,roundRobin.getZero)
-  when(io.output.CYC && selector.orR){
-    maskLock := selector
+  when (roundRobin.orR) {
+    maskLock := roundRobin
   }
 
   //Implement the selector for the output slave signals
--- a/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneArbiterTester.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneArbiterTester.scala
@@ -20,6 +20,31 @@ class WishboneArbiterComponent(config : WishboneConfig,size: Int) extends Compon
   val arbiter = WishboneArbiter(io.busIN,io.busOUT)
 }
 
+case class WishboneMemoryHarness(config: WishboneConfig, size: Int) extends Component {
+  val io = new Bundle{
+    val busIN = Vec(slave(Wishbone(config)), size)
+  }
+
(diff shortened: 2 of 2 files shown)


wifasoi:
On it

wifasoi:
As the specification says the CYC can hang indefinitely, and suggest that if ERR is not implemented in the master this condition is not avoidable.
As RECOMMENDATION 3.10 suggest, maybe we/I need to implement a watchdog on the STB signal and send an ERR/RTY to the master... so ,maibe, it gets the hint to GTFO from the bus (but it need to be implemented in the master logic).

But i like the fairness idea, but I think your implementation (maybe I'm wrong) assume that it's safe to change the master when only `!ACK && CYC`. The problema that It can lent to a problem if the slave dosn't respon in only one clock cycle. So the solution is a flag (for registered classic) or two counter(for pipelined comunication) taht keep tabs on how many ACk and STB are sent/received. Only when the number of STB is equals to the number of ACK we can "yank" the bus from the master, and give control to someone else.

Jeff-Ciesielski:
> But i like the fairness idea, but I think your implementation (maybe I'm wrong) assume that it's safe to change the master when only !ACK && CYC.

This implementation will mask the request with the ack, so there won't be a spurious ack as the 'maskLock' holds the bus until the request is cleared.  To test, I put my boot rom behind a state machine with some wait states and it operates as expected when there is contention:
image

What I'm a little worried about now that I think on it is the case of an asynchronous slave that simply ties ACK = (CYC & STB). Let me see about that one as well

wifasoi:
Maybe the best course of action is:
- drop the fairArbitration feature
- commit the unit test that test this fault (if you have it handy)
- implement the fairArbitration in another PR

Otherwhise, can I ask you to just push the unit test code, so i can test the fairArbitration feature?

Jeff-Ciesielski:
Hey @wifasoi, happy to strip out fair arbitration and submit separately. I actually saw another  failure this week with fairArbitration turned off, so I'm building up some test benches to flush it out.  Will hopefully have something to show for it this weekend (Busy time at work!)

Jeff-Ciesielski:
@wifasoi : Updated with a testbench and a fix that seems to resolve the outstanding issue.

wifasoi:
LGTM, merging!
Thanks for your contribution :D
And thanks to taking the time to hardening the arbiter against greedy masters :D

See you in the next PR

PR #710, opened 2022-05-01
#671 rollback

jijingg:
hi @alexismarquet, I have rollback temporary for reason of do not stop others. When the time is ripe, then we can remove catching code.


Dolu1990:
Thanks :)

PR #718, opened 2022-05-05
Fix execute simulation fail on Windows 11

kazutoiris:
I've tested on Windows 11 Dev, now it works well.


Dolu1990:
Thanks 
I'm now releasing 1.7.0b with this :)

PR #721, opened 2022-05-07
Try to add some formal assertion for common cases.

Readon:
1. for Axi4 address and data channel.
2. for Stream cover cases.


lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -514,8 +514,9 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
    * - After being asserted, valid may only be deasserted once the current payload was acknowleged.
    *
    * @param payloadInvariance Check that the payload does not change when valid is high and ready is low.
+   * @param maxStallCycles Check that the max cycles the interface would hold in stall.
    */
-  def withAsserts(payloadInvariance : Boolean = true)(implicit loc : Location) : this.type = {
+  def withAsserts(payloadInvariance : Boolean = true, maxStallCycles : Int = 0)(implicit loc : Location) : this.type = {
```
Dolu1990, line 519: maxStallCycles isn't implemented right ?

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -534,6 +535,16 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
     this
   }
 
+  def withCovers(back2BackCycles: Int = 1): this.type  = {
```
Dolu1990, line 538: i would say, back2BackCycles should be 0 by default, as in some design, the hardware will never reach maximal throughput. Ex :  Stream.halfPipe

Readon:
I have implement a standalone with timeout asserts and assumes. because the timeout series verification functions are not bind to with asserts.



---Original---
From: ***@***.***&gt;
Date: Mon, May 9, 2022 18:58 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Try to add some formal assertion for commoncases. (PR #721)




 
@Dolu1990 commented on this pull request.
 
 
In lib/src/main/scala/spinal/lib/Stream.scala:
 &gt;     */ -  def withAsserts(payloadInvariance : Boolean = true)(implicit loc : Location) : this.type = { +  def withAsserts(payloadInvariance : Boolean = true, maxStallCycles : Int = 0)(implicit loc : Location) : this.type = {  
maxStallCycles isn't implemented right ?

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -534,6 +535,16 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
...
+    import spinal.core.formal._
+    val hist = History(this.fire, back2BackCycles).reduce(_ && _)
```
Dolu1990, line 540: <3

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -561,6 +572,26 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
 
     (aheadIn, behindIn)
   }
+
+  def withTimeOutAsserts(maxStallCycles : Int = 0) : this.type = {
```
Dolu1990, line 576: withTimeoutAsserts typo ?

Readon: sure, I will fix it soon.

Dolu1990:
What's is the status ? All good ?

Readon:
i am still on the way to add more coupling verification for address channel and data channel. may be should wait for them to be ready.



---Original---
From: ***@***.***&gt;
Date: Wed, May 11, 2022 18:33 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Try to add some formal assertion for common cases. (PR #721)




 
What's is the status ? All good ?

Dolu1990:
It start to be quite a large PR ^^

Readon:
> It start to be quite a large PR ^^ What is the status of the PR ?

I am implementing a helper component HistoryModifyable to keep history records "in order" and the record can also be dropped and altered. 
This is used for maintaining the transaction records created either from AW or W channels, so that the arbitrary timing relationship of payloads in those channels can be verified.

This component is being verified in formal way also.^^

Readon:
@Dolu1990 I have almost done the Axi4WriteOnly bus formal verify tools, please have a review on them.
So that I can fix them also with the ReadOnly implementation.

Readon:
@Dolu1990 one more problem is that, the verification is really too heavy to do every time. Is there any good alternative way to just do the verification manually, but still act as a tester?

Dolu1990:
@Readon We could have a environnement variable which would enable long and slow tests ?

Ex : 
export SPINAL_TEST_ALL=1

Else, the PR seems fine to me ^^

Readon:
> @Readon We could have a environnement variable which would enable long and slow tests ?
> 
> Ex : export SPINAL_TEST_ALL=1
> 
> Else, the PR seems fine to me ^^

Is that possible to split the tester in CI into legacy test part and formal verification part.
So that if they run every time, then it can be checked independently.

Dolu1990:
> Is that possible to split the tester in CI into legacy test part and formal verification part.

What is legacy test part ?

I would say, as long than the test doesn't take too long (not more than 30-60 seconds per test max)

> So that if they run every time, then it can be checked independently.

Yes sure
I guess adding another rule as 
https://github.com/SpinalHDL/SpinalHDL/blob/a09db4b3649c12aafa50170e949b59faf976c8be/.github/workflows/scala.yml#L82

and having both with testOnly *xxxx*      would do the trick, and having different packages for different set of tests to run them independently, that would be good

Readon:
> > Is that possible to split the tester in CI into legacy test part and formal verification part.
> 
> What is legacy test part ?
> 
> I would say, as long than the test doesn't take too long (not more than 30-60 seconds per test max)
> 
> > So that if they run every time, then it can be checked independently.
> 
> Yes sure I guess adding another rule as
> 
> https://github.com/SpinalHDL/SpinalHDL/blob/a09db4b3649c12aafa50170e949b59faf976c8be/.github/workflows/scala.yml#L82
> 
> and having both with testOnly _xxxx_ would do the trick, and having different packages for different set of tests to run them independently, that would be good

I have split the jobs by copy the existing workflow.

Dolu1990:
Ahhh interresting i didn't knew about the Tag thing in scala test.

Nice ^^
Thanks for the PR :)

Readon:
I am implementing a helper component to keep history in order and its record can be dropped and altered. This is useful for maintaining the records created either from AW or W channels.


This component is being verified in formal way also.^^





---Original---
From: ***@***.***&gt;
Date: Wed, May 18, 2022 20:19 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Try to add some formal assertion for common cases. (PR #721)




 
It start to be quite a large PR ^^

Readon:
the default value now is to cover there is one fire.



---Original---
From: ***@***.***&gt;
Date: Mon, May 9, 2022 19:01 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Try to add some formal assertion for common cases. (PR #721)




 
@Dolu1990 commented on this pull request.
 
 
In lib/src/main/scala/spinal/lib/Stream.scala:
 &gt; @@ -534,6 +535,16 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas      this    }   +  def withCovers(back2BackCycles: Int = 1): this.type  = {  
i would say, back2BackCycles should be 0 by default, as in some design, the hardware will never reach maximal throughput. Ex :  Stream.halfPipe

PR #722, opened 2022-05-07
regif address hit naming optmize

jijingg:
regif  bus write and read hit naming optimize.

image

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/regif/RegInst.scala
+++ b/lib/src/main/scala/spinal/lib/bus/regif/RegInst.scala
@@ -317,10 +317,10 @@ abstract class RegBase(name: String, addr: Long, doc: String, busif: BusIf) {
   def readErrorTag = Rerror
   def getFields = fields.toList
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
I'm not sure it is applicable in your case, but in case of : 

```scala
    val bus = new Area {
      val write = Bool()
      val address = UInt(8 bits)
    }
    class RegInst(address : Int) extends Area{
      val isWriting = bus.write && bus.address === address
    }
    val myReg = new RegInst(42)
```

=> 

```verilog
  wire                bus_write;
  wire       [7:0]    bus_address;
  wire                myReg_isWriting;

  assign myReg_isWriting = (bus_write && (bus_address == 8'h2a));
```

jijingg:
> I'm not sure it is applicable in your case, but in case of 

I know about Area, I didn't use much before.  But I have realized recently that Area and Composit is a good tools  to solve naming issue in generated verilog especially on temp signal name .  thanks @Dolu1990

PR #737, opened 2022-05-15
Fix VPI include flags and path string on Windows

kleinai:
Added backslash space escaping for Linux systems but haven't tested it.


Dolu1990:
Something blew up ^^

Dolu1990:
Thanks :)

PR #738, opened 2022-05-15
Add missing MemSymbolsTag in the VHDL emitter

kleinai:


kleinai:
@Dolu1990 can you review this when you have a second? This adds the missing memory symbols tag for simulation like the Verilog emitter has.

Dolu1990:
Ahhh funny.
Seems ok to me, was it tested ?

kleinai:
Yes I tested it locally. I'll update the tests to include VHDL later. Should have done that before

PR #739, opened 2022-05-16
Add memory readSync register resets 

kleinai:
I'm working through adding reset registers for memory synchronous read ports. I'm looking for feedback as to if this way is good or if there's a simpler way I'm not seeing yet.

### DIFF
--- a/core/src/main/scala/spinal/core/Mem.scala
+++ b/core/src/main/scala/spinal/core/Mem.scala
@@ -281,20 +281,58 @@ class Mem[T <: Data](val wordType: HardType[T], val wordCount: Int) extends Decl
     data.assignFromBits(readBits)
   }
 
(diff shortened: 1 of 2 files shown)


kleinai:
I confirmed that the VHDL output will infer BRAM and use the hard register reset signal.

The Verilog memory generator will need more work due to the read first code.

Dolu1990:
Thanks ^^

One question is : 
Do we realy want to add this feature (given the additional complexity / headache / potential issues) ?

Personnaly, i'm not 100% sure XD

kleinai:
Good question. After looking at both memory emitters I'm partial to fully reworking them. For example there's dead VHDL initialize code in the Verilog emitter

kleinai:
Ah never mind. I starting looking into adding a new Mem phase and it got really complex. The other Mem fixes are still valid for VHDL sims.

Dolu1990:
No worries,
Thanks anyway :)

PR #740, opened 2022-05-17
Add readWriteSyncPort to MemPimped

tanhongze:


lib/src/main/scala/spinal/lib/Mem.scala
```diff
@@ -133,12 +133,25 @@ class MemPimped[T <: Data](mem: Mem[T]) {
     ret.data := ClockDomain.current.withRevertedClockEdge()(mem.readSync(ret.address))
     ret
   }
+
+  def readWriteSyncPort : MemReadWritePort[T] = {
```
Dolu1990, line 137: We may want to add  () after the function name, as in the future we may add optional arguements (ex read during write behaviour, and other stuff)

Dolu1990: (same for readWriteSyncPortWithMask)

lib/src/main/scala/spinal/lib/Mem.scala
```diff
@@ -174,4 +187,18 @@ case class MemReadPortAsync[T <: Data](dataType : T,addressWidth : Int) extends
...
\ No newline at end of file
+}
+
+case class MemReadWritePort[T <: Data](dataType : T, addressWidth : Int, useMask : Boolean = false) extends Bundle with IMasterSlave{
+  val address = UInt(addressWidth bit)
+  val rdata   = cloneOf(dataType)
+  val wdata   = cloneOf(dataType)
+  val enable  = Bool()
+  val write   = Bool()
+  val mask    = ifGen(useMask)(Bits(widthOf(dataType) bits))
```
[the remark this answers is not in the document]
Dolu1990: Also, 1 mask bit per word bit isn't properly inferable in most FPGA

tanhongze:
Got it.
By the way, maybe the description for write mask should be put into the Spinal documentation?

Dolu1990:
> By the way, maybe the description for write mask should be put into the Spinal documentation?

Done ^^

tanhongze:
Done.
I am too busy recently, sorry.

Dolu1990:
thanks, no worries :)

PR #741, opened 2022-05-17
[Axi4SlaveFactory] Added function to specify prohibited adresses

sabbari:
send SLVERR on b.resp or r.resp when attempring to do write on a readonly adress or reading from a write only register respectively


sabbari:
Undo pull request

PR #750, opened 2022-05-25
Bitwise operators for Vec

numero-744:
Builds but untested.


Dolu1990:
Good do me ^^

Dolu1990:
ready for merge ?

numero-744:
Well, I have not tested it yet (I have started reading about unit testing with sbt but did not play with it yet).

But if you are sure it is okay, you can merge it :)

Dolu1990:
Hmm we may want to add an assert to ensure the size of both vec is the same ?
Zipped is a bit too relaxed ^^

Dolu1990:
Thanks ^^

PR #752, opened 2022-05-27
Add pendingMax parameter to Axi4CrossbarFactory

tlupick:
This change allows for the pendingMax parameter in the subcomponent Axi4 Decoders to be changed from the default value of 7.


Dolu1990:
Thanks ^^

PR #759, opened 2022-06-04
AFix: Switch to spinal sim, upgrade some rounds

kleinai:
Much faster this time. For some reason I upgraded `floorToZero` and `ceilToInf` to the new pattern.


Dolu1990:
Thanks :D

PR #760, opened 2022-06-04
Add a more convenient repeat data API.

wswslzp:
Add implicit conversion from a tuple of data to `TupleBundle`. Now a straightforward tuple assignment can be done through a tuple bundle like
```scala
  case class tt() extends Module {
    val a = in UInt(3 bit)
    val b = in Bits(14 bit)
    val c = in SInt(16 bit)

    val d, e = out Bits(16 bit)
    val f = Bits(4 bit)

    // assign {d, e, f} = {{2{a}}, b, c};
    (d, e, f) := (B(a, 2), b, c)

    val g = B(a, B(b, 2), c)
    val gg = cloneOf(g)
    // assign gg = {a, {2{b}}};
    gg := B(a, B(b, 2), c)

    val oa, ob = out UInt(4 bit)
    val iab = in UInt(8 bit)
    // assign {aa,bb} = iab;
    (oa, ob) := ia
  }
```

Also, fix a typo.


Dolu1990:
All good, but about  B(a, 2),

What's about having R(a, 2) instead ? (a shorter  form to the already existing Repeat(a, 2)) . 
I'm a bit scared of adding this specific functionality to the B functions.

wswslzp:
From my perspective, repeating a bit string and concatenating a bunch of bit strings are similar operations and should have the same position. In Verilog, these two operations use the same operator:

Concatenate: 
```verilog
wire [1:0] a, b;
wire [3:0] ab;
assign ab = {a,b};
```
and Repeat:
```verilog
wire [1:0] a;
wire [3:0] aa;
assign aa = {2{a}}; //  SpinalHDL cannot generate this kind of verilog operation now.
```
However in SpinalHDL, `Cat` is a function in `spinal.core` while `Repeat` is in `spinal.lib`.

So, I intended to unify them as one single API. First I try to add the repeating function to `Cat`, but soon I found that one `apply` function of object `B` has also implemented `Cat` function in a more straightforward way 
```scala
def apply(head: Data, tail: Data*) : Bits = Cat((head +: tail).reverse)
```
so I want to add the bit repeating function to `B` as well.

wswslzp:
> > but soon I found that one apply function of object B has also implemented Cat function in a more straightforward way
> 
> Ahhh i forgot that one XD
> 
> So, i'm all good with the PR, but why not B(a, 2) instead ? that way it would follow other B constructors ( the data first, arguments after)

OK, I have pushed new commits.

wswslzp:
Hi @Dolu1990 I think this pr can be merged now.

Dolu1990:
Thanks :)

PR #763, opened 2022-06-05
fix speed of the Verilog component emitter

jonnykl:
Nested foreach causes `O(n^2)` complexity in the Verilog generation. This bug did not affect VHDL. For my design this fix reduces the time from ~1h to ~3.5min. VHDL generation is still faster (~2min vs. ~3.5min), so further Verilog generation speedup may be possible.

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -73,14 +73,13 @@ class ComponentEmitterVerilog(
 
 
   def emitEntity(): Unit = {
-    component.getOrdredNodeIo
-      .filterNot(_.isSuffix)
-      .foreach{baseType =>
+    val ios = component.getOrdredNodeIo.filterNot(_.isSuffix)
+    ios.foreach{baseType =>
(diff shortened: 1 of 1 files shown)


Dolu1990:
Nice thanks :)

What kind of hardware are you generating ? how big is it ? Can you share it for me to give a try to see where are bottlenecks ?

jonnykl:
Sorry for my late reply!

You just need to generate a lot of in-/outputs:

```scala
import spinal.core._

object Foo {
    val n = args(0).toInt
    //SpinalConfig().generateVhdl(Foo(n))
    SpinalConfig().generateVerilog(Foo(n))
  }
}


case class Foo(n: Int) extends Component {
  val io = new Bundle {
    val data = out Vec(Bool(), n)
  }

  io.data.foreach(_ := False)
}
```

You can see very clearly how the time consumed by the `Generate Verilog` phase increases quadratically. Test it for `n=10000`, `n=20000` and `n=40000`.

I also fixed my design so that that much in-/outputs are not needed. I'm not sure if anyone will ever use more than 10k ports in a component. In my case these were unused and I did not explicitly deactivate them which caused ~26k ports. So I think there are reasons to optimize for such a large number of ports.

Dolu1990:
Ahhh ok ^^
Right, the number of io was assumed not to be as much :)

PR #767, opened 2022-06-07
AFix updates and improvements

kleinai:
Trait improvements 
- [X] Add Num[T] functions
I have questions about the rounding `aligned` parameters...
- [X] ~MinMaxProvider~ MinMaxDecimalProvider
I think this one is free as AFix already tracks max and min range.
- [x] BitwiseOp[T]
Need to figure out range computations.

General improvements
- [x] Review and rework comparator compile time optimizations
- [x] Rework negate operator
- [x] Upgrade half rounds to the new rounding pattern


Dolu1990:
> I have an important question about your rounding operations in AFix. Almost all the rounding functions in AFix will round off the fraction part and preserve only the integer part. But the most of time we just round off some LSBs to control the bit length, not the entire fraction part (like pi=3.14159, a rounded pi could be 3.14, not 3.0).

That's something i started reworking a bit : 
https://github.com/SpinalHDL/SpinalHDL/commit/754913774e2167bdab7b4bcbb1ac26f98485ffb6

Where you specify at which exp you want to round it, avoiding having to shift left then round then shift right.

Feel free to PR updates for other roundings :)

> There's already a set of rounding functions implemented in UInt/SInt. I think you don't have to rebuild the wheel.

Ahhh right, that could be usefull too

wswslzp:
Also, there's still a weird problem. 

AFix has two constructor parameters: `minValue` and `maxValue` of type `BigInt`, instead of `BigDecimal`. It's weird that a fixed-point number has an integer boundary.

Dolu1990:
minValue and maxValue are relative to the internal storage

real_value = value * e^exp

We could rename those in something less confusing ?

wswslzp:
The original constructor parameters `maxValue` and `minValue` can be renamed as something like `maxRawIntValue` and `minRawIntValue`. 

And implement `maxValue` and `minValue` as two methods defined in the trait `MinMaxProvider`.

Dolu1990:
hmmm maxRaw / minRaw to keep things short ?

kleinai:
Yes, I should have looked a `SInt` and `UInt` for their rounding functions. I was using `XFix` originally and it didn't have rounding implemented so I did it myself. I'll back port the additional rounds to `SInt` and `UInt`.

`minValue` and `maxValue` are `BigInt`s not `BigDecimal` so we'd lose all of the precision. Maybe we need a decimal variant of MinMaxProvider?

Mathematically rounding happens on the decimal point. It was expected that you could virtual shift the decimal point, round, then shift the decimal back. We can change this behavior if it's easier.

Thanks for the fixed-point sim classes!

wswslzp:
> hmmm maxRaw / minRaw to keep things short ?

it looks good to me

wswslzp:
> `minValue` and `maxValue` are `BigInt`s not `BigDecimal` so we'd lose all of the precision. Maybe we need a decimal variant of MinMaxProvider?

Maybe a generic one `MinMaxProvider[T<:ScalaNumber]`?

> Mathematically rounding happens on the decimal point. It was expected that you could virtual shift the decimal point, round, then shift the decimal back. We can change this behavior if it's easier.

In practice, rounding can not only happen on the decimal point but also on any decimal place.

Dolu1990:
> Maybe a generic one MinMaxProvider[T<:ScalaNumber]

Issue of ScalaNumber alone is that it provide kind of no aritmetic API, just convertion functions. Maybe there is other trait which could be added in the mix ?

Another way could be to have : 
maxBigInt, minBigInt, maxBigDecimal, minBigDecimal. In order to be explicit.

> In practice, rounding can not only happen on the decimal point but also on any decimal place available.

+1

kleinai:
How about `MinMaxDecimalProvider` just to make it a trait?

Dolu1990:
> MinMaxDecimalProvider 

Yes, why not ^^

core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -649,17 +692,21 @@ class AFix(val maxValue: BigInt, val minValue: BigInt, val exp: Int) extends Mul
   def floor(exp : Int): AFix = {
     val drop = exp-this.exp
     if(drop < 0) return CombInit(this)
-    val res = new AFix(this.maxValue >> drop, this.minValue >> drop, 0)
+    val res = new AFix(this.maxRaw >> drop, this.minRaw >> drop, 0)
```
wswslzp, line 695: Should `res` be 
```
val res = new AFix(this.maxValue >> drop, this.minValue >> drop, exp)
```
?

kleinai: That's what I get for copy and pasting. Will fix

wswslzp: Thanks! LGTM :)

kleinai:
Thanks for your comments!

PR #768, opened 2022-06-08
Call compositAssignFrom on Vec elements to propagate compositeAssign hook

### LINKED ISSUE #748: Does the assignment of bundle unzip the embedded Vec?
When a bundle is [assigned as inStreams(x).payload by a record](https://github.com/SpinalHDL/SpinalHDL/blob/4d88cb390567127be094d35679dd1643259e10b1/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4WriteOnly.scala#L223), the generated code would not assign [strbs](https://github.com/SpinalHDL/SpinalHDL/blob/4d88cb390567127be094d35679dd1643259e10b1/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4WriteOnly.scala#L97) which is a Vec in a bundle.
``` scala
      when(awValid) {
        hist.io.inStreams(awId).payload := awRecord
        hist.io.inStreams(awId).valid := awValid
      }
```

The generated code would be
``` Verilog
    .io_inStreams_0_payload_strbs_0          (inputAssumes_checker_oRecord_strbs_0[7:0]                        ), //i
    .io_inStreams_0_payload_strbs_1          (inputAssumes_checker_oRecord_strbs_1[7:0]                        ), //i
    .io_inStreams_0_payload_strbs_2          (inputAssumes_checker_oRecord_strbs_2[7:0]                        ), //i
    .io_inStreams_0_payload_strbs_3          (inputAssumes_checker_oRecord_strbs_3[7:0]                        ), //i
```
where the oRecord is a default value of all payload, so the code of assignments seems does not work at all.

When I change the code to [assign each element of strbs separately](https://github.com/SpinalHDL/SpinalHDL/blob/4d88cb390567127be094d35679dd1643259e10b1/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4WriteOnly.scala#L201) then the code seems right.

``` scala
      when(awValid) {
        hist.io.inStreams(awId).strbs.zip(awRecord.strbs).map { case (x, y) => x := y }
        hist.io.inStreams(awId).payload := awRecord
        hist.io.inStreams(awId).valid := awValid
      }
```

``` Verilog
    .io_inStreams_0_payload_strbs_0          (inputAssumes_checker_hist_io_inStreams_0_payload_strbs_0[7:0]    ), //i
    .io_inStreams_0_payload_strbs_1          (inputAssumes_checker_hist_io_inStreams_0_payload_strbs_1[7:0]    ), //i
    .io_inStreams_0_payload_strbs_2          (inputAssumes_checker_hist_io_inStreams_0_payload_strbs_2[7:0]    ), //i
    .io_inStreams_0_payload_strbs_3          (inputAssumes_checker_hist_io_inStreams_0_payload_strbs_3[7:0]    ), //i
```
Then the payload can be assigned from awRecord which is the only one I changed the code.

Dijky:
A `compositeAssign` hook currently has no effect on the elements of a `Vec` because `Vec::assignFromImpl` calls `assignFromImpl` on its elements directly rather than via `compositAssignFrom`.  
This change aligns `Vec`'s behaviour with that of [`Bundle::assignFromImpl`](https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/Bundle.scala#L164).

I believe this fixes #748.

### DIFF
--- a/core/src/main/scala/spinal/core/Vec.scala
+++ b/core/src/main/scala/spinal/core/Vec.scala
@@ -279,7 +279,7 @@ class Vec[T <: Data](var _dataType : HardType[T], val vec: Vector[T]) extends Mu
       case that: Vec[T] =>
         if (that.vec.size != this.vec.size) throw new Exception("Can't assign Vec with a different size")
         for ((to, from) <- (this.vec, that.vec).zipped) {
-          to.assignFromImpl(from, to, kind)
+          to.compositAssignFrom(from, to, kind)
         }
       case _            => throw new Exception("Undefined assignment")
     }


Readon:
nice! it works.

kleinai:
Checks look good. Let's merge in it. Thanks!

PR #770, opened 2022-06-08
Add fixed point simulation interface

wswslzp:
Add simulation interface for fixed-point number type `UFix`, `SFix` and `AFix`. You can now read and write fixed-point signal directly.

### DIFF
--- a/core/src/main/scala/spinal/core/sim/package.scala
+++ b/core/src/main/scala/spinal/core/sim/package.scala
@@ -435,6 +435,84 @@ package object sim {
     }
   }
 
+  /**
+   * Add implicit function to UFix/SFix/AFix
+   */
+  abstract class SimFix[T <: XFix[_, _]](bt: T) {
+    val fractionLength = -bt.minExp
+    val maxRawIntValue : BigInt
+    val minRawIntValue : BigInt
+    private def maxValue = maxRawIntValue.doubleValue() / scala.math.pow(2, fractionLength)
+    private def minValue = minRawIntValue.doubleValue() / scala.math.pow(2, fractionLength)
+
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -435,6 +435,84 @@ package object sim {
...
+
+    def #= (that: BigDecimal): Unit = {
+      val rhs = (that * scala.math.pow(2, fractionLength)).toBigInt
+      require(rhs <= maxRawIntValue, s"$that is overflow. Max value allowed is $maxDouble")
+      require(rhs >= minRawIntValue, s"$that is underflow.Min value allowed is $minDouble")
+      bt.raw #= rhs
+    }
+    def #= (that : Double): Unit = this #= BigDecimal(that)
+    def randomize(): BigDecimal = {
+      var rhs = Random.nextDouble()
```
Dolu1990, line 505: This is problematic, if the AFix / XFix are bigger than 52 bits, then the LSB will always be 0.
I think the best would be to randomize a BigInt value of the size of the "raw", and check it is in the bounds, else retry.

wswslzp: I saw that this branch has been merged into @kleinai 's work and he has reworked the `randomize` part of AFix. For XFix part, I will fix it soon.

Dolu1990: Ahhhh cool ^^, so i guess the klenai one will endup into SpinalHDL upstream too, so all good :)

wswslzp:
Now that this branch has been merged into another branch, I will close this pr.

Dolu1990:
Thanks anyway :)

PR #771, opened 2022-06-09
rename formal functions to distinct Master and Slave ends.

Readon:
withXXXAsserts should include only assert properties, and also withXXXAssumes.
The Master and Slave replace XXX above can specify the target that asserts and assume is target to.


Dolu1990:
Thanks ^^

PR #774, opened 2022-06-12
add formal verify for axi4readonly datas

Readon:


Dolu1990:
let's me know when you are done pushing ^^

Readon:
> let's me know when you are done pushing ^^

I think it works now. I will push other PR when prove is done.

Readon:
It's really a long journey to get the SMT engine pass in prove mode.
However, it's there now, for preview.

Dolu1990:
Thanks ^^

Readon:
How about this one? It also fixes the bugs of existing Axi4ReadOnly formal verification facility.
It has to be one on Axi4WriteOnly version, later.

Dolu1990:
Hi,

So personnaly i realy have issues reviewing PR that are bigger than 100 lines. I'm running out of times / busy schedule XD
Getting drowned into bug fix / PR / questions XD (no blame on anybody)

Readon:
> Hi,
> 
> So personnaly i realy have issues reviewing PR that are bigger than 100 lines. I'm running out of times / busy schedule XD Getting drowned into bug fix / PR / questions XD (no blame on anybody)

Ahhh, I can understand that, how about review on the changes that made on the design code, remains formal verification code there, it can partially be checked by formal tools.

Then the code would be remains less than 100 lines.

Dolu1990:
Ahh, like object Util ?

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Channel.scala
```diff
@@ -533,8 +544,11 @@ case class FormalAxi4Record(val config: Axi4Config, maxStrbs: Int) extends Bundl
   def checkLen(): Bool = {
```
Dolu1990, line 544: Alternative syntax to avoid to have to manualy name signal is : 

```scala
def checkLen() = new Composite(this, "checkLen"){ ... ; val value = xxx}.value
```

Readon: > Alternative syntax to avoid to have to manualy name signal is :
> 
> ```scala
> def checkLen() = new Composite(this, "checkLen"){ ... ; val value = xxx}.value
> ```

Good.

PR #776, opened 2022-06-13
Afix update, some bug fix and some features are added.

wswslzp:
Hi, @kleinai and @Dolu1990 ,

I update AFix implementation by
* fix rounding bugs: mismatch width problem happens in `roundToOdd` and `roundToEven`
* fix the wrong maximum boundary of new rounded result
* bring back rounding alignment (saturate the rounding result if the rounding generates a carry bit)
* add `fixTo` function
* add literal value support (conversion, assignment and initialization)
* add a truncation scope that enables global (or local) control of truncation methods.


wswslzp:
Also, I met a problem of initialization, 
My initialization works well when the code is like
```scala
val input = in(AFix.SQ(8 bit, 8 bit))
val regg =  Reg(input) init AF(23.4, 8 bit, 8 bit, true))
regg := input
```
But if it changed to
```scala
val input = in(AFix.SQ(8 bit, 8 bit))
val reggg =  RegNext(input, init = AF(23.4, 8 bit, 8 bit, true))
```
Spinal will report assignment overlap error like
```
[info] - assign overlap *** FAILED ***
[info]   spinal.core.SpinalExit: Error detected in phase PhaseCheck_noLatchNoOverride
[info] ********************************************************************************
[info] ********************************************************************************
[info]
[info] ASSIGNMENT OVERLAP completely the previous one of (toplevel/reggg :  Bits[17 bits])
[info]     spinal.tester.scalatest.SimFixTest$Expanding.<init>(SimFixTest.scala:131)
[info]     spinal.tester.scalatest.SimFixTest$$anonfun$3$$anonfun$apply$1.apply(SimFixTest.scala:144)
[info]     spinal.tester.scalatest.SimFixTest$$anonfun$3$$anonfun$apply$1.apply(SimFixTest.scala:144)
[info]     spinal.sim.JvmThread.run(SimManager.scala:51)
[info]
[info]
[info] ********************************************************************************
[info] ********************************************************************************
```

wswslzp:
> Also, I met a problem of initialization, My initialization works well when the code is like
> 
> ```scala
> val input = in(AFix.SQ(8 bit, 8 bit))
> val regg =  Reg(input) init AF(23.4, 8 bit, 8 bit, true))
> regg := input
> ```
> 
> But if it changed to
> 
> ```scala
> val input = in(AFix.SQ(8 bit, 8 bit))
> val reggg =  RegNext(input, init = AF(23.4, 8 bit, 8 bit, true))
> ```
> 
> Spinal will report assignment overlap error like
> 
> ```
> [info] - assign overlap *** FAILED ***
> [info]   spinal.core.SpinalExit: Error detected in phase PhaseCheck_noLatchNoOverride
> [info] ********************************************************************************
> [info] ********************************************************************************
> [info]
> [info] ASSIGNMENT OVERLAP completely the previous one of (toplevel/reggg :  Bits[17 bits])
> [info]     spinal.tester.scalatest.SimFixTest$Expanding.<init>(SimFixTest.scala:131)
> [info]     spinal.tester.scalatest.SimFixTest$$anonfun$3$$anonfun$apply$1.apply(SimFixTest.scala:144)
> [info]     spinal.tester.scalatest.SimFixTest$$anonfun$3$$anonfun$apply$1.apply(SimFixTest.scala:144)
> [info]     spinal.sim.JvmThread.run(SimManager.scala:51)
> [info]
> [info]
> [info] ********************************************************************************
> [info] ********************************************************************************
> ```

Already fixed.

core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -760,183 +759,213 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
...
     }
   }
 
   def saturated(): AFix = this.truncated(saturation = true, overflow = false)
 
   def truncated(saturation: Boolean = false,
                 overflow  : Boolean = true,
-                rounding  : RoundType = RoundType.FLOOR) : AFix = {
+                rounding  : RoundType = RoundType.FLOOR,
+                align     : Boolean = true) : AFix = {
```
Dolu1990, line 960: What is the align argument ?

core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -760,183 +759,213 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
...
     if (trunc.isDefined) {
-      this._round(trunc.get.rounding)
+      this._round(trunc.get.rounding, exp, aligned)
     } else {
-      roundHalfToInf(exp)
+      roundToInf(exp, aligned)
     }
   }
 
+  def fixTo(m: Int, exp: Int, roundType: RoundType): AFix = this.sat(m)._round(roundType, exp)
```
Dolu1990, line 927: This function seems dangerous, i mean, its signature is dangerous.
Int, Int, RoundType => could be used in other contextes

exp : ExpNumber  would already protect a bit, but still the first arguement being the number of bit that we would saturate seems arbitrary.

Why not instead doing myAFix.sat(x).fixTo(-8 exp, RoundingType.xxx) ?

core/src/main/scala/spinal/core/AFix.scala
wswslzp: It's a parameter used in all the rounding methods in Num[T], except for floor. 

It decides whether the rounding is aligned. If not aligned, the rounding result will prepend a carry bit at MSB, resulting in one bit longer than expected. Say rounding the last N bit from an M bit number without alignment will produce (M-N+1) bits instead of (M-N) bits (normal result).

Dolu1990: >  without alignment will produce (M-N+1) bits instead of (M-N) bits (normal result).

Shouldn't it be tracked via the maxRaw / minRaw ?
For instance see the "step-1" on : 
https://github.com/SpinalHDL/SpinalHDL/blob/703a281be9fc1e918e1f49431e3e22591b561c2c/core/src/main/scala/spinal/core/AFix.scala#L727

the "M-N+1" is to acount for rounding overflow ? (0xffff + 1) ?

core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -720,23 +719,23 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
...
    */
-  def ceil(exp : Int): AFix = {
+  def ceil(exp : Int): AFix = ceil(exp, getTrunc.align)
+
+  def ceil(exp: Int, aligned: Boolean): AFix = {
     val drop = exp-this.exp
     if(drop < 0) return CombInit(this)
     val step = BigInt(1) << drop
-    val res = new AFix((this.maxRaw+step-1) >> drop, (this.minRaw+step-1) >> drop, exp)
+    val res = new AFix(((this.maxRaw+step-1) >> drop)-1, (this.minRaw+step-1) >> drop, exp)
```
Dolu1990, line 728: Are you sure of the : 
((this.maxRaw+step-1) >> drop)-1
instead of 
(this.maxRaw+step-1) >> drop

?

I mean, let's take the example of a Q0.16 that we want to ceil to remove 4 lsb bits.
if we ceil the max value 0.FFFF, it would go to 1.000 => Q1.12 => 4096 maxRaw

(this.maxRaw+step-1) >> drop would give (0xFFFF+0x10-1) >> 4 = 4096  => OK
while 
((this.maxRaw+step-1) >> drop)-1 would give 4095 => KO

If i understand well the "aligned" i think it should be taken in acount there too. isn't it ?

kleinai:
This is some good discussion I'll digest later. Just wanted to pop in to say we already have range aware saturation we should use.

kleinai:
I don't like using the UInt/SInt alignment. AFix already tracks bit width. If it needs to be saturated it can be tagged using `.truncated` then assigned to a smaller AFix. Although I do see how this breaks pattern from UInt/SInt.

If we want similar aligned behavior it should be done by wrapping the AFix round with an AFix sat like `.roundX(exp).sat(newAlignRange)`.

It looks like some of the issues stem from rounding ranges. I opened a PR to add tests to verify the resulting rounded ranges.
See PR #777

core/src/main/scala/spinal/core/AFix.scala
wswslzp: Yes, this signature is kind of misleading. Maybe we can rework it or just remove it?

core/src/main/scala/spinal/core/AFix.scala
wswslzp: I got it. This part is rolled back.

core/src/main/scala/spinal/core/AFix.scala
Dolu1990: I would say let's remove that one, and update the few others which has exp to use the ExpNumber type

Dolu1990:
@kleinai 

> If we want similar aligned behavior it should be done by wrapping the AFix round with an AFix sat like .roundX(exp).sat(newAlignRange)

I agree on the principe, it may be better to compose instead of trying to crunch too many feature into a single thing.
But for this case,  rounding with saturation can be implemented better as a single chunk than a composed one.


For instance, here is a combined ceil/sat : 

```scala
val (high, low) = input.slpitAt(xxx)
output := high + U(low.orR && !high.andR)  //  && !high.andR  does the saturation
```


Composing it with a xxx.ceil(xxx).sat(yyy) has two disaventage : 
- The saturation has more overhead
- It impose the user to provide yyy

So i guess we realy need to have  def ceil(exp : ExpNumber, saturate :Int)

@wswslzp

What's about instead of naming this "align" we name this saturate ?

wswslzp:
> Composing it with a xxx.ceil(xxx).sat(yyy) has two disaventage :
> 
> * The saturation has more overhead


Sorry for not being an expert at fixed point hardware design. I have some questions:
what's the difference between the composing style `.roundX(exp).sat(newAlignRange)` and combination style `ceil(exp : ExpNumber, saturate :Int)`, in terms of performance or area of the generated hardware circuit? 
What's the exact overhead introduced by a separated saturation method? 

Do you mean every rounding method can have its own unique saturation method that outperforms the general saturation? If yes, maybe we can **create a new PR** to rework all these rounding methods. 

> * It impose the user to provide yyy

Even with `ceil(exp : ExpNumber, saturate :Int)`, it seems that the user still needs to provide yyy to `saturate`, if I don't misunderstand. 

> What's about instead of naming this "align" we name this saturate?

Renaming is OK for me. In the latest commits, I just pass the truncation tag's `saturation` parameter (boolean) to these `align` parameters.

wswslzp:
@kleinai 

Speaking of the range, I'm a bit curious that why there isn't a constructor like `AFix(min: BigDecimal, max: BigDecimal, exp: ExpNumber)`, instead of `AFix(min: BigInt, max: BigInt, exp: ExpNumber)`. 

When I first looked at `AFix(16, 8, 2 exp)`, I thought that can represent `8, 12, 16` instead of `32, 36, 40, 44, 48, 52, 56, 60, 64`. I will concern more with the value range of fixed-point numbers than the raw data range.

kleinai:
@wswslzp
> Speaking of the range, I'm a bit curious that why there isn't a constructor like `AFix(min: BigDecimal, max: BigDecimal, exp: ExpNumber)`, instead of `AFix(minRaw: BigInt, maxRaw: BigInt, exp: ExpNumber)`.
> 
> When I first looked at `AFix(16, 8, 2 exp)`, I thought that can represent `8, 12, 16` instead of `32, 36, 40, 44, 48, 52, 56, 60, 64`. I will concern more with the value range of fixed-point numbers than the raw data range.

It's easier to keep AFix managing integer number ranges. The only decimals fixed-point can accurately represent are power of 2 so it's best to avoid building the class around managing decimal ranges. For instance even `0.1` is an infinitely repeating binary decimal.

kleinai:
I just finished #777 (barring any bugs) which adds rounded range tests and moves round ranges to a BigDecimal computation. It's much simpler code.

Dolu1990:

wswslzp:
Great thanks to @kleinai for the hard work on solving the rounding problems! I have merged the newest dev branch and it looks good to me now.
(in fact range problems are not that important for me because they are normally solved at the algorithm level XD)

Maybe this PR can be merged and we can move to the next step?

core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -1178,11 +1222,32 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
...
+                       val rounding  : RoundType
+                       ) extends SpinalTag{
   override def duplicative = true
   override def canSymplifyHost: Boolean = true
-}
\ No newline at end of file
+}
+
+object AFixTruncatedScope extends ScopeProperty[TagAFixTruncated] {
+  override def default: TagAFixTruncated = new TagAFixTruncated(true, false, RoundType.ROUNDTOINF)
```
Dolu1990, line 1244: Maybe the default should be different, and aim at the most raw way : 
saturation = false,
overflow = false, 
rounding = RoundType.FLOOR

core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -1025,37 +1041,43 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
...
+      case RoundType.ROUNDTOEVEN => this.roundToEven(exp, align)
+      case RoundType.ROUNDTOODD  => this.roundToOdd(exp, align)
+      case RoundType.SCRAP        => this.scrap(exp)
     }
   }
 
-  def saturated(): AFix = this.truncated(saturation = true, overflow = false)
+  def saturated(rounding  : RoundType = RoundType.FLOOR): AFix = this.truncated(saturation = true, overflow = false, rounding = rounding)
 
   def truncated(saturation: Boolean = false,
```
Dolu1990, line 1082: If we provide AFixTruncatedScope as a default, i guess the default value of that function should fetch AFixTruncatedScope current value ?

core/src/main/scala/spinal/core/AFix.scala
Dolu1990: Anyway in general, doesn't it seems to be dangerious to have AFixTruncatedScope  ?
For instance, you developpe some hardware with a given default, and instanciate some sub block which was designed with the idea in mind that AFixTruncatedScope isn't altered => broken hardware.

Also, one way to see things is that saturation / overflow / rounding should be explicitly asked by the user, else missmatched would throw you an error.

Dolu1990: hmm i may have missunderstood the usage of AFixTruncatedScope 
So, it is only to provide some sort of implicit parameters to some user called function right ?

core/src/main/scala/spinal/core/AFix.scala
Dolu1990: hmm, seems like it is trying to do too many thing at one.

Why the saturated function would have rounding ?
I mean saturation is a specific thing.

def truncated is maybe missnamed, but is meant to be the function with all options.
Maybe we should rename def truncated into def reformated ?

core/src/main/scala/spinal/core/AFix.scala
wswslzp: Yes, `AFixTruncationScope` is to provide some implicit parameter to only rounding functions. Users can always pass an explicit parameter to o override it. You can check out the `getTrunc` function. 
Also, the scope does not influence the assignment. You have to add a `TagAFixTruncated` to do truncation when doing an assignment. 

Speaking of the default rounding options, maybe we can add them to SpinalConfig to allow users to define the default?

core/src/main/scala/spinal/core/AFix.scala
wswslzp: "truncated" can be regarded as truncating both LSB and/or MSB, I think? `rounded` and `saturated` are just two shortcuts of truncation with different settings.

core/src/main/scala/spinal/core/AFix.scala
Dolu1990: Ahhhhh, ok, i think i was confused in part because it reuse the TagAFixTruncated class, while it has kind of nothing to do the the original usage of the TagAFixTruncated.

So, i think that's good to merge, i will probably just create a separated class TagAFixTruncated -> RoundingDefault or something in that kind.

All good on your side ?

Dolu1990: > Speaking of the default rounding options, maybe we can add them to SpinalConfig to allow users to define the default?

I would say it isn't necessary, i mean, the concept of ScopeProperty is that you could set them where you want, including outside the Spinal context (before SpinalVerilog(xxx)). So SpinalConfig does not have to be coupled with it.

wswslzp: > Also, still, it can be quite dangerous to change the default implicit rounding of a given design, especially when doing the integration of multiple things

If users wants to avoid changing or misusing the default rounding, they can wrap their code into a afix truncation scope or pass rounding parameters explicitly.

PR #777, opened 2022-06-13
Add AFix rounding range tests

kleinai:
This models the expected new bounds after rounding and compares that to what ranges AFix generates when rounding.

This test is currently broken pending review of the rounding ranges calculations.


kleinai:
Unsurprisingly using BigDecimals to calculate the range will work. I do think it's best we keep the backing range as BigInts to avoid having to stuff non-trivially representable numbers (stuff like 0.1) into fixed points.

Dolu1990:
Yes, BigDecimal scare me XD Would realy be better to stay as much as possible with BigInt to keep things "pure" ^^

kleinai:
I'll rewrite the rounds to work on BigInt instead.  These new tests have already caught one rounding bug.

kleinai:
After much pain, I have converted the clean BigDecimal code into a monster of BigInt code

kleinai:
I think it's good to go. I'll bring it in.

PR #784, opened 2022-06-22
fix axiMemorySim data.drop() bug

hanm2019:
In AxiMemorySim.scala  the  `writeArray` function got fault result when write data to multi-page, since the `data.drop(bytesWritten)` didn't change object data, but return a new object, so, the second page will be written by the origin data.


Dolu1990:
@c-thaler ^^

PR #788, opened 2022-06-27
lib: Add 8b10b encoding

dnltz:
This IP to en-/decode a 8b10b sequence is ported from Verilog to
SpinalHDL [1]. This implementation only uses gates and does not rely on
look-up tables.

1: https://github.com/freecores/1000base-x/tree/master/rtl/verilog

Not sure if `spinal.lib.com.linecode` is the correct package. Line codes are mostly used in communication interfaces


kleinai:
Very nice!  I was going to write one myself eventually but probably using look-up tables.  This is far more complete!

kleinai:
This all looks good and has tests which is great!

Are there any copyright headers or considerations that we need from the referenced work?

Should there be a way to force or invert disparity for a character? I'm primarily familiar with Xilinx Serdes which have an interface for this.

dnltz:
Thanks! I mostly ported the Verilog code and extended it with tests. Well, the license is a good point. The source is licensed under LGPL and can't be moved to MIT. It would be possible to link LGPL as a library to Spinal, but the Crypto package uses MIT as well.

Dolu1990:
Thanks ^^

So, i realy have no idea how the license of a ported code should be handled, + at the end, nearly everything is inspired of something by some ways or another.
Maybe the best is to contact the original autor, to ask if that's ok to have it in the SpinalHDL lib as MIT, with the port header file clearly stating the origine of the code ?

kleinai:
I have no problem with the code and I'd merge it as-is but just that we get the licensing correct.

I'm not a lawyer but we may be okay to use as this long as we keep the original LGPL headers on this class specifically. We already provide the source code so there's no issue there.

Of course that would leave end users with figuring out what code they used is under which license.

Side note, we probably should start enforcing copyright headers on source files. I know I'm guilty of forgetting them.

dnltz:
> I'm not a lawyer but we may be okay to use as this long as we keep the original LGPL headers on this class specifically. We already provide the source code so there's no issue there.
> 
> Of course that would leave end users with figuring out what code they used is under which license.
> 
> Side note, we probably should start enforcing copyright headers on source files. I know I'm guilty of forgetting them.

The entire project is under MIT license and therefore each individual file. Adding source code with a different license is, in my opinion, hard to detect without checking the code in detail and might bring people into a bad situation. I will try to contact the author regarding a license change, as this is the best solution. 

I will also add more information from the source header to this file and will check the possibility to force a disparity.

Todos before merging:
- [ ] Add information from the source header (Author, patens, ...)
- [ ] Ask to change the license
- [ ] Allow changing the disparity through IOs

dnltz:
@kleinai @Dolu1990 - I sent a mail to the Author on the 29th of June. His company mail does not exist anymore and he did not reply from the other mail. Moreover, the original code is 11 years old. I assume he's available for questions anymore.

In the header, he claims this code is based on an IBM publication [1]. It seems this code is indeed based on this paper. The referenced US patent [2] is expired now.

So, I think we can safely publish this code under BSD because the original author's work is based on an expired patent. What do you think?


1: https://hifpga.com/upfiles/15950439066254198.pdf
2: https://patents.google.com/patent/US4486739A/en

Dolu1990:
I guess we could merge then.

Dolu1990:
Thanks :D

PR #791, opened 2022-07-09
Add AvalonST basic support

kleinai:
Avalon Streaming has special timing requirements and allowances which make it incompatible with Spinal streams. This PR will provide a way to model and adapt Avalon-ST between IP and Spinal streams.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/avalon/AvalonST.scala
+++ b/lib/src/main/scala/spinal/lib/bus/avalon/AvalonST.scala
@@ -28,18 +28,18 @@ case class AvalonSTConfig(dataWidth: Int,
 }
 
 case class AvalonSTPayload(config: AvalonSTConfig) extends Bundle {
-  val data: Bits    = if (config.useData) Bits(config.dataWidth bit) else null
-  val channel: UInt = if (config.useChannels) UInt(config.channelWidth bit) else null
-  val error: Bits   = if (config.useError) Bits(config.errorWidth bit) else null
-  val empty: UInt   = if (config.useEmpty) UInt(config.emptyWidth bit) else null
-  val eop: Bool     = if (config.useEOP) Bool() else null
-  val sop: Bool     = if (config.useSOP) Bool() else null
+  val data: Bits    = config.useData generate Bits(config.dataWidth bit)
+  val channel: UInt = config.useChannels generate UInt(config.channelWidth bit)
+  val error: Bits   = config.useError generate Bits(config.errorWidth bit)
+  val empty: UInt   = config.useEmpty generate UInt(config.emptyWidth bit)
+  val eop: Bool     = config.useEOP generate Bool()
+  val sop: Bool     = config.useSOP generate Bool()
 }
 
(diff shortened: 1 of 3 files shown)


Dolu1990:
Cool ^^
What is errorDescriptor: Array[String] = Array(), ?

lib/src/main/scala/spinal/lib/bus/avalon/AvalonST.scala
```diff
@@ -0,0 +1,192 @@
...
+                          emptyWithinPacket: Boolean = false,
+                          errorDescriptor: Array[String] = Array(),
+                          symbolEndianness: Endianness = LITTLE,
+                          readyLatency: Int = 0,
+                          readyAllowance: Int = 0) {
+  assert(!(useReady || useValid), "AvalonST must have at least ready OR valid enabled!")
+}
+
+case class AvalonSTPayload(config: AvalonSTConfig) extends Bundle {
+  val data: Bits    = if (config.useData) Bits(config.dataWidth bit) else null
```
Dolu1990, line 31: An shorter way is : 

```scala
val data = config.useData generate Bits(config.dataWidth bits) //:)
```

lib/src/main/scala/spinal/lib/bus/avalon/AvalonST.scala
```diff
@@ -0,0 +1,192 @@
...
+  val sop: Bool     = if (config.useSOP) Bool() else null
+}
+
+case class AvalonST(config: AvalonSTConfig) extends Bundle with IMasterSlave {
+
+  val ready: Bool = if (config.useReady) Bool() else null
+  val valid: Bool = if (config.useValid) Bool() else null
+  val payload: AvalonSTPayload = AvalonSTPayload(config)
+
+  lazy val latencyDelay: Bool = Delay(this.ready, config.readyLatency)
```
Dolu1990, line 45: Hmm lazy val in a bundle, i'm realy not sure SpinalHDL will like it.
The goal of those signals is just to cache some values if used right ? Not realy to carry them along the bundle right ?

kleinai: Missed this. Yeah those are internal for ready tracking. Avalon as some weird timing relationships. I can un-lazy them if you think there might be problems.
Completely forgot up to now that I was working in a Bundle. XP

kleinai:
I think this is ready.  I don't personally use Avalon streams but Intel IP prefers Avalon so I think this is a nice to have as a start.

lib/src/main/scala/spinal/lib/bus/avalon/AvalonST.scala
Dolu1990: Hi ^^
There is still some lazy val, can you un-lazy them ?
If you want, there is the "signalCache" API.

kleinai: Finally done. Let me know if I'm using signalCache right.

Dolu1990:
Good to me ^^
you can merge when you feel it.

kleinai:
Will do. Thanks!

PR #801, opened 2022-07-22
fix the error generated by xsim emulation of xilinx ip for different devices

liuwei9:
Because the original xsim backend did not specify information about the xilinx chip when generating the vivado project. Therefore, a default xc7vx485tffg1157 project will be generated, and an error will be reported when we use xci files for chips other than this series. As shown below.
I have two xilinx adder ip's for xc7vx485tffg1157-1 and xc7k325tffv900-2.

When I use the adder ip of 325t, the following error is reported.
image

image
[VRFC 10-2063] Module <addAdd> not found while processing module instance <add>

When I use the adder ip of 485t, it is correct.

I have therefore added a method to specify the xilinx device to be used in the simulation.
It works now at both 325t and 485t.

Use the following directions to specify the xilinx device in the added withXilinxDevice method.
image

The simulation results are as follows.
image

### DIFF
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -155,7 +155,7 @@ object SpinalVerilatorSim {
 }
 
 class SpinalVpiBackendConfig[T <: Component](val rtl               : SpinalReport[T],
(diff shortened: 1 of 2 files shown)


kleinai:
Works as described. Thanks!

PR #802, opened 2022-07-22
move SpinalFormalFunSuite to spinal lib.

### LINKED ISSUE #797: It's best to put SpinalFormalFunSuite.scala into lib or sim to help project test with formal.
The existing SpinalFormalFunSuite is a wrap of FunSuite for formal verification.
However, this can only be used to verify SpinalHDL itself.
Move SpinalFormalFunSuite to lib or sim subproject to provide the accessibility of projects using Spinal is necessary.

The problem is where to put this file. 
It seems that lib subproject is not suitable for test related code.
However, sim subproject is also not very fit for that, because it's mainly target to simulation.

Readon:
fix #797


Dolu1990:
Thanks ^^

PR #808, opened 2022-07-26
Add formal verification cases on StreamTransactionCounter and StreamTransactionExtender

Readon:
1. simplify the logic and fix the working signal
2. fix the bug that working signal did not cover the first cycle while noDelay is true.
3. introduce available signal to Counter to indicate if it's ready for new control fire trigger.


Dolu1990:
Let's me know when you want review ^^

Readon:
> Let's me know when you want review ^^

I think it's already for review now.

Dolu1990:
Thanks ^^

Issue #796, opened 2022-07-21
「BUG?」When Merging RTL


dreamflyings:
1. If spinalsystemverilog is used, Verilog is generated after mergertlsource instead of SystemVerilog
2. If spinalconfig is configured as onefilepercomponent, and only report blackboxesSourcesPaths. Add, but without mergertlsource, the files in blackbox addrtlpath will not be copied.
THX！

Dolu1990:
Hi,
Do you have a simple self contained example to reproduce ?

dreamflyings:
> Hi, Do you have a simple self contained example to reproduce ?
Thank you!
Examples of recurrence are as follows:
```scala
import spinal.core._

object MergeRTL0 extends App{
  val report = SpinalSystemVerilog{
    new Module{
      val a = in UInt(8 bits)
      val b = in UInt(8 bits)
      val c = out UInt(8 bits)
      val subMod = new BlackBox {
        setDefinitionName("Sub")
        noIoPrefix()
        val a = in UInt(8 bits)
        val b = in UInt(8 bits)
        val c = out UInt(8 bits)
        setInlineVerilog(
          s"""module Sub(
              | input [7:0] a,
              | input [7:0] b,
              | output[7:0] c
              |);
              | assign c = a + b + (a[1] ^ b[1]);
             |endmodule
             |""".stripMargin)
      }
      subMod.a := a
      subMod.b := b
      c := subMod.c
    }.setDefinitionName("Top")
  }
}

object MergeRTL1 extends App{ // todo Problem 1
  val report = SpinalSystemVerilog{
    class SubMod extends BlackBox {
      setDefinitionName("Sub")
      noIoPrefix()
      val a = in UInt(8 bits)
      val b = in UInt(8 bits)
      val c = out UInt(8 bits)
      addRTLPath("./Sub.sv")
    }
    new Module{
      val a = in UInt(8 bits)
      val b = in UInt(8 bits)
      val c = out UInt(8 bits)
      val subMod = new SubMod
      subMod.a := a
      subMod.b := b
      c := subMod.c
    }.setDefinitionName("Top")
  }
  report.blackboxesSourcesPaths.add("./Top.sv")
  report.mergeRTLSource("Top1")
}

object MergeRTL2 extends App{// todo Problem 1
  val report = SpinalSystemVerilog{
    new Module{
      val a = in UInt(8 bits)
      val b = in UInt(8 bits)
      val c = out UInt(8 bits)
[...]

PR #810, opened 2022-08-01
RTL merge fixes

kleinai:
Fixes for #796

### DIFF
--- a/core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
@@ -23,7 +23,7 @@ package spinal.core.internals
 import spinal.core._
 
 import scala.collection.mutable
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 
 
 class PhaseVerilog(pc: PhaseContext, report: SpinalReport[_]) extends PhaseMisc with VerilogBase {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
```diff
@@ -103,19 +103,21 @@ class PhaseVerilog(pc: PhaseContext, report: SpinalReport[_]) extends PhaseMisc
...
-                outFile = new java.io.FileWriter(targetFilePath)
-                outFile.write(VhdlVerilogBase.getHeader("//", pc.config.rtlHeader, c, config.headerWithDate, config.headerWithRepoHash))
-                outFile.write("`timescale 1ns/1ps ")
-                outFile.write(str)
-                outFile.flush()
-                outFile.close()
-                fileList.write(targetFilePath.replace("//", "/") + "\n")
-                bbImplStrings += str
+            case bb: BlackBox =>
+              bb.listRTLPath.foreach(rtlPath => fileList.write(rtlPath.replace("//", "/") + "\n"))
```
Dolu1990, line 107: One issue seems that if one blackbox is instanciated two time, it will add each time the files in the filelist.
Maybe the filelist should first be a LinkedHashMap of String, which is only at the end written into a file ?

kleinai:
Added the file name hash list. I also noticed the VHDL black box implementation wasn't flushed out when in one file mode. Verilog does this so shouldn't VHDL as well?

Dolu1990:
Hooo you are right, that is missing.

Dolu1990:
All good ?

kleinai:
Yes, I don't think there's anything left to add.

PR #812, opened 2022-08-05
Axi4: Add Axi4IdRemover and tests

kleinai:
Adds a component to perform ID reflection to Axi buses without ID support. Integral part of the translation to Axi4Lite.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4IdRemover.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4IdRemover.scala
@@ -70,13 +70,13 @@ class Axi4SharedIdRemover(config: Axi4Config) extends Component {
 
     txnActive clearWhen(readDone || writeDone)
 
-    io.input.r.arbitrationFrom(io.output.r.continueWhen(txnActive && !txnWrite))
+    io.input.r.arbitrationFrom(io.output.r)
     io.input.r.payload.assignSomeByName(io.output.r.payload)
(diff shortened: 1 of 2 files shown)


lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4IdRemover.scala
```diff
@@ -0,0 +1,136 @@
...
+  } else {
+    val txnActive = RegInit(False)
+    txnActive setWhen(io.input.ar.fire)
+    val txnId = RegNextWhen(io.input.ar.id, io.input.ar.fire)
+
+    io.output.ar << io.input.ar.continueWhen(!txnActive)
+
+    txnActive clearWhen(io.output.r.last && io.output.r.fire)
+
+    io.input.r.arbitrationFrom(io.output.r.continueWhen(txnActive))
```
Dolu1990, line 104: I'm not sure why continueWhen here. I mean, r stream should always come at least one cycle after ar right ? (AXI spec)

kleinai: Yeah I'm not sure why I put that in there. I'm still playing around with this for Axi to Lite conversion.

Dolu1990:
@kleinai What is the status ^^ ?

kleinai:
I believe it's done. It has tests and I rebased it onto dev without issue.

Dolu1990:
thanks ^^

PR #814, opened 2022-08-06
feat(lib): add Stream Unpacker

dokleina:
Current progress for #787


lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -1859,3 +1860,69 @@ class StreamTransactionExtender[T <: Data, T2 <: Data](
...
+    map_layout.keys.zip(unpacker.io.outputs) foreach { case(data, comp_out) =>
+      data.assignFromBits(comp_out.asBits)
+    }
+    unpacker
+  }
+}
+
+class StreamUnpacker[T <: Data](
+  dataType: HardType[T],
+  layout: mutable.LinkedHashMap[Data, Int]
```
Dolu1990, line 1881: Hi ^^

passing Data as arguements of a Component is problematic, as that data would be in the parent component.

The best here i think is to extends Area, that way it is all good, you can directly pass instances of Data and access them.

dokleina: Is there a way to pass in the HardTypes for the Data? I couldn't figure out how to get those from regular Data. It isn't as easy, unlike getting a Stream's `.payloadType`

Dolu1990: It complexify things i would say. you could have layout: mutable.LinkedHashMap[HardType[Data], Int]

But overall, i think the best is to use Area instead of Component, and directly work / assign the Data parameters.

dokleina: I gave up on the HardType map. I couldn't make a Vec of HardTypes but Data works fine.

I am going to move forward with Area instead of Component. Thanks for the suggestion!

dokleina:
Contrary to the lack of activity, this isn't abandoned!

I've got Packing and Unpacking supported and a basic testbench written. It might need to support more cases.

I was thinking of trying to integrate the pack functionality from #889 if that makes sense.

Dolu1990:
> I was thinking of trying to integrate the pack functionality from https://github.com/SpinalHDL/SpinalHDL/pull/889 if that makes sense.

As much as possible, to avoid duplicating features :)

dokleina:
Would it be too much if I went ahead and added some word-centric methods to the pack functions?
I see this as being a shortcut to bit positions based on some word selection. The words themselves wouldn't mean anything until they're handed to a un/packer who has a defined word width to base them on.

If that's too confusing, perhaps the word width could be declared in the PackedBundle before use.

numero-744:
@Dolu1990

dokleina:
Yes! I was thinking about something like this:

```scala
class TestBundlePacked extends PackedBundle(8 bits, BIG) {
  val r = UInt(5 bits) word 0
  val g = UInt(6 bits) word 1
  val b = UInt(5 bits) word 2
  val a = UInt(16 bits) words(
    3 -> (7 downto 0),
    4 -> (15 downto 8)
  )
  val invert = Bool() at 0 word 5
  val ignore = Bool() at 1 word 5
}
```

Where I'd make some changes to `PackedBundle` to optionally take a word width and an endianness. I think under the hood it'll do the tagging that `PackedBundle` already does, but with the knowledge of word widths. I think I'll still need `packTo` and `packFrom` but they'll apply within a word.

dokleina:
Closing this since I've gotten enough feedback. Will open a non-draft PR later with updates.

Dolu1990:
Hi ^^

@kleinai Some changes target PackedBundle, can you take a look ?


About the API,

```scala

    val layout = List(
      input.r -> (0 + offset),
      input.g -> (8 + offset),
      input.b -> (16 + offset)
    ) ++ {
      if (contiguousLayout) {
        List(
          input.a -> 24
        )
      } else {
        List(
          input.a( 7 downto 0) -> 24,
          input.a(15 downto 8) -> 32
        )
      }
    }

    val packer = StreamPacker[Bits](
      io.outStream,
      layout
    )
```

I think that's nice ^^

kleinai:
I have no problems with the changes. The new builder does lend itself to programmatic construction.

Dolu1990:
So ready for merge all ^^ ?

dokleina:
I have nothing more to add. It's all good to me, at least.

kleinai:
Merged. Thanks!

PR #815, opened 2022-08-07
Temperal fix the aggressive prune signal feature of yosys.

Readon:
Yosys would aggressively remove signals while it's wreduce procedure, which lead some non-trivial signals missing.
Add a switch can help to skip the wreduce command while sby command executing.

the withOutWireReduce function can be called from FormalConfig when the wreduce cause verification failed.


Dolu1990:
Thanks :D

PR #821, opened 2022-08-15
feat(lib): add Axi4Unburster

kleinai:
Create an Axi4Unburster which unburstifies the entire Axi4 bus.  There's a little more overhead in generating new last signals and accumulating write response errors.

I also added more event functions to the Axi4 monitors and enriched the existing ones. While I don't strictly need some of these now I will need them in the future for in-progress Axi4 to Lite converter tests.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Unburster.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Unburster.scala
@@ -4,9 +4,11 @@ import spinal.core._
 import spinal.lib._
 
 object Axi4Unburster {
-  def apply(axi: Axi4): Axi4 = {
-    val roUnburstifier = new Axi4ReadOnlyUnburster(axi.config)
-    val woUnburstifier = new Axi4WriteOnlyUnburster(axi.config)
+
+  def apply(axi: Axi4): Axi4 = apply(axi, 3, 3)
+  def apply(axi: Axi4, pendingDepth: Int, pendingWidth: Int): Axi4 = {
+    val roUnburstifier = new Axi4ReadOnlyUnburster(axi.config, pendingDepth, pendingWidth)
+    val woUnburstifier = new Axi4WriteOnlyUnburster(axi.config, pendingDepth, pendingWidth)
 
     roUnburstifier.io.input << axi.toReadOnly()
     woUnburstifier.io.input << axi.toWriteOnly()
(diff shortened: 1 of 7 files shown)


Dolu1990:
Thanks ^^

I would that by default, AXI interconnect element should be assumed to support multple outstanding transactions, as they normaly target "peformance".

This implementation will only allow 1 pending transaction right ? (if yes, that's fine too, maybe just put one comment to make it clear ^^)

kleinai:
Yes as it is now it only supports one transaction. The original intent was to support Lite conversion.
I'll give increasing the performance some thought but otherwise it's already useful.

Dolu1990:
No worries ^^
I guess to support multiple pending transaction you would have to save the context via a FIFO between the AR -> R      and AW -> B

Readon:
> 

This is the cases that StreamTransactionExtender fits.
However, what is the difference between this unburster and existing Axi4AxUnburstified?

Dolu1990:
> However, what is the difference between this unburster and existing Axi4AxUnburstified?

Mostly, it cover the whole bus, not just the AR AW channels.

kleinai:
Axi4AxUnburstified does the major work here but there's an extra step for generating R last or accumulating B resp values.

I'm working on a possible solution for multiple pending transactions with and without ID support.

kleinai:
I have code supporting multiple transactions without IDs.

I'm not sure how to handle ID pending transactions. I'm thinking some number of FIFOs for ID slots which block after they're all allocated to IDs or full of that ID. How's that sound?

Readon:
> I have code supporting multiple transactions without IDs.
> 
> I'm not sure how to handle ID pending transactions. I'm thinking some number of FIFOs for ID slots which block after they're all allocated to IDs or full of that ID. How's that sound?

Pending transaction function might need one FIFO per possible id. 
How about specifying depth for each FIFO?
So if depth equals to 0 then the pending function should be disabled.

Dolu1990:
What i was generaly doing, is to have a fifo for read, a fifo for write, push the input ID in it, and force the output ID to zero.

Upside :  is is light a support multiple outstanding transactions
Downside : It prevent  out of order completion on the slave side, as output id is always zero

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Unburster.scala
```diff
@@ -0,0 +1,295 @@
...
+
+    val unburstified = io.input.ar.continueWhen(manager.io.axIdLen.ready).unburstify
+    manager.io.axIdLen.valid := io.input.ar.fire
+    (config.useId) generate { manager.io.axIdLen.id := io.input.ar.id }
+    manager.io.axIdLen.len := io.input.ar.len
+
+    io.output.ar.arbitrationFrom(unburstified)
+    io.output.ar.payload.assignSomeByName(unburstified.fragment)
+
+    val rFifo = io.output.r.queueLowLatency(8).continueWhen(manager.io.retIdResp.ready)
```
Dolu1990, line 237: I think that fifo shouldn't be integrated in the bridge, to keep things lite, and potentialy the user can added it externaly, or maybe via a parameter.
Mainly, if you use queueLowLatency, on FPGA without distributed ram, it will be very costly, as it will be flip flop based

numero-744:
Hi @kleinai, what is the state of this PR?

kleinai:
I removed the low latency FIFOs.  This component is complete with unit tests.  It might be a little heavy supporting multiple IDs but that leaves open more opportunities for improvement :P

Dolu1990:
I think this can be merged, and maybe in the future, if somebody need it, an option + implementation can be added to the existing component to provide smaller area / no out of order support. Then, maybe that day the default implemention would switch the the lighter implemention.

Readon:
This changes the existing interface of Axi4ReadOnlyMonitor which breaks my existing code .......
Changes of API is really a pain.

Readon:
It might be better to define a compatible interface both.

Take onWriteByte as an example :
``` scala 
    onWriteByte(address: BigInt, data: Byte) //legacy, it can be worked as id equals to 0.
    onWriteByte(address: BigInt, data: Byte, id: Int) //new
```

Readon:
Also, a onLast method has been removed.

Dolu1990:
> This changes the existing interface of Axi4ReadOnlyMonitor which breaks my existing code .......
Changes of API is really a pain.

I understand the paine.

For the changes done, it only "expend" the functionalities ? so add "id : Int" to onWriteByte  and changed onLast into a onResponse with last : Boolean argument.

Issue #813, opened 2022-08-05
Logic to generating pastValid would cause error when prove.


Readon:
The generated code for pastValid in SpinalHDL now is
``` verilog
always @(posedge clk or posedge reset) begin
   if(reset) begin
     formal_with_past <= 1'b0
   end else begin
     formal_with_past <= (!reset)
   end
```
The logic works good when doing BMC check. 
However, when it is in prove mode, the status of formal_with_past could be any value do not obey the asserts and assumes.
When the formal_with_past is initially set to 0 then prove logic would fail.

According to the code that ZipCpu [opened](https://github.com/ZipCPU/wb2axip/blob/ea0fe7fc386353451515a7d760df525c7d617b06/bench/formal/faxil_master.v#L172), use initial keyword to set the formal_with_past seems would solve the problem.

Dolu1990:
Ahhh, so we could adding a xxx.initial(False)  would fix it right ?
Could you give a try ?

Readon:
> 

I think it is related with the generating method of pastValid.
Adding xxx.initial(False) is also a more generic solution.

I will write a minimal example to test it first.

Readon:
I have test with [rtcclock/rtcstopwatch](https://github.com/ZipCPU/rtcclock/blob/0bc3c7fe83fbb17d7e2fbc0dbf8166e5ab2b6734/bench/formal/rtcstopwatch.sby#L1)
by replace the f_past_valid logic with above then validation failed.

The correct version should be
``` verilog
	reg	f_past_valid;
	initial	f_past_valid = 1'b0;
	always @(posedge i_clk)
	   f_past_valid <= 1'b1;
```

So there would be two targets.
1. modify generating mechanism of pastValid to generate correct formal_with_past signal.
2. add initial method for datatypes to fit every signal requires specific initial status.

I am not familiar with those topic, do you might add these? @Dolu1990

Dolu1990:
Shouldn't we also include that on reset f_past_valid should be cleared ? in addition to the initial clear, to be very exaustive and not too specific ?

So : 
```scala
  def pastValid() : Bool = signalCache("formal.pastValid")(past(!ClockDomain.current.isResetActive) init(False) initial(False) setWeakName("formal_with_past"))
``` 

The "initial" feature is already implemented :)

Doesn't it fix all the issues ?

Readon:
> ```scala
>  initial(False)
> ```

[...]

PR #824, opened 2022-08-17
fix past valid definition.

Readon:
So the pastValid is only to indicate there are past values exist.
pastValidAfterReset is defined for synchronous logic with reset logic.
Answer the issue #813 , but not directly fix it.


Dolu1990:
Thanks ^^

PR #826, opened 2022-08-18
skip formal verification while CI.

### LINKED ISSUE #825: sbt test jobs run formal verification also, which slow down the CI.
I don't find the reason [testOption](https://github.com/SpinalHDL/SpinalHDL/blob/8bc3f8539777141b38fef1359f8a7d15c770128e/build.sbt#L207) setting does not work.

How about add a new command alias testWithoutFormal as below
``` scala
addCommandAlias("testWithoutFormal", "testOnly * -- -l spinal.tester.formal")
```
By adding the command into github's workflow definition, the formal verification can be skipped when running the CI.

Readon:
exclude arguments for test are not working now, add testWithoutFormal to reduce the test time while CI.
fix #825

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -19,7 +19,7 @@ jobs:
       run: sbt "+test:compile"
 
     - name: Run tests
-      run: sbt "++2.11.12; test"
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks :)

PR #827, opened 2022-08-18
add no delay function to transaction extender

Readon:
so the extender can output at the cycle that input stream fire.


Dolu1990:
Cool ^^
Thanks !

PR #830, opened 2022-08-21
docs: add links to RTD for Counter

numero-744:
I did not test ScalaDoc to check that the links are correct


numero-744:
Links are correct, can be merged

PR #831, opened 2022-08-21
XSim: Overhaul XSI interface and make it signal width aware

### LINKED ISSUE #828: Erroneous assignment when simulating with `Xsim`
It seems as if many of the least significant bits are discarded.
test.scala
```scala
package test

import spinal.core._

  val io = new Bundle {
    val dataIn  = in UInt (512 bits)
    val dataOut = out UInt (512 bits)
  }

  val reg = RegNext(io.dataIn)
  io.dataOut := reg
}
```
testbench.scala
```scala
package test

import spinal.core.sim._

object Testbench {
    SimConfig.withWave.withXsim
      .compile(new Test)
      .doSim { dut =>
        {
          dut.clockDomain.forkStimulus(period = 10)
          val io = dut.io

          for (_ <- 0 until 100) {
            dut.clockDomain.waitSampling()
            io.dataIn #= BigInt("FFFFFFFF", 16)
            println(io.dataOut.toBigInt)
          }
        }
      }
  }
}
```
With Verilator(removing the `withXsim`) the message is `4294967295`  which is correct, but with `Xsim` this only prints `0`.

kleinai:
Fixes #828 

This is a rewrite of how the XSI interface accesses signal data. I believe the problems in #828 are caused by uninitialized memory or some other issue. On top of that, conversion with `.toInt` and `.toLong` could crash XSim if the underlying signal was not properly sized. Sure this could have been fixed in the JNI interface but I prefer a more robust C++ interface.

### DIFF
--- a/sim/src/main/resources/XSIIface.cpp
+++ b/sim/src/main/resources/XSIIface.cpp
@@ -8,11 +8,6 @@
 #include "XSIIface.hpp"
 #include "xsi_loader.h"
 
-union byte_union {
-    uint32_t word;
-    int8_t byte[4]; 
-};
-
 int64_t pow10(int64_t a) {
     if (a < 0)
         return pow10(-a);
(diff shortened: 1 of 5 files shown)


Dolu1990:
Thanks :D

PR #840, opened 2022-08-31
move wait sampling while pending_reads is empty. partially fix #839

### LINKED ISSUE #839: AxiMemorySim could not simulate continuous read.
The waveform shows that there is an extra idle cycle of data channel's valid, so that the root cause is from AxiMemorySim.
I think it is caused by [this line](https://github.com/SpinalHDL/SpinalHDL/blob/6e128db5ac0a1a7a109439503037108f53fc850b/lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/AxiMemorySim.scala#L404).
There exist one extra waiting sample after one transaction.
It should be moved to "else" block when "pending_reads" is empty.

Readon:


Dolu1990:
Thanks ^^

PR #847, opened 2022-09-06
Add GlobalClock and an example for async verfication.

Readon:


Dolu1990:
Hi ^^

So if i understand well, GlobalClock is a tool to have one global clock generating syncronous sub clocks right ?

Readon:
> Hi ^^
> 
> So if i understand well, GlobalClock is a tool to have one global clock generating syncronous sub clocks right ?

Yes, all the other clocks and resets are retrieved from the global clock. It's the minimum granularity time unit.

Dolu1990:
Ahhh cool ^^
What is the domain.clock.addAttribute("gclk") for ?

Readon:
> Ahhh cool ^^ What is the domain.clock.addAttribute("gclk") for ?

That is a yosys related syntax, to tell yosys that the clock is the global one.

It's a must for async design verification. I have tried to have an input clock instead, but yosys does not know it's global property.

Readon:
Is there any concerns about this PR? @Dolu1990

Dolu1990:
No all good ^^
I just losed track of the PR.

PR #854, opened 2022-09-16
FormalMuxTester

andyspace:
add FormalMuxTester.scala
passed cover, BMC and prove


Dolu1990:
Seems good to me.
@Readon Can you take a look ^^ ?

PR #855, opened 2022-09-16
FormalDeMuxTester

andyspace:
add FormalDeMuxTester.scala
passed cover, BMC and prove


Dolu1990:
Seems good to me too.
@Readon Can you take a look too ?

Issue #873, opened 2022-09-26
Feature: UIntToOh for faster simulation?


tanhongze:
Now, `UIntToOh` is implemented as
```
  def apply(value: UInt, width : Int): Bits = {
    val ret = Bits(width bits)
    for(i <- 0 until width){
      ret(i) := value === i
    }
    ret
  }
```
By using an alternative implement, we could make this simulation much faster(with no impact to synthesis):
```
  def apply(value: UInt, width : Int): Bits = {
    B(1, width bits) |<< value
  }
```

numero-744:
Can you open a PR, please? Else I can do it.

andreasWallner:
Out of curiosity, with which synthesis toolset did you compare the two?
This kind of reminds me of this blog entry: https://andy-knowles.github.io/one-hot-mux/ where Andy Knowles compared multiple HDL implementations of OH encoders.

tanhongze:
I just tested it under Vivado.
For a 5-32 decoder 10-1024 decoder, these implementations of `UIntToOh` would be mapped to exactly same synthesized design. (Thus, same net delay and same logic delay)

Dolu1990:
Merged

PR #874, opened 2022-09-26
perf: make UIntToOh faster in simulation

tanhongze:


numero-744:
Fixes #873

tanhongze:
A detail: what if `width=0`?

Readon:
> A detail: what if `width=0`?

Then there would be also possibilities that value is larger than width.XD

numero-744:
Maybe `width=0` is a special case because `B(1, width bits)` would fail?

If value > width > 0 it would just return 0 I guess?

numero-744:
I think `require(width > 0)` is okay to manage too little width

For this the following part, isn't the default case already okay?

```scala
    if(width > (1<<widthOf(value))){
      val ret = B(1, 1<<widthOf(value) bits) |<< value
      return ret.resize(width)
    }
```

I mean, wouldn't this work with same behavior?

```scala
  def apply(value: UInt, width : Int): Bits = {
    require(width > 0)
    B(1, width bits) |<< value
  }
```

numero-744:
I was going to say that there is an undriven signal with `width<=0`, well there cannot be design errors if there is nothing at all ^^'

LGTM if you have tested it :wink:

Dolu1990:
So, i tested on yosys + ice40, and results seems the same.

Thanks for the PR ^^

PR #880, opened 2022-10-03
gitignore: ignore project/{project,metals.sbt}

numero-744:
I don't know for you, but in my installation it creates these files. (Sometimes it creates `project/project/project` :thinking: :upside_down_face: )

I think they can be ignored.


Dolu1990:
Sure ^^

Issue #829, opened 2022-08-21
[VCS API] When I use VCS as a backend for emulation, it seems to have problems with Z-state warnings and VCS API does not seem to support oneFilePerComponent config


AlenLqx:
Hi， I encountered some error when do VCS simulation，Please take a look. Thank you.

1）Z-State Warning
Take a simple adder as an example：
![(TWA67AQ3HV~A372UIOLV2X](https://user-images.githubusercontent.com/84823360/185778834-d946699a-2692-457e-b116-6274fa6a8b95.png)
Its running results are as follows：
![)XPSKVD5}JC9ACH{(HO@TFS](https://user-images.githubusercontent.com/84823360/185778888-6034e96e-fdab-48b6-ba3a-75af55933aef.png)
It seems to give a Z-state warning that shouldn't be there
2）oneFilePerComponent
Use the following code to simulate：
![FXWP LEC$QBGD1YPDMHFQJH](https://user-images.githubusercontent.com/84823360/185779238-313f8a12-7169-4ba3-a533-3ffbc2b92ca4.png)
I encountered this error after running：
![@4}KY7 M_@@5 {FS_96 Q8C](https://user-images.githubusercontent.com/84823360/185779293-7d4e281c-291b-4e61-8437-ebb15cc1132c.png)

Dolu1990:
Hi :)

To send the code in text format you can do : 

\`\`\`scala
Your code here
\`\`\`

It make things easier for us to reproduce.

Also which version of VCS do you have ?

Charles-Lqx:
Okay.

1) In the Z-State warning case, the example as follow:

/* scala code */
import spinal.lib._
import spinal.core._
import spinal.core.sim._

import scala.util.Random._

case class Demo() extends Component {
  val io = new Bundle {
    val a     = in UInt (8 bits)
    val b     = in UInt (8 bits)
    val ret = out UInt (9 bits)
  }

  io.ret := RegNext(io.a +^ io.b).init(U(0))
}
object demoTest extends App {
  SimConfig.withVCS(VCSFlags(elaborateFlags = List("-LDFLAGS -Wl,--no-as-needed")))
    .withFSDBWave.workspacePath(s"./simWorkSpace").allOptimisation.compile(Demo()).doSim { dut =>
    dut.io.a #= 0
    dut.io.b #= 0
    dut.clockDomain.forkStimulus(10)
    dut.io.a #= nextInt(256)
[...]

PR #882, opened 2022-10-04
Fix bug: only top rtl present in rtl source paths when oneFilePerComponent enable.

wswslzp:
Related to #829


Dolu1990:
Nice thanks ^^

PR #889, opened 2022-10-09
Lib: Add PackedBundle

kleinai:
Simple standard bundle but with support for placing values at specified bit positions. Supports spare packing and endian alignment when data fields need to be truncated.

Long example:
```scala
object PackedExample {
    SpinalVhdl(new Component {
      val packedBundle = new PackedBundle {
        val f1 = UInt(4 bit).pack(15 downto 0, LITTLE)
        val f2 = UInt(8 bit)
        val f3 = UInt(1 bit).packTo(31)
        val f4 = UInt(8 bit).pack(34 downto 32, LITTLE)
        val f5 = UInt(8 bit).pack(38 downto 34, BIG)
        val f6 = UInt(4 bit).pack(48 downto 40, BIG)
      }

      packedBundle.f1.setAll()
      packedBundle.f1.allowOverride
      packedBundle.f4.setAll()
      packedBundle.f4.allowOverride
      packedBundle.f5.setAll()
      packedBundle.f5.allowOverride

      val unpackedBundle = new Bundle {
        val f1 = UInt(4 bit)
        val f2 = UInt(8 bit)
        val f3 = UInt(1 bit)
        val f4 = UInt(8 bit)
        val f5 = UInt(8 bit)
        val f6 = UInt(4 bit)
      }
      unpackedBundle.assignAllByName(packedBundle)

      val io = new Bundle {
        val i = in(packedBundle.asBits.clone())
        val o = out(unpackedBundle.asBits.clone())
      }

      packedBundle.assignFromBits(io.i)
      io.o := unpackedBundle.asBits
    })
  }
}
```

### DIFF
--- a/lib/src/main/scala/spinal/lib/PackedBundle.scala
+++ b/lib/src/main/scala/spinal/lib/PackedBundle.scala
@@ -29,13 +29,13 @@ class PackedBundle extends Bundle {
     elements.map(_._2).map(d => {
       val r = d.getTag(classOf[TagBitPackExact]) match {
         case t: Some[TagBitPackExact] =>
-          t.x.range
+          t.get.range
         case None =>
           (lastPos+d.getBitsWidth) downto (lastPos+1)
       }
       lastPos = r.high
       (r, d)
-    })
+    }).toSeq
   }
 
   override def asBits: Bits = {
@@ -44,7 +44,7 @@ class PackedBundle extends Bundle {
     val packed = B(0, maxWidth bit)
     for ((range, data) <- mappings) {
       val endianness: Endianness = data.getTag(classOf[TagBitPackExact]) match {
-        case t: Some[TagBitPackExact] => t.x.endianness
+        case t: Some[TagBitPackExact] => t.get.endianness
         case _ => LITTLE
       }
       endianness match {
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks ^^
How works UInt(8 bit).pack(38 downto 34, BIG)  for instance ?
i mean  there is 8 bits to put into 5 bits ?

Also endianness is per byte ?

kleinai:
The endian here anchors the MSB or LSB into the packing range. Storing the 8 bit uint into a 5 bit range will drop the top or bottom 3 bits depending on the endian mode.

Dolu1990:
Ahhh ok, i like that it is still kinda a regular Bundle but with tunned  bits conversions ^^

Hmm maybe packAt and packTo are kind of the same meaning without having the same behaviour, which is tricky ?
I mean, i don't see a easy way to rename them XD
Maybe packFrom and packTo ?  from meaning from lsb bit.
As in a sentence we could say from bit 4 to bit 8, but not from bit 8 to bit 4 (instead would be "down to"). ?

Another way would be to have to postfix as : 

xxx.packTo(16 lsb) or something like it

numero-744:
- [x] Scaladoc written
- [x] RTD issue open
- [ ] Unit tested
- [ ] Waiting for review of 2d51f21 by Dolu

Dolu1990:
Seems good to me ^^

numero-744:
If you feel adding unit tests are not required for this, I let you merge this @Dolu1990

kleinai:
While this is fairly simple I should add unit tests. Something basic at the very least.

kleinai:
Provided this passes, this is ready for merge.

Issue #869, opened 2022-09-22
About safe StreamMux and StreamDeMux.


Readon:
According to the verification of Mux and DeMux, the output transaction would be interrupted by the changes of selection.
It might be good to provide a safe version of those facilities.
The question is whether a Stream based selection version or delayed switch version is prefered.

Dolu1990:
Maybe it could be one function for each, i mean : 

```scala
val myDemux = new StreamDemux(..)
val selector = myDemux.io.createStreamSelector()
//alternatively :
// val selector = myDemux.io.createLockedSelector()
```

?

Readon:
> Maybe it could be one function for each, i mean :
> 
> ```scala
> val myDemux = new StreamDemux(..)
> val selector = myDemux.io.createStreamSelector()
> //alternatively :
> // val selector = myDemux.io.createLockedSelector()
> ```
> 
> ?

Good idea, it can keep the compatibility and add function at the same time.

Readon:
> Maybe it could be one function for each, i mean :
> 
> ```scala
> val myDemux = new StreamDemux(..)
> val selector = myDemux.io.createStreamSelector()
> //alternatively :
> // val selector = myDemux.io.createLockedSelector()
> ```
> 
> ?

How about add an function createReg or driveReg of Stream, which can be used as 
```scala
   val readyWhen: Bool = xxx // condition allow register be modified.
   val reg = streamA.driveReg(readyWhen)
```
So that the reg can be terminated by the Reg.
It seems that there are cases that Stream is terminated by registers.

Dolu1990:
> How about add an function createReg or driveReg of Stream, which can be used as

In some ways it already exists as : 

```scala
   val readyWhen: Bool = xxx // condition allow register be modified.
   val reg = streamA.continueWhen(readyWhen).toFlow.toReg
```

So, we could add a toReg toStream directly which would drive the ready High, to avoid the toFlow verbosity. But i think it is better to keep the arbitration (readWhen) outside of the register function, and use continueWhen / haltWhen instead.

In other words, avoiding to much aggregation of different functionalitty, and instead promote composition.
I mean, i'm not sure where the right balance is.

Readon:
> > How about add an function createReg or driveReg of Stream, which can be used as
> 
> In some ways it already exists as :
> 
> ```scala
>    val readyWhen: Bool = xxx // condition allow register be modified.
[...]

Issue #899, opened 2022-10-13
Suggestion of labels for development flow / issue & doc management


numero-744:
* issues start with "bug" or "idea" or "feature"
* "idea" can become "feature" when an API is defined
* "doc required" when no scaladoc OR (missing RTD AND no tracking issue in RTD)
* "doc required" for PRs too, blocking to merge
* need more people with rights to manage labels, to keep them up to date

Dolu1990:
Yes currently label are underused, could be extended.
About having more people, recently a few additional maintainers were added, but so far, nobody is "full time" on SpinalHDL.

numero-744:
Labels updated

PR #901, opened 2022-10-14
add createSelector to Mux and Demux IO.

Readon:
with corresponding verification.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -844,6 +844,13 @@ object StreamMux {
     c.io.select := select
     c.io.output
   }
+
+  def apply[T <: Data](select: Stream[UInt], inputs: Vec[Stream[T]]): Stream[T] = {
+    val c = new StreamMux(inputs(0).payload, inputs.length)
+    (c.io.inputs, inputs).zipped.foreach(_ << _)
+    select >> c.io.createSelector()
+    c.io.output
+  }
 }
 
 class StreamMux[T <: Data](dataType: T, portCount: Int) extends Component {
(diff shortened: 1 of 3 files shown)


Readon:
fix #869

Readon:
I think the doc needed label should be assigned after the PR is merged, due to there are would be PR will not be merged.

Readon:
> Hmm, maybe, instead of having doc needed tag in the SpinalHDL repo, another issue should be open in the SpinalDoc-RTD repo. I don't know what is the best aproache
> 
> The issue i see is that this will start to create many many issue / to track stuff, and endup making the whole flow cumbersome XD

Emm, that's really a problem, is there best practice that GitHub recommanded?

numero-744:
Yes the idea is to have tracking issues.

This is why the description (if you let your mouse on the label) is that there is no documentation or tracking issue. So you can either open an issue so that we do not forget it misses documentation (and document it soon™); or open a PR with the documentation, and do not merge it before the code PR is merged.

numero-744:
> The issue i see is that this will start to create many many issue / to track stuff, and endup making the whole flow cumbersome XD

IMO having tracking issues is useful to not forget things and see how much things are not documented, to act and keep things documented for users. Having features not documented for a public project is as useful as not implementing these features.

numero-744:
> Cross repository PR is really not friendly.

I agree. We have to do two pull requests which have a dependency in time (do not merge docs before the code) because RTD is in a distinct repository. To be honest I think there is no better way to have an RTD, and having an RTD is user-friendly (discussed in https://github.com/SpinalHDL/SpinalDoc-RTD/issues/26).

Dolu1990:
In some ways, the SpinalDoc-RTD documentation often has to come later, when the feature has been used a bit and stabilised after some time. 
To me, if the functions added aren't self explenatory enough by their own name / parameter name, adding some scala comments / scala doc is enough. I guess ?

So, i'm not sure coupling too strongly SpinalHDL and SpinalDoc is a good thing.

Also, i think for a PR if this kind, in code doc is good enough.

numero-744:
> Ahhh, the topic is far away from the original one, I will merge if it is good enough.

True XD
If you want, feel free to re-open #899 to discuss docs management.

PR #903, opened 2022-10-15
Rework on VCS Flags

wswslzp:
Hi all, 

Since VCS is supported in SpinalHDL, some folks (e.g. one of my colleagues) complain about the default VCS flags that were embedded in the source code, and cannot be changed by API. Thus, I have reworked the VCSFlag class and provided some capabilities to overwrite the default flags, except for the flags that are generated automatically by SpinalHDL for compiling VPI module, etc. 

Though the VCSFlags class is reworked, it's still backward-compatible. It should not impact the old code that uses old API.

During my reworking, I also refactored the source codes of several simulator backends. In the original source codes, those simulator backend classes are written in the same VPIBackend.scala. Now each of them is separated into a single class file. This process is done by IDEA IntelliJ's refactoring functionality so this part should not occur errors. 

Feel free to leave your comments and suggestions here.


sim/src/main/scala/spinal/sim/VCSBackend.scala
```diff
@@ -0,0 +1,374 @@
...
+
+import org.apache.commons.io.FileUtils
+import spinal.sim.vpi.SharedMemIface
+
+import java.io.{File, PrintWriter}
+import java.nio.file.{Files, Paths}
+import scala.sys.process.{Process, ProcessLogger}
+
+import scala.collection.mutable
+case class VCSFlags(
```
numero-744, line 11: Thanks for the contribution!

Idea: can splitting flags this way be useful for other simulators? It could look like this:

```scala
trait SimulatorFlags {
  val xxxFlags: mutable.ArrayBuffer[String] = mutable.ArrayBuffer[String]()

  final def withXxxFlags(flags: String*): this.type = ...
  final def withDefaultXxxFlags: this.type = this.withXxxFlags(defaultXxxFlags)...
  final def withDefaultFlags: this.type = this.withDefaultXxxFlags...

  protected def defaultXxxFlags: mutable.ArrayBuffer[String]
}
```

And each backend would define its own implementation:

```scala
case class VerilatorFlags extends SimulatorFlags {
  def defaultVlogFlags = ... // Flags for cc?
  def defaultVhdlFlags = ??? // cannot be used as Verilog cannot build Vhdl
  def defaultElabFlags = ... // Flags for ld?
  ...
```

sim/src/main/scala/spinal/sim/IVerilogBackend.scala
```diff
@@ -0,0 +1,171 @@
...
+
+import org.apache.commons.io.FileUtils
+import spinal.sim.vpi.SharedMemIface
+
+import java.io.{File, PrintWriter}
+import java.nio.file.{Files, Paths}
+import scala.sys.process.Process
+
+
+/* README first!
```
numero-744, line 11: I think it should be turned into:

```scala
@deprecated("This backend ...")
```

So that it appears on user side

sim/src/main/scala/spinal/sim/VCSBackend.scala
```diff
@@ -0,0 +1,374 @@
...
+                     vloganFlags    : mutable.ArrayBuffer[String],
+                     vhdlFlags    : mutable.ArrayBuffer[String],
+                     elaborateFlags    : mutable.ArrayBuffer[String],
+                     runFlags    : mutable.ArrayBuffer[String]
+                   ) {
+  // todo: 1. has default flags
```
numero-744, line 17: Is it blocking for the PR?

numero-744 (commented):
Also, I think it was a good thing to split this file :smile:

sim/src/main/scala/spinal/sim/VCSBackend.scala
wswslzp: It's case by case, I think. This can be discussed further.

sim/src/main/scala/spinal/sim/IVerilogBackend.scala
wswslzp: Ahh, for this one, we should discuss it with @Dolu1990 

For me, IVerilogBackend works well. Why does it not work?

sim/src/main/scala/spinal/sim/VCSBackend.scala
wswslzp: Will remove.

sim/src/main/scala/spinal/sim/VCSBackend.scala
Dolu1990: To me, simulator flags are all different stuff for different environnement, i'm not sure there is much gain trying to make stuff common.
But my opinon isn't realy strong.

sim/src/main/scala/spinal/sim/IVerilogBackend.scala
Dolu1990: We realy need a working backend for IVerilog, @numero-744 what is the replacement ?

sim/src/main/scala/spinal/sim/VCSBackend.scala
numero-744: Was just an idea, one™ should compare how simulators work.

Maybe not for this PR, I'll do that maybe later, maybe never ^^'

wswslzp:
Reopen a new PR #911 that merge the conflicts.

PR #904, opened 2022-10-17
refactor: RegIf Overhaul and Clean Up

kleinai:
RegIf has potential but is messy to read through and doesn't lend itself to adding more instance types (RAM, streams, etc.).
I think this code needs clean up and abstraction. There's a lot of things shared between the classes that could easily be derived from inheritance.

Edit: Reduce scope to clean up and extend from BusSlaveFactory


Readon:
The stream interface of RegIf might be complicate.
If the slave's ready have to wait a long time while valid is asserted, then the bus would stall, which is unacceptable.

A flow might be better in this case, user have to take the responsibility to ensure the slave is ready to be read or write. 
it can be achieved by status regs. However, that would be a using pattern not the lib itself.

kleinai:
I'd like to leverage my `PackedBundle` from #889 to offload the register packing to something reusable. It looks like `readBits` is marked as ToDo from years ago.

wswslzp:
How about merging the BusSlaveFactory into RegIf or building RegIf upon the BusSlaveFactory? Like #352 do. Something like block memory R/W has been implemented.

Dolu1990:
(so, not necessarly a merge, but a composition)

wswslzp:
Yes, that's the idea. RegIf doesn't have to implement most R/W logic itself but utilizes the API from BusSlaveFactory. What RegIf API should do is wrap BusSlaveFactory and provide some additional functions like interrupts design and documentation generation. And also keep the RegIf API behavior unchanged.

kleinai:
I agree it would be easier to rework RegIf to use the old factory code. RegIf doesn't add anything new that couldn't be planted on top of BusSlaveFactory. This would save having to make two bus interfaces.

kleinai:
How about this plan, cut RegIf into pieces.  Integrate the documentation into BusSlaveFactory which will give us all types (streams, flows, registers, whatever).  Then move all the interrupt code into some sort of interrupt controller(?)

In this way we keep all the work in the documentation generation which exceeds BusSlaveFactory's while leveraging the pre-existing bus generation code which RegIf lacks.

kleinai:
Scratch the last plan. I directly ported the code to `BusSlaveFactory` and it was waaay easier.

Had to implement `readFire` and `writeFire` for each factory type to get the examples to compile.

NA handling is an issue and generates -1 bit widths. Still needs work.

kleinai:
@jijingg You might want to know I'm reworking your code.

jijingg:
@kleinai  As I said before, BusSlaveFactory and BusIf have completely different ways of collecting and checking the validity of registers. BusSlaveFactory will one by one generate the circuit, while BusIf needs to collect all the registers for validity check first, and then generate the circuit generation document. What you can see is that the bus interfaces are similar. If you try to merge, they will be neither fish nor fowl. So at present, it seems a little imperfect, but refactoring it means that new problems will be brought.
1. For compatibility problems caused by APIs, If regif  removed, now the new version will cause other people's old code to crash. This is unacceptable
2. After merging, there will be multiple APIs, which will cause confusion.

We might as well optimize from these directions:
1. Generate HTLML/CHEADER/JSON independently.  I realized that this problem was necessary, and had time to practice it due to the progress of the project
2. Create a new register API on the BusSlavefactory  with valid-check, UVM-25 accessType, and keep API simple and intuitive ,  If be ready , please review the API before starting.
3. Add FIFO/ RAMS Aceess API

In this way, our old code can work, and it is possible to discuss a new and better register interface API or work flow

jijingg:
> The stream interface of RegIf might be complicate. If the slave's ready have to wait a long time while valid is asserted, then the bus would stall, which is unacceptable.
> 
> A flow might be better in this case, user have to take the responsibility to ensure the slave is ready to be read or write. it can be achieved by status regs. However, that would be a using pattern not the lib itself.

In addition, I agree that do not try to add a Stream to ReigsterInterface. This will make things more complicated and unnecessary. The register read/write timing is generally fixed, and there is no need to shake hands.

kleinai:
> @kleinai As I said before, BusSlaveFactory and BusIf have completely different ways of collecting and checking the validity of registers. BusSlaveFactory will one by one generate the circuit, while BusIf needs to collect all the registers for validity check first, and then generate the circuit generation document. What you can see is that the bus interfaces are similar. If you try to merge, they will be neither fish nor fowl. So at present, it seems a little imperfect, but refactoring it means that new problems will be brought.
> 
> 1. For compatibility problems caused by APIs, If regif  removed, now the new version will cause other people's old code to crash. This is unacceptable

We'll keep RegIf around but as a component riding on top of BusSlaveFactory for code reuse. It would mean changing a bit of the instantiation APIs but maybe we can work out a wrapper.

> 2. After merging, there will be multiple APIs, which will cause confusion.
> 
> We might as well optimize from these directions:
> 
> 1. Generate HTLML/CHEADER/JSON independently.  I realized that this problem was necessary, and had time to practice it due to the progress of the project

A lot of the core of this is already merged when we consolidated the data traits. These new changes will have little to no impact for the generation APIs.

> 2. Create a new register API on the BusSlavefactory  with valid-check, UVM-25 accessType, and keep API simple and intuitive ,  If be ready , please review the API before starting.

We can keep RegIf separate for now. It's easier to slowly chip away at the problem. It's fairly simple to port to the bus factory functions as it is now.

> 3. Add FIFO/ RAMS Aceess API
> 
> In this way, our old code can work, and it is possible to discuss a new and better register interface API or work flow

Readon:
> > The stream interface of RegIf might be complicate. If the slave's ready have to wait a long time while valid is asserted, then the bus would stall, which is unacceptable.
> > A flow might be better in this case, user have to take the responsibility to ensure the slave is ready to be read or write. it can be achieved by status regs. However, that would be a using pattern not the lib itself.
> 
> In addition, I agree that do not try to add a Stream to ReigsterInterface. This will make things more complicated and unnecessary. The register read/write timing is generally fixed, and there is no need to shake hands.

There is already toReg method according to #869. so FIFO interface

kleinai:
@jijingg I think this is ready for review.

I diff'd the outputs.  The documentation generation is exactly the same.  The HDL output is of course different and even a little smaller when ran on the RegIf example.

Are there any unit tests for RegIf?

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/WishboneBusInterface.scala
```diff
@@ -1,48 +1,16 @@
...
-  if (readSync) {
-    readError.setAsReg() init False
-    readData.setAsReg() init 0
-  } else {
-    readError := False
-    readData := 0
+import spinal.lib.bus.wishbone.{Wishbone, WishboneSlaveFactory}
+
+object WishboneBusInterface {
+  @deprecated
```
numero-744, line 7:
```suggestion
  @deprecated("Use BusInterface(WishboneSlaveFactory(bus), sizeMap, regPre) instead")
```

numero-744:
I'm curious, what is the difference between:

* Bus slave factory
* RegIf (Register Interface?)
* BusIf (Bus Interface?)

jijingg:
> I'm curious, what is the difference between:
> 
> * Bus slave factory
> * RegIf (Register Interface?)
> * BusIf (Bus Interface?)

BusSlaveFactory and BusIf are handle of Bus to creat register , they are similar 
RegIf is a lib name for RegisterBankSolution at `spinal.lib.bus.regif` which use BusIf.

jijingg:
> @jijingg I think this is ready for review.
> 
> I diff'd the outputs. The documentation generation is exactly the same. The HDL output is of course different and even a little smaller when ran on the RegIf example.
> 
> Are there any unit tests for RegIf?

Let me do some regression on the old code，In addition, some APIs need to be little update

jijingg:
hi @kleinai 
I found these problems when run on old code

1. simlation fail, read operation fail, because. 
`bus.PSEL(selId)  && !bus.PWRITE` not `bus.PSEL(selId) && bus.PENABLE && !bus.PWRITE`,    it' little different to BusSlaveFactory define.  
2. BusSlaveFactory does not handle errors/PSLVEROR, PRESP 
3. code generate fail
image


This is also the reason why I gave up using BusSlaveInterface at the beginning, because BusIf is only for such simple scenarios as register reading and writing, as long as it is correct. If I forcibly modify the BusSlave Inteface to adapt registerinterface use,I am afraid that the stability of the BusSlave may be damaged

wswslzp:
> hi @kleinai I found these problems when run on old code
> 
> 1. simlation fail, read operation fail, because.
>    `bus.PSEL(selId)  && !bus.PWRITE` not `bus.PSEL(selId) && bus.PENABLE && !bus.PWRITE`,    it' little different to BusSlaveFactory define.
> 2. BusSlaveFactory does not handle errors/PSLVEROR, PRESP
> 3. code generate fail
>    image
> 
> This is also the reason why I gave up using BusSlaveInterface at the beginning, because BusIf is only for such simple scenarios as register reading and writing, as long as it is correct. If I forcibly modify the BusSlave Inteface to adapt registerinterface use,I am afraid that the stability of the BusSlave may be damaged

The first two problems seem to be the bugs/incorrect implementation of BusSlaveFactory? And the second problem may be already fixed by #927.

jijingg:
> The last one looks like an address conflict. The root cause may be in the wrapper layer, not BusSlaveFactory.

This basic demo cannot be compiled

```scala
class RegIfExample0 extends Component {
  val io = new Bundle{
    val apb = slave(Apb3(Apb3Config(16,32)))
    val a, b, c, d, e = in Bool()
  }

  val busif = BusInterface(io.apb,(0x000,1 KiB), 0, regPre = "AP")

  val M_TURBO_EARLY_QUIT    = busif.newReg(doc = "Turbo Hardware-mode register1")
  val early_quit  = M_TURBO_EARLY_QUIT.field(Bool(), RW, 0, doc = "CRC validate early quit enable").asOutput()
  val early_count = M_TURBO_EARLY_QUIT.field(Bits(2 bit), RW, 2, doc = "CRC validate early quit enable").asOutput()
  val M_TURBO_THETA         = busif.newReg(doc = "Turbo Hardware-mode register2")
  val theta       = M_TURBO_THETA.field(Bits(4 bit), RW, 0x6, doc = "Back-Weight, UQ(4,3), default 0.75").asOutput()
  val bib_order   = M_TURBO_THETA.field(Bool(), RW, 1, doc = "bib in Byte, defalut. \n 0:[76543210] \n 1:[01234567]").asOutput()
  val M_TURBO_START         = busif.newReg(doc = "Turbo Start register")
  val turbo_start = M_TURBO_START.field(Bool(), W1P, 0, doc = "turbo start pulse").asOutput()
  val M_TURBO_MOD           = busif.newReg(doc = "Turbo Mode")
  val umts_on   = M_TURBO_MOD.field(Bool(), RW, 1, doc = "1: umts mode\n0: emtc mode").asOutput()
  val M_TURBO_CRC_POLY      = busif.newReg(doc = "Turbo CRC Poly")
  val crc_mode  = M_TURBO_CRC_POLY.field(Bool(), RW, 1, doc = "0: CRC24; 1: CRC16").asOutput()
  val crc_poly  = M_TURBO_CRC_POLY.field(Bits(24 bit), RW, 0x864cfb, doc = "(D24+D23+D18+D17+D14+D11+D10+D7+D6+D5+D4+D3+D+1)").asOutput()
  val M_TURBO_K             = busif.newReg(doc = "Turbo block size")
  val K         = M_TURBO_K.field(Bits(12 bit), RW, 32, doc="decode block size \n max: 4032").asOutput()
  val M_TURBO_F1F2          = busif.newReg(doc = "Turbo Interleave Parameter")
  val f1        = M_TURBO_F1F2.field(Bits(9 bit), RW, 0x2f, doc="turbo interleave parameter f1").asOutput()
  M_TURBO_F1F2.reserved(7 bits)
  val f2        = M_TURBO_F1F2.field(Bits(9 bit), RW, 0x2f, doc="turbo interleave parameter f2").asOutput()
  val M_TURBO_MAX_ITER      = busif.newReg(doc = "Turbo Max Iter Times")
  val max_iter  = M_TURBO_MAX_ITER.field(Bits(6 bits), RW, 0, doc="Max iter times 1~63 avaliable").asOutput()
  val M_TURBO_FILL_NUM      = busif.newReg(doc = "Turbo block-head fill number")
  val fill_num  = M_TURBO_FILL_NUM.field(Bits(6 bits), RW, 0, doc="0~63 avaliable, Head fill Number").asOutput()
  val M_TURBO_3G_INTER_PV   = busif.newReg(doc = "Turbo UMTS Interleave Parameter P,V")
  val p     =  M_TURBO_3G_INTER_PV.field(Bits(9 bits), RW, 0,doc="parameter of prime").asOutput()
  M_TURBO_3G_INTER_PV.reserved(7 bits)
  val v     =  M_TURBO_3G_INTER_PV.field(Bits(5 bits), RW, 0,doc="Primitive root v").asOutput()
  val M_TURBO_3G_INTER_CRP  = busif.newReg(doc = "Turbo UMTS Interleave Parameter C,R,p")
  val Rtype     =  M_TURBO_3G_INTER_CRP.field(Bits(2 bits), RW, 0,doc="inter Row number\n0:5,1:10,2:20,3:20other").asOutput()
  val CPtype    =  M_TURBO_3G_INTER_CRP.field(Bits(2 bits), RW, 0,doc="CP relation\n0: C=P-1\n1: C=p\n2: C=p+1").asOutput()
  val KeqRxC    =  M_TURBO_3G_INTER_CRP.field(Bool(), RW, 0,doc="1:K=R*C else 0").asOutput()
  M_TURBO_3G_INTER_CRP.reserved(3 bits)
  val C         =  M_TURBO_3G_INTER_CRP.field(Bits(9 bits), RW, 0x7,doc="interlave Max Column Number").asOutput()
  val M_TURBO_3G_INTER_FILL = busif.newReg(doc = "Turbo UMTS Interleave Fill number")
  val fillPos   =  M_TURBO_3G_INTER_FILL.field(Bits(9 bits), RW, doc="interlave start Column of fill Number").asOutput()
  M_TURBO_3G_INTER_FILL.reserved(7 bits)
  val fillRow   =  M_TURBO_3G_INTER_FILL.field(Bits(2 bits), RW, doc="interlave fill Row number, 0~2 avaliable\n n+1 row").asOutput()

  busif.interruptFactory("T", io.a, io.b, io.c, io.d, io.e)

  busif.accept(CHeaderGenerator("header", "AP"))
  busif.accept(HtmlGenerator("regif", "AP"))
  busif.accept(JsonGenerator("regif"))
  busif.accept(RalfGenerator("regif"))
}

```

jijingg:
@kleinai First of all, I would like to thank your PR very much. Your suggestion is very good, but compared with the benefits, I think this revision is more risky. It did not bring about a particularly substantial improvement, even though BusInterface and BusSlavefactor seem to be a little repetitive and imperfect. But it's not that bad. 

Busif has a large amount of code, and the uncertainty brought by the revision of this scale is unacceptable to the risk of existing projects. Unless we can perform strict simulation testing on those new code, It's also a huge job.

Again, the return is not proportional to the risk.

kleinai:
> This basic demo cannot be compiled. and I found that readlogic was completely rewritten. What is the purpose？

The read logic was removed because all reads should be handled directly by BusSlaveFactory.

The bug here was that RO doesn't specify a range when reading and instead reads the whole bus. I could go either way as to if this should be an error but the factory doesn't like it so it's fixed.
There was a second bug in the upgrade of WB but I've patched that too.

> Busif has a large amount of code, and the uncertainty brought by the revision of this scale is unacceptable to the risk of existing projects. Unless we can perform strict simulation testing on those new code, It's also a huge job.

This would be a huge plus and should be done. I take it there aren't any self checking tests, are there?

numero-744:
If code can break without changing the tests it is just that there was not enough testing, and we have to fix it anyway. It will require method, maybe it is a big work but IMHO if it can bring **better code and tests** (ie. lead to something easier to maintain), and if we have someone who wants to work on this, it is worth!

kleinai:
I'm working on basic unit tests to verify both this and the mainline implementation. I didn't see any RegIf unit tests in the mainline already.

kleinai:
Waiting on #945

jijingg:
I will make some modifications and add special test, which may take a little time. These tests will add Spinal Tests for regression use

PR #539, opened 2021-12-06
Use scalafmt to format code

jiegec:

Issue #535, opened 2021-11-23
Consider using code formatting tools 


jiegec:
The current code is pretty inconsistent with regard to indents, spaces etc. Scalafmt can be used to format the scala sources.

Dolu1990:
We can try it, I'm a bit scared that as as SpinalHDL is a idsl, it may trigger quite some false positive

jiegec:
Closed by #539

Dolu1990:
> formatting all files at once; might be annoying if there are currently big PRs modifying lots of existing files (related to https://github.com/SpinalHDL/SpinalHDL/pull/539#issuecomment-1083185452 I guess)

I think this is a no go, it would also make others people forks quite hard to merge

> formatting files not modified by currently open PRs (I hope there is a way to automate this)

This doesn't has the visbility on fork not being PR yet.

> make the lint return warnings so that GitHub sorts them out, to only show those related to the current PR (I'm not 100% sure about it, and having a hard pass/fail would be better IMO).

This would be ok i think, as long as it isn't too rigid and allows hand crafted code alignment, i guess.

>  (IMO 120 chars per line is a lot, the maximum is usually between 80 and 100)

I think here there is a few factors which make 120 necessary : 
- Hardware description is a lot about nested scope
- Hardware description require often a lot of chained doted notation

"Random" example from the NaxRiscv LSU : 
```scala
          setup.cacheStore.cmd.mask :=  AddressToMask(setup.cacheStore.cmd.address, size, widthOf(setup.cacheStore.cmd.mask))
```

numero-744:
I 80% agree your first point, and 100% the second one.

> This would be ok i think, as long as it isn't too rigid and allows hand crafted code alignment, i guess.

It would just be a check, we can always ignore it. If it is set to be a warning, I think it will just print stuff in the diff and not even put a red cross on the actions at the end of PR discussion. About code alignment, if you are talking about things like aligning assignment tokens, IMHO it is really useful for verbose languages like VHDL and Verilog, but not for modern DRY languages like Scala (and SpinalHDL).

To make the code base slowly improve, we could imagine formatting files as we modify them, sometimes, as a side-effect of other PRs, (in a specific commit so that it is cleaner and easier to review), for those who want/care. Because if someone else modifies the same file, there will probability be conflicts anyway. I mean, there is always an increased risk of conflict but here as the initial risk is > 0 then the increase is lower XD

[...]

PR #906, opened 2022-10-17
Allow braces

numero-744:
Blocking point: missing documentation for 2 methods.

### DIFF
--- a/core/src/main/scala/spinal/core/Data.scala
+++ b/core/src/main/scala/spinal/core/Data.scala
@@ -466,7 +466,7 @@ trait Data extends ContextUser with NameableByComponent with Assignable with Spi
     addTag(allowDirectionLessIoTag)
   }
 
-  // TODO doc
+  /** Allow a register to be partially assigned */
   def allowPartialyAssigned(): this.type = {
     addTag(AllowPartialyAssignedTag)
   }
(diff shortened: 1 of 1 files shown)


kleinai:
So you've added braces to a few functions but then reformatted all of Data and renamed a functions? Maybe we should move the formatting and renaming changes to another PR? I think a few people have tried to reformat the code before.

numero-744:
Well actually, I have the "auto-format" reflex (just like the "save" reflex XD), which is not an issue if the code is kept formatted using the official style / program.

I'm gonna split this PR in 3:

* allow braces
* rename 2 private functions
* add a GH action to check that code is correctly formatted

So that we can discuss each of them.

Readon:
there are disscussions on format tools in #535 and #539 .

numero-744:
I cannot find documentation for those 2 items:

* `allowPartialyAssigned`: should it be linked to a new RTD page about `PARTIALLY ASSIGNED REGISTER` design error? [Phase.scala:2025](https://github.com/SpinalHDL/SpinalHDL/blob/d12c8b917eb2a3680f5eb15dac4d1a34b1bbcdcd/core/src/main/scala/spinal/core/internals/Phase.scala#L2025))
* `noBackendCombMerge`: merge what?

Dolu1990:
@numero-744 

> allowPartialyAssigned

Ahh yes it isn't documented, yes it is to skyp the PARTIALLY ASSIGNED REGISTER check for the given signal.
would need some doc ^^

> noBackendCombMerge

With that option enable, each combinatorial signal which require a always/process block (HDL) will have its own one. While if the option is disabled, then SpinalHDL will merge always/process block which use the same when statements ( but because HDL sucks, it may create endless simulation loop in some specific cross dependence cases ).

Dolu1990:
If tests pass it can be merged ^^

PR #907, opened 2022-10-17
remove storage/ + todo.md

numero-744:
Given what `storage/` looks like I think we can delete XD

Also, `todo.md` seems really outdated (and should be replaced by tracking issues if needed)

(Noticed while creating `CONTRIBUTING.md`)


Dolu1990:
Yes, quite some legacy stuff XD

PR #908, opened 2022-10-17
Hide ci

numero-744:
I'm not sure if `travis.sh` can be removed


Dolu1990:
Travis stuff can be removed, that's dead.

But tools.sh isn't realy github specific, so shouldn't be moved in .github i think.

Dolu1990:
Hmmm
"Collection of shell functions used to install EDA tooling"

PR #910, opened 2022-10-18
refactor: isEquals -> isEqualTo

numero-744:
My internal grammar checker was feeling really bad about this XD

It is private so it is not a breaking change (API not changed).


Dolu1990:
Peace to your grammar checker ^^

PR #911, opened 2022-10-18
Rework on VCS Flags

wswslzp:
Merge the conflicts. 

Original PR #903


sim/src/main/scala/spinal/sim/IVerilogBackend.scala
```diff
@@ -0,0 +1,171 @@
...
+    val verilogSourcePaths = rtlSourcesPaths
+      .filter { s =>
+        (s.endsWith(".v") ||
+        s.endsWith(".sv") ||
+        s.endsWith(".vl"))
+      }
+      .mkString(" ")
+
+    val simulationDefSource = s"""
+                               |`timescale 1ns/1ns
```
kleinai, line 100: Missed the timescale update

sim/src/main/scala/spinal/sim/VCSBackend.scala
```diff
@@ -0,0 +1,377 @@
...
+
+  def withDefaultVloganFlags(): this.type = {
+    vloganFlags ++= List(
+      "-nc",
+      "-full64",
+      "-sverilog",
+      "+v2k",
+      "-ntb",
+      "-debug_access+all",
+      "-timescale=1ns/1ps",
```
kleinai, line 32: Should we be setting a default precision (and scale)? Shouldn't the sim take the highest precision from the source files?

wswslzp: oops, I forgot to remove these lines.

wswslzp:
I wonder why regression tests do not work when I push new commits.

wswslzp:
> The status is that if it pass the regression test it can get merged right ?

Yes if @kleinai doesn't have any other problems.

sim/src/main/scala/spinal/sim/VCSBackend.scala
numero-744: Removed in https://github.com/SpinalHDL/SpinalHDL/pull/911/commits/179e86f91bfe6a9a768a7aea26e1b02682400e02#diff-7d9bf2ef5d69afec7b529e62bdc45c2de4fddaaf7ecd156895a5a4740cd6c5d5

kleinai:
Looks good. Thanks!

PR #913, opened 2022-10-19
docs: remove old version

numero-744:
Update: #918 will push automatically `scaladoc` from `dev` branch and releases into one folder per branch/release. The outdated content can be deleted.

<details><summary>Previous version</summary>

While cleaning branches (removed old branches without commits ahead `dev`) I noticed a `gh-pages`. Do we want to keep it updated (to be automated with GH actions)?

If yes:

Idea: like for RTD, add one folder for each tag / branch (done only for `dev` for now)

Known issues for now:

* `sim` package is missing (`build.sbt` configuration needs an update?).
* template is really 2000-ish ^^'

Preview available here: https://numero-744.github.io/SpinalHDL/dev/

</details>


Dolu1990:
> Do we want to keep it updated (to be automated with GH actions)?

I would say yes. (but personnaly i'm never using it)

> Idea: like for RTD, add one folder for each tag / branch (done only for dev for now)

Yes i think that's how it should be done for the best result.

> sim package is missing (build.sbt configuration needs an update?).

Yes, in build.sbt, here is the filter: 
    unidocProjectFilter in (ScalaUnidoc, unidoc) := inProjects(lib, core)
    
Also, commands to build the doc seems to be at the bottom of the build.sbt XD

> template is really 2000-ish ^^'

Maybe this can be updated by updating project/plugin.sbt :
addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.1")

numero-744:
> unidocProjectFilter in (ScalaUnidoc, unidoc) := inProjects(lib, core)

Updated locally and fixed, thanks :smile:

> addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.1")

Updated locally and did not fix. I think it just merges docs and if I generate docs without merging `sbt doc` it also has this old template… I will investigate

numero-744:
For the template the issue was that used Scala version was old (2.11). Replaced by 2.12 (incremented the index of chosen version among selected versions).

numero-744:
@Dolu1990 for review and merge if approved

Dolu1990:
Looking at the diff size, i'm asking myself if having the Scala doc in the SpinalHDL repo is realy a good idea.
I mean, this will create a lot of data on a repo which is normaly just for code.
Which overtime may mess things up with oversized git.

Maybe the best is to have a separated repo for it ?

numero-744:
What do you think about the `single-commit` option of https://github.com/JamesIves/github-pages-deploy-action (action used in #918). This option can be added to #918? I don't know if it actually removes data or if it just unreferences the commit from branches…

numero-744:
@Dolu1990

Dolu1990:
@numero-744 I think single-commit is good :D

PR #914, opened 2022-10-19
build: fix simple warnings

numero-744:
Not all warnings are fixed, this is just a first pass.

I'll continue but this first pack of fixes is consistent alone too.


Dolu1990:
Thanks ^^

PR #915, opened 2022-10-20
fix: regif PSLVERROR bug fix

jijingg:
When a read error occurs, the PSLVERROR will be pulled up until the next correct read, which will lead to intermediate write operations and PSLVERROR errors.  Fixed


lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
```diff
@@ -22,6 +22,8 @@ case class AhbLite3BusInterface(bus: AhbLite3, sizeMap: SizeMapping, regPre: Str
   val doRead      = bus.HREADY & askRead
 
   val addressDelay = RegNextWhen(bus.HADDR, askRead | askWrite)
+  readError.clearWhen(askWrite)
```
Dolu1990, line 25: The error signal is for the data phase of AHB right ? or is it for the address phase ?

numero-744: Error signal is HRESP, and is for data phase. It is currently set to `False` (never errors).

Btw I think this implementation does not match AHB protocol. Should I PR it?

jijingg: Yes, the AHBLite3BusInterface HRESP not completed here .
In a word, my requirement is that if an error (HRESP/PSLVERROR) occurs in a read/write operation, it should not continue to the next access. The APB3bus interface has such a bug before. At present, it seems that the APB3 problem has been fixed. 

> Btw I think this implementation (outside the scope of this PR) does not match AHB protocol. Should I PR it?

But I haven't finished AHB yet. If you are willing to help me finish it, I will be very grateful

numero-744:
Please tell me where I can PR, i will be happy to help. To dev or you have a branch with your work in progress somewhere?

Le 21 octobre 2022 16:58:48 UTC, jijingg ***@***.***> a écrit :
***@***.*** commented on this pull request.

jijingg:
hi @numero-744  @Dolu1990 . 

Whether it is the APB or AHB Error-response signal, it is enough to only implement the read error signal at present. For the write out of bounds error, considering that it is necessary to judge whether all registers hit, this will be a resource overhead, and the error information is optional, so I do not plan to support the Write error.  

What's your opinion

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
```diff
@@ -18,21 +18,20 @@ case class AhbLite3BusInterface(bus: AhbLite3, sizeMap: SizeMapping, regPre: Str
...
+  bus.HRESP     := readError
   bus.HRDATA    := readData
 
   def readAddress():UInt   = bus.HADDR
   def writeAddress(): UInt = addressDelay
 
-  def readHalt(): Unit  = bus.HREADY === False
-  def writeHalt(): Unit = bus.HREADY === False
+  def readHalt(): Unit  = bus.HREADYOUT === False
+  def writeHalt(): Unit = bus.HREADYOUT === False
```
numero-744, line 34: Expression at statement position. Do you mean `HREADYOUT := False`?

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
```diff
@@ -18,21 +18,20 @@ case class AhbLite3BusInterface(bus: AhbLite3, sizeMap: SizeMapping, regPre: Str
...
-  val doRead      = bus.HREADY & askRead
+  val doWrite     = bus.HREADYOUT & RegNext(askWrite, False)
+  val doRead      = bus.HREADYOUT & askRead
 
   val addressDelay = RegNextWhen(bus.HADDR, askRead | askWrite)
-  readError.clearWhen(askWrite)
 
   bus.HREADYOUT := True
-  bus.HRESP     := False
+  bus.HRESP     := readError
```
numero-744, line 27: Depending on how `readError` is driven, this implementation might not match the protocol.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
```diff
@@ -18,21 +18,20 @@ case class AhbLite3BusInterface(bus: AhbLite3, sizeMap: SizeMapping, regPre: Str
 
   val askWrite    = bus.HSEL & bus.HTRANS(1) === True &  bus.HWRITE
   val askRead     = bus.HSEL & bus.HTRANS(1) === True & !bus.HWRITE
-  val doWrite     = bus.HREADY & RegNext(askWrite, False)
-  val doRead      = bus.HREADY & askRead
+  val doWrite     = bus.HREADYOUT & RegNext(askWrite, False)
+  val doRead      = bus.HREADYOUT & askRead
```
numero-744, line 22: `HREADY` should be used instead of `HREADYOUT` as the transfer can be delayed ("stuck in address phase") by wait states from the previous transfer (during its data phase).

numero-744 (commented):
About AHB, see also #920

(Btw I found code duplication while working on it, with duplicated implementation errors)

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
jijingg: at the view of Master  we should see the slave' READYOUT to find is it ready to write or read , right?
earlier we use HREADY which from master it' not reasonable ,  it should be a bug, right?

kleinai:
Can we get some of this moved over to the `BusSlaveFactory` class? I'd like to consolidate the bus interface/slave components so we don't have two separate implementations. See #904

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
numero-744: HREADY comes from the multiplexor (section 2.5).

HREADY is driven by the HREADYOUT of the currently selected slave. HREADY is sent to the master and all the slaves (section 4.2).

So master actually reads HREADY, not HREADYOUT, it is just that during a transaction with a slave, HREADY is connected to HREADYOUT, so the behavior is the same unless if two consecutive transactions are not on the same slave.

Master does not have an HREADY-like output (section 2.2) but it can use HTRANS to insert wait states in the middle of a burst with the BUSY value (b01) (section 3.2).

jijingg: > HREADY is driven by the HREADYOUT of the currently selected slave. HREADY is sent to the master and all the slaves (section 4.2).

you right,  The AHB protocol is special, which is a little different from the common handshake bus.

jijingg:
> Can we get some of this moved over to the `BusSlaveFactory` class? I'd like to consolidate the bus interface/slave components so we don't have two separate implementations. See #904

The collection of registers by BusSlaveFactory is completely different from that of BusIF. If they are merged, it may be a difficult task. I tried but gave up eventually, which will make things more complicated.
Of course, regIF does have room for improvement. In fact, I want to completely separate the generation of HTLM, Json, and so that BusSlaveFactory can use it or customize it.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
numero-744: You removed the line for `HREADYOUT`, I guess because of the `otherwise` line 38, but kept the one for HRESP. I don't understand.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
```diff
@@ -16,22 +16,35 @@ case class AhbLite3BusInterface(bus: AhbLite3, sizeMap: SizeMapping, regPre: Str
...
+  val askWrite    = bus.HSEL & bus.HTRANS(1) &  bus.HWRITE
+  val askRead     = bus.HSEL & bus.HTRANS(1) & !bus.HWRITE
+  val doWrite     = bus.HREADY & RegNext(askWrite, False)
+  val doRead      = bus.HREADY & askRead
 
   val addressDelay = RegNextWhen(bus.HADDR, askRead | askWrite)
 
-  bus.HREADYOUT := True
   bus.HRESP     := readError
   bus.HRDATA    := readData
```
numero-744, line 27: Why not just `val readData = bus.HRDATA`?

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
```diff
@@ -16,22 +16,35 @@ case class AhbLite3BusInterface(bus: AhbLite3, sizeMap: SizeMapping, regPre: Str
...
+    bus.HRESP     := False
+  }
+
   def readAddress():UInt   = bus.HADDR
   def writeAddress(): UInt = addressDelay
 
-  def readHalt(): Unit  = bus.HREADYOUT === False
-  def writeHalt(): Unit = bus.HREADYOUT === False
+  def readHalt(): Unit  = bus.HREADY === False
+  def writeHalt(): Unit = bus.HREADY === False
```
numero-744, line 47: This is wrong because here it should be HREADYOUT not HREADY, and the action is to drive it. Here you just build a wire which is `!bus.HREADY` and forget it.

@jijingg Have you checked #920?

build.sbt
```diff
@@ -7,7 +7,7 @@ val defaultSettings = Defaults.coreDefaultSettings ++ xerial.sbt.Sonatype.sonaty
   organization := "com.github.spinalhdl",
   version      := SpinalVersion.all,
   crossScalaVersions := SpinalVersion.compilers,
-  scalaVersion := SpinalVersion.compilers(0),
+  scalaVersion := SpinalVersion.compilers(1),
```
numero-744, line 10: This will be done in https://github.com/SpinalHDL/SpinalHDL/pull/918/commits/4c2aa37fa1a767eec8e7ab0771c1407f95887279

numero-744:
This branch is still active so I set the PR as draft

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
jijingg: In fact, these two functions are not used in Busif

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
jijingg: ?? readData is data read from Register need put on bus.HRDATA ?   what is `val readData = bus.HRDATA` ?

numero-744: `val a = b` is just aliasing `a` as another name for `b`. It does not create a new wire.

So you can `a := x` and `y := a` and it is the same as `b := x` and `y := b`.

jijingg:
@numero-744 I think the implementation of AHB/APB now is OK, ready for merge，   Do you have any questions

numero-744:
For AHB part only (for the other parts I'm not competent ^^'):

* Error management seems good **iff** you guarantee that `readError` is True 0 or 1 cycle, not more (I guess it is what `readError.clearWhen(readError)` means).
* You reverted the changes for the functions which do nothing but is not written as `= {}` so there is no issue in the changes of the PR XD

AHB changes look good to me.

PR #916, opened 2022-10-20
Add some utilities that can help analyze the circuit structure

wswslzp:
Hi folks, 

This PR provides two new tools that may help you analyze your RTL. 

**DataAnalyzer** enables you to find all the fanin signals and/or fanout signals of one specific signal.
**ModuleAanlyzer** enables you to get all the input/output ports, get the wanted clocks/registers/nets/cells/cell pins according to the condition you provide. 

With these tools, designers don't have to understand the internal database of SpinalHDL to grab the design structure. For someone who wants to report the quality of design, automate the EDA tools flow or do some fancy meta-HDL things, they can make things easier.

### DIFF
--- a/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
+++ b/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
@@ -1,0 +1,40 @@
+package spinal.lib.tools
+import spinal.core._
+import spinal.core.internals._
+
+import scala.language.implicitConversions
+
+class DataAnalyzer(data: BaseType) {
+  def allFanIn: Set[BaseType] = {
+    val ret = Set.newBuilder[BaseType]
+    data.foreachStatements {st=>
+      if (!(st.isInstanceOf[InitAssignmentStatement] || st.isInstanceOf[InitialAssignmentStatement])) {
+        st.foreachDrivingExpression{
+          case bt: BaseType=> ret += bt
+          case _=>
+        }
+        st.walkParentTreeStatementsUntilRootScope{tree=> tree.walkDrivingExpressions{
+          case bt: BaseType=> ret += bt
+          case _=>
+        }}
+      }
+    }
+    ret.result()
+  }
+  def getFanIn(cond: BaseType=> Boolean) = allFanIn.filter(cond)
+  def walkFanIn(cond: BaseType=> Boolean)(func: BaseType=> Unit): Unit = getFanIn(cond).foreach(func)
+
+  def allFanOut: Set[BaseType] = {
+    import ModuleAnalyzer._
+    data.globalData.toplevel.getNets{bt=>
+      val e = new DataAnalyzer(bt)
+      e.allFanIn.contains(data)
+    }
+  }
+  def getFanOut(cond: BaseType=> Boolean): Set[BaseType] = allFanOut.filter(cond)
+  def walkFanOut(cond: BaseType=> Boolean)(func: BaseType=> Unit): Unit = getFanOut(cond).foreach(func)
+}
+
+object DataAnalyzer {
+  implicit def toAnalyzer(data: BaseType): DataAnalyzer = new DataAnalyzer(data)
+}
--- a/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
+++ b/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
@@ -1,10 +1,21 @@
 package spinal.lib.tools
+
 import spinal.core._
 import spinal.core.internals._
 
 import scala.language.implicitConversions
 
+/**
+ * Base type data analyzer. It provides some utilities that help designer analyze the
+ * circuit at the runtime.
+ * @param data - the data to be analyzed. The analyze unit is BaseType signal like Bits/UInt/SInt/Bool
+ */
 class DataAnalyzer(data: BaseType) {
+
+  /**
+   * Return all the fan-ins signals of this data
+   * @return - a set of fan-ins signals.
+   */
   def allFanIn: Set[BaseType] = {
     val ret = Set.newBuilder[BaseType]
     data.foreachStatements {st=>
@@ -21,9 +32,25 @@ class DataAnalyzer(data: BaseType) {
     }
     ret.result()
   }
-  def getFanIn(cond: BaseType=> Boolean) = allFanIn.filter(cond)
+
+  /**
+   * Filter the fan-ins according to the condition.
+   * @param cond - a predicate to filter the fan-in
+   * @return a set of fan-ins signals
+   */
+  def getFanIn(cond: BaseType=> Boolean): Set[BaseType] = allFanIn.filter(cond)
+
+  /**
+   * Iterate on the filtered fan-ins.
+   * @param cond -  a predicate to filter the fan-in
+   * @param func - the function applied on the data
+   */
   def walkFanIn(cond: BaseType=> Boolean)(func: BaseType=> Unit): Unit = getFanIn(cond).foreach(func)
 
+  /**
+   * Return all the fan-outs signals of this data
+   * @return - a set of fan-outs signals.
+   */
   def allFanOut: Set[BaseType] = {
     import ModuleAnalyzer._
     data.globalData.toplevel.getNets{bt=>
@@ -31,7 +58,19 @@ class DataAnalyzer(data: BaseType) {
       e.allFanIn.contains(data)
     }
   }
+
+  /**
+   * Filter the fan-outs according to the condition.
+   * @param cond - a predicate to filter the fan-out
+   * @return a set of fan-outs signals
+   */
   def getFanOut(cond: BaseType=> Boolean): Set[BaseType] = allFanOut.filter(cond)
+
+  /**
+   * Iterate on the filtered fan-outs.
+   * @param cond -  a predicate to filter the fan-outs
+   * @param func - the function applied on the data
+   */
   def walkFanOut(cond: BaseType=> Boolean)(func: BaseType=> Unit): Unit = getFanOut(cond).foreach(func)
 }
 
--- a/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
+++ b/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
@@ -1,6 +1,7 @@
 package spinal.lib.tools
 
 import spinal.core._
+
 import scala.language._
 
 /**
@@ -10,12 +11,36 @@ import scala.language._
  * @param module the module being analyzed
  */
 class ModuleAnalyzer(module: Module) {
+  /**
+    * Return all input port of the module
+    * @return set of base type
+    */
   def allInputs: Set[BaseType] = module.getAllIo.filter(_.isInputOrInOut).toSet
+
+  /**
+    * Filter the input ports
+    * @param filter - the predicate to filter the input ports
+    * @return set of filtered input ports
+    */
   def getInputs(filter: BaseType => Boolean): Set[BaseType] = allInputs.filter(filter)
 
+  /**
+    * Get all output ports
+    * @return set of output base type
+    */
   def allOutputs: Set[BaseType] = module.getAllIo.filter(_.isOutputOrInOut).toSet
+
+  /**
+    * Filter the output ports
+    * @param filter - the predicate to filter the output ports
+    * @return set of filtered output ports
+    */
   def getOutputs(filter: BaseType=> Boolean): Set[BaseType] = allOutputs.filter(filter)
 
+  /**
+    * Get all the clock domains inside the module
+    * @return set of the clock domains
+    */
   def allClocks: Set[ClockDomain] = {
     val ret = Set.newBuilder[ClockDomain]
     module.walkComponents {comp=>
@@ -27,8 +52,18 @@ class ModuleAnalyzer(module: Module) {
     }
     ret.result()
   }
+
+  /**
+    * Filter the clock domains
+    * @param filter - the predicate to filter the clock domains
+    * @return set of clock domain
+    */
   def getClocks(filter: ClockDomain=> Boolean): Set[ClockDomain] = allClocks.filter(filter)
 
+  /**
+    * Get all the registers inside the module
+    * @return set of register of base type
+    */
   def allRegisters: Set[BaseType] = {
     val ret = Set.newBuilder[BaseType]
     module.dslBody.walkDeclarations {
@@ -38,8 +73,19 @@ class ModuleAnalyzer(module: Module) {
     }
     ret.result()
   }
+
+  /**
+    * Filter the registers
+    * @param filter  - the predicate to filter the clock domains
+    * @return set of filtered registers
+    */
   def getRegisters(filter: BaseType=> Boolean): Set[BaseType] = allRegisters.filter(filter)
 
+  /**
+    * Get the submodule instances that meet the condition
+    * @param cond - the condition that instance should meet.
+    * @return - set of sub module instances
+    */
   def getCells(cond: Module=> Boolean): Set[Module] = {
     val ret = Set.newBuilder[Module]
     module.walkComponents{m=>
@@ -47,6 +93,11 @@ class ModuleAnalyzer(module: Module) {
     }
     ret.result()
   }
+  /**
+    * Get the sub-blackbox instances that meet the condition
+    * @param cond - the condition that instance should meet.
+    * @return - set of sub blackbox instances
+    */
   def getLibCells(cond: BlackBox=> Boolean): Set[BlackBox] = {
     val ret = Set.newBuilder[BlackBox]
     module.walkComponents{
@@ -56,6 +107,12 @@ class ModuleAnalyzer(module: Module) {
     }
     ret.result()
   }
+
+  /**
+    * Get the wire/net inside the module
+    * @param cond the filtering condition
+    * @return set of the base type nets
+    */
   def getNets(cond: BaseType=> Boolean): Set[BaseType] = {
     val ret = Set.newBuilder[BaseType]
     module.walkComponents{m=>
@@ -67,6 +124,12 @@ class ModuleAnalyzer(module: Module) {
     }
     ret.result()
   }
+
+  /**
+    * Get pins of the sub-module instance inside the module.
+    * @param cond the filtering condition
+    * @return set of the pin type
+    */
   def getPins(cond: BaseType=> Boolean): Set[BaseType] = {
     val ret = Set.newBuilder[BaseType]
     module.walkComponents{m=>
@@ -78,6 +141,11 @@ class ModuleAnalyzer(module: Module) {
     }
     ret.result()
   }
+  /**
+    * Get pins of the sub-blackbox instance inside the module.
+    * @param cond the filtering condition
+    * @return set of the pin type
+    */
   def getLibPins(cond: BaseType=> Boolean): Set[BaseType] = {
     val ret = Set.newBuilder[BaseType]
     val cellTrue = (_: BlackBox) => true
@@ -87,6 +155,12 @@ class ModuleAnalyzer(module: Module) {
     }
     ret.result()
   }
+
+  /**
+    * Get the toplevel module's ports
+    * @param cond the filtering condition
+    * @return set of the ports
+    */
   def getPort(cond: BaseType=> Boolean): Set[BaseType] = {
     val e = new ModuleAnalyzer(module.globalData.toplevel)
     (e.allInputs ++ e.allOutputs).filter(cond)
--- a/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
+++ b/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
@@ -1,0 +1,101 @@
+package spinal.lib.tools
+
+import spinal.core._
+import scala.language._
+
+/**
+ * Module topology analyzer. It provides some methods that return the input or output pins,
+ * all sub-modules or sub-blackboxes, all clocks inside, and apply condition to the returning
+ * results.
+ * @param module the module being analyzed
+ */
+class ModuleAnalyzer(module: Module) {
+  def allInputs: Set[BaseType] = module.getAllIo.filter(_.isInputOrInOut).toSet
+  def getInputs(filter: BaseType => Boolean): Set[BaseType] = allInputs.filter(filter)
+
+  def allOutputs: Set[BaseType] = module.getAllIo.filter(_.isOutputOrInOut).toSet
+  def getOutputs(filter: BaseType=> Boolean): Set[BaseType] = allOutputs.filter(filter)
+
+  def allClocks: Set[ClockDomain] = {
+    val ret = Set.newBuilder[ClockDomain]
+    module.walkComponents {comp=>
+      comp.dslBody.walkDeclarations {ds=>
+        ds.foreachClockDomain{cd=>
+          ret += cd
+        }
+      }
+    }
+    ret.result()
+  }
+  def getClocks(filter: ClockDomain=> Boolean): Set[ClockDomain] = allClocks.filter(filter)
+
+  def allRegisters: Set[BaseType] = {
+    val ret = Set.newBuilder[BaseType]
+    module.dslBody.walkDeclarations {
+      case ds: BaseType if ds.isReg=>
+        ret += ds
+      case _=>
+    }
+    ret.result()
+  }
+  def getRegisters(filter: BaseType=> Boolean): Set[BaseType] = allRegisters.filter(filter)
+
+  def getCells(cond: Module=> Boolean): Set[Module] = {
+    val ret = Set.newBuilder[Module]
+    module.walkComponents{m=>
+      if (cond(m)) ret += m
+    }
+    ret.result()
+  }
+  def getLibCells(cond: BlackBox=> Boolean): Set[BlackBox] = {
+    val ret = Set.newBuilder[BlackBox]
+    module.walkComponents{
+      case bb: BlackBox if cond(bb) =>
+        ret += bb
+      case _=>
+    }
+    ret.result()
+  }
+  def getNets(cond: BaseType=> Boolean): Set[BaseType] = {
+    val ret = Set.newBuilder[BaseType]
+    module.walkComponents{m=>
+      m.dslBody.walkDeclarations {
+        case ds: BaseType if cond(ds)=>
+          ret += ds
+        case _=>
+      }
+    }
+    ret.result()
+  }
+  def getPins(cond: BaseType=> Boolean): Set[BaseType] = {
+    val ret = Set.newBuilder[BaseType]
+    module.walkComponents{m=>
+      m.dslBody.walkDeclarations {
+        case ds: BaseType if (ds.isInputOrInOut || ds.isOutput) && cond(ds) =>
+          ret += ds
+        case _=>
+      }
+    }
+    ret.result()
+  }
+  def getLibPins(cond: BaseType=> Boolean): Set[BaseType] = {
+    val ret = Set.newBuilder[BaseType]
+    val cellTrue = (_: BlackBox) => true
+    getLibCells(cellTrue).foreach{bb=>
+      val e = new ModuleAnalyzer(bb)
+      ret ++= e.getPins(cond)
+    }
+    ret.result()
+  }
+  def getPort(cond: BaseType=> Boolean): Set[BaseType] = {
+    val e = new ModuleAnalyzer(module.globalData.toplevel)
+    (e.allInputs ++ e.allOutputs).filter(cond)
+  }
+
+}
+
+object ModuleAnalyzer {
+  implicit def toElaborator(module: Module): ModuleAnalyzer = new ModuleAnalyzer(module)
+  val cellTrue = (_: Component) => true
+  val dataTrue = (_: Data) => true
+}
--- a/sim/src/main/scala/spinal/sim/VCSBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VCSBackend.scala
@@ -237,10 +237,10 @@ class VCSBackend(config: VCSBackendConfig) extends VpiBackend(config) {
     val verilogSourcePaths = rtlSourcesPaths.filter { s =>
       s.endsWith(".v") || s.endsWith(".sv") || s.endsWith(".vl") || s.endsWith(".vh") || s.endsWith(".svh") || s
         .endsWith(".vr")
-    } // .mkString(" ")
+    }
     val vhdlSourcePaths = rtlSourcesPaths.filter { s =>
       s.endsWith(".vhd") || s.endsWith(".vhdl")
-    } // .mkString(" ")
+    }
 
     config.rtlSourcesPaths
       .filter { s =>


numero-744:
My dream is an interactive thing for that

Dolu1990:
Nice :D

One thing to be carefull about, is to avoid using regular Set / Map, but instead always using the LinkedHashSet / LinkedHashMap to ensure that everything is deterministic.

Ex :  instead of toSet, there is distinctLinked which is added in the spinal.lib. Maybe we could add toLinkedSet.

Readon:
Good work.

wswslzp:
> Nice :D
> 
> One thing to be carefull about, is to avoid using regular Set / Map, but instead always using the LinkedHashSet / LinkedHashMap to ensure that everything is deterministic.
> 
> Ex : instead of toSet, there is distinctLinked which is added in the spinal.lib. Maybe we could add toLinkedSet.

Could you explain more about the difference between regular Set/Map with LinkedHashSet/Map? The former two are immutable while the latter two are mutable. @Dolu1990

numero-744:
If I understand correctly it is that when you iterate on them the other for Linked is deterministic (["The iterator and all traversal methods of this class visit elements in the order they were inserted."](https://www.scala-lang.org/api/current/scala/collection/mutable/LinkedHashSet.html), else [it is not specified](https://www.scala-lang.org/api/current/scala/collection/Set.html) so can be considered pseudo-random.

Also, notice that Set can be [mutable](https://www.scala-lang.org/api/current/scala/collection/mutable/Set.html) or [immutable](https://www.scala-lang.org/api/current/scala/collection/immutable/Set.html).

wswslzp:
So the deterministic here means a definite order of traversing the set. But so far I don't see the necessity of having a determined order of circuit elements in a set. For RTL designers, we may not have such restrictions as the internal database of Spinal.

Readon:
> So the deterministic here means a definite order of traversing the set. But so far I don't see the necessity of having a determined order of circuit elements in a set. For RTL designers, we may not have such restrictions as the internal database of Spinal.

IMO, it's important to keep the consistency of generated docs to make git friendly.
For example, If we generate an architecture diagram based on this function, then keep the generated doc elements in order is necessary.

wswslzp:
I got this. But I found that in `component.scala`, method `getAllIo` of `Component` is 

https://github.com/SpinalHDL/SpinalHDL/blob/b5b5df1df80a0c90b8a42f66b435f9cde7ec6abf/core/src/main/scala/spinal/core/Component.scala#L310

I will change the return type from mutable.Set to mutable.LinkedHashSet, since some of my code relies on this method.

wswslzp:
If there are no further suggestions, this PR can be merged now.

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,176 @@
...
+    * @return set of output base type
+    */
+  def allOutputs: mutable.LinkedHashSet[BaseType] = module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isOutputOrInOut)
+
+  /**
+    * Filter the output ports
+    * @param filter - the predicate to filter the output ports
+    * @return set of filtered output ports
+    */
+  def getOutputs(filter: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = allOutputs.filter(filter)
```
numero-744, line 39:
```suggestion
  def getOutputs(filter: BaseType => Boolean = _ => true): mutable.LinkedHashSet[BaseType] = allOutputs.filter(filter)
```

numero-744:
Can you run `scalafmt` on the two new files, please?

<details><summary>FYI I get this</summary>

```diff
diff --git a/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala b/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
index 47f536732..464b710fc 100644
--- a/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
+++ b/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
@@ -6,73 +6,68 @@ import spinal.core.internals._
 import scala.language.implicitConversions
 import scala.collection.mutable
 
-/**
- * Base type data analyzer. It provides some utilities that help designer analyze the
- * circuit at the runtime.
- * @param data - the data to be analyzed. The analyze unit is BaseType signal like Bits/UInt/SInt/Bool
- */
+/** Base type data analyzer. It provides some utilities that help designer analyze the
+  * circuit at the runtime.
+  * @param data - the data to be analyzed. The analyze unit is BaseType signal like Bits/UInt/SInt/Bool
+  */
 class DataAnalyzer(data: BaseType) {
 
-  /**
-   * Return all the fan-ins signals of this data
-   * @return - a set of fan-ins signals.
-   */
+  /** Return all the fan-ins signals of this data
+    * @return - a set of fan-ins signals.
+    */
   def allFanIn: mutable.LinkedHashSet[BaseType] = {
     val ret = mutable.LinkedHashSet.newBuilder[BaseType]
-    data.foreachStatements {st=>
+    data.foreachStatements { st =>
       if (!(st.isInstanceOf[InitAssignmentStatement] || st.isInstanceOf[InitialAssignmentStatement])) {
-        st.foreachDrivingExpression{
-          case bt: BaseType=> ret += bt
-          case _=>
+        st.foreachDrivingExpression {
+          case bt: BaseType => ret += bt
+          case _            =>
+        }
+        st.walkParentTreeStatementsUntilRootScope { tree =>
+          tree.walkDrivingExpressions {
+            case bt: BaseType => ret += bt
+            case _            =>
+          }
         }
-        st.walkParentTreeStatementsUntilRootScope{tree=> tree.walkDrivingExpressions{
-          case bt: BaseType=> ret += bt
-          case _=>
-        }}
       }
     }
     ret.result()
   }
 
-  /**
-   * Filter the fan-ins according to the condition.
-   * @param cond - a predicate to filter the fan-in
-   * @return a set of fan-ins signals
-   */
-  def getFanIn(cond: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = allFanIn.filter(cond)
+  /** Filter the fan-ins according to the condition.
+    * @param cond - a predicate to filter the fan-in
+    * @return a set of fan-ins signals
+    */
+  def getFanIn(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allFanIn.filter(cond)
 
-  /**
-   * Iterate on the filtered fan-ins.
-   * @param cond -  a predicate to filter the fan-in
-   * @param func - the function applied on the data
-   */
-  def walkFanIn(cond: BaseType=> Boolean)(func: BaseType=> Unit): Unit = getFanIn(cond).foreach(func)
+  /** Iterate on the filtered fan-ins.
+    * @param cond -  a predicate to filter the fan-in
+    * @param func - the function applied on the data
+    */
+  def walkFanIn(cond: BaseType => Boolean)(func: BaseType => Unit): Unit = getFanIn(cond).foreach(func)
 
-  /**
-   * Return all the fan-outs signals of this data
-   * @return - a set of fan-outs signals.
-   */
+  /** Return all the fan-outs signals of this data
+    * @return - a set of fan-outs signals.
+    */
   def allFanOut: mutable.LinkedHashSet[BaseType] = {
     import ModuleAnalyzer._
-    data.globalData.toplevel.getNets{bt=>
+    data.globalData.toplevel.getNets { bt =>
       val e = new DataAnalyzer(bt)
       e.allFanIn.contains(data)
     }
   }
 
-  /**
-   * Filter the fan-outs according to the condition.
-   * @param cond - a predicate to filter the fan-out
-   * @return a set of fan-outs signals
-   */
-  def getFanOut(cond: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = allFanOut.filter(cond)
+  /** Filter the fan-outs according to the condition.
+    * @param cond - a predicate to filter the fan-out
+    * @return a set of fan-outs signals
+    */
+  def getFanOut(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allFanOut.filter(cond)
 
-  /**
-   * Iterate on the filtered fan-outs.
-   * @param cond -  a predicate to filter the fan-outs
-   * @param func - the function applied on the data
-   */
-  def walkFanOut(cond: BaseType=> Boolean)(func: BaseType=> Unit): Unit = getFanOut(cond).foreach(func)
+  /** Iterate on the filtered fan-outs.
+    * @param cond -  a predicate to filter the fan-outs
+    * @param func - the function applied on the data
+    */
+  def walkFanOut(cond: BaseType => Boolean)(func: BaseType => Unit): Unit = getFanOut(cond).foreach(func)
 }
 
 object DataAnalyzer {
diff --git a/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala b/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
index 36f52e854..bca83cb36 100644
--- a/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
+++ b/lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
@@ -5,48 +5,45 @@ import spinal.core._
 import scala.language._
 import scala.collection.mutable
 
-/**
- * Module topology analyzer. It provides some methods that return the input or output pins,
- * all sub-modules or sub-blackboxes, all clocks inside, and apply condition to the returning
- * results.
- * @param module the module being analyzed
- */
+/** Module topology analyzer. It provides some methods that return the input or output pins,
+  * all sub-modules or sub-blackboxes, all clocks inside, and apply condition to the returning
+  * results.
+  * @param module the module being analyzed
+  */
 class ModuleAnalyzer(module: Module) {
-  /**
-    * Return all input port of the module
+
+  /** Return all input port of the module
     * @return set of base type
     */
-  def allInputs: mutable.LinkedHashSet[BaseType] = module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isInputOrInOut)
+  def allInputs: mutable.LinkedHashSet[BaseType] =
+    module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isInputOrInOut)
 
-  /**
-    * Filter the input ports
+  /** Filter the input ports
     * @param filter - the predicate to filter the input ports
     * @return set of filtered input ports
     */
   def getInputs(filter: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allInputs.filter(filter)
 
-  /**
-    * Get all output ports
+  /** Get all output ports
     * @return set of output base type
     */
-  def allOutputs: mutable.LinkedHashSet[BaseType] = module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isOutputOrInOut)
+  def allOutputs: mutable.LinkedHashSet[BaseType] =
+    module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isOutputOrInOut)
 
-  /**
-    * Filter the output ports
+  /** Filter the output ports
     * @param filter - the predicate to filter the output ports
     * @return set of filtered output ports
     */
-  def getOutputs(filter: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = allOutputs.filter(filter)
+  def getOutputs(filter: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allOutputs.filter(filter)
 
-  /**
-    * Get all the clock domains inside the module
+  /** Get all the clock domains inside the module
     * @return set of the clock domains
     */
   def allClocks: mutable.LinkedHashSet[ClockDomain] = {
     val ret = mutable.LinkedHashSet.newBuilder[ClockDomain]
-    module.walkComponents {comp=>
-      comp.dslBody.walkDeclarations {ds=>
-        ds.foreachClockDomain{cd=>
+    module.walkComponents { comp =>
+      comp.dslBody.walkDeclarations { ds =>
+        ds.foreachClockDomain { cd =>
           ret += cd
         }
       }
@@ -54,115 +51,108 @@ class ModuleAnalyzer(module: Module) {
     ret.result()
   }
 
-  /**
-    * Filter the clock domains
+  /** Filter the clock domains
     * @param filter - the predicate to filter the clock domains
     * @return set of clock domain
     */
-  def getClocks(filter: ClockDomain=> Boolean): mutable.LinkedHashSet[ClockDomain] = allClocks.filter(filter)
+  def getClocks(filter: ClockDomain => Boolean): mutable.LinkedHashSet[ClockDomain] = allClocks.filter(filter)
 
-  /**
-    * Get all the registers inside the module
+  /** Get all the registers inside the module
     * @return set of register of base type
     */
   def allRegisters: mutable.LinkedHashSet[BaseType] = {
     val ret = mutable.LinkedHashSet.newBuilder[BaseType]
     module.dslBody.walkDeclarations {
-      case ds: BaseType if ds.isReg=>
+      case ds: BaseType if ds.isReg =>
         ret += ds
-      case _=>
+      case _ =>
     }
     ret.result()
   }
 
-  /**
-    * Filter the registers
+  /** Filter the registers
     * @param filter  - the predicate to filter the clock domains
     * @return set of filtered registers
     */
-  def getRegisters(filter: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = allRegisters.filter(filter)
+  def getRegisters(filter: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allRegisters.filter(filter)
 
-  /**
-    * Get the submodule instances that meet the condition
+  /** Get the submodule instances that meet the condition
     * @param cond - the condition that instance should meet.
     * @return - set of sub module instances
     */
-  def getCells(cond: Module=> Boolean): mutable.LinkedHashSet[Module] = {
+  def getCells(cond: Module => Boolean): mutable.LinkedHashSet[Module] = {
     val ret = mutable.LinkedHashSet.newBuilder[Module]
-    module.walkComponents{m=>
+    module.walkComponents { m =>
       if (cond(m)) ret += m
     }
     ret.result()
   }
-  /**
-    * Get the sub-blackbox instances that meet the condition
+
+  /** Get the sub-blackbox instances that meet the condition
     * @param cond - the condition that instance should meet.
     * @return - set of sub blackbox instances
     */
-  def getLibCells(cond: BlackBox=> Boolean): mutable.LinkedHashSet[BlackBox] = {
+  def getLibCells(cond: BlackBox => Boolean): mutable.LinkedHashSet[BlackBox] = {
     val ret = mutable.LinkedHashSet.newBuilder[BlackBox]
-    module.walkComponents{
+    module.walkComponents {
       case bb: BlackBox if cond(bb) =>
         ret += bb
-      case _=>
+      case _ =>
     }
     ret.result()
   }
 
-  /**
-    * Get the wire/net inside the module
+  /** Get the wire/net inside the module
     * @param cond the filtering condition
     * @return set of the base type nets
     */
-  def getNets(cond: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = {
+  def getNets(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
     val ret = mutable.LinkedHashSet.newBuilder[BaseType]
-    module.walkComponents{m=>
+    module.walkComponents { m =>
       m.dslBody.walkDeclarations {
-        case ds: BaseType if cond(ds)=>
+        case ds: BaseType if cond(ds) =>
           ret += ds
-        case _=>
+        case _ =>
       }
     }
     ret.result()
   }
 
-  /**
-    * Get pins of the sub-module instance inside the module.
+  /** Get pins of the sub-module instance inside the module.
     * @param cond the filtering condition
     * @return set of the pin type
     */
-  def getPins(cond: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = {
+  def getPins(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
     val ret = mutable.LinkedHashSet.newBuilder[BaseType]
-    module.walkComponents{m=>
+    module.walkComponents { m =>
       m.dslBody.walkDeclarations {
         case ds: BaseType if (ds.isInputOrInOut || ds.isOutput) && cond(ds) =>
           ret += ds
-        case _=>
+        case _ =>
       }
     }
     ret.result()
   }
-  /**
-    * Get pins of the sub-blackbox instance inside the module.
+
+  /** Get pins of the sub-blackbox instance inside the module.
     * @param cond the filtering condition
     * @return set of the pin type
     */
-  def getLibPins(cond: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = {
+  def getLibPins(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
     val ret = mutable.LinkedHashSet.newBuilder[BaseType]
     val cellTrue = (_: BlackBox) => true
-    getLibCells(cellTrue).foreach{bb=>
+    getLibCells(cellTrue).foreach { bb =>
       val e = new ModuleAnalyzer(bb)
       ret ++= e.getPins(cond)
     }
     ret.result()
   }
 
-  /**
-    * Get the toplevel module's ports
+  /** Get the toplevel module's ports
     * @param cond the filtering condition
     * @return set of the ports
     */
-  def getPort(cond: BaseType=> Boolean): mutable.LinkedHashSet[BaseType] = {
+  def getPort(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
     val e = new ModuleAnalyzer(module.globalData.toplevel)
     (e.allInputs ++ e.allOutputs).filter(cond)
   }
```

</details>

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
wswslzp: Same as `allOutputs`.

Readon:
then it might not necessary to define all outputs?



---Original---
From: ***@***.***&gt;
Date: Sun, Oct 23, 2022 21:55 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] feat: add some utilities that can helpanalyze the circuit structure (PR #916)




 
@wswslzp commented on this pull request.
 
 
In lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala:
 &gt; +    * @return set of filtered input ports +    */ +  def getInputs(filter: BaseType =&gt; Boolean): mutable.LinkedHashSet[BaseType] = allInputs.filter(filter) + +  /** +    * Get all output ports +    * @return set of output base type +    */ +  def allOutputs: mutable.LinkedHashSet[BaseType] = module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isOutputOrInOut) + +  /** +    * Filter the output ports +    * @param filter - the predicate to filter the output ports +    * @return set of filtered output ports +    */ +  def getOutputs(filter: BaseType=&gt; Boolean): mutable.LinkedHashSet[BaseType] = allOutputs.filter(filter)  
Same as allOutputs.

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
numero-744: True, my issue is that `getOutputs` is documented "Filter the output ports" and not "Get the outputs" as we would expect. So if I want to get the output, I type `getOutputs`, I notice it misses a parameter.

What do you think about this renaming?

```scala
def outputs: ... = ...
def outputsWhere(cond: ... => Boolean)
```

Readon:
IMO, getOutputs is a good action+target way to name method. How about define get outputs method return all by default, and accept filter for specific usage.



---Original---
From: ***@***.***&gt;
Date: Sun, Oct 23, 2022 22:06 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] feat: add some utilities that can helpanalyze the circuit structure (PR #916)




 
@numero-744 commented on this pull request.
 
 
In lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala:
 &gt; +    * @return set of filtered input ports +    */ +  def getInputs(filter: BaseType =&gt; Boolean): mutable.LinkedHashSet[BaseType] = allInputs.filter(filter) + +  /** +    * Get all output ports +    * @return set of output base type +    */ +  def allOutputs: mutable.LinkedHashSet[BaseType] = module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isOutputOrInOut) + +  /** +    * Filter the output ports +    * @param filter - the predicate to filter the output ports +    * @return set of filtered output ports +    */ +  def getOutputs(filter: BaseType=&gt; Boolean): mutable.LinkedHashSet[BaseType] = allOutputs.filter(filter)  
True, my issue is that getOutputs is documented "Filter the output ports" and not "Get the outputs" as we would expect. So if I want to get the output, I type getOutputs, I notice it misses a parameter.
 
What do you think about this renaming?
 def outputs: ... = ... def outputsWhere(cond: ... =&gt; Boolean)

numero-744:
Ah @Readon do you mean?

```scala
def getOutputs(): mutable.LinkedHashSet[BaseType] = ... allOutputs
def getOutputs(filter: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = getOutputs.filter(filter)
```

wswslzp:
Good suggestions. So in short, renaming `all*` as `get*` methods w/o filters. 

The original design consideration is that the API can be 1-1 mapped to standard SDC commands (though not exactly mapping). 
In SDC constraints, `get_*` command always need a *pattern* to filter out the results while `all_*` command returns all of the results. For RTL designers who write sdc on their own it might be familiar to them.

numero-744:
Ah thanks for the explanation. It makes sense with this context.

Do you think it is overkill to keep a Scala API by renaming `all*`->`get*`, and add a tcl-like API with names `get_*` and `all_*`, which would just call the function from the more Scala-ish API?

I'm taking a look at the generated docs (thank you so much for having written one :pray:), can you remove the `-` in `@param someparam - some description`, please?

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+            case _            =>
+          }
+        }
+      }
+    }
+    ret.result()
+  }
+
+  /** Filter the fan-ins according to the condition.
+    * @param cond - a predicate to filter the fan-in
```
numero-744, line 39:
```suggestion
    * @param cond a predicate to filter the fan-ins
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    * @return a set of fan-ins signals
```
numero-744, line 40:
```suggestion
    * @return a set of fan-in signals
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    */
+  def getFanIn(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allFanIn.filter(cond)
+
+  /** Iterate on the filtered fan-ins.
+    * @param cond -  a predicate to filter the fan-in
```
numero-744, line 45:
```suggestion
    * @param cond a predicate to filter the fan-ins
```

Readon:
It seems that TCL do not have default value for function arguments, that's why it is not as descriptive as Scala. 😆 :laugh:

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    * @param func - the function applied on the data
```
numero-744, line 46:
```suggestion
    * @param func the function applied on the data
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    */
+  def walkFanIn(cond: BaseType => Boolean)(func: BaseType => Unit): Unit = getFanIn(cond).foreach(func)
+
+  /** Return all the fan-outs signals of this data
```
numero-744, line 50:
```suggestion
  /** Get all the fan-out signals of this data
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    * @return - a set of fan-outs signals.
```
numero-744, line 51:
```suggestion
    * @return a set of fan-out signals.
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    import ModuleAnalyzer._
+    data.globalData.toplevel.getNets { bt =>
+      val e = new DataAnalyzer(bt)
+      e.allFanIn.contains(data)
+    }
+  }
+
+  /** Filter the fan-outs according to the condition.
+    * @param cond - a predicate to filter the fan-out
+    * @return a set of fan-outs signals
```
numero-744, line 63:
```suggestion
    * @return a set of fan-out signals
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+  def allFanOut: mutable.LinkedHashSet[BaseType] = {
+    import ModuleAnalyzer._
+    data.globalData.toplevel.getNets { bt =>
+      val e = new DataAnalyzer(bt)
+      e.allFanIn.contains(data)
+    }
+  }
+
+  /** Filter the fan-outs according to the condition.
+    * @param cond - a predicate to filter the fan-out
```
numero-744, line 62:
```suggestion
    * @param cond a predicate to filter the fan-outs
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    * @return a set of fan-outs signals
+    */
+  def getFanOut(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allFanOut.filter(cond)
+
+  /** Iterate on the filtered fan-outs.
+    * @param cond -  a predicate to filter the fan-outs
```
numero-744, line 68:
```suggestion
    * @param cond a predicate to filter the fan-outs
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    * @param func - the function applied on the data
```
numero-744, line 69:
```suggestion
    * @param func the function applied on the data
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    */
+  def walkFanOut(cond: BaseType => Boolean)(func: BaseType => Unit): Unit = getFanOut(cond).foreach(func)
+}
+
+object DataAnalyzer {
```
numero-744, line 74:
```suggestion
object DataAnalyzer {
  /** Build DataAnalyzer to analyze this */
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
+package spinal.lib.tools
+
+import spinal.core._
+
+import scala.language._
+import scala.collection.mutable
+
+/** Module topology analyzer. It provides some methods that return the input or output pins,
+  * all sub-modules or sub-blackboxes, all clocks inside, and apply condition to the returning
```
numero-744, line 9:
```suggestion
  * all sub-modules or sub-blackboxes, all clocks inside, and filter the returned
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+  * results.
+  * @param module the module being analyzed
+  */
+class ModuleAnalyzer(module: Module) {
+
+  /** Return all input port of the module
```
numero-744, line 15:
```suggestion
  /** Get all input ports of the module
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @return set of base type
+    */
+  def allInputs: mutable.LinkedHashSet[BaseType] =
+    module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isInputOrInOut)
+
+  /** Filter the input ports
+    * @param filter - the predicate to filter the input ports
```
numero-744, line 22:
```suggestion
    * @param filter the predicate to filter the input ports
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+            case bt: BaseType => ret += bt
+            case _            =>
+          }
+        }
+      }
+    }
+    ret.result()
+  }
+
+  /** Filter the fan-ins according to the condition.
```
numero-744, line 38:
```suggestion
  /** Get the fan-ins matching the condition
```

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    */
+  def allFanOut: mutable.LinkedHashSet[BaseType] = {
+    import ModuleAnalyzer._
+    data.globalData.toplevel.getNets { bt =>
+      val e = new DataAnalyzer(bt)
+      e.allFanIn.contains(data)
+    }
+  }
+
+  /** Filter the fan-outs according to the condition.
```
numero-744, line 61:
```suggestion
  /** Get the fan-outs matching the condition
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+  */
+class ModuleAnalyzer(module: Module) {
+
+  /** Return all input port of the module
+    * @return set of base type
+    */
+  def allInputs: mutable.LinkedHashSet[BaseType] =
+    module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isInputOrInOut)
+
+  /** Filter the input ports
```
numero-744, line 21:
```suggestion
  /** Get the input ports matching the condition
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    */
+  def getInputs(filter: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allInputs.filter(filter)
+
+  /** Get all output ports
+    * @return set of output base type
+    */
+  def allOutputs: mutable.LinkedHashSet[BaseType] =
+    module.getAllIo.asInstanceOf[mutable.LinkedHashSet[BaseType]].filter(_.isOutputOrInOut)
+
+  /** Filter the output ports
```
numero-744, line 33:
```suggestion
  /** Get the output ports matching the condition
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @param filter - the predicate to filter the output ports
```
numero-744, line 34:
```suggestion
    * @param filter the predicate to filter the output ports
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+      comp.dslBody.walkDeclarations { ds =>
+        ds.foreachClockDomain { cd =>
+          ret += cd
+        }
+      }
+    }
+    ret.result()
+  }
+
+  /** Filter the clock domains
```
numero-744, line 54:
```suggestion
  /** Get the clock domains matching the condition
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @param filter - the predicate to filter the clock domains
```
numero-744, line 55:
```suggestion
    * @param filter the predicate to filter the clock domains
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    val ret = mutable.LinkedHashSet.newBuilder[BaseType]
+    module.dslBody.walkDeclarations {
+      case ds: BaseType if ds.isReg =>
+        ret += ds
+      case _ =>
+    }
+    ret.result()
+  }
+
+  /** Filter the registers
```
numero-744, line 73:
```suggestion
  /** Get the registers matching the condition
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @param filter  - the predicate to filter the clock domains
```
numero-744, line 74:
```suggestion
    * @param filter the predicate to filter the registers
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @return set of filtered registers
+    */
+  def getRegisters(filter: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = allRegisters.filter(filter)
+
+  /** Get the submodule instances that meet the condition
+    * @param cond - the condition that instance should meet.
```
numero-744, line 80:
```suggestion
    * @param cond the condition that instance should meet.
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @return - set of sub module instances
```
numero-744, line 81:
```suggestion
    * @return set of sub module instances
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+  def getCells(cond: Module => Boolean): mutable.LinkedHashSet[Module] = {
+    val ret = mutable.LinkedHashSet.newBuilder[Module]
+    module.walkComponents { m =>
+      if (cond(m)) ret += m
+    }
+    ret.result()
+  }
+
+  /** Get the sub-blackbox instances that meet the condition
+    * @param cond - the condition that instance should meet.
```
numero-744, line 92:
```suggestion
    * @param cond the condition that instance should meet.
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @return - set of sub blackbox instances
```
numero-744, line 93:
```suggestion
    * @return set of sub blackbox instances
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+    ret.result()
+  }
+
+  /** Get pins of the sub-blackbox instance inside the module.
+    * @param cond the filtering condition
+    * @return set of the pin type
+    */
+  def getLibPins(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
```
numero-744, line 141: I don't understand `getLibPins` vs `getPins`

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+      ret ++= e.getPins(cond)
+    }
+    ret.result()
+  }
+
+  /** Get the toplevel module's ports
+    * @param cond the filtering condition
+    * @return set of the ports
+    */
+  def getPort(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
```
numero-744, line 155:
```suggestion
  def getPorts(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
```

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    val e = new ModuleAnalyzer(module.globalData.toplevel)
+    (e.allInputs ++ e.allOutputs).filter(cond)
+  }
+
+}
+
+object ModuleAnalyzer {
+  implicit def toElaborator(module: Module): ModuleAnalyzer = new ModuleAnalyzer(module)
```
numero-744, line 163: Why not `toAnalyzer`?

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -0,0 +1,166 @@
...
+    * @return set of the ports
+    */
+  def getPort(cond: BaseType => Boolean): mutable.LinkedHashSet[BaseType] = {
+    val e = new ModuleAnalyzer(module.globalData.toplevel)
+    (e.allInputs ++ e.allOutputs).filter(cond)
+  }
+
+}
+
+object ModuleAnalyzer {
```
numero-744, line 162: Missing doc for 3 items here

numero-744:
I've submitted many comments but @wswslzp please mind that I would be so happy if all SpinalHDL code was documented as you did :heart:

Dolu1990:
@wswslzp 

> Could you explain more about the difference between regular Set/Map with LinkedHashSet/Map? The former two are immutable while the latter two are mutable. 

numero-744  got it right ^^

> So the deterministic here means a definite order of traversing the set. But so far I don't see the necessity of having a determined order of circuit elements in a set. For RTL designers, we may not have such restrictions as the internal database of Spinal.

One example :
- A RTL design may use your functions to analyse the hardware and proceduraly generate additional hardware out of it, for instance he may ask the API to return all the Mem to blackbox them. If the order to iterate on it is pseudo random, the issue is that the netlist will change from one generation to the other. Also, buggy user code may start to show up in a non-deterministic manner.

So as much as possible, always Linked Set/Map :)

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
wswslzp: `getLibPins` returns only the blackbox instances' pins, while `getPins` returns all instances' pins.

numero-744: I do not understand the relationship between "lib" and "BlackBox", maybe because I don't know `sdc` enough XD

wswslzp: "lib" here means a library containing IP blocks/macros. Those things in a "lib" are represented as `BlackBox` in SpinalHDL.

numero-744: Okay I better understand, thanks!

lib/src/main/scala/spinal/lib/tools/ModuleAnalyzer.scala
```diff
@@ -154,13 +154,16 @@ class ModuleAnalyzer(module: Module) {
...
 
 }
 
 object ModuleAnalyzer {
-  implicit def toElaborator(module: Module): ModuleAnalyzer = new ModuleAnalyzer(module)
+  /** Implicitly convert Module/Component into ModuleAnalyzer */
+  implicit def toAnalyzer(module: Module): ModuleAnalyzer = new ModuleAnalyzer(module)
+
+  /** handy always-true filter */
   val cellTrue = (_: Component) => true
```
numero-744, line 167: Do you have a use-case for these two `val`?

Since the API allows to not filtering by providing no condition at all, I don't see when they should be used.

numero-744:
Thanks for the update, LGTM. I wait for the tests to end.

wswslzp:
So if we don't propose any new suggestions, I think this PR is ready to be merged.

Readon:
It seems that this PR still need documentation. :)

numero-744:
@Readon 

There is https://spinalhdl.github.io/SpinalHDL/dev/spinal/lib/tools/DataAnalyzer.html, do you think we should add things in RTD? If so, please open an issue in RTD repository.

PR #917, opened 2022-10-20
ci: update vendor actions

numero-744:
It removes warnings in GitHub actions: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


numero-744:
@Dolu1990 for review and merge if approved

PR #920, opened 2022-10-21
fix: ahb BusIf

numero-744:


numero-744:
Was to show implementation example, not meant to be merged.

PR #921, opened 2022-10-21
fix(sim): Verilator time precision

### LINKED ISSUE #919: the high time precision breaks when old verilator is used.
While using "sbt test" to test my design below message prompt.
``` log
../verilator/VAddressExpander__spinalWrapper.cpp: In constructor ‘Wrapper_1::Wrapper_1(const char*)’:
../verilator/VAddressExpander__spinalWrapper.cpp:192:41: error: ‘timeprecision’ is not a member of ‘Verilated’
  192 |       this->time_precision = Verilated::timeprecision();
      |                                         ^~~~~~~~~~~~~
```

My verilator version is: Verilator 4.028 2020-02-06 rev v4.026-92-g890cecc1
This is the default version Ubuntu 20 included.

kleinai:
Would have used VPI to get the time precision, but I couldn't get that to compile correctly. This solution looks the version and generates code accordingly.

Fixes #919

### DIFF
--- a/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
@@ -93,7 +93,7 @@ class VerilatorBackend(val config: VerilatorBackendConfig) extends Backend {
                 WaveFormat.NONE
               }
 
-  def genWrapperCpp(): Unit = {
+  def genWrapperCpp(useTimePrecision: Boolean = true): Unit = {
     val jniPrefix = "Java_" + s"wrapper_${workspaceName}".replace("_", "_1") + "_VerilatorNative_"
     val wrapperString = s"""
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


numero-744:
LGTM if it passes checks

sim/src/main/scala/spinal/sim/VerilatorBackend.scala
```diff
@@ -644,7 +645,7 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
         FileUtils.copyDirectory(workspaceCacheDir, workspaceDir)
       }
 
-      genWrapperCpp()
+      genWrapperCpp(verilatorVersionDeci > BigDecimal("4.034"))
```
numero-744, line 648: I'm curious, how have you found this value `4.034`?

kleinai: First tagged commit after the time precision function was added to Verilator

numero-744: So isn't is `>=`?

numero-744:
@Readon do you want to test it before we merge it?

Readon:
This works but #821 changes API which still breaks the existing code.

PR #924, opened 2022-10-22
Axi4 sim agent fixes

### LINKED ISSUE #923: breaking changes in sim/Agent api
```diff
-def onWriteByte(address: BigInt, data: Byte): Unit
+def onWriteByte(address: BigInt, data: Byte, id: Int): Unit
-def onLast(id: Int): Unit
```

Reported by @Readon
Assigned to @kleinai

kleinai:
Restore original function patterns. Fixes #923 
I needed a few new agent functions to test the unburster (and some subsequently un-PR'd AxiLite4 conversion). I believe these original functions share the same transaction life cycle stage as before.


Readon:
This might be not a good solution, because the subclass still need to implement onWirteBytes with three arguments.

kleinai:
Shouldn't have to. There's a default empty function. You'd override it if you need it.

PR #935, opened 2022-10-24
feat: add port keyword

numero-744:
Status: I think code is good, please approve it so that I rework git history and merge.

### Context

I have identified 3 syntaxes to define ports. A port can be `in`, `out`, `inout`, `master` or `slave`. Let `$dir` be one of these 5 keywords.

The 3 syntaxes are:

* `$dir($ports)` for already-defined wires (useful for `IMasterSlave`)
* `val $a = $dir $Type ($type_params)` for `Type` in a restricted set of types. It uses the Scala infix operator notation so it is formatted with a space between `Type` and `(`.
* `val $a = $dir($Type($type_params))` (useful for other types but braces are annoying for a DSL)

### This PR adds the following syntax, using Scala infix operator (no braces) and compatible with all possible types

`val $a = $dir port $Type($type_params)`

For instance:

```scala
val a = in port Rgb
val b = out port Rgb(8 bits)
val c = master port AhbLite3(cfg)
```

It is still possible to use the previous syntaxes:

```scala
val d = in Bits (8 bits)
val e = out(Bits(8 bits))
```

It is syntaxically of the equivalent of `ClockDomain.on` for ports:

`val a = cd(comp)` => `val a = cd on comp`
`val a = in(type)` => `val a = in port type`

### This PR brings documentation for `IODirection` & `IMasterSlave` + seals these two traits

I guess these two traits were not meant to be overloaded so it it almost non-breaking? @Readon :grin:

### This PR brings native management of `null`

Do not use `inWithNull`, just use `in` :smiley:

### DIFF
--- a/lib/src/main/scala/spinal/lib/MasterSlave.scala
+++ b/lib/src/main/scala/spinal/lib/MasterSlave.scala
@@ -91,7 +91,6 @@ trait MSFactory {
   *
   *   - `Flow`
   *   - `Stream`
-  *   - `Event`
   *
   * The "spaceful" syntax is generic and beatiful, but more verbose.
   *
@@ -143,16 +142,6 @@ sealed trait MS {
       port(interface)
     }
   }
-
-  object event extends EventFactory {
-    override def postApply(interface: IMasterSlave): Unit = {
-      super.postApply(interface)
-      port(interface)
-    }
-  }
-
-  // TODO: why not just renaming event above as Event?
-  def Event: Event = port(spinal.lib.Event)
 }
 
 /** Declare a master port
--- a/lib/src/main/scala/spinal/lib/MasterSlave.scala
+++ b/lib/src/main/scala/spinal/lib/MasterSlave.scala
@@ -27,14 +27,14 @@ trait IMasterSlave extends Data {
 
   /** Set as master interface */
   final def setAsMaster(): Unit = {
-    _isMasterInterface = Some(true)
     asMaster()
+    _isMasterInterface = Some(true)
   }
 
   /** Set a slave interface */
   final def setAsSlave(): Unit = {
-    _isMasterInterface = Some(false)
     asSlave()
+    _isMasterInterface = Some(false)
   }
 
   /** Override it to define port directions for a master interface.
--- a/lib/src/main/scala/spinal/lib/MasterSlave.scala
+++ b/lib/src/main/scala/spinal/lib/MasterSlave.scala
@@ -3,7 +3,7 @@ package spinal.lib
 import spinal.core.{Data, HardType}
 
 /** Master/slave interface */
-trait IMasterSlave extends Data {
+trait IMasterSlave {
 
   /** Are port directions set for a Master interface? */
   final def isMasterInterface: Boolean = _isMasterInterface == Some(true)
@@ -116,7 +116,7 @@ sealed trait MS {
     *
     * See [[MS]] for other syntax.
     */
-  def port[T <: IMasterSlave](i: HardType[T]): T = port(i())
+  def port[T <: Data with IMasterSlave](i: HardType[T]): T = port(i())
 
   /** Declare a port with braces
     *
@@ -124,7 +124,7 @@ sealed trait MS {
     */
   def apply[T <: IMasterSlave](i: T): T = port(i)
 
-  def apply[T <: IMasterSlave](data: HardType[T]): T = apply(data())
+  def apply[T <: Data with IMasterSlave](data: HardType[T]): T = apply(data())
 
   /** Declare existing interfaces as ports, variadic syntax */
   def apply(is: IMasterSlave*): Unit = is.foreach(port(_))
--- a/lib/src/main/scala/spinal/lib/MasterSlave.scala
+++ b/lib/src/main/scala/spinal/lib/MasterSlave.scala
@@ -1,88 +1,182 @@
 package spinal.lib
 
-import spinal.core.Data
+import spinal.core.{Data, HardType}
 
-trait IMasterSlave {
+/** Master/slave interface */
+trait IMasterSlave extends Data {
 
-  var isMasterInterface: Boolean = false
+  /** Are port directions set for a Master interface? */
+  final def isMasterInterface: Boolean = _isMasterInterface == Some(true)
 
-  def asMaster(): Unit
-  def asSlave(): Unit = master(this).asInstanceOf[Data].flip()
-}
+  /** Are port directions set for a Master interface? */
+  final def isSlaveInterface: Boolean = _isMasterInterface == Some(false)
 
-//trait MasterSlave extends IMasterSlave{
-//  def asMaster() : this.type = {
-//    this
-//  }
-//}
+  private var _isMasterInterface: Option[Boolean] = None
 
-//trait MasterSlave extends IMasterSlave{
-//  def applyMaster() : Unit
-//  def asMaseter() : this.type = {
-//    applyMaster()
-//    this
-//  }
-//}
-
-trait MSFactory {
-  def postApply(interface : IMasterSlave) : Unit = {}
-}
-
-trait MS{
-  def apply[T <: IMasterSlave](i: T) : T
-  def apply(a: IMasterSlave,b: IMasterSlave,c: IMasterSlave*) : Unit = {
-    apply(a)
-    apply(b)
-    for(e <- c) apply(e)
+  /** Convert into master */
+  final def intoMaster(): this.type = {
+    setAsMaster()
+    this
   }
 
+  /** Convert into slave */
+  final def intoSlave(): this.type = {
+    setAsSlave()
+    this
+  }
+
+  /** Set as master interface */
+  final def setAsMaster(): Unit = {
+    _isMasterInterface = Some(true)
+    asMaster()
+  }
+
+  /** Set a slave interface */
+  final def setAsSlave(): Unit = {
+    _isMasterInterface = Some(false)
+    asSlave()
+  }
+
+  /** Override it to define port directions for a master interface.
+    *
+    * @deprecated This method must be overriden but not called. Calling this
+    * method is not correct. Call `setAsMaster()` or `intoMaster()` instead.
+    *
+    * This method is named `asXxx` but it does not return `Xxx`.
+    *
+    * This method does not update `isMasterInterface` and `isSlaveInterface`.
+    */
+  def asMaster(): Unit
+
+  /** Override it to define port directions for a master interface.
+    *
+    * If not overriden, defaults to the opposite port directions of `asMaster()`.
+    *
+    * @deprecated This method can be overriden but not called. Calling this
+    * method is not correct. Call `setAsSlave()` or `intoSlave()` instead.
+    *
+    * This method is named `asXxx` but it does not return `Xxx`.
+    *
+    * This method does not update `isMasterInterface` and `isSlaveInterface`.
+    */
+  def asSlave(): Unit = intoMaster().asInstanceOf[Data].flip()
+}
+
+/** Something which can create master/slave interfaces */
+trait MSFactory {
+
+  /** Called on IMasterSlave creation (eg: to apply master/slave-ness) */
+  def postApply(interface: IMasterSlave): Unit = {}
+}
+
+/** Declare a port as `master` or `slave`
+  *
+  * There are 4 available syntaxes, which are all equivalent:
+  *
+  * {{{
+  * val braces = master(Flow(Bool))
+  *
+  * val short = master Flow (Bool)
+  *
+  * val spaceful = master port Flow(Bool)
+  *
+  * val variadic = Flow(Bool)
+  * master(variadic)
+  * }}}
+  *
+  * The "braces" syntax is short and generic, but it uses braces.
+  *
+  * The "short" syntax is short, but it is formatted with a space between the
+  * type and its parameters, and it can be used only with:
+  *
+  *   - `Flow`
+  *   - `Stream`
+  *   - `Event`
+  *
+  * The "spaceful" syntax is generic and beatiful, but more verbose.
+  *
+  * The "variadic" syntax can be used with any number of interfaces, but can
+  * be used only if the interfaces are already declared.
+  *
+  * @see [[master]] [[slave]]
+  */
+sealed trait MS {
+
+  /** Override it to define how to apply port specification on a non-null IMasterSlave */
+  protected def applyIt[T <: IMasterSlave](i: T): T
+
+  /** Declare a port without braces, spaceful syntax
+    *
+    * See [[MS]] for other syntax.
+    */
+  def port[T <: IMasterSlave](i: T): T =
+    if (i != null) applyIt(i)
+    else i
+
+  /** Declare a port without braces, spaceful syntax
+    *
+    * See [[MS]] for other syntax.
+    */
+  def port[T <: IMasterSlave](i: HardType[T]): T = port(i())
+
+  /** Declare a port with braces
+    *
+    * See [[MS]] for other syntaxes.
+    */
+  def apply[T <: IMasterSlave](i: T): T = port(i)
+
+  def apply[T <: IMasterSlave](data: HardType[T]): T = apply(data())
+
+  /** Declare existing interfaces as ports, variadic syntax */
+  def apply(is: IMasterSlave*): Unit = is.foreach(port(_))
+
   object Flow extends FlowFactory {
-    override def postApply(interface : IMasterSlave) : Unit = {
+    override def postApply(interface: IMasterSlave): Unit = {
       super.postApply(interface)
-      MS.this.apply(interface)
+      port(interface)
     }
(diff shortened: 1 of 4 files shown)


numero-744:
@Dolu1990

numero-744:
Yes I just had to add exactly one keyword which would be the "infix operator" XD

Do you have an example of case where you really want to give to `in` the responsibility to crash if using `null`?

Dolu1990:
> Do you have an example of case where you really want to give to in the responsibility to crash if using null?

No, i don't have any ^^
It is more about not having a javascript like API, which try by all ends to not crash.

On the other end, as i don't see a real case where this would be problematic, i guess that's ok to have in(null) being ok. (aswell out slave master inout)

So, ok overall i think.

numero-744:
I agree with the js-like argument, I was thinking to it after you sent me previous message actually XD. But I think in this specific case this pattern addresses a real use-case. I mean, enough to have written a variant to handle this, so why not just manage it natively, without adding verbose things?

Can you help me to fix the TODOs in the diff, please?

Readon:
> I agree with the js-like argument, I was thinking to it after you sent me previous message actually XD. But I think in this specific case this pattern addresses a real use-case. I mean, enough to have written a variant to handle this, so why not just manage it natively, without adding verbose things?
> 
> 我同意类似 js 的论点，在你发给我之前的消息之后，我就想到了 XD。但是我认为在这个特定的案例中，这个模式解决了一个真正的用例。我的意思是，已经编写了足够的变体来处理这个问题，那么为什么不直接管理它，而不添加冗长的东西呢？
> 
> Can you help me to fix the TODOs in the diff, please?
> 
> 你能帮我修理一下差速器里的待命命令吗？



> I agree with the js-like argument, I was thinking to it after you sent me previous message actually XD. But I think in this specific case this pattern addresses a real use-case. I mean, enough to have written a variant to handle this, so why not just manage it natively, without adding verbose things?
> 
> Can you help me to fix the TODOs in the diff, please?

Emm, this will introduce a new coding paradigm.
Often, having too many ways to do the same thing isn't that great.
Different coding styles throw a question to the user: "Which one should I use?"
Users must define their own coding style internally, and this problem exists in SpinalHDL's development as well.

numero-744:
I agree that too much choice (#perl) can be confusing (eg: `Component` vs. `Module`).

But:

* I believe this change actually improves the syntax for all non-standard types (types not supported by `in Type (n bits)`) and even for standard types (formatting is more consistent, especially for `in Bool ()`) as it does not mess formatting, while being lightweight (no braces, which is a good thing for a DSL) even if it is one keyword longer.
* It is universal so it avoids some confusion: I think all Scala/SpinalHDL newbies (including me) have been wondering while discovering SpinalHDL: "Why does `in UInt(5 bits)` work and not `in Rgb(5 bits)`?"
* The benefits and drawbacks are described in the doc comments to help to choose
* It is too early to start deprecating other syntaxes, we have to "taste" things first to adjust guidelines and, maybe later, start deprecating things so for now we have to keep all syntaxes.
* If we want to not break things and not having several things that look alike, we remove absolutely all hope of evolution to improve things.

IMHO maybe it could deprecate `val a = in(Type())` (heavy syntax with braces) and maybe even `val a = in SupportedType ()` (confusing syntax) in the future, so there would be 2 non-deprecated syntaxes with a clear distinction:

```scala
val onePort = in port AnyPossibleType
in(several, ports, eg, contents, of, imasterslave, bundles)
```

And, magic, 2 (this possibly future situation) < 3 (current situation) so in the long term this change could eventually fix the current "too much choice" issue.

But as I mentioned, it is reaaaaaaally too early to deprecate things and constraint evolution, so I would see how things evolve to choose later (in at least a few months).

Dolu1990:
> But I think in this specific case this pattern addresses a real use-case

Yes i agree

> val a = in(Type()) (heavy syntax with braces) 

Yes that always enoyed me XD

> depreciation

I would say, no depreciation ever on those things.

numero-744:
> I would say, no depreciation ever on those things.

Yes, guidelines could be enough.

I have questions in TODOs, please.

Dolu1990:
hmm i have some kind of issue with the comments added in the PR.
Issue is that they seems to me to have a lot of redondancy. Also, the ratio comment / code is kind of way too high :

Ex : 

```scala
  /** Declare a port with braces
    *
    * {{{
    * val input = in(Vec(UInt(5 bits), 10))
    * }}}
    *
    * or
    *
    * {{{
    * val input = Vec(UInt(5 bits), 10)
    * in(input)
    * }}}
    *
    * See [[IODirection]] for other syntaxes.
    */
  def apply[T <: Data](data: HardType[T]): T = port(data())
```

Also, one question is how to scala doc the following fictive code : 

```scala
def setAsInput(value : UInt) ...
def setAsInput(value : SInt) ...
def setAsInput(value : BInt) ...
```

unless i miss something, scala doc would require to add a comment for each of those function, which is very wastfull, instead of regular : 

```scala
// Set value as a input of its host component.
def setAsInput(value : UInt) ...
def setAsInput(value : SInt) ...
def setAsInput(value : BInt) ...
``` 

Just trying to find a not too verbose way to document code in some usefull way XD

lib/src/main/scala/spinal/lib/MasterSlave.scala
```diff
@@ -2,87 +2,199 @@ package spinal.lib
...
+    * }}}
+    */
+  def apply(is: IMasterSlave*): Unit = is.foreach(port(_))
+
+  // TODO: this is really differen than IODirection
+  // Maybe we can compare the two means and choose the best,
+  // or document why it is not the same in comments.
   object Flow extends FlowFactory {
-    override def postApply(interface : IMasterSlave) : Unit = {
+    // TODO: can't it just be replaced by port(this)
```
Dolu1990, line 151: What would be remplaced ? postApply ?

lib/src/main/scala/spinal/lib/MasterSlave.scala
```diff
@@ -2,87 +2,199 @@ package spinal.lib
...
+      port(interface)
     }
   }
 
-  def Event : Event = {
-    val ret = spinal.lib.Event
-    this.apply(ret)
-    ret
-  }
+  // TODO: why not just renaming event above as Event?
```
Dolu1990, line 172: I think object event is never ever used. So could be deleted likely

numero-744:
> Issue is that they seems to me to have a lot of redondancy

Please point to an example in the diffs

> Also, the ratio comment / code is kind of way too high :

From other official APIs I have read (many in Rust more than Scala actually), this ratio does not seem to be an issue.

> scala doc would require to add a comment for each of those function, which is very wastfull

Actually in your example you define three times the same function I guess? Else I don't know, I can look for it.

```scala
/** Some documentation for a unique interface */
def setAsInput(value: BitVector) ...
```

lib/src/main/scala/spinal/lib/MasterSlave.scala
numero-744: Maybe just deprecate it for now, to avoid breaking changes?

Dolu1990:
> Actually in your example you define three times the same function I guess?
def setAsInput(value: BitVector) ...

That's not the point ^^

> Please point to an example in the diffs

For instance

```scala

  /** Declare a port with braces
    *
    * {{{
    * val input = in(Vec(UInt(5 bits), 10))
    * }}}
    *
    * or
    *
    * {{{
    * val input = Vec(UInt(5 bits), 10)
    * in(input)
    * }}}
    *
    * See [[IODirection]] for other syntaxes.
    */
  def apply[T <: Data](data: T): T = port(data)

  /** Declare a port with braces
    *
    * {{{
    * val input = in(Vec(UInt(5 bits), 10))
    * }}}
    *
    * or
    *
    * {{{
    * val input = Vec(UInt(5 bits), 10)
    * in(input)
    * }}}
    *
    * See [[IODirection]] for other syntaxes.
    */
  def apply[T <: Data](data: HardType[T]): T = port(data())
``` 

https://levelup.gitconnected.com/youre-commenting-your-code-too-much-and-other-controversial-thoughts-on-documentation-1ee617ed46af

In some way, this could be reduced to : 

```scala
  // Provide in/out/inout declaration to the end user. Ex : in(Bool())
  def apply[T <: Data](data: T): T = port(data)
  def apply[T <: Data](data: HardType[T]): T = port(data())
``` 

> From other official APIs I have read (many in Rust more than Scala actually), this ratio does not seem to be an issue.

*joke* Some put pineapple on pizza *joke*

lib/src/main/scala/spinal/lib/MasterSlave.scala
numero-744: Yes, I have grepped it and it is overridden only here so these `super.postApply` seem to be a no-op.

It seems really related to `Fragment`, I'm not sure I understand the mechanism behind it. It seems like a complex mechanism to apply the same master/slave-ness to another interface; which could be more simple to implement right in `IMasterSlave`?

Dolu1990:
By the way, don't get me wrong, i like pineapple on pizza, but with no ham.

numero-744:
For the example you are pointing at, you're right, there is duplication there. Maybe we could document the most used and document the other as "Same as XXX for XXX type"?

Maybe, even better, both should have a one-line docstring and refer to the examples of different syntaxes in the trait `IODirection`?

Dolu1990:
> I'm speaking about the documentation, which is not the same. Documentation explains the "what", and is for the users of the API.

For me, this code if for developper only.
I don't think we can expect users to navigate their way until that place.
That's too deep the rabbit hole i would say. (that's why there is the SpinalDoc-RTD)

> "Same as XXX for XXX type"?
> Maybe, even better, both should have a one-line docstring and refer to the examples of different syntaxes in the trait IODirection?

I would realy prefer having a "chunky boy" comment on the trait IODirection, as it can give a much more global view about "wtf is that thing" in details ^^

lib/src/main/scala/spinal/lib/MasterSlave.scala
Dolu1990: > Yes, I have grepped it and it is overridden only here so these super.postApply seem to be a no-op

Yes they are no-op. The super.postApply(interface) was placed there in case of a day it would not be a no-op anymore.

The usage of postApply is that : 
For a given hardware signal factory (ex StreamFactory), being able to apply some procedural modifier via the execution of postApply (ex setting a stream as slave when called in the context of slave Stream(xxx)

So, should not be in IMasterSlave.

lib/src/main/scala/spinal/lib/MasterSlave.scala
Dolu1990: I realy think we can just delete it, as  is can currently only be used with : 
slave.event.apply
which is kind of very broken XD

numero-744: Ok wilco

lib/src/main/scala/spinal/lib/MasterSlave.scala
numero-744: Thanks I'll document this a little bit

numero-744:
> > "Same as XXX for XXX type"?
> > Maybe, even better, both should have a one-line docstring and refer to the examples of different syntaxes in the trait IODirection?
> 
> I would realy prefer having a "chunky boy" comment on the trait IODirection, as it can give a much more global view about "wtf is that thing" in details ^^

Okay so I go for the second option, then

numero-744:
Remaining TODOs in this PR:

### `MS`

there is no such function for `MS` but there is one for `IODirection`

```scala
def apply[T <: IMasterSlave](data: HardType[T]): T = port(data())
```

### `IODirection`

I don't know how these functions are used (for the second one I guess it is just brace syntax? So I should add it to `port`?).

```scala
def cloneOf[T <: Data](that: T): T = port(spinal.core.cloneOf(that))
def apply(senum: SpinalEnum): SpinalEnumCraft[senum.type] = port(senum.craft())
```

Why isn't `Bool` overriding anything from `BaseTypeFactory`, unlike other things? `BaseTypeFactory` extends the empty trait `BoolFactory`, is there a link with it?

Dolu1990:
@numero-744 

> MS; there is no such function for MS but there is one for IODirection
def apply[T <: IMasterSlave](data: HardType[T]): T = port(data())

Ahh yes it can be added.

> def cloneOf[T <: Data](that: T): T = port(spinal.core.cloneOf(that))

Kind of never realy used, it was for when we were passing hardware signal as component parameter to provide a type parameter.
Now we can use HardType[T] which is much cleaner.

> def apply(senum: SpinalEnum): SpinalEnumCraft[senum.type] = port(senum.craft())

Ahh yes, this should also be added to port. The usage is  : in(MySpinalEnumDefinition)

> Why isn't Bool overriding anything from BaseTypeFactory, unlike other things? BaseTypeFactory extends the empty trait BoolFactory, is there a link with it?

likely it is to avoid conflict between the Bool defined in the spinal.core package object.

numero-744:
>  def apply[T <: IMasterSlave](data: HardType[T]): T = port(data())

Ah seems not possible because `IMasterSlave` does not extend `Data` (requirement for type `HardType`). Maybe `IMasterSlave` should extend `Data`?

---

>  Now we can use HardType[T] which is much cleaner.

Do I deprecate it? Is the replacement message "use `apply(HardType(T))` instead"?

---

> Ahh yes, this should also be added to port. The usage is : in(MySpinalEnumDefinition)

:heavy_check_mark: 

---

Should port have this `def port[T <: Data](data: HardType[T]): T = port(data())`, too?

---

> likely it is to avoid conflict between the Bool defined in the spinal.core package object.

But there is no issue with other Spinal types? :thinking:

numero-744:
```
[warn] /home/come/Development/SpinalHDL/core/src/main/scala/spinal/core/Data.scala:515:16: match may not be exhaustive.
[warn] It would fail on the following inputs: inWithNull, outWithNull
[warn]           that.dir match {
[warn]                ^
```

I think `inWithNull` and `outWithNull` were not a good thing! XD

Dolu1990:
@numero-744 

> Ah seems not possible because IMasterSlave does not extend Data (requirement for type HardType). Maybe IMasterSlave should extend Data?

Ahh right. I think the type T should be changed to : T <: Data

> Do I deprecate it? Is the replacement message "use apply(HardType(T)) instead"?

Naaa, don't need to depreciate it, as it is still a functional thing ^^

> Should port have this def port[T <: Data](data: HardType[T]): T = port(data()), too?

Yes

> But there is no issue with other Spinal types? 

Right. Bool is kind of a special case. It is both a Type and a function that can be used to create a Bool. And this create all sort of issues across scala versions XD
Overall, now i'm avoiding using Bool to create hardware, but instead i use Bool()

> I think inWithNull and outWithNull were not a good thing! XD

Ahh ??

numero-744:
> Ahh right. I think the type T should be changed to : T <: Data

So it will work because `IMasterSlave` is always mixed-in with `Bundle` which extends `Data`? So `IMasterSlave` should not extend `Data`?

> Naaa, don't need to depreciate it, as it is still a functional thing ^^

So `cloneOf` is "Declare port with same type as `that`"?

Btw, I am never sure, does cloning cut all assignments?

> Ahh ??

Yes, as shown in the warning message, `IODirection` is used for matching the direction so extending it with anything else than `in`, `out` or `inout` triggers warnings.

For Scala, these things, which were expected to just be helpers, are actually new "direction kinds", which is not expected.

numero-744:
@Dolu1990 The errors show it is not that simple to replace the T this way. Extending Data seems to work.

Dolu1990:
> So it will work because IMasterSlave is always mixed-in with Bundle which extends Data? So IMasterSlave should not extend Data?

Yes right.

> So cloneOf is "Declare port with same type as that"?

Yes

> Btw, I am never sure, does cloning cut all assignments?
right
It create a totaly independent signal with the same width. Other properties aren't preserved (direction, assignements, tags, isReg, and so on)

> Yes, as shown in the warning message, IODirection is used for matching the direction so extending it with anything else than in, out or inout triggers warnings.
For Scala, these things, which were expected to just be helpers, are actually new "direction kinds", which is not expected.

Ahhhh got it XD

Dolu1990:
Another solution to avoid trait IMasterSlave extends Data would have been :

def port[T <: IMasterSlave with Data](i: HardType[T]): T = port(i()) 

Maybe ?

numero-744:
I think it would work, but how better is it compared to `IMasterSlave` extending `Data`?

Dolu1990:
I would say, avoiding the diamond inheritence stuff is a good thing (avoiding class hearchy madness), also, allows IMasterSlave  to be used in different contextes

numero-744:
@Dolu1990 I think code is good now, please tell me if you approve it so that I rework git history and merge

Dolu1990:
Seems good to me ^^
So if it pass the regression test, it can be merged.
Thanks for the PR :D

Dolu1990:
@numero-744 All good for you ?

numero-744:
Rebased and cleaned history, I merge if it passes tests

PR #936, opened 2022-10-25
ci: add scalafmtCheck step

### LINKED ISSUE #535: Consider using code formatting tools

numero-744:
Fixes #535 

### This PR avoids double workflow

Double workflow makes annotations appear twice if PR and source branch are on the same repository, and consume twice more energy.

### This PR updates `sbt-scalafmt` plugin and configures it

Only modified files (compared to `dev` branch) are checked

### This PR adds a step to Scala CI workflow to check formatting against `dev` and put annotations on files

Example of annotated PR: https://github.com/numero-744/SpinalHDL/pull/6/files#diff-9c3e1dfcdddbdc33403d5e3c6ac4b6728f8f86ba6f7448c81d47906c8f7895f6

Notice [at the end of the diff](https://github.com/numero-744/SpinalHDL/pull/6/files#diff-9c3e1dfcdddbdc33403d5e3c6ac4b6728f8f86ba6f7448c81d47906c8f7895f6R62-R66), there are 3 blank lines, which GitHub detects correctly as a problem with severity "warning" (see the Checks log), but decides to not put the annotation on the file, I don't know why.

About authorship: I could not make work the stuff I found on other projects so I decided to do my own things, so I'm responsible for any bug which can occur XD

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -1,6 +1,12 @@
 name: Scala CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+    branches:
+      - dev
 
 jobs:
   test:
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -47,3 +47,21 @@ jobs:
 
     - name: Build scaladoc
       run: sbt unidoc
+
+  fmt-lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: 'recursive'
+    - uses: ./.github/actions
+    - name: Fetch dev to be able to check only modified files
+      run: |
+        git fetch --depth=1 origin dev
+        git branch dev origin/dev
+    - name: Setup annotations
+      run: echo "::add-matcher::.github/matcher.json"
+    - name: Check formatting and pass
+      run: bash .github/checkfmt.bash
--- a/.github/checkfmt.bash
+++ b/.github/checkfmt.bash
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+print() {
+  local line=$1
+  local text=$2
+
+  echo "$file:$line: `basename $file` $message"
+  echo "$text"
+}
+
+blank='^[[:space:]]*$'
+
+sbt -no-colors clean scalafmtCheck | while IFS= read ll
+do
+
+  # Keep all lines in logs
+  echo "$ll"
+
+  # Skip lines which are not [warn]
+  if ! [[ "$ll" =~ ^\[warn\] ]]; then
+    continue
+  fi
+  # Remove this warn to ease process
+  l=$(sed 's/^\[warn\] //' <<< "$ll")
+
+  # Match start of mis-formatted file
+  if [[ "$l" =~ ^scalafmt:\ $PWD/([^[:space:]]+)\ (isn\'t\ formatted\ properly\!)$ ]]
+  then
+    file="${BASH_REMATCH[1]}"
+    message="${BASH_REMATCH[2]}"
+  fi
+
+  # Debug
+  #echo "debug $linea:$lineb:$l"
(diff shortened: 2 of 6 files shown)


numero-744:
Reworded second commit from `fix:` to `ci:`, ready for review.

Dolu1990:
> Scala CI / fmt-lint (pull_request) Successful in 1m

Rawrrrrrrr

hmm, maybe we can include a dummy example file in the PR to show how things behave on dirty code :D ?

Dolu1990:
Here is a lot of dirty code <3 :
https://github.com/SpinalHDL/SpinalHDL/blob/dev/tester/src/main/scala/spinal/tester/code/Play2.scala

numero-744:
So it does not show all annotations but in the Check run panel they are correctly detected by GitHub… :angry:  I have reduced verbosity in the example PR and I get the same results so this is not the issue. So much time spent to be finally BugHit

 https://github.com/community/community/discussions/9144

We could also throw my work about parsing diffs and convert them to "problems" with blank lines adjustments, and just have an action which fails if there is a bug… As it only checks the modified files it will not fail in `dev` anyway. And the user can still read the diff in the check result.

Dolu1990:
As you feel it, don't burn yourself ^^
If that's more pain than fun, don't need to do it.

numero-744:
So maybe we can merge it and see how it behaves in real situation, then adjust later?

.github/workflows/scala.yml
```diff
@@ -1,6 +1,12 @@
 name: Scala CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - dev
```
Dolu1990, line 6: Why only dev ?
I would say should be for all branches ?

numero-744: The `pull_request` case after matches it

Dolu1990: I mean, both should be for all branches i would say ?

numero-744: In this case for PRs you get the flow run twice, displaying annotations twice and consuming twice as much energy.

If we do not care about energy we can do `if: github.event_name != "pull_request" || github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name` for the lint job to avoid annotations twice?

Ref: https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/

Dolu1990: Ahhh damned XD
Right, so we do care about it as the regression take quite a bit of processing power.

So the case where things run twice is when sombody do a PR, each of his commit will trigger the regression on his repo and the SpinalHDL repo ?

numero-744: No, it runs on each push, only on the last commit / top of the branch (I think we can't do anything against this, just don't `git push` if you don't want to share for a peer-review).

It is more that for a given pushed commit, the workflow is run twice, once as a `push` on the author repository, and once as a `pull_request` on the target repository. If they are the same, currently there are 2 workflows started to do exactly the same thing on exactly the same commit, which make annotations displayed twice.

With the suggested modification it will always run on `dev` (if you push on `dev` so that others don't see what you are doing XD or after merging a PR) and run on branches which have an open PR to `dev`.

Dolu1990:
> No, it runs on each push, only on the last commit / top of the branch (I think we can't do anything against this, just don't git push if you don't want to share for a peer-review).

Yes, that's ok

> It is more that for a given pushed commit, the workflow is run twice, once as a push on the author repository, and once as a pull_request on the target repository.

Ahhh yes

> With the suggested modification it will always run on dev (if you push on dev so that others don't see what you are doing XD or after merging a PR) and run on branches which have an open PR to dev.

Good to me then

numero-744:
@Dolu1990 I have rebased on `dev` to fix conflicts, please merge when it is okay for you.

Dolu1990:
Thanks ^^

Issue #937, opened 2022-10-25
Let `randomize` calls return the new value


andreasWallner:
I often encounter the situation that I randomize a signal in simulation but would need the new value, like with
```
...
someBool.randomize()
// won't work since we need a delta cycle to read the correct value
if(someBool.toBoolean) {...}
```

I'd propose to modify the Bool/Bits/UInt/SInt/... randomize methods to return the new value so that we could do

```
if(someBool.randomize()) {...}
```

The classes I currently wouldn't know how to handle would be Data & Bundles...

Or did I overlook functionality that also works around this issue so this is not needed? If not and you think it's fine to have this I'd open a PR.

Readon:
Do you need delayed logic as Dolu90 mentioned in #655 ?

numero-744:
Question: for `UInt` and `SInt` it would return `BigInt` instead of `Int`?

numero-744:
For `Bundle` what do you think about returning a kind of dictionary? (I don't know how Scala would handle this)

numero-744:
Actually it would be really fun to turn the Bundle into another custom type with these same user-defined fields from the Bundle (via reflection?) but with other types (Bool -> Boolean, etc.).

I've found this but I don't know how an IDE auto-completion would react XD
https://www.scala-lang.org/api/current/scala/Dynamic.html

numero-744:
Got this draft working but I don't like it much… Returning `Any` is :vomiting_face: + no auto-completion or type-checking in IDE, just notice assertions fail if you wrote `assert(someBool == 5)`.

```scala
import scala.collection.mutable.Map

object SimBundle {
  import spinal.core.sim._

  def apply(bundle: Bundle): SimBundle = {
    val map = Map[String, Any]()

    bundle.elements.foreach {
      case (name, b: Bool)         => map += ((name, b.toBoolean))
      case (name, bits: BitVector) => map += ((name, bits.toBigInt))
      case (name, bundle: Bundle)  => map += ((name, apply(bundle)))
    }

    new SimBundle(map)
  }
}

class SimBundle(map: Map[String, Any]) extends Dynamic {
  def selectDynamic(name: String) = map.get(name).get
}

object TestSimBundle extends App {
  class Dut extends Component {
    val io = new Bundle {
      val someBool = out Bool ()
      val someUInt = out UInt (16 bits)
    }
    io.someBool := True
    io.someUInt := 6
  }

[...]

PR #940, opened 2022-10-26
feat(sim): let randomize() return the new value

andreasWallner:
Fix for #937 - opening this so that we have code to discuss over.

I considered adding functions that would return the value, leaving `randomize` as it is but a) these additional similar functions would have made the API much unclearer b) the impact on `randomize` (performance) seemed minimal, considering a switch to C code happens there c) I could not find a name that I liked (`randomizedLong` is out of the question...).
That's why I landed on adding the BigInt return type (which can easily be converted if needed as int/long)

What I like about this: makes many of my simulation models simpler.
What I do not like / am not sure about:
- There is currently no solution for `Bundle` (which should be the only case handled by the `SimDataPimper`) - but I also do not see any apart from using Dynamic (I guess we'd also have to cache some stuff there, dynamic is quite slow - and with changing `randomize` that would be felt in quite a few places)
- UFix/SFix/AFix `randomize` returns a `Double` in my implementation - maybe a BigDecimal would be better here...

### DIFF
--- a/core/src/main/scala/spinal/core/sim/package.scala
+++ b/core/src/main/scala/spinal/core/sim/package.scala
@@ -328,8 +328,10 @@ package object sim {
 
     def #=(value: Boolean) = setLong(bt, if(value) 1 else 0)
 
-    def randomize(): Unit = {
-      bt #= Random.nextBoolean()
+    def randomize(): Boolean = {
+      val b = Random.nextBoolean()
+      bt #= b
+      b
     }
   }
 
@@ -356,81 +358,62 @@ package object sim {
     }
   }
 
-  /**
-    * Add implicit function to Bits
-    */
-  implicit class SimBitsPimper(bt: Bits) {
+  protected abstract class RandomizableBitVector(bt: BitVector, longLimit: Int) {
+    protected val width = bt.getWidth
 
-    def randomize(): Unit = {
-      val width = bt.getWidth
-      if(width < 64){
-        bt #= Random.nextLong() & ((1l << width) - 1)
-      }else {
-        bt #= BigInt(width, Random)
+    def randomize(): BigInt = {
+      if (width < longLimit) {
+        val l = randomizedLong()
+        bt #= l
+        l
+      } else {
+        val bi = randomizedBigInt()
+        bt #= bi
+        bi
       }
     }
 
-    def randomizedBigInt() = {
-      val width = bt.getWidth
-      BigInt(width, Random)
-    }
+    def randomizedBigInt() = BigInt(width, Random)
+
     def randomizedLong() = {
-      val width = bt.getWidth
       assert(width < 64)
       Random.nextLong() & ((1l << width) - 1)
     }
+
     def randomizedInt() = {
-      val width = bt.getWidth
       assert(width < 32)
       Random.nextInt() & ((1 << width) - 1)
     }
   }
 
   /**
-    * Add implicit function to UInt
-    */
-  implicit class SimUIntPimper(bt: UInt) {
+   * Add implicit function to Bits
+   */
+  implicit class SimBitsPimper(bt: Bits) extends RandomizableBitVector(bt, 64)
 
-    def randomize(): Unit = {
-      val width = bt.getWidth
-      if(width < 64){
-        bt #= Random.nextLong() & ((1l << width) - 1)
-      }else {
-        bt #= BigInt(width, Random)
-      }
-    }
-
-    def randomizedBigInt() = {
-      val width = bt.getWidth
-      BigInt(width, Random)
-    }
-    def randomizedLong() = {
-      val width = bt.getWidth
-      assert(width < 64)
-      Random.nextLong() & ((1l << width) - 1)
-    }
-    def randomizedInt() = {
-      val width = bt.getWidth
-      assert(width < 32)
-      Random.nextInt() & ((1 << width) - 1)
-    }
-  }
+  /**
+   * Add implicit function to UInt
+   */
+  implicit class SimUIntPimper(bt: UInt) extends RandomizableBitVector(bt, 64)
 
 
   /**
-    * Add implicit function to SInt
-    */
-  implicit class SimSIntPimper(bt: SInt) {
-
-    def randomize(): Unit = {
-      val width = bt.getWidth
-      if(width <= 64){
-        val shift = 64 - width
-        bt #= (Random.nextLong() << shift) >> shift
-      }else {
-        bt #= BigInt(width, Random) - (BigInt(1) << width-1)
-      }
+   * Add implicit function to SInt
+   */
+  implicit class SimSIntPimper(bt: SInt) extends RandomizableBitVector(bt, 65) {
+    override def randomizedLong(): Long = {
+      assert(width <= 64)
+      val shift = 64 - width
+      (Random.nextLong << shift) >> shift
     }
+
+    override def randomizedInt(): Int = {
+      assert(width <= 32)
+      val shift = 32 - width
+      (Random.nextInt() << shift) >> shift
+    }
+
+    override def randomizedBigInt(): BigInt = BigInt(width, Random) - (BigInt(1) << width - 1)
   }
 
   /**
@@ -442,8 +425,10 @@ package object sim {
 
     def #=(value: SpinalEnumElement[T]) = setBigInt(bt, bt.encoding.getValue(value))
 
-    def randomize(): Unit ={
-      bt #= bt.spinalEnum.elements(Random.nextInt(bt.spinalEnum.elements.length))
+    def randomize(): SpinalEnumElement[T] = {
+      val e = bt.spinalEnum.elements(Random.nextInt(bt.spinalEnum.elements.length))
+      bt #= e
+      e
     }
   }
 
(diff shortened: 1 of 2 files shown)


numero-744:
I agree with all your arguments here.

For `MultiData` (`Vec` and `Bundle`) in general, maybe we can just consider that the user can:

* randomize fields one by one
* randomize the whole thing then re-randomize a specific field to get the randomized value. After all it is just putting random values XD
* Create specific `case class`es and override `randomize` if it is really needed

<details><summary>Another bad idea that I put here just for completeness</summary>
Another idea (again): I guess `#=` uses some "force" mechanism. Maybe we could do something like `def toXxx = if (forced) forcedValue else readValue`? Issue: it is a breaking change if the sim code was using delta-cycle propagation to keep reading previous value after having changed the drive.
</details>

Is there a specific reason why you keep UInt, SInt and Bits apart, keeping code duplication? Can't it be merged into a `SimBitVectorPimper`?

Also, if I understand correctly what these types are, I agree that for precision's sake Double should not be used.

wswslzp:
From my experience as a verification engineer, randomization does not happen on the signal level but transaction level. 

In a typical testbench, we have a stimulus model that generates all the constrained random transactions and these transactions all will be converted to signal values to drive the interface. 

Therefore, the `randomize` API provided by SpinalHDL is not necessary from my point of view.

numero-744:
Hmm even if it does not enable to perform complete verification (UVM?) tests, maybe sometimes the designer wants to use `randomize()` while doing simple/basic troubleshooting on a design?

andreasWallner:
@wswslzp Most of my tests for SpinalHDL hardware look quite different from what I might do at work...
I think I find the `randomize` calls here so useful because of the often done split between driver/monitor, which alleviates the need to prepare transactions "beforehand". And in my models I could for sure use other ways of generating randomness, but `randomize` has the advantage to me that it generates data bounded to the signal width - less thinking/work to cover generic widths in test cases. But I guess to each their own ;-)

@numero-744 Thanks for the hint - I actually did not want to extend the API for a `BitVector`. Main reason for that is that a BitVector does not have an interpretation of the data and I've never seen them used in user code. I also wanted to make sure that you wouldn't though some API by accident cast down from e.g. an `SInt` to `BitVector` and get seemingly weird behavior.

I'll change the implementation for the fixed-width types and update the PR

Readon:
> @numero-744 Thanks for the hint - I actually did not want to extend the API for a `BitVector`. Main reason for that is that a BitVector does not have an interpretation of the data and I've never seen them used in user code. I also wanted to make sure that you wouldn't though some API by accident cast down from e.g. an `SInt` to `BitVector` and get seemingly weird behavior.

According to the [docs](https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Data%20types/index.html), the BitVector is the base class of Bits, UInt, SInt so that if you pimp on BitVector, it's not necessarily pimp on both of its subclasses individually.

numero-744:
Still do not want to merge these classes into one for `BitVector`? It would avoid code duplication (easier to maintain).

andreasWallner:
@numero-744 That actually does not work, since the Pimpers are not derived from each other, so I couldn't call e.g. the `randomizedLong` method of `SimUIntPimper` from `SimBitVectorPimper`. If I'm in `SimBitVectorPimper` the type of the data is already casted down to BitVector, so the `SimUIntPimper` implicit wouldn't work.
We also can't move e.g. `randomizedLong` to `SimBitVectorPimper` since the implementation depends on whether we have a signed/unsigned type (or decimal type).
Ok, we could move the `Bits` and `UInt` ones to `BitVector`, but I'm not sure that actually improves readability (since there is no derivation involved)
Also I'd still like to avoid getting weird results if e.g. an SInt is down-casted to a BitVector somwehere. (in which case you could get inconsistent results w/o noticing for a while)

Readon:
Yes, if there is a call to SInt's randomize, the sign has to be considered.
However, is that possible to define a common randomize method to BitVector which is not sensitive to sign bits.
And add sign aware version of randomize method to only SInt like types.

If possible, why we should check the bit width first to decide calling on randomizeLong or randomizeBigInt when the return value is always BigInt? Is that design for performance? As I know, the branch might cause some performance loss when branch predict fall which lead to more time to run.

One more thing, it seems that randomizeInt is not used at all, what is it for?

btw, good work!

andreasWallner:
I'll update the PR with a an interface to be shared between `Bits`, `UInt` and `SInt` - this way we all get what we want (I think ;-) )

@Readon ad `BitVector`: I think that it should not have a `randomize` method, just like it is now. It is not used by library users and only a possible source for confusion.

ad long vs BigInt: I'd assume so, took that from the original code

ad `randomizedInt`: as a user-facing API if you want an `Int`?

tester/src/test/scala/spinal/tester/scalatest/SpinalSimRandomizeTester.scala
```diff
@@ -0,0 +1,57 @@
...
+        val b = dut.io.b.randomize()
+        val ui = dut.io.ui.randomize()
+        val uil = dut.io.uil.randomize()
+        val si = dut.io.si.randomize()
+        val sil = dut.io.sil.randomize()
+        val e = dut.io.e.randomize()
+        val uf = dut.io.uf.randomize()
+        val sf = dut.io.sf.randomize()
+
+        dut.clockDomain.waitSampling()
```
numero-744, line 44: Could be just `sleep(10)`, to avoid tricks to add an unused clock domain?

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -356,50 +358,49 @@ package object sim {
...
-      if(width < 64){
-        bt #= Random.nextLong() & ((1l << width) - 1)
-      }else {
-        bt #= BigInt(width, Random)
-      }
-    }
-
+   * Add implicit function to UInt
+   */
+  implicit class SimUIntPimper(bt: UInt) extends RandomizableBitVector(bt, 64) {
```
numero-744, line 403: `extends SimBitsPimper(bt)` with no body? Or even, as suggested above, defining defaults for functions in the base class, and override functions only for `SInt`?

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -356,50 +358,49 @@ package object sim {
     }
   }
 
-  /**
-    * Add implicit function to Bits
-    */
-  implicit class SimBitsPimper(bt: Bits) {
+  protected abstract class RandomizableBitVector(bt: BitVector, longLimit: Int) {
```
numero-744, line 361: Hmm how does `protected` behave in a package? Is it like `private`?

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -356,50 +358,49 @@ package object sim {
...
+        bt #= bi
+        bi
       }
     }
 
+    def randomizedBigInt(): BigInt
+
+    def randomizedLong(): Long
+
+    def randomizedInt(): Int
```
numero-744, line 380: What about providing a default implementation of all functions as for unsigned, and then only `SInt` would override by substraction?

```scala
override def randomizedBigInt(): BigInt = {
  super.randomizedBigInt() - (BigInt(1) << width)
}
```

Also, isn't it possible to just have the base class `implicit`, not `abstract` (is the word for it "concrete"? "actual"? "instantiable"? "instantiatable"?) and public, and having another implicit class only for `SInt`?

numero-744 (commented):
Good job, the "using Long or BigInt" logic has been de-duplicated and the API declarations are made uniform. I think more code can be de-duplicated, see suggestions below.

Dolu1990:
> UFix/SFix/AFix randomize returns a Double in my implementation - maybe a BigDecimal would be better here...

Hmm Maybe there the best may be to be explicit about the expected return type.  randomizeBd  randomizeBi, randomizeDouble, as multiple usage may lead to different needs ?

Dolu1990:
So, staying explicit to avoid future API conflicts ?

andreasWallner:
@Dolu1990 The "issue" with that is the easy mixup between `randomizedBigInt` and `randomizeBigInt`, which was the initial reason I didn't want to do that. Both would return a value, but only one would modify the signal...

tester/src/test/scala/spinal/tester/scalatest/SpinalSimRandomizeTester.scala
andreasWallner: didn't think of that since that would be my first comb-only SpinalHDL component ;-)

core/src/main/scala/spinal/core/sim/package.scala
andreasWallner: sure, I forgot to commit that before I pushed and had it lying around here

core/src/main/scala/spinal/core/sim/package.scala
andreasWallner: changed to private, don't know why I used protected...
The difference, after looking it up: protected classes can be returned by functions to outside of the package, private can not

andreasWallner: Edit: changed by to protected, if the class is private then it is not allowed to escape the context at all, so a public class can´t derive from it ... the blog where I looked it up was a bit...wrong

Dolu1990:
Ahh right XD
Also, if the user want BigInt he can use the xFix.raw.randomize

numero-744 (approved):
Good work, LGTM

I have added side-notes but I also do approve this PR as-is

I'll merge it in a few days if there is no activity / if nobody is against it, so if anyone finds an issue with this PR, let me know. Also feel free to merge it if you are okay with it

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -356,80 +358,65 @@ package object sim {
...
-    * Add implicit function to SInt
-    */
-  implicit class SimSIntPimper(bt: SInt) {
+   * Add implicit function to SInt
+   */
+  implicit class SimSIntPimper(bt: SInt) extends RandomizableBitVector(bt, 65) {
+    override def randomizedLong(): Long = {
+      assert(width <= 64)
+      val shift = 64 - width
+      (Random.nextLong << shift) >> shift
```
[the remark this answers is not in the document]
andreasWallner: I'm not sute I understand what you mean. We need an assert here since it's still user facing code, and the width to assert here differs from the unsigned case?

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -502,6 +501,7 @@ package object sim {
     val fractionLength = bt.fracWidth
     val maxRawIntValue = bt.maxRaw
     val minRawIntValue = bt.minRaw
+    private val factor = scala.math.pow(2, fractionLength)
```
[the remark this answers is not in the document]
andreasWallner: For a single bit being set a double would actually precise xD - and the rest depends on how BigDecimal division is implemented

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -465,35 +454,45 @@ package object sim {
...
   implicit class SimUFixPimper(bt: UFix) extends SimFix(bt){
+    private val factor = scala.math.pow(2, fractionLength)
     override val maxRawIntValue = bt.raw.maxValue
     override val minRawIntValue: BigInt = 0
 
     override protected def rawAssign(that: BigInt): Unit = bt.raw #= that
     override def toBigDecimal: BigDecimal = {
-      BigDecimal(bt.raw.toBigInt) / scala.math.pow(2, fractionLength)
+      BigDecimal(bt.raw.toBigInt) / factor
+    }
```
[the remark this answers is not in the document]
andreasWallner: I'll change it here and in another location further up in the evening

core/src/main/scala/spinal/core/sim/package.scala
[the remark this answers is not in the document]
numero-744: `super.randomized…` already performs the same assertion doesn't it?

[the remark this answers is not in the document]
numero-744: If `longLimit` is inclusive it is possible to use it but anyway there is the same issue with `randomizedInt`.

andreasWallner:
@numero-744 from my POV this could be merged

numero-744:
Thanks for the ping

Issue #85, opened 2017-11-09
Feature Proposal: allowing latches


cristianmacario:
Hi,
talking about latches (let's say today is Latch day ;) ), even if they are rarely used, they might be useful (for instance in describing clock gating cells).
I would propose to allow the intentional introduction of Latches, using tags, just as we do with combinatorial loops (with the "noCombinatorialLoopCheck" tag)
For instance, introducing a tag called "noLatchCheck" or something similar, we could define latches like follws:
```scala
  val d = Bool
  val e = Bool
  val l = Bool addTag(noLatchCheck)
  when (e) {
    l := d
  }
```
This would need to be translated in Verilog as:
```verilog
  wire e;
  wire d;
  reg l;
  always @ (e or d)
  begin
    if(e)begin
      l <= d;
    end
  end
```
Let me point out that in Verilog (I don't know about VHDL) latches should be assigned using the non-blocking assignment ("<=") instead of the blocking assignment ("=").
Just a proposal, let me know what you think.
Thanks!

Dolu1990:
Hi,

So until now, the idea was "If you want latches, you have to instanciate them through blackboxes"

But now it could be possible to integrate the feature in reworkDev branch, be before trying it, i would like to release the reworkDev branch into the master one, see if everything is fine, get some feedbacks, and later adding new feature of this order (latches).  

Until that time, is the blaboxing way enough ? 
Maybe those blackboxes could be integrated into devRework, and be instanciated via a wrapper. Then when the feature is nativly supported it would not need any changes in the user code base.

cristianmacario:
Sure, latches are not so essential and blackboxing is OK.
Future/long term feature is totally fine.

numero-744:
For the design I think a `Latch(T)` would feel more natural (like `Reg(T)`). With `LatchWhen(...)`, etc.

So your example would be:

```scala
val d = Bool
val e = Bool
val l = Latch(Bool)
when (e) {
    l := d
}
```

Or

```scala
val d = Bool
val e = Bool
val l = LatchWhen(d, e)
```

PR #944, opened 2022-11-02
feat: add `Latch` and `LatchWhen`

dokleina:
I took a pass at implementing #85 because I saw it under the milestone and I had bandwidth.


core/src/main/scala/spinal/core/Latch.scala
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+
+object LatchWhen {
+
+  def apply[T <: Data](next: T, cond: Bool): T = {
+    val latch = next.clone()
+      .setCompositeName(next, "latchWhen", true)
+      .addTag(noLatchCheck)
+      .asInstanceOf[T]
```
numero-744, line 16:
```suggestion
  def apply[T <: Data](next: HardType[T], cond: Bool): T = {
    val latch = Latch(next).setCompositeName(next, "latchWhen", true)
```

Is it possible to re-use code like this? See https://github.com/SpinalHDL/SpinalHDL/blob/603bfcc3a234f092242e37721450f406fae066b5/core/src/main/scala/spinal/core/Reg.scala#L60

core/src/main/scala/spinal/core/Latch.scala
```diff
@@ -0,0 +1,23 @@
+package spinal.core
+
+object Latch {
+
+  def apply[T <: Data](dataType: HardType[T]): T = {
+    dataType.apply().addTag(noLatchCheck)
```
numero-744, line 6:
```suggestion
    dataType().addTag(noLatchCheck)
```

You can just `apply` with `()`. This is done with `HardType` at other places in the code and I think it is better to remain consistent and use just `()`.

Dolu1990:
Ahhhhhh realy nice :D :D
Once the regression test execution is done, it can be merged i think.

Thanks ^^

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
```diff
@@ -0,0 +1,39 @@
+package spinal.tester.scalatest
+
+import org.scalatest.funsuite.AnyFunSuite
+import spinal.core._
+import spinal.core.sim._
+
+class LatchTester extends AnyFunSuite {
+  test("bad latch") {
+    SimConfig.compile(new Component {
```
numero-744, line 9: I guess this test will fail, and the failure should be asserted here.

Maybe using something like:

```scala
var goodFail = false
try {
  SpinalVerilog(new Component …)
} catch {
  case e: ExpectedErrorType => goodFail = true
}
assert(goodFail)
```

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
```diff
@@ -0,0 +1,39 @@
...
+      val latchVal = Bits(8 bits)
+
+      when(myCond) {
+        latchVal := inVal
+      }
+    })
+  }
+
+  test("allowed latch") {
+    SimConfig.compile(new Component {
```
numero-744, line 21: Maybe we can add a simulation test to check the actual behavior of generated code? @Dolu1990 do you think it is needed?

numero-744 (commented):
Thanks a lot!

I think that this + `port` keyword can make a great v1.8 release?

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
Dolu1990: Yes it would be good. Also to see how verilator will hate us <3

numero-744:
About the docs I think it is self-explanatory, no need Scaladoc for now. I'll open an issue in RTD to document it. I think it will be possible to add a kind of "added in vX.X" tag in RTD for new features?

numero-744:
Opened issue https://github.com/SpinalHDL/SpinalDoc-RTD/issues/136 for documentation

Dolu1990:
Yes i think 1.8 is fair.

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
numero-744: It failed as expected, I got this working:

```scala
  test("bad latch") {
    var goodFail = false

    try {
      SpinalVerilog(new Component {
        val myCond = Bool
        val inVal = Bits(8 bits)
        val latchVal = Bits(8 bits)

        when(myCond) {
          latchVal := inVal
        }
      })
    } catch {
      case e: SpinalExit if e.getMessage().contains("LATCH DETECTED") =>
        goodFail = true
    }

    assert(goodFail)
  }
```

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
dokleina: Verilator hates it. There's a warning about adding linter comments:
```
%Warning-LATCH: Users/[...]/SpinalHDL/tmp/job_1/unamed.v:14:3: Latch inferred for signal 'unamed.latchVal' (not all control paths of combinational always assign a value)
                                                                : ... Suggest use of always_latch for intentional latches
   14 |   always @(*) begin
      |   ^~~~~~
                ... Use "/* verilator lint_off LATCH */" and lint_on around source to disable this message.
%Error: Exiting due to 1 warning(s)
```
How should this be resolved? (I'm not familiar with Verilator)

core/src/main/scala/spinal/core/Latch.scala
dokleina: I recall I had the `apply()` because IntelliJ underlined it for some reason. I'll drop it

core/src/main/scala/spinal/core/Latch.scala
dokleina: I wasn't aware that would work. It seems it does so I'll take it

dokleina:
Thanks for the feedback everyone! I made some changes with your recommendations.

I managed to work around Verilator's warnings about latches with a comment on the Component definition.

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
```diff
@@ -4,36 +4,99 @@ import org.scalatest.funsuite.AnyFunSuite
...
-      val latchVal = LatchWhen(inVal, myCond)
+      definition.addComment("verilator lint_off LATCH")
+
+      val io = new Bundle {
+        val myCond = in Bool()
+        val inVal = in Bits (8 bits)
+        val outVal = out Bits (8 bits)
+      }
+      io.outVal := LatchWhen(io.inVal, io.myCond)
+    }).doSim(dut => {
```
numero-744, line 80: Idea to avoid test code duplication.

Here we are testing two different configurations of an entity, in VHDL words.

(This is draft code written in GitHub comments, did not test it)

```scala
// In LatchTester
class Dut extends Component {
  val io = new Bundle {
    val cond = in Bool ()
    val input = in Bits (8 bits)
    val output = out Bits (8 bits)
  }
}

class DutWhen extends Dut {
  val latch = Latch(Bits(8 bits))
  when (io.cond) { latch := io.input }
  io.output := latch
}

class DutLatchWhen extends Dut {
  io.output := LatchWhen(io.input, io.cond)
}

def protocol(dut: Dut): Unit = {
  // the simulation code here
}

test("Latch") {
  SimConfig.compile(DutConfigWhen).doSim(protocol)
}

test("LatchWhen") {
  SimConfig.compile(DutConfigLatchWhen).doSim(protocol)
}
```

Note: could this go in https://github.com/SpinalHDL/SpinalDoc-RTD/issues/131?

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
```diff
@@ -4,36 +4,99 @@ import org.scalatest.funsuite.AnyFunSuite
...
+    }
+
+    assert(didRaise)
   }
 
   test("allowed latch") {
     SimConfig.compile(new Component {
-      val myCond   = Bool
-      val inVal    = Bits(8 bits)
+      definition.addComment("verilator lint_off LATCH")
```
numero-744, line 33: Having to do a manual fix is not really user-friendly…

* can it be more local to the latch (signal definition?)
* can it be added automatically with `Latch` and `LatchWhen`?

@Dolu1990 any idea?

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
dokleina: That largely worked as you expected it to. The only hang up I had was with the `.doSim(protocol)`. I had to do either `.doSim(protocol _)` or `.doSim(protocol(_))`. You'll see it in the commit I made. (I also found `simDriver` to be a more descriptive name for the method)

Thanks for the suggestion! I knew there had to be a way to keep it DRY

numero-744: Ah yes, maybe defining the protocol as below would have allowed the `.doSim(protocol)`. But that's not important. Thanks again for your work!

```scala
val protocol: Dut => Unit = { dut =>
  // the simulation code here
}
```

tester/src/test/scala/spinal/tester/scalatest/LatchTester.scala
```diff
@@ -0,0 +1,87 @@
...
+    val oldInput = dut.io.input.toInt
+    var newInput = dut.io.input.randomizedInt()
+    while(newInput == oldInput) newInput = dut.io.input.randomizedInt()
+    dut.io.input #= newInput
+
+    // Wait a bit
+    sleep(1)
+
+    // Check that the latch did not take the value
+    assert(dut.io.input.toInt != dut.io.output.toInt)
```
numero-744, line 77:
```suggestion
    assert(dut.io.output.toInt == oldInput)
```

Not only we want to check that the latch is not in "capture" mode, we also want it to be in "keep" state.

numero-744:
Okay so I am waiting for a feedback from @Dolu1990 about:

* the fact that the user has to add a comment to avoid Verilator warning (could it be done another way?)
* what about testing with other tools / languages?

Then, before the PR is merged, @dokleina could you auto-format the new files, please?

Your favorite IDE or `sbt` should be able to do that:

```sbt
core/scalafmtOnly src/main/scala/spinal/core/Latch.scala
```

OR, if run from shell

```bash
sbt 'core/scalafmtOnly src/main/scala/spinal/core/Latch.scala'
```

(LatchTester is already correctly formatted but FYI the command would be `tester/scalafmtOnly test/scala/spinal/tester/scalatest/LatchTester.scala`)

dokleina:
@numero-744 Do the files need the Spinal copyright header? I can add it real quick if they do.

numero-744:
LGTM, I am waiting for:

* the tests to pass
* @Dolu1990 (sry I'm spamming you) to answer to:
    * the fact that the user has to add a comment to avoid Verilator warning (could it be done another way?)
    * what about testing with other tools / languages?
    * question about copyright header

Dolu1990:
@numero-744 No worries, i missed your initial ping.

> the fact that the user has to add a comment to avoid Verilator warning (could it be done another way?)

Yes sure, we can add a transformation phase to patch the latches automaticaly when compiling for verilator.
See : 
https://github.com/SpinalHDL/SpinalHDL/blob/591c51b1b9ce81e1f69003bf1c85cee57486d4c8/core/src/main/scala/spinal/core/sim/SimBootstraps.scala#L872

Which does something kinda similar but to enforce signal to be public.
Idealy, we can replace it with a phase which does both a once (latch + sim public features).

Do you wanna try implement it ? else i can do it.

> what about testing with other tools / languages?

So, GHDL and IVerilog are supported by SpinalSim, they can be used for it.
So idealy, instead of manualy using SpinalSim, you can use the "SpinalSimTester" which is used to run the test on all 3 supported simulators (verilator, ghdl, iverilog)


Grosso modo : 
```scala
  SpinalSimTester { env =>
    import env._

    test("miaou) {
    }
  }    
````

> question about copyright header

Personnly i do not botther, don't need to be added.

Dolu1990:
(the the import env._ will provide a SimConfig already configured with the proper simulator, and the landa function providing env =>  will be call for each of the simulator)

Dolu1990:
Thanks ^^

numero-744:
@Dolu1990 don't we have to re-enable the warning after the component?

Also is it okay to add several times the disabling comment for the same component?

Dolu1990:
@numero-744 

> Also is it okay to add several times the disabling comment for the same component?

It is only added once (hashset)

> don't we have to re-enable the warning after the component?

Would have been ideal, but hoo well, not that important at the end, as everything was already linted by SpinalHDL itself.

Good to merge ?

numero-744:
> It is only added once (hashset)

Ah yes `HashSet` not `HashMap` :smile: 

> Would have been ideal, but hoo well, not that important at the end, as everything was already linted by SpinalHDL itself.

Yes why not

> Good to merge ?

Ahah a little bit too late to ask XD

Dolu1990:
> Ahah a little bit too late to ask XD

Ahhh right XD

PR #950, opened 2022-11-07
feat(lib): debug over Intel Virtual JTAG

LYWalker:


numero-744:
@LYWalker Hello, thanks for the PR!

Do you have an example of code using this new feature, please? I will help us understanding the modifications in the PR to review them, and then to document the feature.

numero-744:
Thanks for the link. Now I understand what it is for.

We will still have to document the feature on its own, but it can be done later :smiley:

I think [`lib/src/main/scala/spinal/lib/blackbox/altera/vjtag.scala`](https://github.com/SpinalHDL/SpinalHDL/blob/c89af6d97ec8317339d137a997f9db9214049996/lib/src/main/scala/spinal/lib/blackbox/altera/vjtag.scala) should be formatted.

```sh
sbt 'lib/scalafmtOnly src/main/scala/spinal/lib/blackbox/altera/vjtag.scala'
```

@Dolu1990 can you review this PR? Else do you know who can?

LYWalker:
@Dolu1990 I'll happily rename the bundle to VjatgIO, but I'm not sure why the blackbox was named VJTAG in the first place. Altera have no such component, so it won't compile unless it's manually supplied (itself instantiating an sld_virtual_jtag component)

Dolu1990:
@LYWalker Hoo, about VJTAG, i have no idea. It isn't ideal, but let's keep VJTAG blackbox and have VjtagIo :)

Dolu1990:
Thanks ^^
I'm merging a bit early, as the feature was already merged in VexRiscv. I will patch the vjtagio

PR #953, opened 2022-11-09
ci: do not check fmt lint on dev

numero-744:


numero-744:
Wtf I have used git integrated in VSCodium and it introduced a commit without asking me :angry:… I'll fix using CLI, which is more reliable!

numero-744:
Fixed

numero-744:
Too see if it works, let's see if it passes here: https://github.com/numero-744/SpinalHDL/commit/6447af234191338256bb65982c4564644f254929

numero-744:
Ok it passed so @Dolu1990 you can merge this PR if it is okay for you.

Context: I noticed that the check was passing for PRs but not on branch `dev`. I have found that the error is the setup phase of the job which is fetching `dev` twice, so now on `dev` this specific job is skipped.

Dolu1990:
thanks ^^

Issue #956, opened 2022-11-11
Suggestion: Bus monitor direction


kleinai:
Does Spinal have a monitor bus type? It could be useful for in-system bus debug components.

Maybe something like this is sufficient?
```scala
object monitor extends MS {
  def apply[T <: IMasterSlave](i: T) = {
    i.asInstanceOf[Data].asInput()
    i.isMasterInterface = false
    i
  }
}
```

numero-744:
There is not (yet). Is it really useful in hardware description? Not just a simulation thing?

kleinai:
I've used Xilinx bus monitor IP in hardware

numero-744:
Oh, was it on a blackbox?

I can PR this feature quickly if others agree we should implement it (and document it later™)

Readon:
Is it an online monitor logic that inject into the design?

kleinai:
Like a stream monitor to capture data for debug.

I've made a PR #957

numero-744:
While reviewing the PR, I have found this line which triggered thinking:

[`def asMonitor(): Unit = this.asInstanceOf[Data].asInput()`](https://github.com/SpinalHDL/SpinalHDL/pull/957/files#diff-0d895d626825770c57d288b933d6deaf6eedbbf2f67105a2ca1bbed11b485457R74)

What more can we expect from this function? I mean, this set of `asXxx` functions is meant to be overridden with directions which are protocol-specific. Here I am not sure it is useful, maybe `monitor` should just set inputs.

Also [line 47](https://github.com/SpinalHDL/SpinalHDL/pull/957/files#diff-0d895d626825770c57d288b933d6deaf6eedbbf2f67105a2ca1bbed11b485457R47) it is defining itself as a slave, which is not really true.

So here is my opinion on this topic: maybe, instead of adding a word in the "bus" vocabulary, `in` should be used? It is still available, and seems to represent what you want. No need to PR and update documentation, just know that the solution already exists.

```scala
val io = new Bundle {
  val monitored = in port MyBus(arguments)
}
```

kleinai:
I rescind my PR. `in(...)` does work as intend. I thought for sure this wouldn't work in the past.

numero-744:
Anyway thanks for the PR @kleinai, it was useful to have concrete things to think about!

PR #957, opened 2022-11-11
MasterSlave: Add input-only monitor direction

kleinai:


lib/src/main/scala/spinal/lib/MasterSlave.scala
```diff
@@ -37,6 +42,11 @@ trait IMasterSlave {
     _isMasterInterface = Some(false)
   }
 
+  final def setAsMonitor(): Unit = {
+    asMonitor()
+    _isMasterInterface = Some(false)
```
numero-744, line 47: This is saying "I am a slave", so the types extending this trait can assume that. It could lead to unexpected behavior because it is not really a slave. Maybe it should not be declared as master nor slave because it isn't any of them?

```suggestion
```

lib/src/main/scala/spinal/lib/MasterSlave.scala
```diff
@@ -160,6 +172,14 @@ object slave extends MS {
   def applyIt[T <: IMasterSlave](i: T) = i.intoSlave()
 }
 
+/** Declare a monitor port
```
numero-744, line 175:
```suggestion
/** Declare a monitor port (all inputs)
```

lib/src/main/scala/spinal/lib/MasterSlave.scala
```diff
@@ -60,6 +70,8 @@ trait IMasterSlave {
     * This method does not update `isMasterInterface` and `isSlaveInterface`.
     */
   def asSlave(): Unit = intoMaster().asInstanceOf[Data].flip()
+
+  def asMonitor(): Unit = this.asInstanceOf[Data].asInput()
```
numero-744, line 74: See https://github.com/SpinalHDL/SpinalHDL/issues/956#issuecomment-1312164776

numero-744 (commented):
Actually I am questioning the need of this new notion instead of just using `in`. See https://github.com/SpinalHDL/SpinalHDL/issues/956#issuecomment-1312164776

PR #965, opened 2022-11-16
feat: SpinalSpec (draft)

numero-744:
# Context, Motivation & Description

It can be useful to have testbenches as a scalatests, but using `scalatest` as-is can be redundant for simulation tests (need to repeat the test name in `doSim` and `test`, need to define the `dut` in each test suite, etc.)

This draft PR shows that it possible to do things like:

```scala
// SpinalFunSuite is the new feature in this PR
// Identity is the Component to test

class SimpleIdentityTester extends SpinalFunSuite(new Identity) {
  // There is a default `SimConfig` but I can provide my one
  override val config = SimConfig.withWave

  // I list all my tests
  test("random tests") { dut =>
    for (i <- 0 to 99) {
      val v = dut.io.input.randomize()
      sleep(1)
      assert(dut.io.output.toInt == v)
    }
  }

  test("all tests") { dut =>
    for (v <- 0 to 255) {
      dut.io.input #= v
      sleep(1)
      assert(dut.io.output.toInt == v)
    }
  }
}
```

`test` mixes `test` from `scalatest` and `doSim` from SpinalHDL. Running these tests results in two wave files named after the tests.

It also brings a way to have preSim hooks for stuff we do not want to put in each test, like forking clock domain.

There is a variant which enables abstracting the Dut (for instance with #879).

Here is a dummy example on the same `Identity` component as above.

```scala
class AbstractIdentityTester extends SpinalFunSuiteTransform(new Identity) {
  type SoftDut = SoftIdentity

  def preSimTransform(dut: Identity): SoftIdentity = SoftIdentity(dut)

  override val config = SimConfig.withWave

  test("random tests") { dut =>
    for (i <- 0 to 99) {
      val v = dut.dut.io.input.randomizedInt()
      assert(dut(v) == v)
    }
  }

  test("all tests") { dut =>
    for (v <- 0 to 255)
      assert(dut(v) == v)
  }
}

case class SoftIdentity(dut: Identity) {
  import dut._

  def write(v: Int): Unit = io.input #= v
  def read(): Int = io.output.toInt
  def apply(v: Int): Int = {
    write(v)
    sleep(1)
    read()
  }
}
```

The goal of this draft PR is to discuss if we want such a feature or not.

The goal of the suggestion is not to re-implement all the `scalatest` variants of `Fun` or `Spec`, only to provide a helper to make it easier and more *DRY* to unit test components.

Issue #734, opened 2022-05-12
Wavedrom testbench: idea for a far future


numero-744:
Hello,

A feature idea for a far future:

Designers can use [Wavedrom](https://github.com/wavedrom/wavedrom) to [easily build waveforms](https://wavedrom.com/editor.html) with inputs and outputs of the component. The generated waveform can then be added to the component documentation. Using an editor such as wavedrom, the designer can see the waves as they type.

SpinalHDL could interpret the wavedrom JSON file to generate input stimuli in a testbench, and assert that the outputs are correct.

This way, the designer could easily check / unit test the component against the specification (driven testbench only), before sending the design to the verification team.

PR #968, opened 2022-11-19
feat: SpinalSpec

numero-744:
Continuation of #965 (thanks for your feedback @andreasWallner @Readon)

Scalatest is useful to quickly write unit / regression tests, but it is not really practical for hardware description testing:

* we have to repeat the test name with the Spinal test function
* the official documentation website is almost always down
* me trying to use always starts with: "where is the thing to extend? `import scalatest.funs…` no, `anyfuns…` Err, is it `FlatSuite`?"

This PR mixes Spinal's `doSim` and `doSimUntilVoid` functions with `scalatest` features to provide the user a unified `SpinalSpec`, which is a part of `spinal.core.sim._` so no need to import more things!

### What is a `SpinalSpec`?

A `SpinalSpec` contains at least a `Bench`, which is kind of a `SimCompiled` but with more context to give you neat `scalatest` reports.

```scala
class MyComponentSpec extends SpinalSpec[MyComponent] {
  val myComponent = bench(Identity(cfg))
}
```

The goal of a bench is to run tests, isn't it? Two ways to do so:

```scala
myComponent should "do something" in { dut =>
  // some code
}

myComponent.test("some test") { dut =>
  // some code
}
```

Just created 3 tests. Three? Yes, three:

* `myComponent compiles`
* `myComponent should do something`
* `myComponent test: some test`

So if compilation fails, the failure is reported, and tests on this testbench are cancelled. :sparkles: If it succeeds (why would it fail? :thinking:) it is not re-compiled by default (but it can be configured, see below).

### Configurable components

For a configurable component, I suggest to use `ProtocolCfg`

```scala
val testProtocol = ProtocolCfg((cfg: MyConfig) => bench(MyComponent(cfg))) { (cfg, it) =>
  it should "do stuff" in { dut =>
    // dam dam dee-dam
  }

  // Tests can be enabled conditionally over the current configuration
  if (cfg.size >= 8) {
    it should "lack of inspiration" in { dut => }
  }
}

val cfgs = for (n <- 5 to 7) yield MyConfig(n)
testProtocol.runWithAll(cfgs)
```

Oh, btw, if you just want to check that a configuration is valid or not, you can!

```scala
val goodBench = bench(MyComponent(goodCfg))
goodBench should compile

val badBench = bench(MyComponent(badCfg))
badBench shouldNot compile
```

### Configuration

You can configure:

* `config: SpinalSimConfig` (defaults to `SimConfig`) by overriding it in the spec or when calling `bench`
* `caching: Boolean` (defaults to `true`) to cache `SimCompiled` in a bench, adding a test for bench compilation. Setting it to false removes the test for compilation and re-compiles on each test, so if compilation fails, all tests on this bench fail. Configured by overriding it in the spec or when calling `bench`.
* `seed: Option[Int]` (defaults to `None`) by overriding it in the spec, when calling `bench` or on each test.

### Other features

More information in the file with unit tests.

# Checklist

- [x] Unit tests were added


andreasWallner (commented):
In general this seems nice to use. I also like the little touches, like that compiles the DUT lazily, which should speed of looped tests where many of the tests are filtered out.

The small thing that I'd add is to be able to specify a random seed via the scalatest command line, to reproduce a failure w/o changing the test (via `testOnly ... -- -Dseed=x` or similar - environment variable doesn't catch that since I wouldn't want to restart sbt for every seed change). (and yes, I'm aware of the irony that I'll contradict myself immediately ;-))

My major concern is that this feels like you tried to account for every possibility, rather than provide something for 90% of the cases that is easy to explain. If I want to mess with the used seed then I can: 
- change the environment variable for it
- override the seed member
- pass a seed to the `bench`
- use `withSeed` or pass a seed to `test`
- (if we'd add it pass the seed on the scalatest invocation)
This seems to complicated for such a simple thing, hard to document, explain and to provide help in e.g. the Gitter.
Another one I'd cut back is syntax variants of running a test: I'd choose either the FlatSpec style or the `test` function - no reason to have both?

I think you could should be able to drop the type parameter and simplify this a bit. On some other technical notes I'd try to move `Bench` and (some?) of the other classes out of the copy of the `Spec` to avoid confusion between `bench` and `Bench`

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+    *
+    * @param dut The device to test
+    * @param preSimHook An action to do on `dut` before each test (defaults to nothing)
+    * @param config To override `SpinalSpec.config`
+    * @param seed To override `SpinalSpec.seed`
+    * @param caching To override `SpinalSpec.caching`
+    */
+  def bench(
+      dut: => Dut,
+      preSimHook: Dut => Unit = { _ => },
```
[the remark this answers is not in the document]
andreasWallner: Ok, I was too quick there, preSimHook is called after compilation which makes a difference...
What would you do with the Dut post-compilation?

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,273 @@
+package spinal.core.sim
+
+import org.scalatest.freespec.AnyFreeSpec
+import spinal.core.{Component, Nameable}
+
+/** Define unit / regression tests for a component */
+abstract class SpinalSpec[Dut <: Component] extends AnyFreeSpec with spinal.idslplugin.ValCallback {
```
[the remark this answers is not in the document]
numero-744: Good catch! But now I have to provide 2 type parameters on Protocol (see test file)

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+  val config: SpinalSimConfig = SimConfig
+
+  /** `override val caching = false` to disable caching by default */
+  val caching: Boolean = true
+
+  /** `override val seed = Some(value)` to choose a fixed seed */
+  var seed: Option[Int] = None
+
+  /** Define a protocol that can be run on several benches */
+  case class Protocol[SoftDut](tests: Bench[SoftDut] => Unit) extends Nameable {
```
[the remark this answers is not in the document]
numero-744: Good catch

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+        for (bench <- benches) {
+          bench.getName - {
+            tests(bench)
+          }
+        }
+      }
+    }
+
+    /** Run the protocol on a sequence of benches */
+    def runOnAll(namedBenches: Seq[(Bench[SoftDut], String)]): Unit = {
```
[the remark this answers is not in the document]
numero-744: Renamed all to `runWith`/`runWithAll`

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
numero-744: `preSim` not `preCompile`

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+  override def valCallback[T](ref: T, name: String): T = {
+    ref match {
+      case nameable: Nameable => nameable.setName(name)
+      case _                  =>
+    }
+    ref
+  }
+
+  /** Override it to define the default SpinalSimConfig (else defaults to SimConfig) */
+  val config: SpinalSimConfig = SimConfig
```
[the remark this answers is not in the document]
numero-744: I though it was clear, like shadowing in nested scopes: the more local wins. Well, this is applicable to spec vs bench, but not to bench vs test (all-to-all relation ship, not all-in-one)…

So indeed, for `seed` the precedence is not clear.

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+
+      /** Run a simulation to check that it does the thing it should
+        *
+        * {{{
+        * it should "do something" in { dut =>
+        *   // code to check that it does the thing it should
+        * }
+        * }}}
+        */
+      def in: Body => Unit = sim(txt.itShould(doWhat), seed)
```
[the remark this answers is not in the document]
numero-744: I agree but I couldn't find a better one too…

(Hmm, what about `inSim`/`inSimUntilVoid`?)

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
numero-744: > What would you do with the Dut post-compilation?

I don't understand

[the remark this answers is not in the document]
numero-744: Ah, what for? For instance to fork clock stimulus, set simulation timeout, initialize inputs for t=0, wait first rising edge. All things you would do at the beginning of each test, but here you write them one.

numero-744:
About seed from scalatest command line, can you point me to an example of code getting this value from the Scala code? Never used that so far.

Actually I never set the seed myself, so I don't really know the use case. How do you think should we manage this? Not having the seed in the code at all would be great I think, so maybe just get the optional flag as you suggested, else let Spinal manage the seed with environment variables?

Why having both `should` and `test`? I would say `should` because it is a nice syntax and `test` because sometimes you don't want to express things with `should`. For instance, when I want to do debug tests on a spec with diagrams, I would do `test("Figure 5: The figure 5 title")` and mimic the diagram in the test. (We could even take a Wavedrom diagram but that's another topic #734 XD)

About `bench` vs `Bench`, well I would say it is usual `function` vs `Type` with conventional casing. Also `Bench` is not `case` so that users are less tempted XD Is it okay for you?

About moving stuff outside, I don't want to pollute the namespace, and as far as these are only used in `SpinalSpec`, I think it is better to let them there.

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
numero-744: I fixed it; the 2 type parameters were `Dut` (`HardDut`, the actual `Component` to compile) and `SoftDut` (possibly abstracted what tests are run on, the only thing used in the interface).

So I made `Bench` abstract, moved `Dut` in an internal type so that the only type which remains in the type signature is `SoftDut`, making it possible to have only one type parameter for `Protocol`.

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
numero-744: I updated names of bench's parameters and their documentation to make things more clear. For seed I think I'll remove it and manage it only with an `sbt` flag instead. I also renamed default values in the spec to make things more clear.

numero-744:
I found that, for `-- -Dseed=10`, just need to convert to integer, now :wink:  https://stackoverflow.com/questions/35433177/passing-additional-arguments-to-tests-with-scalatest

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
andreasWallner: Understood, I think I'd still prefer to call a setup function in each test case - might be more lines to write but it's more local and less places to see what a test is doing.

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
numero-744: For seed it is updated too, I just let a `defaultSeed` in the Spec, and override by command option is documented

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
andreasWallner: why not `sim` and `simUntilVoid` (if we want to stick with the `UntilVoid` postfix)?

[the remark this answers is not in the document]
numero-744: Hmm, shorter than `inSim` and `isSimUntilVoid` but the sentence is less `Spec`-like: "it should work sim" instead of "it should work in sim"… I don't know, what do you prefer?

andreasWallner:
> Actually I never set the seed myself, so I don't really know the use case.
My use case is just during debugging: I run with random seed until I start debugging an issue, then I fix it so that it doesn't e.g. just go away because of different timing/data. When that specific bug is fixed I run with random again.

> About moving stuff outside, I don't want to pollute the namespace, and as far as these are only used in `SpinalSpec`, I think it is better to let them there.

As far as I know, the usual way to do that in scala would be to put the classes in a companion object. That way they don't show up in intellisense (if you use such a thing) or the global namespace. They also don't get a closure of the surrounding class.

> I found that, for -- -Dseed=10, just need to convert to integer, now wink https://stackoverflow.com/questions/35433177/passing-additional-arguments-to-tests-with-scalatest

Sorry, I thought you might have seen it when I posted the link on your original PR. I also had to look for it for quite some time: https://github.com/andreasWallner/spinalStuff/blob/dbde3792e46be4230995291ec727e38fffbc31eb/src/test/scala/andreasWallner/SpinalFunSuite.scala#L36

> Why having both should and test?

Fair, I had thought it might avoid confusion to have 2 ways to do the exact same thing, with only minor wording diff.

numero-744:
I'm trying to move `Bench` to a companion object but there are errors with the `in` and `cancel` placed outside the `Spec` and I'm not sure I can fix that.

My bad, I hadn't noticed that part of your code the first time and I didn't think to look at it for `-D`.

About `test`, is `testUntilVoid` name okay?

andreasWallner:
> I'm trying to move `Bench` to a companion object but there are errors with the `in` and `cancel` placed outside the `Spec` and I'm not sure I can fix that.

I had a quick look in the webinterface and only found a single reference to `cancel`, what actually is it?

> My bad, I hadn't noticed that part of your code the first time and I didn't think to look at it for -D.

No worries, I did want to send you hunting though^^

> About test, is testUntilVoid name okay?

Seems ok

Btw. I noticed that the code in `sim` and `should` is pretty similar, just with minor differences, but I guess you'll move that to a common function anyways?

numero-744:
`cancel` is what I use to mark simulation tests as cancelled when the cached compilation test failed.

You are right, `it should compile` can be simplified, but not with sim as in this very function it would put two tests with the same name when caching is enabled, which is not allowed by ScalaTest. I simplified `it should compile` another way, forcing to build the cached bench, as the "compiles" test is already performed in this function.

With both this version and the previous one, it is not allowed to do `it should compile` for an already compiled and cached bench because it would imply running the same test twice. I'm forcing here it for two reasons:

* When the cache is not enabled, I still want this test
* When the cache is enabled, I prefer to make explicit the error of putting the same test twice rather than silently remove the second one.

numero-744:
Rebased on `dev` and cleaned history

@andreasWallner I think I have nothing left to modify, waiting for your approval :smiley:

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,276 @@
...
+  /** Seed can be set with `testOnly MyComponentSpec -- -Dseed=123456789` */
+  final override def beforeAll(configMap: ConfigMap): Unit = {
+    seed = configMap
+      .get("seed")
+      .map(_.asInstanceOf[String].toInt)
+      .orElse(defaultSeed)
+  }
+
+  /** Override it to define the default SpinalSimConfig (else defaults to SimConfig) */
+  val defaultConfig: SpinalSimConfig = SimConfig
```
numero-744, line 38: Should be def to avoid strange behavior because of interior mutability (for instance `bench(finalConfig = defaultConfig.withWave` => oops, enabled waves possibly everywhere.

```suggestion
  def defaultConfig: SpinalSimConfig = SimConfig
```

numero-744:
So should all overridable defaults be `def` then?

numero-744:
@andreasWallner can you review, please?

build.sbt
```diff
@@ -133,6 +135,7 @@ lazy val core = (project in file("core"))
     libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
     libraryDependencies += "com.github.scopt" %% "scopt" % "3.7.1",
     libraryDependencies += "com.lihaoyi" %% "sourcecode" % "0.2.7",
+    libraryDependencies += scalatest,
```
Readon, line 138: I am very sensitive on the dependency of the libraries recently.
In my opinion, the SpinalSpec is not for core, it's for sim.
the core mainly contains facilities that elaborate logic, so my favorite fundamental component Stream have been put to the lib.
So I the sim is focused on simulation where Spec can be layed out.


Btw, the existing Tester are all located in tester directory?
How about keep that consistent?

numero-744: IIUC `sim` is for internals, not intended to be imported at all by the user. Users import `core.sim._`, which is the place where I put `SpinalSpec` here.

Btw I prefer how @andreasWallner added this dependency, so I suggest to merge his PR https://github.com/SpinalHDL/SpinalHDL/pull/970 first, then I solve conflicts by just taking his `build.sbt`.

About where the tests are, it is more compliant with `sbt` structure (@andreasWallner [agrees with it](https://github.com/SpinalHDL/SpinalHDL/pull/970)) so that if I'm seeking tests for `SpinalSpec.scala`, I can easily find them. It could be moved to `tester` but it would be more messy to mix all tests without any structure. I think `tester` is good for integration tests, but here it is more local, it is unit testing.

kleinai: Good point, #970 isn't about sim so why does it rearrange the build libraries?

Dolu1990: Maybe scala-test should have been integrated to scala itself as it is a very very much used library, so is it a big issue to have it as a depedency of SpinalHDL ? @Readon

Readon: I saw some one is using uTest framework instead. 
IMO, If scalatest dependency is not propagated in published version, then it could be good.

numero-744: I need it to provide the feature: https://github.com/SpinalHDL/SpinalHDL/pull/968/files#diff-8958280689950ea4b1f86c34cbcfdc12116d668559bfc108dbd358e70785c6dfR3-R4

Readon: I have noticed that.
emm, if we agreed on instist using scalatest but others, it can be resonable to import that dependency.

numero-744: I want to help users writing tests for hardware without having them to deal with `scalatest`.

Do you think we should base that on another tool than `scalatest`? Or if we should base that on nothing and interact directly with lower-level testing stuff?

andreasWallner: @kleinai I had added scalatest to be able to be able to add test to the core library. I had expected that sbt would support test dependencies, but it that is not the case, e.g. looking at this: https://www.scala-sbt.org/1.x/docs/Testing.html#Basics

andreasWallner:
> Rebased on `dev` and cleaned history
> 
> @andreasWallner I think I have nothing left to modify, waiting for your approval 😃

Sorry for the delay, I was in a workshop the last few days, will have a look in the evening.

numero-744:
@andreasWallner Great! Also, please don't merge it now: the way you modified `build.sbt` is better so let's get your PR merged first :smiley:

build.sbt
andreasWallner: @kleinai I just understood that I could limit the scope of the dependency to tests only via adding `% test` - seen that before, not sure why I forgot. Will do that change in my PR, which means it's still a topic here.

numero-744: Ah so for this reason maybe the approach here with a `val` is better because it makes it possible to add `% test` to all projects other than `core`?

But as I need it as normal dependency for this PR, maybe it is okay too to make it as a normal dependency everywhere?

andreasWallner: @numero-744 there is still no reason to add it as a regular dependency to core though?

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
andreasWallner: Not sure on that: my original concern was that both functions run all the tests over the passed benches, but only one has "all" in it's name. And the difference is the names being overriden, which you couldn't tell from the names (not that you have to though, maybe just have two overrides with the same name?)

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
andreasWallner: also sounds ok to me

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
andreasWallner: Which is btw. also an (to me) interesting difference between `TestGroup` and `TestGroupCfg` - why does one allow for overriding the names and the other doesn´t?

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,276 @@
...
+
+      /** Run a simulation to check that it does the thing it should
+        *
+        * {{{
+        * it should "do something" inSimUntilVoid { dut =>
+        *   // code to check that it does the thing it should
+        * }
+        * }}}
+        */
+      def isSimUntilVoid: Body => Unit = simUntilVoid(txt.itShould(doWhat))
```
andreasWallner, line 247: why do `is...` here and `inSim` above and not e.g. `in...` both times?

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,276 @@
...
+    def should(doWhat: String): Should = new Should(doWhat)
+
+    /** `it should compile` runs a test to check that `it` compiles */
+    def should(c: Compile): Unit = prepareCachedBench(force = true)
+
+    /** `it should compile` runs a test to check that `it` does not compile */
+    def shouldNot(c: Compile): Unit = {
+      txt.notCompiles in {
+        if (caching && triedCompile) {
+          assert(cachedBench.isEmpty)
```
andreasWallner, line 260: maybe provide a more useful error message here, otherwise it would just state that an assertion in code that the user has not written has failed, same just below?

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,276 @@
...
+            tests(cfg, benchBuilder(cfg))
+          }
+        }
+      }
+    }
+  }
+
+  /** Create a test-bench
+    *
+    * @param dut The device to test
```
andreasWallner, line 107: strictly speaking its not a component/device that is passed here ;-) not sure it should be mentioned/explained though...

But that's a fun thing, didn't know that scala would delay evaluation like this here... I was initially a bit stumped by the `Identity(IdentityConfig(0))` example below and why it didn't blow up immediately, so it might be worth highlighting that somewhat in the documentation later.

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  /** Create a test-bench with an abstraction of the `dut`
+    *
+    * @param dut The device to test
+    * @param preSimTransform Function taking the `dut` and returning your abstracted dut
+    * @param finalConfig To override `SpinalSpec.config`
+    * @param finalCaching To override `SpinalSpec.caching`
+    */
+  def benchTransform[HardDut <: Component, SoftDut](
```
andreasWallner, line 128: Sorry for the long reply...

To use-case for this function to me initially seemed it was to allow writing kind of generic tests for two implementations that "do the same thing" but have different implementations, i.e. a parallel adder and a bitserial adder (just to explain what I mean).

Something like:
```
case class ParallelAdder() ... { val io = { val a = UInt(...), ... }}
case class SerialAdder() ... { val io = { val a = Bool(), ... }}
abstract class Adder { def doAdd(a: Int, b: Int): Int = ??? }
case class ParallelAdderWrapper(dut: ParallelAdder) extends Adder { 
  override def add(a: Int, b: Int): Int = { dut.io.a #= a; ...; sleep(1); dut.io.result.toInt }
}
case class SerialAdderWrapper(dut: SerialAdder) extends Adder {
  override def add(a: Int, b: Int): Int = {
    for(i <- a.range) {
      // set one bit here
      dut.clockDomain.waitSampling()
      // get one bit here
    }
    // return final result
}
def makeAdder(dut: SerialAdder): Adder = SerialAdderWrapper(dut)
def makeAdder(dut: ParallelAdder): Adder = ParallelAdderWrapper(dut)

...

benchTransform(?dut?, makeAdder) { it => 
  it should "add correctly" {
    assert(it.add(24, 18), 42)
  }
}
```
But that doesn't work since one still needs to pass the DUT... (kind of a DUT-independent TestGroup)
So I'm wondering what use case you had in mind, because currently it looks to me like a complicated way of adding a helper function that we could otherwise get by just making a function that gets the DUT as a parameter, or if we really wanted the syntax use an implicit class? You call it "software abstraction" in a few places, but I'm not sure what you exactly had in mind there.

PS: not saying it should do what I initially though... ;-)

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,276 @@
...
+        for (bench <- benches) {
+          bench.getName - {
+            tests(bench)
+          }
+        }
+      }
+    }
+
+    /** Run the tests on a sequence of benches */
+    def runWithAll(namedBenches: Seq[(Bench[SoftDut], String)]): Unit = {
```
andreasWallner, line 58: as commented in the old thread above^^

build.sbt
numero-744: This way the user would have to add `import spinal.lib._` to use it?

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
numero-744: The initial reason of the `All` suffix here is that Scala considers `T*` and `Seq[T]` to finally be the same type so it prevents overloading (but `T*` cannot be given a `Seq[T]`). Maybe using a `DummyImplicit` and not having the `All` suffix would be a better solution here?

Also `runWith` is variadic (it can take one or more benches) and `runWithAll` is not, it takes a single argument which has to be a user-defined sequence, hence "`runWithAll` elements of the given sequence".

---

For `TestGroup` vs `TestGroupCfg`:

`TestGroup` can iterate on benches which are `Nameable`.

* Its `runWith` is expected to take named (put in their own `val name = ...`) benches.
* Its `runWithAll` is expected to take a sequence of tests which can be generated, so without a `val name = ...` for each bench, so it has to take the names to print additionally.

`TestGroupCfg` can iterate on configurations which are user-defined types so they can't be expected to be `Nameable`; but configurations can be converted to nice text (ex: `IdentityCfg(5)`) by default thanks to `case class`. So there is not this named vs. not named difference.

Btw the user can overload the way the text is generated from each `config`.

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
numero-744: Typo XD

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
numero-744: IMHO and if I understand correctly, this distinction is a little bit advanced (for people beginning with SpinalHDL without a Scala background) Scala stuff from the `=> Dut` type (parameter by name); SpinalHDL makes things transparent for users so that they don't have to worry about this. Else it is pure Scala stuff so it does not have to be documented in SpinalHDL?

Maybe it is more about the delayed "lazy" compilation which makes the error pop when the user does not expect it? In this case maybe the error message is explicit enough? Idk… For instance in RTD it can be documented along with the added `it compiles` test, stating that the given expression is evaluated only during this test?

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
numero-744: The initial use-case of `benchTransform` is for things like (draft code, didn't check buses names XD):

```scala
case class Ahb2ApbBridge() extends Component {
  val io = new Bundle {
    val ahb = slave  port AhbLite(...)
    val apb = master port Apb3(...)
  }
  ...
}

case class SoftAhb2ApbBridge(dut: Ahb2ApbBridge) {
  val ahb = AhbDriver(dut.io.ahb)
  val apb = ApbSlaveSubstitute(dut.io.apb)
}

val bridge = benchTransform(Ahb2ApbBridge(), SoftAhb2ApbBridge)

bridge should "read" { dut =>
  val read = dut.ahb.Read(0xaaaa)
  dut.apb.setOnReads(addr => addr + 1)
  read.run()
  assert(read.value == 0xaaab)
  assert(read.waitCycles = ...)
}
```

Then in a test you have helpers to emulate a slave (with `ApbSlaveSubstitute`) and a master (with `AhbDriver`) so that you can write tests with a transaction-level approach (#879), for instance perform a read at given address with given attributes, await the answer, assert the read value and the number of inserted wait states, etc.

Tbh I'm wondering if the transformation (along with `preSimHook` stuff) should be performed on test groups instead of benches because:

* the hardware compiled is actually the same
* it puts the code run before tests closer to the code of the tests

The drawback would be that it would not be possible to use this feature outside of a test group.

@andreasWallner What do you think about that?

---

### Below I did not notice the two components didn't have the same interface, you can ignore it

Hmm when I was designing the API, I wanted to address a use-case which looks like yours. Below I try to transpose your use case into what I am addressing, tell me if it matches for you:

```scala
// only io interface, no logic here
abstract class Adder extends Component { val io = new Bundle { ... } }
// logic, both using io defined earlier
case class SerialAdder() extends Adder { ... }
case class ParallelAdder() extends Adder { ... }
// software abstraction which is an interface which does not depend on the component implementation
// the main use-case here is to define protocols to perform transaction-level tests
case class SoftAdder(dut: Adder) {
  def add(a: Int, b: Int): Int = ...
  ...
}

val serialAdder = benchTransform(SerialAdder(), SoftAdder)
val parallelAdder = benchTransform(ParallelAdder(), SoftAdder)

val myTestGroup = TestGroup[SoftAdder] { it =>
  it should "perform additions" inSim { dut =>
    assert(dut.add(24, 18), 42)
  }
}
```

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
andreasWallner: @numero-744 yes, I only meant the delayed/lazy compilation, no need to mention the mechanics

numero-744: Okay so I have added an item for that in https://github.com/SpinalHDL/SpinalDoc-RTD/issues/155

build.sbt
andreasWallner: @numero-744 yes, just like with `SpinalFormalFunSuite`

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
andreasWallner: Ah, ok - that makes more sense than what I'd though about. From the descriptions/names I always though about providing a software abstraction for the DUT (e.g. you call it `SoftAhb2ApbBridge` which kind of implies to me that it is a SW version of the bridge; similar with `SoftDut`, whereas the `Soft...` part actually contains part of the testbench).

In that form it makes sense to me, containing drivers, scoreboards, etc. But: that also means that all of those are then accessed via `dut. ...`, but those drivers and scoreboards are not part of the DUT, they are part of the testbench. In your example it reads nice, but in other cases this might range from weird (`dut.rxScoreboard.pushRef(...)`) to confusing (was it `dut.ahb...` or `dut.dut.ahb...`?)

Regarding location: I see why you'd want to move the stuff to a `TestGroup`, but that would also mean that w/o one you can't use it that way (Not sure that'd be a big problem). Come to think of it: For the use-case you described, you could just use the case-class w/o the `benchTransform` - just instantiate it in either a `test`, after the `bench` statement or in a `TestGroup`. That should work just the same, just with the dut and testbench being two separate objects? (and leaves all options open)

Small comment on naming: `Bench` might be easily misunderstood, I checked with some of my digital designer colleagues and they would expect a similar meaning to what Wikipedia mentions:
> The term "test bench" is used in digital design with a [hardware description language](https://en.wikipedia.org/wiki/Hardware_description_language) to describe the test code, which instantiates the DUT and runs the test.
Where only the first part is true here, not the second. (actually less true if you do the move ;-) )
But I see why you used it, and I don't have an alternative I'm really happy with...

andreasWallner:
Small comment on naming: `Bench` might be easily misunderstood, I checked with some of my digital designer colleagues and they would expect a similar meaning to what Wikipedia mentions:
> The term "test bench" is used in digital design with a [hardware description language](https://en.wikipedia.org/wiki/Hardware_description_language) to describe the test code, which instantiates the DUT and runs the test.

Where only the first part is true here, not the second. (actually less true if you do the move ;-) )
But I see why you used it, and I don't have an alternative I'm really happy with...

Dolu1990:
Let's not have scala-test in lib, but in tester ? add this utility in the tester/main and add tester in the releases as a library ?

numero-744:
@Dolu1990 So for the user who wants to use the feature in this PR it implies that:

* we release tester
* user has to add `import spinal.tester`?

build.sbt
Dolu1990: or let's move SpinalSpec and SpinalFormalFunSuite to the spinal tester module ? and add that module into the releases (after some cleaning of the tester main module XD)

Dolu1990:
@numero-744 Yes

Dolu1990:
Also, instead of SpinalSpec, what's about SpinalTest ?

build.sbt
andreasWallner: @Dolu1990 that came to mind yesterday as well, might make sense if we want to skip the dependency on scalatest - I don't think there are many SpinalHDL users that use `core` but not `lib`

Dolu1990: See PR : https://github.com/SpinalHDL/SpinalHDL/pull/980 which implement it.

Dolu1990:
PR about scala-test : 
https://github.com/SpinalHDL/SpinalHDL/pull/980

If you are good with #980 we can merge #980 first i think

numero-744:
@andreasWallner `benchTransform` contains things around the component (hence the initial name `Bench`) but if we move / delay transformation in user's code, maybe `bench` and `Bench` can become `dut`/`Dut`? Then `thisDut should "do something" inSim`, I think it is more consistent than previously :smiley: I'm going on that and I'll seek inspiration to have a nice API for the transformation stuff :thinking:

@Dolu1990 The current name is after `*Spec` from `scalatest` as it makes it possible to use the `it should` syntax, but maybe it is not a really good name… I think `SpinalTest` is too close to the notion of one single test. What do you think about `SpinalTestBench`, if renaming mentioned above is performed? @andreasWallner it would match the definition, isn't it?

I have set this PR as draft, I'll commit forward so that you can identify the new suggestions. I'll rebase when we agree on the result before removing draft status.

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
```diff
@@ -0,0 +1,276 @@
...
+  ) extends Nameable {
+
+    /** Override this function to define how configurations should appear in test results */
+    def cfgToString(cfg: Cfg): String = cfg.toString()
+
+    /** Run the tests with (a) component configuration(s) */
+    def runWith(cfgs: Cfg*): Unit = runWithAll(cfgs)
+
+    /** Run the tests with a sequnece of component configurations */
+    def runWithAll(cfgs: Seq[Cfg]): Unit = {
```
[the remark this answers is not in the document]
numero-744: I explained the current state in https://github.com/SpinalHDL/SpinalHDL/pull/968#discussion_r1032953416 but suggestions for improvements are still welcome!

> `runWith(seqOfCfgs:_*)`

What is it?

Dolu1990:
@numero-744 

> The current name is after *Spec from scalatest as it makes it possible to use the it should syntax, but maybe it is not a really good name… I think SpinalTest is too close to the notion of one single test. 

Good ^^

numero-744:
FYI I have tried turning on parallelism between tests, and it cancelled many tests; maybe because the test-bench cached compilation is not atomic. Also I don't know what Spinal simulation API supports.

Dolu1990:
About atomicity of the compilatoin cache, it is implemented, but maybe there is holes in it.

Do you have a way to more or less reproduce the issue ?

numero-744:
@Dolu1990 For now I think it is a `scalatest` configuration issue, not a Spinal issue: the tests are only registered for later run, and the thing is about controlling in which order they are run, having the `compiles` test to complete before starting the tests using the `Dut`.

numero-744:
FYI short example of use is available in [numero-744/Aes/tb/spinal/aes/AesTestBench.scala](https://github.com/numero-744/Aes/blob/main/tb/spinal/aes/AesTestBench.scala)

Btw I could use `.asHex` but I was missing [BU"x123"](https://github.com/SpinalHDL/SpinalHDL/pull/981) here XD

core/src/main/scala/spinal/core/sim/SpinalSpec.scala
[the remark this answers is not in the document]
numero-744: I resolve this conversation as the API has changed a lot since, feel free to re-open one on the updated code if needed :smiley:

build.sbt
numero-744: According to #980, I'll rebase and move my stuff to `tester` dependency without moving it from `core.sim` package once @andreasWallner validates the suggested updated API.

andreasWallner:
> @andreasWallner `benchTransform` contains things around the component (hence the initial name `Bench`) but if we move / delay transformation in user's code, maybe `bench` and `Bench` can become `dut`/`Dut`? Then `thisDut should "do something" inSim`, I think it is more consistent than previously 😃 I'm going on that and I'll seek inspiration to have a nice API for the transformation stuff 🤔

Not a big fan of DUT since that would be the Component that we are testing itself? What I could offer would be `Environment`, but looking at pictures like the one on https://www.chipverify.com/systemverilog/systemverilog-simple-testbench this is also not a good fit since none of the typical jargon has words for the build environment around it which `bench` is doing. Looking at your AES example though it seems to be a great fit there (and you use it already)

numero-744:
> Not a big fan of DUT since that would be the Component that we are testing itself?

I would say the component class name should be describing what it does, and only in the context of the testbench it becomes the "device under test" so we can say "the dut is a Dut": 

```scala
val dut: Dut[MyComponent] = Dut(MyComponent(cfg))
//  ^^^^^^^^ the dut is a Dut
// but for better test reports it is better to use a good name
val myComponent = Dut(MyComponent(cfg))

myComponent should "work" inSim { dut =>
  // hello
}
// => "myComponent should work" test passed
```

I had taken a look at diagrams like https://www.chipverify.com/systemverilog/systemverilog-simple-testbench to rename the the `SoftMyComponent` as `MyComponentEnv` (used with `myDut withEnv MyComponentEnv`)

numero-744:
Ready for final review and merge if you have time, else it can wait for `v1.8.1` :grin:

Readon:
About the directory, is it more clean that if we have 
SpinalTestBench to main/scala/spinal/tester/sim
SpinalFormalFunSuite to main/scala/spinal/tester/formal
Also, need to get corresponding package statements in those class modified. Maybe to "package spinal.tester.sim/formal"

Because the tester is a standalone package for the user.

numero-744:
I don't know for other items, but for `SpinalTestBench` having it in the same `import` clause as `spinal.core.sim._` can be really handy for the user, to avoid adding one more import line for each single file / component to test. With correct documentation I think it is not a big deal?

If there is a consensus it is really an issue, I'll move it to the `tester` package, and I can add a testing file template to the snippets so that it doesn't imply more typing for the (codium) user as well.

Also, still in the case we move it to `tester` package, I suggest to put this PR to standby the time I sort the tests in tester; it will reveal more testing utilities, then with all the testing utilities we can discuss the package structure of tester?

andreasWallner:
> I don't know for other items, but for `SpinalTestBench` having it in the same `import` clause as `spinal.core.sim._` can be really handy for the user,

To be honest I have no idea how well that plays with other JVM tools, at least for java it would be a no-no to not have the correct library name in the package name. I noticed that scala seems to care less about it.

Apart from that I thought a bit about whether the `Dut` class would be strictly needed, but I think its good the way it is

numero-744:
I don't know… If it is an issue with tools, then isn't having package `spinal.lib` in project `lib` instead of project `spinal` an issue, too?

andreasWallner:
> I don't know… If it is an issue with tools, then isn't having package `spinal.lib` in project `lib` instead of project `spinal` an issue, too?

Sadly I don't know - why not stick with it until some issue pops up?

Readon:
Maybe the library is too big, so mixing the core, simulation, and library together is not a good idea. 
Additionally, the simulation subproject only contains backends for various tools, which can also be put into the core. 
However, they have very clear targets, so it is a good idea not to tightly couple them to the core. 
IMO, It is reasonable to put them into a standalone library.

Readon:
> I don't know for other items, but for `SpinalTestBench` having it in the same `import` clause as `spinal.core.sim._` can be really handy for the user, to avoid adding one more import line for each single file / component to test. With correct documentation I think it is not a big deal?

Yes, the decision is really hard to make.
On one side, it would be easier to write the code to put SpinalTestBench into namespace `spinal.core.sim`.
On other side, name space sometimes is a clue for user to find the location of problem while facing an error. That would lead to a change of existing code, not user-friendly.

Also, as we discussed before, if we organize tester structure the same with things to be tested, it would be convenient, which support the method one.

Readon:
@numero-744 Is there any problem left for this PR?

numero-744:
@Readon before releasing, we should be more sure about where it should be. So it is about the structure of the `tester/main` project, which should be the main discussion in #1001

Readon:
> 

You have put it into the tester/main subproject, so that it is only required in user's test case. I think it's nice.
And the location of this is not related with the other testers which located in tester/test. 
So I am not very confirmed why #1001 is related.

numero-744:
To me it is about things like splitting `tester/main` into packages to separate formal vs non-formal or not.

The relationship with #1001 is that while moving tests, there were testing utilities that were common to different tested things, so I have put them into `tester/main` without more organization of this project into several packages. So I would discuss organization of `tester/main` in #1001 before adding more stuff into `tester/main`.

Readon:
> To me it is about things like splitting `tester/main` into packages to separate formal vs non-formal or not.
> 
> The relationship with #1001 is that while moving tests, there were testing utilities that were common to different tested things, so I have put them into `tester/main` without more organization of this project into several packages. So I would discuss organization of `tester/main` in #1001 before adding more stuff into `tester/main`.

Because the SpinalTestBench is in spinal.core.sim, so we can just use the path it is now as [tester/src/main/scala/spinal/core/sim/SpinalTestBench.scala](https://github.com/SpinalHDL/SpinalHDL/pull/968/files#diff-9ee775fecc2aa6c68a9bda6cd6e3868375c6ae110fe463b334369923aaf3fdbf). 
The only difference that would be is to move SpinalTestBench and SpinalFormalFuncSuite into the same spinal.tester package, which I think is not necessary now.

I think the PR is best to be decoupled, so that things would be simplified.

numero-744:
Also the `scalatest` dependency (needed for this feature) has been removed from `spinal.core/main` so it wouldn't work if we merge it as-is.

PR #969, opened 2022-11-20
Add support for mill build.

Readon:
Closes #878, partially.

Support build SpinalHDL by Mill, it could compile/test/publishLocal the existing modules.
Build through mill can be much faster than sbt, and sbt would be really unavailable sometimes due to network problem.

#### compile library
mill __.compile

#### test library
mill __.test

#### run individual app to play
mill tester.runMain spinal.xxxxx.xxxxx

#### publish to local, work as dev version
mill __.publishLocal

This might be the first step to create a standalone working environment on Windows.
The second step has been done already by generating MinGW64 Yosys and SymbiYosys packages on MSYS2.
The last step is to fix the cocotb testing library on the environment.

# Impact on code generation

No.

# Checklist

- [ ] Unit tests were added


project/Version.sc
```diff
@@ -0,0 +1,17 @@
...
+  private val major = "1.7.2"
+  val all         = if(isDev) "dev" else s"$major$snapshot"
+  val sim         = all
+  val core        = all
+  val lib         = all
+  val ip          = all
+  val debugger    = all
+  val demo        = all
+  val tester      = all
+}
```
kleinai, line 17: Isn't this just a copy of project/Version.scala?

Readon: > Isn't this just a copy of project/Version.scala?

Yes, it's not done yet. I want to use a symbol link instead. 
However, it is not that friendly to Windows. I am finding the way.

kleinai: Can it not load the existing Version file as-is?

Readon: > Can it not load the existing Version file as-is?

To my knowledge, the mill would load only file with extension 'sc' through it's "import $file" syntax.

Readon:
I have done some [test](https://github.com/Readon/SpinalHDL/actions/runs/3511409104) by GitHub action.
It shows that the full test cost only 1 hour and 6 minus, which is much faster than sbt's 1 hour (simulation) plus 12 minutes (formal verification).
image

Just an experiment for the curious.

kleinai:
While I don't have a problem with the symlink usually, I do primarily work on windows which could be an issue. I'll have to test it later.

Readon:
> While I don't have a problem with the symlink usually, I do primarily work on windows which could be an issue. I'll have to test it later.

Can you test this in your environment? I need to confirm it before merging.

kleinai:
It does work on Windows with the cloned `Version.sc`.  I had to do `mill mill.scalalib.GenIdea/idea` and reboot my IDE to get it to play nice with the `build.sc`, and set Ammonite mode.

I'll agree the mill build file is a lot shorter and to the point. Sadly Intellij doesn't yet support it out of the box.

I defer to @Dolu1990 for the final say. Two build systems could get annoying to keep in lock-step.

Readon:
> I defer to @Dolu1990 for the final say. Two build systems could get annoying to keep in lock-step.

I think it might be a plan-B only for developer who want a faster compiling and test while debugging.
So it's not necessarily pay much attention if the symlink problem of Version.sc could be solved.

kleinai:
Fair point

PR #970, opened 2022-11-20
Fix xfix mux

### LINKED ISSUE #947: Some questions about `SFix` bitwidth adaptation
The `accumulator`, `weight` and `activation` are all defined as `SFix(peak = 8 exp, width = 16 bits)`.
1. Why does the following:
```scala
accumulator := accumulator + (io.weight * io.activation).truncated
```
throw the following error:
```
[error] toplevel/accumulator : SFix[peak=2^8 resolution=2^-7] can't be assigned by toplevel/??? : SFix[peak=2^17 resolution=2^-14] because of truncation. You can do x := y.truncated if that's fine.
```
...but this(notice the position of parenthesis):
```scala
accumulator := (accumulator + io.weight * io.activation).truncated
```
is fine?
2. Why does
```scala
accumulator := io.clear ? operandType().getZero | (accumulator + io.weight * io.activation).truncated
```
throws
```
[error] WIDTH MISMATCH (16 bits <- 32 bits) on (toplevel/??? :  SInt[16 bits]) := (Bool ? Bits | Bits)[32 bits]
```
?

andreasWallner:
Closes #947

Fixes multiplexers for both UFix and SFix, and adds truncation to a given format.

# Checklist

- [x] Unit tests were added
- RTD PR coming up


numero-744:
The last merged PR has a conflict to solve XD (joking, not a big deal)

core/src/main/scala/spinal/core/FixedPoint.scala
```diff
@@ -247,6 +248,17 @@ abstract class XFix[T <: XFix[T, R], R <: BitVector with Num[R]](val maxExp: Int
...
+  def truncated(maxExp: ExpNumber, bitCount: BitCount): T = {
+    val t = fixFactory(maxExp, bitCount)
+    t := this.truncated.asInstanceOf[T]
+    t
+  }
+  def truncated(maxExp: ExpNumber, resolution: ExpNumber): T = {
+    val t = fixFactory(maxExp, resolution)
+    t := this.truncated.asInstanceOf[T]
+    t
+  }
```
numero-744, line 260: Off-topic: Here I think Scala really lacks the ability to implement a trait for a type after the type definition (AFAIK). It would make it possible to have something like:

```scala
trait FixRes
implement FixRes for BitCount
implement FixRes for ExpNumber
```

Then you can have only one function to define, document and implement; accepting the two types; and you do not duplicate logic until there is an actual need to access the content of this second argument (using a `case`).

andreasWallner:
If the PR is ok as-is, I'd say we can merge this.

numero-744:
LGTM but I have zero knowledge about this part of Spinal so I can't really review it…

As it is a bug fix, I'll merge it tomorrow afternoon/evening if no changes are requested until then. If there is an issue we can blame previous authors for not exhaustive enough regression tests :sweat_smile: 

I'll merge https://github.com/SpinalHDL/SpinalDoc-RTD/pull/154 just after that when you remove its draft status.

core/src/main/scala/spinal/core/FixedPoint.scala
```diff
@@ -267,6 +279,18 @@ abstract class XFix[T <: XFix[T, R], R <: BitVector with Num[R]](val maxExp: Int
       case _ => SpinalError("Undefined assignment")
     }
   }
+
+  override def getMuxType[M <: Data](list: TraversableOnce[M]): HardType[M] =
+    HardType(fixFactory(maxExp, bitCount).asInstanceOf[M])
```
Dolu1990, line 284: I think that is wrong. It should provide the hardware type able to store every values from the list parameter.
For instance, if you have List(Q8.4, Q12.2) then it should return Q12.4

Dolu1990: And also, eventualy, if one of the input has the truncated flag, then that input width can be filtred out from the logic.

andreasWallner: Now that you say it - makes sense since XFix automatically upcasts - I'll change that.
Doesn't AFix have a similar problem (for different physical widths)?

kleinai:
@andreasWallner are the build.sbt changes required for this PR?

Dolu1990:
> @andreasWallner are the build.sbt changes required for this PR?

For me that's ok, waiting on readon to see if that's ok too.

kleinai:
> > @andreasWallner are the build.sbt changes required for this PR?
> 
> For me that's ok, waiting on readon to see if that's ok too.

It's not functionally a problem, but lumping extra changes on unrelated PRs makes it hard to have the correct discussions about those changes

kleinai:
@andreasWallner I see now your tests here are in core so you will be needing the scala test dependency in core

andreasWallner:
@kleinai yes, but you are correct that I can limit the dependency to tests only, and I'll do so

andreasWallner:
@Dolu1990 Thanks pointing out my oversight, should be fixed now. Interestingly it seems we now break the intellisense engine of IntelliJ which can't figure out that `val muxed2 = io.choose ? io.uf2e_10b.truncated | io.uf2e_8b` is actually valid code... (I guess I'll open open a bugreport there...)

I coded it now such that a multiplexer never looses any bits of non-truncated inputs.

numero-744:
@andreasWallner In `build.sbt` I don't understand why `scalatest` is as a `test` dependency in `core` and as a regular dependency in `lib`.

Also could it be `val scalatest = "org.scalatest" %% "scalatest" % "3.2.5"` then used as `scalatest % "test"`?

andreasWallner:
@numero-744 The reason was to not force scalatest as a dependency on the user if not necessary, it could e.g. clash with a setup where the user is using an incompatible version of scalatest in their own code. And in `core` we are currently only using scalatest in the tests.
In `lib` it has to be "normal" dependency since in the normal library code the formal suite class derives from `AnyFunSuite`.

Yes, that syntax would also work.

Dolu1990:
I think, what we should do about scala test is to have for all test module, and add it  in tester/main. Cleanup tester/main module with only the usefull code, include all the scala test related tools and include it into releases.
remove scala test from spinal.lib, move the lib SpinalFormalFunSuite into tester.

Dolu1990:
Thanks for the PR ^^

Issue #948, opened 2022-11-05
Test seed not being randomized anymore is confusing


andreasWallner:
In this change: https://github.com/SpinalHDL/SpinalHDL/commit/1e94dd2efd95e36036ec9f1789e9d6d11442e57f, because of #657 you changed the randomization behavior in the simulation.
To me this is quite a misleading behavior, I (and some colleagues of mine) assumed that tests are randomized if I don't pass a fixed seed to them. We expected a call to `randomize` would randomize by default, 
It also makes randomized tests in e.g. intellij much more inconvenient, since I can't use the "Play" button beside a test anymore to run it with random seed, except if I started intellij from a command line with the variable set (or it being set globally) - and not from the start menu. It's similar with VSCode & Metals. If one needs a fixed seed it's much easier by just changing the `seed` parameter.

Please reconsider the change, to me it feels like a regression.

If not, we should document the behavior in RTD, the sim environment variables are currently not mentioned there. We might also want to change the SpinalHDL github actions to set this environment variable?

numero-744:
Oooh I did not notice the change. What exactly is the changed behavior?

Does it imply initial values of signals? Does it imply values chosen with `Random` and `randomize()` API?

Side note: pushing on `dev` or `master` does not seem a nice thing, doing PRs is good to discuss this kind of things ^^'

kleinai:
Yes this behavior should be rolled back

Dolu1990:
So, let's change 

```scala
def doSimApi(name: String = "test", seed: Int = sys.env.getOrElse("SPINAL_SIM_SEED_RANDOM", "0").toInt*Random.nextInt(2000000000),
```

Into : 
```scala
  def newSeed(): Unit ={
    sys.env.get("SPINAL_SIM_SEED") match {
      case Some(v) => v.toInt
      case None => sys.env.getOrElse("SPINAL_SIM_SEED_RANDOM", "1") match {
        case "1" => Random.nextInt(2000000000)
        case "0" => 0
      }
    }
  }
def doSimApi(name: String = "test", seed: Int = newSeed(), ...)
```

?

That way we provide a way to fix the seed easily without code modification. Also if a regression fail, you can fix the seed via environnement variable

andreasWallner:
I'm not sure I understand, that would mean:
- setting the seed parameter overrides everything else
- setting SPINAL_SIM_SEED in the terminal allows to fix the seed to a certain value
[...]

PR #971, opened 2022-11-20
Make SpinalSim seed default random again

andreasWallner:
Closes https://github.com/SpinalHDL/SpinalHDL/issues/948

This changes the default behavior for simulation back to random.
As discussed in the Bug, it only uses a single environment variable (SPINAL_SIM_SEED) which can be set to any fixed value.

This will change the behavior for people that are currently using SPINAL_SIM_SEED_RANDOM, but prevent future confusion as agreed in https://github.com/SpinalHDL/SpinalHDL/issues/948#issuecomment-1309904827.

RTD note on this will be added, just not sure where to add it ATM - I think that needs a new page...

### DIFF
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -571,7 +571,7 @@ abstract class SimCompiled[T <: Component](val report: SpinalReport[T]){
     }
   }
 
-  def doSimApi(name: String="test", seed: Int=newSeed(), joinAll: Boolean)(body: T => Unit): Unit = {
+  def doSimApi(name: String = "test", seed: Int = newSeed(), joinAll: Boolean)(body: T => Unit): Unit = {
     Random.setSeed(seed)
     GlobalData.set(report.globalData)
 
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -564,7 +564,14 @@ abstract class SimCompiled[T <: Component](val report: SpinalReport[T]){
 
   def newSimRaw(name: String, seed: Int) : SimRaw
 
-  def doSimApi(name: String = "test", seed: Int = sys.env.getOrElse("SPINAL_SIM_SEED_RANDOM", "0").toInt*Random.nextInt(2000000000), joinAll: Boolean)(body: T => Unit): Unit = {
+  def newSeed(): Int = {
+    sys.env.get("SPINAL_SIM_SEED") match {
+      case Some(v) => v.toInt
+      case None => Random.nextInt(2000000000)
+    }
+  }
+
+  def doSimApi(name: String="test", seed: Int=newSeed(), joinAll: Boolean)(body: T => Unit): Unit = {
     Random.setSeed(seed)
     GlobalData.set(report.globalData)
 


numero-744:
For documentation, maybe a part can be added at the end of https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Simulation/bootstraps.html

core/src/main/scala/spinal/core/sim/SimBootstraps.scala
```diff
@@ -564,7 +564,14 @@ abstract class SimCompiled[T <: Component](val report: SpinalReport[T]){
...
 
-  def doSimApi(name: String = "test", seed: Int = sys.env.getOrElse("SPINAL_SIM_SEED_RANDOM", "0").toInt*Random.nextInt(2000000000), joinAll: Boolean)(body: T => Unit): Unit = {
+  def newSeed(): Int = {
+    sys.env.get("SPINAL_SIM_SEED") match {
+      case Some(v) => v.toInt
+      case None => Random.nextInt(2000000000)
+    }
+  }
+
+  def doSimApi(name: String="test", seed: Int=newSeed(), joinAll: Boolean)(body: T => Unit): Unit = {
```
numero-744, line 574: Style suggestion:

```suggestion
  def doSimApi(name: String = "test", seed: Int = newSeed(), joinAll: Boolean)(body: T => Unit): Unit = {
```

numero-744 (approved):
Optional suggestion, I also approve as-is

numero-744:
Also for docs, if https://github.com/SpinalHDL/SpinalDoc-RTD/issues/151 is approved it would be moved to "10. iii. Random tests".

core/src/main/scala/spinal/core/sim/SimBootstraps.scala
```diff
@@ -564,7 +564,14 @@ abstract class SimCompiled[T <: Component](val report: SpinalReport[T]){
 
   def newSimRaw(name: String, seed: Int) : SimRaw
 
-  def doSimApi(name: String = "test", seed: Int = sys.env.getOrElse("SPINAL_SIM_SEED_RANDOM", "0").toInt*Random.nextInt(2000000000), joinAll: Boolean)(body: T => Unit): Unit = {
+  def newSeed(): Int = {
+    sys.env.get("SPINAL_SIM_SEED") match {
+      case Some(v) => v.toInt
+      case None => Random.nextInt(2000000000)
```
Readon, line 570: I am always curious on each const value, how this constant value is derived? And why it should be a const?
Although, it is from the legacy code here. :)

core/src/main/scala/spinal/core/sim/SimBootstraps.scala
andreasWallner: Had the wrong formatting config set up in intellij... formatted with the scalafmt now

core/src/main/scala/spinal/core/sim/SimBootstraps.scala
andreasWallner: same here, no idea why that number so I kept it ;-)

Dolu1990: I just didn't wanted a negative number, that's why i set it to something arbitrary "big".

Dolu1990:
Thanks ^^

core/src/main/scala/spinal/core/sim/SimBootstraps.scala
numero-744: Ah then do this?

```scala
Random.nextInt(Integer.MAX_VALUE);
```

Dolu1990: Yes you are right ^^

Dolu1990: I pushed Integer.MAX_VALUE in the repo.

PR #979, opened 2022-11-27
fix: SimBigIntPimper toBytes

### LINKED ISSUE #978: `BigInt().toBytes(...)` is inconvenient to use/broken(?) for positive numbers with "sign bit" set
When using the `toBytes` method coming from `SimBigIntPimper` the following behavior surprised me:

``` scala
// fails with assertion due to internal length check
// assertion failed: Original BigInt has more bytes then bits specified.
BigInt(0xFF).toBytes(8, endian=BIG) 
// returns [0x00, 0xff]
BigInt(0xFFFFFFFFL).toBytes(endian=BIG)
```
Due to the sign bit being output when calling `BigInt.toByteArray`.

The same issue occurs in the usage locations of `toBytes` in the library, e.g.:
```scala
class Playground1 extends AnyFunSuite {
  case class test01() extends Component {
    val io = new Bundle {
      val u8 = in UInt (8 bit)
    }
  }
  test(s"blub") {
    SimConfig.compile(test01()).doSim("blub") { dut =>
      dut.io.u8 #= 0xff
      sleep(10)
      println(dut.io.u8.toBytes mkString(" "))
    }
  }
}
```
Aborts the simulation because the same assertion as mentioned above fails.
A similar case can be constructed for the use of it in the `SimBitVectorPimper`.

From the usages I'd say that it looks like we wanted the method to get the bit as they are. In that sense I'd "fix" this by ignoring an initial 0 byte if it is there.
That might lead to some surprises when doing something like (just to illustrate) `S(x, 9 bit).toBigInt().toBytes(8)` which would work as long as the number is positive, get an error if negative. But that is no so different from `U(x, 32 bit).toBigInt().toBytes(8)` which will work as long as the number is small enough.
It would also mean that as long as numbers are positive we'd remove a "leading sign bit 0", but this is actually the behavior that we need for e.g. the simulation example above...

Any thoughts? (I'll open a PR with test that show the implications)

While on the topic: Another thing that comes to mind looking at the code is that while we pass a length in bits to the method, it will internally round up to full bytes which seems a bit inconsistent. It might make sense to check for the actually needed length in bits?

andreasWallner:
Fix zero sign bit being reserved, leading to errors when converting values with MSB of high byte being set (adding an unexpected byte)

Fixes #978

I kind of dislike the helper function in the Unittests, but w/o it they become quite unreadable IMHO...

# Checklist

- [x] Unit tests were added
- Simulation helper are currently undocumented

### DIFF
--- a/core/src/main/scala/spinal/core/sim/package.scala
+++ b/core/src/main/scala/spinal/core/sim/package.scala
@@ -568,15 +568,25 @@ package object sim {
     * Add implicit function to BigInt
     */
   implicit class SimBigIntPimper(x: BigInt) {
+    /**
+      * Convert value to array of bytes
+      *
+      * Checks if a value fit into the given number of bits, but does not reserve a bit for
+      * a possible sign bit if the BigInt is positive.
+      * Raises an error if a specific BigInt value doesn't fit into the given number of bits.
+      * If the value is negative, the sign bit is extended to the given number of bits (if passed)
+      * otherwise to the least number of bytes necessary.
+      */
     def toBytes(bits: Int = -1, endian: Endianness = LITTLE): Array[Byte] = {
-      val raw = x.toByteArray
-      val byteCount = if (bits < 0) raw.length else (bits + 7) / 8
-      assert(raw.length <= byteCount, "Original BigInt has more bytes then bits specified.")
-      val out = Array.fill[Byte](byteCount) { 0 }
-      for (i <- 0 until byteCount) {
-        out(i) = ((x >> i * 8) & 0xff).toByte
+      val requiredLen = x.bitLength + (if(x < 0) 1 else 0)
+      assert(bits < 0 || bits >= requiredLen, "Original BigInt has more bytes then bits specified.")
+      // use full bytes if not instructed otherwise
+      val outLen = if(bits < 0) (requiredLen + 7) & (-8) else bits
+      val out = for (shift <- 0 until outLen by 8) yield {
+        val mask = if(outLen - 8 < shift) 0xff >> (8 - (outLen - shift)) else 0xff
+        ((x >> shift) & mask).toByte
       }
-      if (endian == BIG) out.reverse else out
+      if (endian == BIG) out.reverse.toArray else out.toArray
     }
   }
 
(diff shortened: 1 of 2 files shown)


Dolu1990:
I think all the changes to build.sbt could be reverted, just need to merge upstream spinalHDL which now add scala-test by default.

Else all good i think ^^

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -569,14 +569,15 @@ package object sim {
...
-      val raw = x.toByteArray
-      val byteCount = if (bits < 0) raw.length else (bits + 7) / 8
-      assert(raw.length <= byteCount, "Original BigInt has more bytes then bits specified.")
-      val out = Array.fill[Byte](byteCount) { 0 }
-      for (i <- 0 until byteCount) {
-        out(i) = ((x >> i * 8) & 0xff).toByte
-      }
-      if (endian == BIG) out.reverse else out
+      val requiredLen = x.bitLength + (if(x < 0) 1 else 0)
+      assert(bits < 0 || bits >= requiredLen, "Original BigInt has more bytes then bits specified.")
```
Readon, line 573: The condition bits < 0 would lead the default bits = -1 be asserted as fail always.

andreasWallner: I don´t think so, it asserts that either `bits < 0` (i.e. default - do it automatically) or that bits is large enough for the data as to not cut off data?

Readon:
> > I think all the changes to build.sbt could be reverted, just need to merge upstream spinalHDL which now add scala-test by default.
> 
> you are totally right, when I opened this the PR with the scatatest dependency was not yet merged - removed that stuff.

The only thing I still concern is about the test directory structure as describe in #982. 
What you do is split the original function of the tester subproject into each subproject's test director, as #947 do.
Keep the old way, put all test in tester subproject, or split them into each directory is a question. Although I prefer the last one personally.

Readon:
One more thing, would you please add more tests on integers with more bytes? Like 4 and 5 bytes?

andreasWallner:
@Readon Would you have another look, should be all fixed now.

PR #980, opened 2022-11-28
build: clean and add tester module to releases, move SpinalFormalFunSuite in it

Dolu1990:

Avoid adding scalatest depencies in core / lib

Note, that it change the package of SpinalFormalFunSuite into spinal.tester.formal.SpinalFormalFunSuite

Maybe should be spinal.lib.formal.SpinalFormalFunSuite to avoid complexifying things XD ?


tester/src/test/scala/spinal/tester/code/Debug.scala
```diff
@@ -189,7 +189,26 @@ object Debug2 extends App{
...
+      is(M"1----") { io.o := 4 }
+      default{ io.o.assignDontCare()}
+    }
+
+//    val axis_s =  Stream(Fragment(Bits(32 bits)))
+//    val x = axis_s.payload
+//    println(x.getBitsWidth)
+//    println(widthOf(axis_s.payload))
+//    println(widthOf(axis_s.payload.asBits))
+//    val x = (in(UInt(4 bits)) === 42)
```
numero-744, line 211: Why?

numero-744:
> Maybe should be spinal.lib.formal.SpinalFormalFunSuite to avoid complexifying things XD ?

I'm okay with having the dependency being able to add things in packages of other dependencies if it is clearly documented so I approve that. `SpinalFormalFunSuite` [is undocumented](https://spinalhdl.github.io/SpinalDoc-RTD/master/search.html?q=SpinalFormalFunSuite&check_keywords=yes&area=default) btw so nothing to add yet? XD

tester/src/test/scala/spinal/tester/code/Debug.scala
Dolu1990: Ahhh this is some dirty worksheets ^^

numero-744:
I have pushed some commits, feel free to reset if there is an issue.

Dolu1990:
> btw so nothing to add yet? XD 

Ahh ?

numero-744:
> Ahh ?

$\forall doc \in \\{$ `SpinalFormalFunSuite` documentations $\\}, tester\ dependency \in doc$

is true because

$\\{$ `SpinalFormalFunSuite` documentations $\\} = \emptyset$

Dolu1990:
Should we invoke @Readon for some doc ?
Or mostly SpinalFormalFunSuite is a "private" tool for now ?

numero-744:
As you prefer, I'm not a user of this feature so I couldn't tell.

Readon:
> Should we invoke @Readon for some doc ? Or mostly SpinalFormalFunSuite is a "private" tool for now ?

I will try to complete that later. :)
It just add a tag to formal related test, so that test of spinalhdl library itself by formal verification could be done seperately.
that's the way to support formal job in CI.

However, I think it's nice to move it to tester subproject to avoid scalatest dependency.

Readon:
On the other hand, it can be distributed into the tester module.
Users can build their formal verification only test by this.

Readon:
It seems that this PR has not changing the location of SpinalForamlFunSuite.

numero-744:
It changed the location of the file in `tester/src/main/scala/spinal/lib/formal/SpinalFormalFunSuite.scala` so that it is included only if the `tester` dependency has been added, but the scala path has not been changed to keep current imports working.

Readon:
> It changed the location of the file in `tester/src/main/scala/spinal/lib/formal/SpinalFormalFunSuite.scala` so that it is included only if the `tester` dependency has been added, but the scala path has not been changed to keep current imports working.

Ahh, get it, the similar paths could make it harder to check.

Issue #783, opened 2022-06-22
Possibly add HDL comment to back trace Scala code line:column?


kellenwang1017:
Hi, We noticed that in Chisel it is possible to switch on appendix comment in HDL with annotations of corresponding origional Scala line:column information, syntax like this:

if(a) begin // @[Foo.scala 110:2]
  b <= 1'b1 // @[Foo.scala 111:4]
end

would it be possible to add similar features in SpinalHDL? We think it may help with bug back tracing in practice.

Dolu1990:
So about // @[Foo.scala 111:4]

Basicaly, it can be usefull, but it break down realy fast
as soon things are encapsulated into functions which can be called from multiple place it isn't so usefull
For instance :
myStream.stage()

All the things created into the stage function will be Stream.scala xxx.y which isn't very usefull, you would need kind of the full stack trace XD

that's why in some ways, SpinalHDL focus more on name propagation

Overall, it could be implemented without much paine, there is already somthing kind of similar (see loc : https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/when.scala#L78) , but it isn't a silver bullet as far as i can see.

PR #985, opened 2022-12-01
Add Vhdl/Verilog assignements comments to traceback to the original Scala code

Dolu1990:

Add Vhdl/Verilog assignements comments to traceback to the original Scala code : 

```scala
    val x = True
    val y = False
    when(True){
      y := True
    }

    val a,b,c = UInt(8 bits)
    c := a + b + 1

    val src = in Bool()
    val dst = Bool()

    src <> dst
```

```verilog
module miaou (
  input               src
);

  wire       [7:0]    _zz_c;
  wire                x;
  reg                 y;
  wire                when_Debug_l193;
  wire       [7:0]    a;
  wire       [7:0]    b;
  wire       [7:0]    c;
  wire                dst;

  assign _zz_c = (a + b);
  assign x = 1'b1; // Debug.scala l191
  always @(*) begin
    y = 1'b0; // Debug.scala l192
    if(when_Debug_l193) begin
      y = 1'b1; // Debug.scala l194
    end
  end

  assign when_Debug_l193 = 1'b1; // Debug.scala l193
  assign c = (_zz_c + 8'h01); // Debug.scala l198
  assign dst = src; // Debug.scala l203

endmodule
```

# Impact on code generation

Lot of comments everywhere on my screen

### DIFF
--- a/core/src/main/scala/spinal/core/AFix.scala
+++ b/core/src/main/scala/spinal/core/AFix.scala
@@ -1,5 +1,7 @@
 package spinal.core
 
+import spinal.idslplugin.Location
+
 import scala.collection.mutable.ArrayBuffer
 import scala.math.BigDecimal.RoundingMode
 
@@ -1096,7 +1098,7 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
   def rounded(rounding  : RoundType): AFix = truncated(saturation = false, overflow = false, rounding = rounding)
   def rounded: AFix = this.rounded(getTrunc.rounding)
 
-  override private[core] def assignFromImpl(that: AnyRef, target: AnyRef, kind: AnyRef): Unit = {
+  override private[core] def assignFromImpl(that: AnyRef, target: AnyRef, kind: AnyRef)(implicit loc: Location): Unit = {
     that match {
       case af: AFix =>
         val trunc = af.getTag(classOf[TagAFixTruncated])
--- a/core/src/main/scala/spinal/core/BaseType.scala
+++ b/core/src/main/scala/spinal/core/BaseType.scala
@@ -21,6 +21,8 @@
 package spinal.core
 
 import spinal.core.internals._
+import spinal.idslplugin.Location
+
 import scala.collection.Seq
 import scala.collection.mutable.ArrayBuffer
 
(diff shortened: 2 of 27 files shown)


numero-744:
What do you think about using the `relative/path:line` syntax, used by many compilers?

Also, there is no comment for `assign _zz_c = (a + b);` :thinking:

Dolu1990:
> Then, can't do a+b+c anymore XD

Ahh no not exacly.

The bug i have is more tricky
Syntaxes as : `(io.input.cmd.address + io.input.cmd.length) (selRange)` do not work anymore.
It tell me that 
```
type mismatch;
 found   : scala.collection.immutable.Range.Inclusive
 required: spinal.idslplugin.Location
      (io.input.cmd.address + io.input.cmd.length) (3 downto 0)

```

That's like if the brackets around (io.input.cmd.address + io.input.cmd.length) does not matter

Which is realy realy weird

Dolu1990:
Hoo my hoo my, the reason why the test are failing is funny : 

  assign io_x = 3'b000; // VerilatorCacheTester.scala l22
  
  => verilator compiltion fail with 
  %Error: /media/data/open/SpinalHDL/tmp/job_1/ComponentA.v:4016:25: Unknown verilator comment: '/*verilator CacheTester.scala l22*/'
 4016 |   assign io_x = 3'b000; /*verilator CacheTester.scala l22*/

numero-744:
> Hoo my hoo my, the reason why the test are failing is funny :

You won't get that with relative path :joy:

Also you had suggested `// @[Foo.scala 111:4]` in https://github.com/SpinalHDL/SpinalHDL/issues/783#issuecomment-1163355988

Dolu1990:
> You won't get that with relative path 

There is always a way ^^

> Also you had suggested // @[Foo.scala 111:4]

Sure can be done

numero-744:
So I guess you can `git revert 45fd064eb52a06f3583b3efdada03b3681261645` now?

Dolu1990:
Yes ^^

Dolu1990:
I'm just waiting to learn that the @ symbole isn't permited in VHDL files, even as a comment :P

Who knows !

numero-744:
Let the workflow tell us that XD

jijingg:
This feature is familiar and very "Chisel"  :joy:

and  i think  `// Debug.scala l203`  is better than  `// @[Debug.scala 203:7]`
`@ : []` seems tool much nosie and useless
image

and i suggest set `genLineComments = false`  to keep the code clean for the reason of In most cases SpinalHDL does not have the problem of code tracing.

If there is a guy from Chisel who likes this feature, he can open it himself. By default, I want Spinal to be clean and tidy, which I believe is one of the reasons most people like spinalhdl.

@Dolu1990 @numero-744

jijingg:
by the way
dose SpinalHDL support comment by user themselves.  i tried like this, but nothing happen.
```scala
  class Top extends Component {
    val a, b, c, d = in Bits (8 bit)
    val f = a | b
    f.addTag(new CommentTag( "balabala "))
    this.addComment("bilibili ")
  }
  SpinalVerilog(new Top)
```
If possible we can add comments to variables or expressions, I think it will be a very attractive feature,@Dolu1990

numero-744:
Not having the special prefix `@[` can be an issue for instance with verilator, where it can think depending on the name that it is a command (to disable lints or such things).

I agree to remove the column number and to disable it by default.

I think there is such thing for comments, can't remember what. I don't have time but if you have time you can look for it in the PR about `Latch`, there was a discussion about need to introduce comments.

Dolu1990:
> i suggest set genLineComments = false to keep the code clean for the reason of In most cases SpinalHDL does not have the problem of code tracing.

Ok for me

> dose SpinalHDL support comment by user themselves. i tried like this, but nothing happen.

Ahh currently CommentTag are generated only for the myComponent.definition.
This can be expended

> Not having the special prefix @[ can be an issue for instance with verilator, where it can think depending on the name that it is a command (to disable lints or such things).

maybe just having the @ would be enough, could remove the [ ] ?

> I agree to remove the column number and to disable it by default.

+ 1

> latch

Yes, but it was handeled by adding a comment on the whole component, not on the signal itself : 
`latchesIn.foreach(_.definition.addComment("verilator lint_off LATCH"))`

numero-744:
> maybe just having the @ would be enough, could remove the [ ] ?

:+1:

Dolu1990:
I'm doing the PR

Dolu1990:
Now released : 
https://github.com/SpinalHDL/SpinalHDL/releases/tag/v1.8.0a

PR #990, opened 2022-12-07
Make APB3CCToggle sensitive to PSEL

distributed:

`Apb3CCToggle` does not consider the `PSEL` signal. The internal state machine is activated by bus traffic for other slaves. Under the right timing conditions this can lead to the input side of the CDC replying to transactions too early, giving wrong read data.

The fix is to make `Apb3CCToggle` look at the `PSEL` signal before performing work. 

# Impact on code generation

None, except for slightly changed component logic.

# Checklist

- [x] Unit tests were added 


Dolu1990:
Version.sc and Version.scala  changes need to be reverted ^^

distributed:
I.e. rebase to `dev` like it is now?

PR #994, opened 2022-12-11
fix: fixes getFanOut; adds test for data analyzer

### LINKED ISSUE #991: How to use the getFanOut?
I'm trying to use the function `getFanout` in `spinal.lib.tools.DataAnalyzer`.  Here is my code:

```scala
import spinal.core._
import spinal.lib._

class MyTry extends Component {
  val io = new Bundle {
    val a = in Bool()
    val b = out Bool()
  }
  io.b:=io.a
}

object MyTry extends App {
    SpinalVerilog(new MyTry)
    val mytest=new MyTry
    val analyzer_a=new DataAnalyzer(mytest.io.a)
    println("fanout signal of a:"+analyzer_a.getFanOut)
    val analyzer_b=new DataAnalyzer(mytest.io.b)
    println("fanin signal of b:"+analyzer_b.getFanIn)
}
```

here is the results

```
fanout signal of a:Set()
fanin signal of b:Set((toplevel/io_a : in Bool))
```

Why the fanout signal of a is null?

wswslzp:
Closes #991 

As the title. 

# Impact on code generation

No impact.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
+++ b/lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
@@ -18,18 +18,31 @@ class DataAnalyzer(data: BaseType) {
     */
   def getFanIn: mutable.LinkedHashSet[BaseType] = {
     val ret = mutable.LinkedHashSet.newBuilder[BaseType]
-    data.foreachStatements { st =>
+
+    def traceBaseTypeFromExpression(e: Expression): mutable.LinkedHashSet[BaseType] = {
+      val ret = mutable.LinkedHashSet.newBuilder[BaseType]
+      e.foreachDrivingExpression{
+        case bt: BaseType=> ret += bt
+        case e=> ret ++= traceBaseTypeFromExpression(e)
+      }
+      ret.result()
+    }
+
+
+    data.foreachStatements {st=>
       if (!(st.isInstanceOf[InitAssignmentStatement] || st.isInstanceOf[InitialAssignmentStatement])) {
-        st.foreachDrivingExpression {
-          case bt: BaseType => ret += bt
-          case _            =>
-        }
-        st.walkParentTreeStatementsUntilRootScope { tree =>
-          tree.walkDrivingExpressions {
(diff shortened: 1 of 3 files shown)


tester/src/test/scala/spinal/tester/scalatest/DataAnalyzerTest.scala
```diff
@@ -0,0 +1,49 @@
...
+    val a2 = RegNext(a2_1)
+
+    io.b := (a2, a1, a0).asBits
+
+  }
+
+  test("Gen Test Module") {
+    SpinalVerilog(TestModule()).printRtl()
+  }
+
```
numero-744, line 26:
```suggestion
```

I think this test is not needed

lib/src/main/scala/spinal/lib/tools/DataAnalyzer.scala
```diff
@@ -18,18 +18,31 @@ class DataAnalyzer(data: BaseType) {
     */
   def getFanIn: mutable.LinkedHashSet[BaseType] = {
     val ret = mutable.LinkedHashSet.newBuilder[BaseType]
-    data.foreachStatements { st =>
+
+    def traceBaseTypeFromExpression(e: Expression): mutable.LinkedHashSet[BaseType] = {
+      val ret = mutable.LinkedHashSet.newBuilder[BaseType]
+      e.foreachDrivingExpression{
```
numero-744, line 24: Can you auto-format this file, please?

ZhaokunHu:
Thank you for the PR and the example ！I understand how to use it now and it works good.

wswslzp:
All done, I think it's ok to be merged.

Readon:
seems that current analyzer is designed for generated rtl, according to the test.
is it applicapable for modules before elaboration? where the signals' name are different.

do you think it might be better if there is an test for that use case?

Readon:
seems that current analyzer is designed for generated rtl, according to the test.
is it applicapable for modules before elaboration? where the signals' name are different.

do you think it might be better if there is an test for that use case?

PR #995, opened 2022-12-12
feat: add ECP5 EHXPLLL blackbox and some helper functions

kleinai:

This isn't a complete realization of the EHXPLLL and its features. This blackbox and helper calculator function have been verified on hardware.

The frequency assertions might be a little overbearing. Personally, I'd prefer to catch frequency errors early rather than after all of place and route.

# Impact on code generation

Zero.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/blackbox/lattice/ecp5/EHXPLLL.scala
+++ b/lib/src/main/scala/spinal/lib/blackbox/lattice/ecp5/EHXPLLL.scala
@@ -44,7 +44,9 @@ object EHXPLLLConfig {
     var mDiv: BigDecimal = 1
     var fbDiv: BigDecimal = 1
     var err: BigDecimal = calculateError(inFreq.toBigDecimal, outFreq.toBigDecimal, fbDiv/mDiv)
-    while(mDiv < 128 && fbDiv < 128 && err.abs > tolerance) {
+    while(mDiv < 128 && fbDiv < 128 &&
+          (err.abs > tolerance || (inFreq.toBigDecimal*fbDiv)/mDiv <= VCO_MIN.toBigDecimal) &&
+          (inFreq.toBigDecimal*fbDiv)/mDiv <= VCO_MAX.toBigDecimal) {
       err = calculateError(inFreq.toBigDecimal, outFreq.toBigDecimal, fbDiv/mDiv)
       if (err < 0) {
         mDiv += 1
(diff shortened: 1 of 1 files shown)


kleinai:
That would make the assertions look better! I'm not in a hurry to get this merged in, if you want to go ahead with your PR first.

andreasWallner:
> 

Done: #997

andreasWallner:
@kleinai Sorry to have held you up so long, we just merged the comparison operators :-)

lib/src/main/scala/spinal/lib/blackbox/lattice/ecp5/EHXPLLL.scala
```diff
@@ -0,0 +1,224 @@
...
+    */
+  def singleOutput(inFreq: HertzNumber, outFreq: HertzNumber, tolerance: BigDecimal = 0.02): EHXPLLLConfig = {
+    def calculateError(clki: BigDecimal, clko: BigDecimal, factor: BigDecimal): BigDecimal = {
+      (clko-(clki*factor))/clko
+    }
+
+    var mDiv: BigDecimal = 1
+    var fbDiv: BigDecimal = 1
+    var err: BigDecimal = calculateError(inFreq.toBigDecimal, outFreq.toBigDecimal, fbDiv/mDiv)
+    while(mDiv < 128 && fbDiv < 128 && err.abs > tolerance) {
```
andreasWallner, line 58: Would it make sense to add limit checks in this loop? Otherwise it could find a "best solution" that actually has an invalid internal frequency?

lib/src/main/scala/spinal/lib/blackbox/lattice/ecp5/EHXPLLL.scala
```diff
@@ -0,0 +1,224 @@
+package spinal.lib.blackbox.lattice.ecp5
+
+import spinal.core._
+
+object EHXPLLLConfig {
+
+  trait FeedbackType {
```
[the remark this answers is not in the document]
kleinai: Okay, we'll try that. I think my previous understand of the PLL feedback types was incorrect. I believe INT_CLKOP and friends are accepted values.

lib/src/main/scala/spinal/lib/blackbox/lattice/ecp5/EHXPLLL.scala
kleinai: Yes, it should attempt to keep the VCO within tolerance and valid frequency range.  The config will do the bounds check and it would be annoying if this function returns a VCO that's out of range.

kleinai:
This is good for review. Still works on my ECP5 eval board as expected.

PR #996, opened 2022-12-12
Just fix the CI process failed in some cases.

### LINKED ISSUE #987: bug: CI failed with "Unable to locate command >iverilog<"
The details is log [here](https://github.com/Readon/SpinalHDL/actions/runs/3627510211/jobs/6118569392).

Readon:
Closes #987 

two probelms has met.
1. iverilog compilation reports that "parse.hh" could not be found.
2. On latest Ubuntu, iverilog 11.0 could fail when cpp sanity check stop.

This is not necessarily merged. Just wait to see if there are same problems from other one's repositories.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/Enum.scala
+++ b/core/src/main/scala/spinal/core/Enum.scala
@@ -84,8 +84,11 @@ class SpinalEnum(var defaultEncoding: SpinalEnumEncoding = native) extends Namea
   type C = SpinalEnumCraft[this.type]
   type E = SpinalEnumElement[this.type]
 
-  private[core] var isPrefixEnable = GlobalData.get.config.enumPrefixEnable
-  private[core] var isGlobalEnable = GlobalData.get.config.enumGlobalEnable
+  var forcedPrefixEnable : Option[Boolean] = None
+  var forcedGlobalEnable : Option[Boolean] = None
+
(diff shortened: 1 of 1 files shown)


kleinai:
I have no problems with this change. It's probably better to have docker pegged to an LTS release of Ubuntu rather than the latest.

numero-744:
@Readon I'm curious, if this wasn't the solution, how have you fixed the issue? Btw I'm facing the same issue on my CI.

Readon:
> @Readon I'm curious, if this wasn't the solution, how have you fixed the issue? Btw I'm facing the same issue on my CI.

What I have done is to run the ubuntu branch in my repository once. Then it works while I switch back to upstream dev branch. I think it's just about the cache policy that GitHub applied.

numero-744:
I think the branch fixes the issue and puts the fix in the cache so I would actually merge the branch in `dev` so that the fix is made permanent.

If you go into Actions/Cache you might see with more details what happens?

Readon:
> I think the branch fixes the issue and puts the fix in the cache so I would actually merge the branch in `dev` so that the fix is made permanent.
> 
> If you go into Actions/Cache you might see with more details what happens?

The problem is that we will miss the results of building on Ubuntu 22, if we fix it to Ubuntu 20. That's why I retreat this PR.

PR #997, opened 2022-12-12
feat: add comparison operators to PhysicalNumber

andreasWallner:

In code for checking frequencies/times are in range, e.g. for calculating&checking PLL parameters it's much more readable when PhysicalNumbers can be compared directly and with units, instead of having to go through `toBigDecimal`.

```
assert(fdiv_freq < 420MHz)
//vs
assert(fdiv_freq.toBigDecimal < 420000000L)
``` 

# Impact on code generation

None

# Checklist

- [x] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/Units.scala
+++ b/core/src/main/scala/spinal/core/Units.scala
@@ -80,7 +80,14 @@ abstract class PhysicalNumber[T <: PhysicalNumber[_]](protected val value: BigDe
   def /(that: BigDecimal) = newInstance(this.value / that)
   def %(that: BigDecimal) = newInstance(this.value % that)
 
-  def max(that: T) = newInstance(value.max(that.value))
+  def <(that: T): Boolean = this.value < that.value
+  def <=(that: T): Boolean = this.value <= that.value
+  def >(that: T): Boolean = this.value > that.value
+  def >=(that: T): Boolean = this.value >= that.value
+
+  def abs: T = newInstance(this.value.abs)
+  def max(that: T): T = newInstance(this.value.max(that.value))
+  def min(that: T): T = newInstance(this.value.min(that.value))
 
   def toInt        = value.toInt
   def toLong       = value.toLong
--- a/core/src/test/scala/spinal/core/PhysicalNumbers.scala
+++ b/core/src/test/scala/spinal/core/PhysicalNumbers.scala
@@ -1,0 +1,37 @@
+package spinal.core
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class PhysicalNumbers extends AnyFunSuite {
+  val hz100 = 100 Hz
+  val hz100_2 = 100 Hz
+  val hz200 = 200 Hz
+  val ms2 = 100 sec
+
+  test("Comparisons") {
+    assert(hz100 == hz100)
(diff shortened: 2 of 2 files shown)


kleinai:
I think we can merge this without issue if it passes tests.

kleinai:
I'm going to fix the offending sim code because I wrote it two different ways and I prefer the match-case way instead of the not equal check.

core/src/main/scala/spinal/core/Units.scala
```diff
@@ -80,7 +80,15 @@ abstract class PhysicalNumber[T <: PhysicalNumber[_]](protected val value: BigDe
   def /(that: BigDecimal) = newInstance(this.value / that)
   def %(that: BigDecimal) = newInstance(this.value % that)
 
-  def max(that: T) = newInstance(value.max(that.value))
+  def ==(that: T): Boolean = this.value == that.value
```
numero-744, line 83:
```suggestion
  def ==(that: T): Boolean = this.value == that.value
  def !=(that: T): Boolean = this.value != that.value
```

Hmm maybe `!=` too? Or scala implements it automatically from `==`?

Also I'm not sure of what happens when using the `==` method above, can unit tests be added, please?

I find references defining `canEqual`, `equals` and `hashCode`, maybe that should be done instead?

andreasWallner: Thanks, I actually wanted to remove `==` as well for symmetry, and `!=` was removed for the breakage it caused in the GHDL backend, see #998.
`==` is actually the same as `equals`

andreasWallner:
So... with this implementation `!= null` works as required. Removing it was the right thing to do, since `HertzNumber` and `TimeNumber` are case classes and have a compiler generated, member equality checking, overridden `equals` anyways, so mine would never be called...

core/src/test/scala/spinal/core/PhysicalNumbers.scala
```diff
@@ -0,0 +1,37 @@
...
+
+    assert(hz100 <= hz200)
+    assert(hz100 <= hz100_2)
+    assert(!(hz200 <= hz100))
+
+    assert(hz100 >= hz100_2)
+    assert(hz200 >= hz100)
+    assert(!(hz100 >= hz200))
+
+    assert(hz100.hashCode() == hz100_2.hashCode())
```
numero-744, line 35:
```suggestion
    assert(hz100 == hz100_2)
```

It feels more natural

numero-744:
How do you think it should be documented? (Do you think it should be?)

andreasWallner:
> How do you think it should be documented? (Do you think it should be?)

I don´t think we need specific documentation, `PhysicalUnit` is only mentioned in a small subchapter in the documentation - if there is a user who needs the comparison I guess it will just feel naturally?

core/src/test/scala/spinal/core/PhysicalNumbers.scala
andreasWallner: That is the same as the check further above, here I wanted to explicitly check that hashCode is implemented 'correctly' (for some approximation of correctness ;-))

numero-744: Ah sorry, didn't see

PR #998, opened 2022-12-13
Use match-case when computing GHDL time flag

kleinai:

Formatting match to other simulator time checks. Avoids using `!=` operator.

# Impact on code generation

None.

### DIFF
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -271,8 +271,9 @@ object SpinalGhdlBackend {
   def apply[T <: Component](config: SpinalGhdlBackendConfig[T]) : Backend = {
     val vconfig = new GhdlBackendConfig()
     vconfig.analyzeFlags = config.simulatorFlags.mkString(" ")
-    if (config.timePrecision != null) {
-      vconfig.elaborationFlags = s"--time-resolution=${config.timePrecision.decompose._2}"
+    config.timePrecision match {
+      case null =>
+      case t => vconfig.elaborationFlags = s"--time-resolution=${t.decompose._2}"
     }
     vconfig.runFlags = config.simulatorFlags.mkString(" ")
     vconfig.logSimProcess = config.enableLogging


core/src/main/scala/spinal/core/sim/SimBootstraps.scala
```diff
@@ -271,8 +271,9 @@ object SpinalGhdlBackend {
   def apply[T <: Component](config: SpinalGhdlBackendConfig[T]) : Backend = {
     val vconfig = new GhdlBackendConfig()
     vconfig.analyzeFlags = config.simulatorFlags.mkString(" ")
-    if (config.timePrecision != null) {
-      vconfig.elaborationFlags = s"--time-resolution=${config.timePrecision.decompose._2}"
+    config.timePrecision match {
+      case null =>
+      case t => vconfig.elaborationFlags = s"--time-resolution=${config.timePrecision.decompose._2}"
```
numero-744, line 276:
```suggestion
      case t => vconfig.elaborationFlags = s"--time-resolution=${t.decompose._2}"
```

numero-744 (commented):
To me the question is not really `if` vs `match` (which is IMO in the current case just a way to write the same thing with more text), it is more about `null` vs `Option`: https://alvinalexander.com/scala/scala-null-values-option-uninitialized-variables/

If we change internals to switch to `Option`, let's not modify the API: let only one `internal = if (arg == null) None else Some(arg)` in the builder?

kleinai:
> To me the question is not really `if` vs `match` (which is IMO in the current case just a way to write the same thing with more text), it is more about `null` vs `Option`: https://alvinalexander.com/scala/scala-null-values-option-uninitialized-variables/
> 
> If we change internals to switch to `Option`, let's not modify the API: let only one `internal = if (arg == null) None else Some(arg)` in the builder?

I agree we should use option types more however this was intended to be a small code style change to help along #997

andreasWallner:
While I agree with the point on `Option`, I did a quick check and the original issue is caused by me incorrectly implementing `==`. Looking at [scaladoc](https://www.scala-lang.org/api/current/scala/Any.html#==(x$1:Any):Boolean) I should have included a check against null... (it also between the lines suggests to do null checks with `x eq null` instead of `x ==null` which is also mirrored by the scala standard library in many places)

numero-744:
Ah so the description of == in Scaladoc explains why people use a 
canEqual / equals / hashCode mechanism to override that.Message ID: 
***@***.***>

numero-744:
IIUC with correct implementation of equality with the 3 methods, this PR is not needed and `xxx != null` can be used?

numero-744:
Closed as not needed: https://github.com/SpinalHDL/SpinalHDL/pull/997#issuecomment-1356565628

PR #1000, opened 2022-12-18
ci: fix missing iverilog error

### LINKED ISSUE #987: bug: CI failed with "Unable to locate command >iverilog<"

numero-744:
Closes #987

### DIFF
--- a/tools.sh
+++ b/tools.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 install_verilator(){
-  sudo apt install -y git make autoconf g++ flex libfl-dev bison  # First time prerequisites
+  sudo apt install -y git make autoconf build-essential flex libfl-dev bison  # First time prerequisites
   git clone http://git.veripool.org/git/verilator   # Only first time
   unset VERILATOR_ROOT  # For bash
   cd verilator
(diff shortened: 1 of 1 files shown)


numero-744:
It passed, I squashed and re-pushed, and it failed :thinking: Cleaning caches and trying again.

numero-744:
Here it just passed twice :shrug:

Readon:
Ahh, you just install iverilog from Ubuntu's repository. 
I don't know the reason to build and cache iverilog in the CI, so I did not alternate that part.

Readon (commented):
@Dolu1990 need to confirm if build from source can be ignored.

Dolu1990:
@Readon It was built from source, because icaruse verilog had bugs in quite a few version. + there was issue with the cocotb interopability.

Using apt-get packet manager is oppening the way for out of control breaks : 
- Unkown / untested version of iverilog will be installed
- Version will change over time

I think we should keep building it from source as much as possible. That would save us future time, pain and suffering.
If that's realy to much of a pain to build from source => ok for apt-get

Readon:
> @Readon It was built from source, because icaruse verilog had bugs in quite a few version. + there was issue with the cocotb interopability.
> 
> Using apt-get packet manager is oppening the way for out of control breaks :
> 
> * Unkown / untested version of iverilog will be installed
> * Version will change over time
> 
> I think we should keep building it from source as much as possible. That would save us future time, pain and suffering. If that's realy to much of a pain to build from source => ok for apt-get

The existing code for CI would fail to build from start up. You can see it when the cached packages are cleaned.

numero-744:
Okay I'll try to adapt things tomorrow

PR #1001, opened 2022-12-18
Move tests

numero-744:
TL;DR: gl hf

This PR moves tests to their standard location according to SBT. To do so:

* "test" dependencies between the different subprojects were added (`see build.sbt`)
* The things that I had to put common (or I thought I had to) were moved to `tester`, and its organization in packages should be discussed (eg: sim vs formal)
* I could not associate some tests with a specific item / file so for them, they are in `integration` folder of the appropriate sub-project.

Closes #982

PR #1002, opened 2022-12-19
feat(lib): Axi4 to AxiLite4 conversions

kleinai:

Adds simple conversion methods to go from Axi4 to AxiLite4 and vice versa.
ID remover (reflection) and unbursting (conversion) applied as needed.

# Impact on code generation

# Checklist

- [ ] Unit tests were added


Readon:
Good work.

Btw, AxiLite4 is a special case of Axi4 with 
``` scala
      useId = false,
      useRegion = false,
      useBurst = false,
      useLock = false,
      useCache = false,
      useSize = false,
      useQos = false,
      useLen = false,
```
Do you think it's worth to unify them into one by different config?
If we do so, it is not necessary to convert them.
Maybe the most complex work is to fit all facilities that existing.

kleinai:
> Good work.

Thanks!

> Do you think it's worth to unify them into one by different config? If we do so, it is not necessary to convert them. Maybe the most complex work is to fit all facilities that existing.

I believe this conversation has been had before and the general consensus is no.  I don't have a preference either way so long as there are easy conversion methods.

kleinai:
Also, I have tests I wrote while building these methods if we want those too.  Although this includes some rework of the AxiLite4 sim driver.

Readon:
> Also, I have tests I wrote while building these methods if we want those too. Although this includes some rework of the AxiLite4 sim driver.

That would be helpful to have a test. Does the unmodified AxiLite4 sim driver works?
If a rework on AxiLite4 Sim driver is required, how about make a PR on that first?

numero-744:
About tests, don't forget #1001

Readon:
> About tests, don't forget #1001

I think it's in draft status still. ^^

numero-744:
Yep, because I wanted to fix the CI first; and feedback is needed for the structure of the test directory.

Readon:
Thanks, ^^.

PR #1010, opened 2023-01-10
Add ethernet udp tx/rx module

jjyy-Huang:
The udp project is [here](https://github.com/jjyy-Huang/SpinalHDL-ethernet.git).
- The goal of this PR is to create a UDP/IP Tx/Rx module with SpinalHDL for Xilinx 100GbE Ethernet-Subsystem (MRMAC/CMAC).

- the module support following features:
  - UDP Tx/Rx without checksum
  - IP Tx/Rx
  - Ethernet Frame Layer2 (pre-add MAC adn EtherType)
  - 256b data width
  - Fully pipeline

# Impact on code generation

Zero

# Checklist

- [ ] Unit tests were added

# TODO
- [ ] Redesign Rx module control logic
- [ ] Add ARP request module
- [ ] Add ICMP module
- [ ] I/O data support 384b/512b
- [ ] Use cocotb for verification

### DIFF
--- a/lib/src/main/scala/spinal/lib/ethernet/Axi4StreamConditionalJoin.scala
+++ b/lib/src/main/scala/spinal/lib/ethernet/Axi4StreamConditionalJoin.scala
@@ -1,0 +1,42 @@
+package spinal.lib.ethernet
+
+import spinal.core._
+import spinal.lib.Stream
+import spinal.lib.bus.amba4.axis.Axi4Stream.Axi4StreamBundle
+
+object Axi4StreamConditionalJoin {
+
+  /** Convert a tuple of streams into a stream of tuples
+    * source stream1 has higher priority
+    */
+  def apply[T1 <: Axi4StreamBundle, T2 <: Axi4StreamBundle](
+      source1: Stream[T1],
+      source2: Stream[T2],
+      source1Select: Bool,
+      source2Select: Bool
+  ): Stream[TupleBundle2[T1, T2]] = {
+    val sources = Seq(source1, source2)
+    val combined = Stream(
+      TupleBundle2(
+        source1.payloadType,
+        source2.payloadType
+      )
+    )
+    combined.valid := (sources(0).valid && sources(1).valid) |
+      ((source1Select) && sources(0).valid)
+    sources(0).ready := combined.fire
+    sources(1).ready := source2Select & combined.fire
+    //    if (useKeep/useLast/useUser)
+    combined.payload._1.keep := source1.fire ? source1.payload.keep | 0
+    combined.payload._2.keep := source2.fire ? source2.payload.keep | 0
+    if (source1.user != null & source2.user != null) {
+      combined.payload._1.user := source1.fire ? source1.payload.user | 0
+      combined.payload._2.user := source2.fire ? source2.payload.user | 0
+    }
+    combined.payload._1.data := source1.fire ? source1.payload.data | 0
+    combined.payload._2.data := source2.fire ? source2.payload.data | 0
+    combined.payload._1.last := source1.fire ? source1.payload.last | False
+    combined.payload._2.last := source2.fire ? source2.payload.last | False
+    combined
+  }
+}
--- a/lib/src/main/scala/spinal/lib/ethernet/IPv4Header.scala
+++ b/lib/src/main/scala/spinal/lib/ethernet/IPv4Header.scala
@@ -1,0 +1,58 @@
+package spinal.lib.ethernet
+
+import spinal.core._
+import EthernetProtocolConstant._
+
+import scala.collection.mutable
+
+object IPv4Header {
+  def apply(array: Array[Bits]): Array[Bits] = {
+    val gen = new IPv4Header
+    require(
+      array.length == gen.header.length,
+      s"Initializing parameters not enough! Require ${gen.header.length} but gave ${array.length}"
+    )
+    gen.header.zipWithIndex.foreach { case (data, idx) =>
+      data := array(idx)
+    }
+    gen.header
+  }
+
+  def unapply(header: Bits): (Array[String], Array[Bits]) = {
+    val gen = new IPv4Header
+    val bitWidth = gen.frameFieldInit.values.sum
+    require(
+      bitWidth == header.getWidth,
+      s"Initializing parameters not enough! Require ${bitWidth} but gave ${header.getWidth}"
+    )
+    val asLittleEnd = header.subdivideIn(bitWidth / 8 slices).reduce(_ ## _)
+    val tmp = asLittleEnd
+      .sliceBy(gen.frameFieldInit.values.toList.reverse)
+      .reverse
+      .toArray
+    gen.header.zipWithIndex.foreach { case (data, idx) =>
+      data := tmp(idx)
+    }
+    (gen.frameFieldInit.keys.toArray, gen.header)
+  }
+}
+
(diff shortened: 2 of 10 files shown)


lib/src/main/scala/spinal/lib/ethernet/Axi4StreamConditionalJoin.scala
```diff
@@ -0,0 +1,37 @@
...
+import spinal.core._
+import spinal.lib.Stream
+import spinal.lib.bus.amba4.axis.Axi4Stream.Axi4StreamBundle
+
+object Axi4StreamConditionalJoin {
+
+  /** Convert a tuple of streams into a stream of tuples
+    * source stream1 has higher priority
+    */
+  def apply[T1 <: Axi4StreamBundle, T2 <: Axi4StreamBundle](
```
Dolu1990, line 12: One way to deal wth the select is to stream1.haltWhen(!source1Select) , same for source2 :)

``` scala
    combined.valid := (sources(0).valid && sources(1).valid) |
      (source1Select && sources(0).valid)
    sources(0).ready := combined.fire
    sources(1).ready := source2Select & combined.fire
```

This is some kind of tricky arbitration. Maybe the utlity should be more neutral ? Or more documented about the edge case : ex !source1.valid && source2.valid && source2Select

Seems like this could be a regular Data instead of Axi4StreamBundle type paramater ?

lib/src/main/scala/spinal/lib/ethernet/IPv4Header.scala
```diff
@@ -0,0 +1,36 @@
...
+  }
+  def unapply(header: Bits): (Seq[String], Seq[Bits]) = {
+    val extract = IPv4Header()
+    this.extract(header, extract)
+  }
+}
+
+case class IPv4Header() extends FrameHeader {
+  val frameFieldInit = mutable.LinkedHashMap(
+    "protocolVersion" -> IP_VERSION_WIDTH,
```
Dolu1990, line 21: An alternative of using String would be to use Object : 

```scala
trait Key
object Keys{
  val protocolVersion = new Key
  val internetHeaderLength = new Key
  ..
}
  val frameFieldInit = mutable.LinkedHashMap(
    Keys.protocolVersion -> IP_VERSION_WIDTH,
```

lib/src/main/scala/spinal/lib/ethernet/Utils.scala
```diff
@@ -0,0 +1,78 @@
+package spinal.lib.ethernet
+
+import spinal.core._
+
+import UserConfiguration._
+
+object rotateLeftByte {
```
Dolu1990, line 7: rotate operation could be added to the Bits data type directly :)
Maybe in a more general form ?
data: Bits, bias: UInt, sliceWidth : BitCount  ?

lib/src/test/scala/spinal/lib/ethernet/LoopbackCheck.scala
```diff
@@ -0,0 +1,214 @@
...
+      def loadMeta(sendDataBytes: Int): Unit = {
+        dut.io.metaIn.payload.dstMacAddr #= simCfg.dstMacAddr.asHex
+        dut.io.metaIn.payload.dstIpAddr #= simCfg.dstIpAddr.asHex
+        dut.io.metaIn.payload.dstPort #= simCfg.dstPort.asHex
+        dut.io.metaIn.payload.srcMacAddr #= simCfg.srcMacAddr.asHex
+        dut.io.metaIn.payload.srcIpAddr #= simCfg.srcIpAddr.asHex
+        dut.io.metaIn.payload.srcPort #= simCfg.srcPort.asHex
+        dut.io.metaIn.payload.dataLen #= sendDataBytes
+        dut.io.metaIn.valid #= true
+        dut.clockDomain.waitRisingEdge()
```
Dolu1990, line 182: Prefer waitSampling to waitRisingEdge, as it take in acount the reset status and the clock edge kind.

lib/src/main/scala/spinal/lib/ethernet/Axi4StreamConditionalJoin.scala
Dolu1990: Maybe this utility should be more like : 

```scala
object SubStreamJoin{
  def apply[T1 <: Data, T2 <: Data](
      main: Stream[T1], sub: Stream[T2],
      withSub: Bool
  ): Stream[TupleBundle2[T1, T2]] = {
}
```

?

Where main stream can flow throw if !withSub
Else, sub has to be valid and will be merged

If the user want to hold the main stream, he can just use main.haltWhen(xxx) outside this utility

?

Dolu1990:
Hi ^^

So, i think for the doc, as it is a specific thing with quite a few abstraction, it would need to have some on the RTD

lib/src/main/scala/spinal/lib/ethernet/Axi4StreamConditionalJoin.scala
jjyy-Huang: You are right! 

Because the function I want to achieve here is to merge two streams in some condition, using one stream as primary and another one as a sub could be better.

I will improve it in the next commit.

lib/src/main/scala/spinal/lib/ethernet/IPv4Header.scala
jjyy-Huang: Got it. It seems more reasonable.

lib/src/test/scala/spinal/lib/ethernet/LoopbackCheck.scala
jjyy-Huang: Got it : )

jjyy-Huang:
> Hi ^^
> 
> So, i think for the doc, as it is a specific thing with quite a few abstraction, it would need to have some on the RTD

Sorry for the late reply.

The doc will be added with more details of implementation in the next commit

lib/src/main/scala/spinal/lib/ethernet/Utils.scala
jjyy-Huang: Ummm...

Actually, there is `rotateLeft()` in BitVector,
```Scala
def rotateLeft(that: UInt): T = {
    ...
}
```

I wonder if it would be better to use `switch case` instead of `bits shift` one by one.

Dolu1990: > I wonder if it would be better to use switch instead of bits shift one by one.

Hmm, the main challenge is to ensure the synthesis tool can infer a proper barrel shifter using log2up(width) stages, that's why it is done in shifts sequence.

I would say the main limitation of the current implementation is that there is now parameter to specify how many bits to rotate per unit.

jjyy-Huang:
> Hi ^^
So, i think for the doc, as it is a specific thing with quite a few abstraction, it would need to have some on the RTD

I'm sorry for waiting on this document for a long time.

The following link is the doc about this module design.

https://github.com/jjyy-Huang/SpinalHDL-ethernet/blob/udp/README.md

Dolu1990:
Thanks :)
I will check it, just have quite a high stack of work at the moment

jjyy-Huang:
Hi,

Sorry for the late reply, I changed my email and can't receive news from GitHub.

I appreciate your suggestion. I will integrate it into my work process and continue to make improvements as I move forward.

Thanks again for your input, and please let me know if you have any further suggestions or feedback.

PR #1015, opened 2023-01-12
feat: add loadHex function for SparseMemory

### LINKED ISSUE #1014: add loadHex function for SparseMemory
how about add a `loadHex` function for SparseMemory in `lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/AxiMemorySim.scala`, so that the memory can be inited both in bin and hex file.

hanm2019:
Closes #1014 

the `SparseMemory` in `lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/AxiMemorySim.scala` has a function `loadbin()` so that in the simulation, we can init the memory through bin file. It may be more convenient that offer a function `loadHex()`, so that we can directly init the memory by hex file

# Impact on code generation

Has no impact on code generation, this function impact the simulation of AxiMemorySim

However, I am not sure the implement of `loadHex()` is right, would you please review this code, I have test this function on a hex [file](https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/resources/hex/C.ADD.elf.hex). 

# Checklist

- [x] Unit tests were added

### DIFF
--- a/tester/src/test/scala/spinal/tester/scalatest/SpinalSimBmbLengthSpliterTester.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/SpinalSimBmbLengthSpliterTester.scala
@@ -1,12 +1,70 @@
 package spinal.tester.scalatest
 
 import org.scalatest.funsuite.AnyFunSuite
-import spinal.core.sim.SimConfig
+import spinal.core.sim.{SimConfig, SimTimeout}
 import spinal.lib.bus.bmb.sim.BmbBridgeTester
 import spinal.lib.bus.bmb.{BmbAccessParameter, BmbAlignedSpliter, BmbParameter, BmbSourceParameter}
 
+
 class SpinalSimBmbLengthSpliterTester extends AnyFunSuite {
-  for(w <- List(false, true); r <- List(false, true);   if w || r) {
+//  test("failed to stop case: only read " ) {
+//    SimConfig.compile {
+//      val c = BmbAlignedSpliter(
+//        ip = BmbAccessParameter(
+//          addressWidth = 16,
+//          dataWidth = 32
+//        ).addSources(16, BmbSourceParameter(
+//          lengthWidth = 6,
+//          contextWidth = 3,
+//          alignmentMin = 0,
+//          canRead = true,
+//          canWrite = false,
+//          alignment = BmbParameter.BurstAlignement.WORD
+//        )).toBmbParameter(),
+//        lengthMax = 4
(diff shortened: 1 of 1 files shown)


lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/AxiMemorySim.scala
```diff
@@ -205,6 +205,13 @@ case class SparseMemory() {
     println(s"Loading 0x${byteArray.length.toHexString} bytes from ${file} to 0x${address.toHexString}")
   }
 
+  def loadHex(offset: Int, file: String) : Unit = {
```
numero-744, line 208:
```suggestion
  def loadHex(file: String, offset: Int = 0) : Unit = {
```

Wouldn't this order be more conventional (eg to match `readHexFile`) and having `offset` defaulting to 0 be a good thing?

lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/AxiMemorySim.scala
```diff
@@ -205,6 +205,13 @@ case class SparseMemory() {
...
+    import spinal.lib.misc.HexTools.readHexFile
+    readHexFile(file, offset, (address, data) => {
+      writeBigInt(address, data,2)
```
numero-744, line 211:
```suggestion
      writeBigInt(address, data, 2)
```

Btw, wondering what this 2 stands for.

hanm2019:
I also update the readHexFile function in HexTools.
In my opinion, the HexOffset may bigger than `0x80000000`, If the datatype of HexOffset is int, this value will exceed the value range of int. Maybe use `Long` is better. 
I also update the `hToI` function with the same reason.

for example. In hex file, there is a line:
```
:02 00 00 04 8000 7A
```
this line means that the offset will change to `0x8000` << 16. If the `hToI` use `Integer.parseInt` to parse this string, It will get `-2147483648`, not the `2147483648`.

kleinai:
Makes sense. Is this ready to merge now?

numero-744 (approved):

Readon:
Is that possible to add some test for this? It would be good in two points. 
1. help user to know how to use it.
2. ensure code is working as expected.

hanm2019:
> Is that possible to add some test for this? It would be good in two points.
> 
> 1. help user to know how to use it.
> 2. ensure code is working as expected.

I have add some test for loadHex

PR #1016, opened 2023-01-12
build: upgrade dependency versions

### LINKED ISSUE #1012: Dependency Review and Upgrade
I'd like to request, and help preform, a review of the dependencies and upgrade them as necessary. I'd like to try to use Spinal at my office and they would prefer more up-to-date libraries than what are being used now.

kleinai:
Closes #1012

Flat upgrading all the dependencies seems to have worked well with only one incompatibility.

Upgrade dependencies to keep current.

# Impact on code generation

None expected.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import sbt.Keys._
 import sbt._
 import sbt.Tests._
 
-val scalatestVersion = "3.2.5"
+val scalatestVersion = "3.2.14"
 val defaultSettings = Defaults.coreDefaultSettings ++ xerial.sbt.Sonatype.sonatypeSettings ++ Seq(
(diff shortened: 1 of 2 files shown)


numero-744 (approved):

numero-744:
Thanks for the PR!

It seems GitHub has network issues (see https://github.com/SpinalHDL/SpinalHDL/pull/1015/checks), I'm waiting for them to be fixed before I merge this PR (to avoid failed workflow on `dev`).

Readon:
This might affect the projects based on this library like NaxRiscv, VexRiscv and a bundle of other works, which would need an update on dependency also.
I still suggest to wait most of other projects be tested first.

PR #1028, opened 2023-01-26
Implement addRunFlag(), needed by GHDL.

likewise:
GHDL requires simulation run flags separate from elaboration flags, our approach did not allow GHDL these. Fix hat.

vconfig.runFlags = config.simulatorFlags.mkString(" ")

https://umarcor.github.io/ghdl/using/InvokingGHDL.html#run-r

Closes #

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -168,6 +168,7 @@ class SpinalVpiBackendConfig[T <: Component](val rtl               : SpinalRepor
                                              val waveDepth        : Int,
                                              val optimisationLevel: Int,
                                              val simulatorFlags   : ArrayBuffer[String],
+                                             val runFlags         : ArrayBuffer[String],
                                              val usePluginsCache  : Boolean,
                                              val pluginsCachePath : String,
                                              val enableLogging    : Boolean,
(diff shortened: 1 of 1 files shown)


likewise:
`addSimulatorFlag()` is not yet documented, but typically contains the elaboration flags for the simulator.
`addRunFlag()` contains the run flags for the simulator that do not influence elaboration.

For GHDL the run flags come at the end of the line, so we cannot use `addSimulatorFlag()` like with other simulators.
See https://umarcor.github.io/ghdl/using/InvokingGHDL.html#run-r
We already had runFlags in the backend, you just could not set them.

This feature allows to debug more efficiently with these GHDL features:
```
.addRunFlag("--unbuffered").addRunFlag("--disp-tree=inst")
.addRunFlag("--ieee-asserts=disable").addRunFlag("--assert-level=none")
.addRunFlag("--backtrace-severity=warning")
```

core/src/main/scala/spinal/core/sim/SimBootstraps.scala
```diff
@@ -274,7 +280,7 @@ object SpinalGhdlBackend {
     if (config.timePrecision != null) {
       vconfig.elaborationFlags = s"--time-resolution=${config.timePrecision.decompose._2}"
     }
-    vconfig.runFlags = config.simulatorFlags.mkString(" ")
+    vconfig.runFlags = config.runFlags.mkString(" ")
```
kleinai, line 283: Should we completely squash simulator flags for GHDL?

likewise: The current implementation squashes both simulatorFlags and runFlags into simulatorFlags.

My implementation keeps those apart and introduces a way to specify runFlags to the simulator.

So my patch does quite the opposite from squashing.

kleinai: I see I missed the line above this that sets elaboration flags from sim flags

kleinai:
Looks good to me. Thanks!

PR #1033, opened 2023-02-09
Add GHDL sim backend support for different waveform formats.

likewise:
GHDL supports different kinds of waveforms; add support for this in our backend.

This does not affect code generation.


Dolu1990:
Thanks :)

PR #1039, opened 2023-02-21
[regif] fix readData hold issue for the reason of security

jijingg:
add defaultValue on readData else branch avoid security risk 
```verilog
if(readAsk) begin
  ......
else  begin   
   readData  <= defaultValue;
   readError <= False       ;
end
```


likewise:
Two AXI4-Lite handshakes must occur; one on the AR channel, one on the R channel.

Commit https://github.com/SpinalHDL/SpinalHDL/commit/230ab60ba549be816929661a27d7d4d7f08f1b54 is wrong, in that it resets the readData if not in the AR handshake. However, it should have waited for the R handshake to have completed instead.

PR #1041, opened 2023-02-21
Axis width fixes

kleinai:

Fixes a bug when tkeep is disabled in the width adapter.

Fixes a gross inefficiency when generating the sparse compactor mux table. The previous method could generate very large files. I've seen up to 500MB of netlist! This new method uses **_only_** 2^N matches which is an improvement but not by much.

# Impact on code generation

# Checklist

- [X] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4StreamSparseCompactor.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4StreamSparseCompactor.scala
@@ -29,21 +29,42 @@ class Axi4StreamSparseCompactor(config: Axi4StreamConfig) extends Component {
 
   io.axis_m << outStage
 
-  def indexOfBoolN(bits: Bits, n: Int): UInt = {
-    val index = UInt(log2Up(widthOf(bits)+1) bit)
-    index := U(widthOf(bits))
-
-    def walkLowBits(width: Int, n: Int): Seq[BigInt] = {
-      if (width <= 0 || n <= 0) {
-        Seq(0)
-      } else {
-        (n-1 until width).flatMap { i =>
-          walkLowBits(i, n - 1).map { o =>
-            (BigInt(1) << i) + o
-          }
+  /**
+    * Generates all possible permutations of N bits which fit within the width
+    *
+    * For example: walkLowBits(4, 2) would generate:
+    *     11
+    *    101
+    *    110
+    *   1001
+    *   1010
+    *   1100
+    * @param width Maximum bit width
+    * @param n Number of set bits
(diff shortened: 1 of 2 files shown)


lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4StreamSparseCompactor.scala
```diff
@@ -29,51 +29,48 @@ class Axi4StreamSparseCompactor(config: Axi4StreamConfig) extends Component {
...
-  def indexOfBoolN(bools: Seq[Boolean], n: Int): Int = {
-    var boolCount = 0
-    for ((b, i) <- bools.zipWithIndex) {
-      if (b) {
-        boolCount += 1
+  def indexOfBoolN(bits: Bits, n: Int): UInt = {
+    val index = UInt(log2Up(widthOf(bits)+1) bit)
+    index := U(widthOf(bits))
+
+    def walkLowBits(width: Int, n: Int): Seq[BigInt] = {
```
andreasWallner, line 36: there is no functional reason to define `walkLowBits` inside `indexOfBoolN` - I looked for a bit which variable of the closure it used...
Also: a short comment about the sequence it generates would have been nice ;-)

andreasWallner (approved):
LGTM, only two small notes

lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4StreamSparseCompactor.scala
```diff
@@ -29,51 +29,48 @@ class Axi4StreamSparseCompactor(config: Axi4StreamConfig) extends Component {
...
   io.axis_m << outStage
 
-  def indexOfBoolN(bools: Seq[Boolean], n: Int): Int = {
-    var boolCount = 0
-    for ((b, i) <- bools.zipWithIndex) {
-      if (b) {
-        boolCount += 1
+  def indexOfBoolN(bits: Bits, n: Int): UInt = {
+    val index = UInt(log2Up(widthOf(bits)+1) bit)
+    index := U(widthOf(bits))
```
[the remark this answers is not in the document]
kleinai: Done

lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4StreamSparseCompactor.scala
kleinai: I originally had the function there to keep from cluttering the higher scope.  I've moved it out and commented its usage and output.

andreasWallner (approved):

PR #1043, opened 2023-02-23
[Fix] fix typo

chenbo-again:
Closes #

a small typo which can be ignored because of implicit conversion.

# Impact on code generation
nothing

# Checklist

- [ -] Unit tests were added
- [ -] API changes are or will be documented:


kleinai:
Yes this could be a problem if the offset is wider than a standard Int. Thanks for the PR!

PR #1051, opened 2023-03-05
Fix subdivideIn(..., n bit) for corner cases

### LINKED ISSUE #1049: subdivideIn function gets wrong width of the slices
hi, i tried to split the `val i = Bits(8 bits)` with `i.subdivideIn(5 bits, false)`, 
which i was expecting to get a `Vec` like this: `Vec(Bits(5 bits), Bits(3 bits))`.

but the SpinalHDL will actually split it into `Vec(Bits(4 bits), Bits(4 bits))`.

here is the code i tested:

```scala
  val io = new Bundle {
    val i = in Bits(8 bits)
    val o = out Vec(Bits(5 bits), 2)
  }
  val temp = Vec(Bits(5 bits), Bits(3 bits))
  val tmp = io.i.subdivideIn(5 bits, false)
  io.o(0) := tmp(0).resized
  io.o(1) := tmp(1).resized
}
```

and the generated `.v` file:

```verilog
// Generator : SpinalHDL v1.8.0b    git head : 761a30e521263983ddf14de3592f7a9f38bf0589
// Component : Test
// Git hash  : 98857617cfac48aac841f1f5f0fd2e4e33b6b0d3

`timescale 1ns/1ps

module Test (
  input      [7:0]    io_i,
  output     [4:0]    io_o_0,
  output     [4:0]    io_o_1
);

  wire       [4:0]    temp_0;
  wire       [2:0]    temp_1;
  wire       [3:0]    tmp_0;
  wire       [3:0]    tmp_1;

  assign tmp_0 = io_i[3 : 0]; // @ BaseType.scala l299
  assign tmp_1 = io_i[7 : 4]; // @ BaseType.scala l299
  assign io_o_0 = {1'd0, tmp_0}; // @ CodeGen.scala l31
  assign io_o_1 = {1'd0, tmp_1}; // @ CodeGen.scala l32

endmodule
```

it seems that in [https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/BitVector.scala#L322](https://github.com/SpinalHDL/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/BitVector.scala#L322), 
only calculate the `sliceCount` is not enough to get the correct width of the slice, 
and will split the 8 bits to two 4 bits in my situation.

andreasWallner:
When splitting 8 into 5 the old algorithm would generate an inconsistent result with other cases as it would generate two slices of 4 bit. In most other cases subdivideIn would generate the expected width as long as possible, and have a smaller last element.

If this is actually expected behavior I'd update the documentation to mention it.

Closes #1049 

# Impact on code generation

None

# Checklist

- [x] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/BitVector.scala
+++ b/core/src/main/scala/spinal/core/BitVector.scala
@@ -299,27 +299,41 @@ abstract class BitVector extends BaseType with Widthable {
   }
 
   /**
-    * Split the BitVector into x slice
-    * @example {{{ val res = myBits.subdiviedIn(3 slices) }}}
-    * @param sliceCount the width of the slice
-    * @return a Vector of slices
-    */
-  def subdivideIn(sliceCount: SlicesCount, strict : Boolean): Vec[T] = {
-    require(!strict || this.getWidth % sliceCount.value == 0)
-    val sliceWidth = widthOf(this) / sliceCount.value + (if(widthOf(this) % sliceCount.value != 0) 1 else 0)
-    val w = getWidth
-    Vec((0 until sliceCount.value).map(i => this(i * sliceWidth, (w-i * sliceWidth) min sliceWidth bits).asInstanceOf[T]))
+   * Split the BitVector into x slice
+   *
+   * @example {{{ val res = myBits.subdivideIn(3 slices) }}}
+   * @param sliceCount the width of the slice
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks :)

PR #1052, opened 2023-03-05
Initial state for timeout

andreasWallner:

Add an `initialState` parameter to `Timeout` so that we can start out with an already elapsed timeout.
This comes in handy if we use the timeout to e.g. block some action, but want to be accepting directly after start.

# Impact on code generation

No impact on code generation, change is backwards compatible.

# Checklist

  - on [RTD](https://github.com/SpinalHDL/SpinalDoc-RTD)? -> coming if accepted

### DIFF
--- a/lib/src/main/scala/spinal/lib/Utils.scala
+++ b/lib/src/main/scala/spinal/lib/Utils.scala
@@ -582,15 +582,19 @@ class Counter(val start: BigInt,val end: BigInt) extends ImplicitArea[UInt] {
 }
 
 object Timeout {
-  def apply(cycles: BigInt) : Timeout = new Timeout(cycles)
-  def apply(time: TimeNumber) : Timeout = new Timeout((time*ClockDomain.current.frequency.getValue).toBigInt)
-  def apply(frequency: HertzNumber) : Timeout = Timeout(frequency.toTime)
+  def apply(cycles: BigInt): Timeout = apply(cycles, initialState = false)
+  def apply(time: TimeNumber): Timeout = apply(time, initialState = false)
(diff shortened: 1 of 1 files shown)


andreasWallner:
I just realized that I could also use `init` like in `val t = Timeout(100) init True` which works, but has some quite misleading verilog output:
```
case class TestX() extends Component {
  val o = out Bool()
  val t = Timeout(16) init True
  o := t
}
```
which will generate
```
...
    if(reset) begin
      _zz_o <= 1'b0;
      _zz_3 <= 4'b0000;
      _zz_o <= 1'b1;
...
```
where both of the initializations are visible. This might actually be the better solution...

Readon:
> I just realized that I could also use `init` like in `val t = Timeout(100) init True` which works, but has some quite misleading verilog output:
> 
> ```
> case class TestX() extends Component {
>   val o = out Bool()
>   val t = Timeout(16) init True
>   o := t
> }
> ```
> 
> which will generate
> 
> ```
> ...
>   always @(posedge clk or posedge reset) begin
>     if(reset) begin
>       _zz_o <= 1'b0;
>       _zz_3 <= 4'b0000;
>       _zz_o <= 1'b1;
> ...
> ```
> 
> where both of the initializations are visible. This might actually be the better solution...

I think the init value is for `state` register in Timeout not the `counter`.
Because Timeout(16) returns it's `state` as implicit value which is a register.

andreasWallner:
@Readon yes it is, but that is also what I'd be initializing with my PR?
I'm still unsure about whether this PR makes sense - on the one hand it would produce cleaner HDL, on the other hand the functionality is already there and I could just ignore the HDL ;-)

andreasWallner:
I think that using init is a better solution then the change here, closing.

PR #1036, opened 2023-02-13
Move cocotb outputs to seperate folder

andreasWallner:
This is currently a try for #1029, since I can't easily get cocotb to work on my machine.

# Context, Motivation & Description

Less clutter in the root directory

# Impact on code generation

None

# Checklist

- ~[ ] Unit tests were added~
- ~[ ] API changes are or will be documented:~

PR #1053, opened 2023-03-06
Retide the generated files by tester/test cases.

### LINKED ISSUE #1029: `sbt test` generates .bin, .v and .vhd files in repository root folder
When running
``` bash
sbt test
```
for the library while developing.

files like RomTester3.v_toplevel_rom_symbol0.bin would be placed in project directory.
and "rm: cannot remove 'tester/src/test/python/spinal/RomTester3/RomTester3.v_toplevel_rom_symbol0.bin': No such file or directory" is reported.

There are also many .v and .vhd files in the project's root.
I guess the new CI test code does not solve the path for generated files.

@numero-744

Readon:
Closes #1029 

1. Avoid to generate intermediate files in the root directory while running "sbt test".
2. Move temp files for cocotb into individual cocotbWorkspace directory.
3. Create a SpinalAnyFunSuite to wrap all spinal tester specific default settings.
4. Expose SpinalTesterCocotbBase and SpinalTesterGhdlBase to user who can inherit and use it if they want to use cocotb and ghdl correspondingly.

A lot of changes are based on @andreasWallner 's PR #1036. Thanks a lot.

# Impact on code generation

None

# Checklist

- [ ] ~~Unit tests were added~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,7 @@ sim/xsim.dir/
 *.lst
 
 simWorkspace/
+cocotbWorkspace/
 tmp/
 null/
 command
--- a/core/src/main/scala/spinal/core/Spinal.scala
+++ b/core/src/main/scala/spinal/core/Spinal.scala
@@ -131,7 +131,7 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         defaultConfigForClockDomains   : ClockDomainConfig = ClockDomainConfig(),
                         onlyStdLogicVectorAtTopLevelIo : Boolean = false,
                         defaultClockDomainFrequency    : IClockDomainFrequency = UnknownFrequency(),
-                        targetDirectory                : String = ".",
+                        targetDirectory                : String = SpinalConfig.defaultTargetDirectory,
                         oneFilePerComponent            : Boolean = false,
                         netlistFileName                : String = null,
                         dumpWave                       : DumpWaveConfig = null,
@@ -262,6 +262,8 @@ object SpinalConfig{
       case None         => ???
     }
   }
+
+  var defaultTargetDirectory: String = "."
 }
 
 
--- a/core/src/main/scala/spinal/core/Spinal.scala
+++ b/core/src/main/scala/spinal/core/Spinal.scala
@@ -268,7 +268,7 @@ object SpinalConfig{
     }
   }
 
-  var defaultTargetDirectory: String = "."
+  var defaultTargetDirectory: String = System.getenv().getOrDefault("SPINAL_TARGET_DIR", ".")
 }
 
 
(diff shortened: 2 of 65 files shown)


andreasWallner:
Thanks for continuing this, I was also playing around and modified the changes to the `SpinalTesterGhdlBase` a bit more: GHDL will ignore the workspace for executables - currently the only way to get them into the directoy is via executing from there (checked that with the GHDL crowd).
So my additional change modified the `SpinalTesterGhdlBase` such that GHDL is always run in the `workspaceRoot` (+ a few typo fixes - especially if we want advertise these to users)
I can't modify you PR, so I opened a PR to your repo... https://github.com/Readon/SpinalHDL/pull/1

Readon:
It seems that this works, anyone has concerns about these changes? @Dolu1990 
Especially, the SpinalAnyFunSuite definition. That class is expected to be used in tester/test always.

Dolu1990:
Thanks :)

So, i would say https://github.com/SpinalHDL/SpinalHDL/pull/1053/files#diff-2ee64c46c6773621611f3c349e88c170f4c4936af74ca242a0cba5e9b80192a1R271 is tricky (global var)

The best maybe instead that https://github.com/SpinalHDL/SpinalHDL/pull/1053/files#diff-b24e8f0cef8d42962a85e30fbf6c73b7534412321c9d567a29efbe3835af4556R8 provide a function with a preinitialized instance of SpinalConfig ?

Readon:
> 

I have tried many ways to solve this, the problem is that there are cases of using SpinalConfig there with various arguments, which make the definition extremely complex. I think there are commits for that try.

Readon:
The previous version is [here](https://github.com/SpinalHDL/SpinalHDL/pull/1053/commits/a0343048c69871628e6ba62cf815b5895cb00676#diff-3d91de9d0d5f58e34ebc822225729c5ab047bcc324ab14f87559d8189619adfeR10)

Dolu1990:
Hmm
What's about using environnement variables ?

Readon:
> 

The problem is that if environment variable is using as default target directory, the package in tester could not take effect when environment variable is set.
Here is the argument targetDirectory's definition.
```
targetDirectory                : String = System.getenv().getOrDefault("SPINAL_TARGET", "."),
```
Here is the define in tester/package.scala
```
System.setProperty("SPINAL_TARGET", simWorkspacePath)
```

However, it does not work.

Dolu1990:
> do you have better way to change the environment variable of the same process? @Dolu1990

No, not realy.

> SpinalConfig.defaultTargetDirectory

So, fondamentaly, would have been better to avoid it, as it is kind of a cursed design, but at the end, as it is very corner case, i think that's fine, can be merged.

Feel free to merge when all good for you.

PR #1056, opened 2023-03-13
fix bugs that cause engines missing while copy.

Readon:

While do formal verification with different engine like z3.
The existing engines might not be copied for the object who calls compiledCloned method, in compile function of SpinalFormalConfig class.
So the default yices engine are always used, which act differently with the original design.

# Impact on code generation

# Checklist

- [ ] Unit tests were added
~~- [ ] API changes are or will be documented:~~


Readon:
done by #1055 , this is not necessary.

PR #1059, opened 2023-03-18
Set the correct reset signal type (polarity) in QSys IP file

tlupick:
The current QSysify only identifies the reset signal in the HW IP as an active high reset, even if the reset is active low.
This will cause Quartus Platform Designer to erroneously insert reset inverters if the top level reset signal is an active low reset.

This change selects the correct reset signal type based on the clock domain config.

# Impact on code generation
Emits the correct reset signal type in the HW IP tcl file.

# Checklist

- [ ] Unit tests were added


Dolu1990:
Thanks :D

PR #1064, opened 2023-03-26
[Fix] polish AhbLite3Interconnect example in comment which can not co…

chenbo-again:
there are many examples in comment, I think we should make these example compilable, but in this file, the comment can not compile because the compiler can't find ahbSlaves(0), it can only find io.ahbSlaves(0).

# Impact on code generation
nothing

# Checklist

- [-] Unit tests were added
- [-] API changes are or will be documented:


Dolu1990:
thanks :)

PR #1065, opened 2023-03-26
Fix cocotb test cases

andreasWallner:
Any deposit to a pad signal breaks the analog tests since these act as forces on the signal. Therefore driving 'z' fixes the pad level to 'z'. `Release` also does not improve the situation - so simply never touch those signals.


Dolu1990:
<3

PR #1073, opened 2023-03-31
Iconnectable

### LINKED ISSUE #1071: BundleWithAssign trait
I always use `<<` or `>>` to connect the bus. However, I must to implement these functions in my custom bundles.
I tried to design a trait.
```scala
trait BundleWithAssignTrait[T <: BundleWithAssignTrait[T]] {
  def connectFrom(that: T): T
  def <<(that: T): T = connectFrom(that)
  def >>(into: T): T = {
    into << this.asInstanceOf[T]
    into
  }
}
```
It worked for me, and I want to open a PR for this trait.

xueweiwujxw:
Closes #1071 

Add IConnectable trait to connect bus

Example:
```scala
object BundleWithAssignTraitTest extends App {
  case class TestBundle()
      extends Bundle
      with IMasterSlave
      with IConnectable[TestBundle] {
    val valid = Bool
    val ready = Bool

    def asMaster(): Unit = {
      inWithNull(ready)
      outWithNull(valid)
    }

    def connectFrom(that: TestBundle): TestBundle = {
      that.ready := this.ready
      this.valid := that.valid
      that
    }
  }

  case class TestModule() extends Component {
    val io = new Bundle {
      val s = slave(TestBundle())
      val m = master(TestBundle())
    }

    val save = Reg(TestBundle())
    io.s >> save
    io.m << save
  }

  SpinalConfig().generateVerilog(TestModule())
}
```
Generated Verilog code:
```Verilog
// Generator : SpinalHDL v1.7.2    git head : 08fc866bebdc40c471ebe327bface63e34406489
// Component : TestModule
// Git hash  : ff0242cd879a6e8623afd4a1817f525fb45aada3

`timescale 1ns/1ps

module TestModule (
  input               io_s_valid,
  output              io_s_ready,
  output              io_m_valid,
  input               io_m_ready,
  input               clk,
  input               reset
);

  reg                 save_valid;
  reg                 save_ready;

  assign io_s_ready = save_ready;
  assign io_m_valid = save_valid;
  always @(posedge clk) begin
    save_valid <= io_s_valid;
    save_ready <= io_m_ready;
  end


endmodule
```
This PR designs a trait to connect bus using `>>` and `<<`. 
Users do not need to implement these functions for custom bundles

# Impact on code generation

It is a feature and it would not impact the code without IConnectable trait.

# Checklist

- [ ] Unit tests were added


Readon:
How about define the IMasterSlave a child of IConnectable?
There should be always to support "<<" and ">>" for IMasterSlave.

Readon:
One more thing, the ```save``` in the example is not a Stream and have been converted automatically into a stream.
This might be convenient and also might cause problem while there is no solid check on that. 
After all, explicit is better than implicit.
``` scala
val save = Reg(TestBundle())
```

kleinai:
> One more thing, the `save` in the example is not a Stream and have been converted automatically into a stream. This might be convenient and also might cause problem while there is no solid check on that. After all, explicit is better than implicit.
> 
> ```scala
> val save = Reg(TestBundle())
> ```

The example code bundle _is_ a stream so this is approximately a register stage.

I do find myself trying to use `<<` as if it should magically work for all master/slave interfaces.

If we're making this change this should also bring in existing streams and buses too?  Or maybe a different issue for to track that?

xueweiwujxw:
> How about define the IMasterSlave a child of IConnectable? There should be always to support "<<" and ">>" for IMasterSlave.

I think it's a good idea because I just want to conveniently connect the bus. 
```scala
val save = Reg(TestBundle())
```
And it is indeed an inappropriate example. Perhaps an example that connects two components would be better.

xueweiwujxw:
upgrade the example
```scala
object IMasterSlaveConnectableTest extends App {
  case class TestBundle()
      extends Bundle
      with IMasterSlaveConnectable[TestBundle] {
    val valid = Bool
    val ready = Bool

    def asMaster(): Unit = {
      inWithNull(ready)
      outWithNull(valid)
    }

    def connectFrom(that: TestBundle): TestBundle = {
      that.ready := this.ready
      this.valid := that.valid
      that
    }
  }

  case class TestModule() extends Component {
    val io = new Bundle {
      val s = slave(TestBundle())
      val m = master(TestBundle())
    }

    val tmp = TestBundle()
    io.s >> tmp
    io.m << tmp
  }

  SpinalConfig().generateVerilog(TestModule())
}

```

xueweiwujxw:
> > One more thing, the `save` in the example is not a Stream and have been converted automatically into a stream. This might be convenient and also might cause problem while there is no solid check on that. After all, explicit is better than implicit.
> > ```scala
> > val save = Reg(TestBundle())
> > ```
> 
> The example code bundle _is_ a stream so this is approximately a register stage.
> 
> I do find myself trying to use `<<` as if it should magically work for all master/slave interfaces.
> 
> If we're making this change this should also bring in existing streams and buses too? Or maybe a different issue for to track that?

I found that the existing streams and buses basically have implementations of connectFrom, `<<`, and `>>` functions.
This is also the inspiration behind my design of this trait.

kleinai:
> I found that the existing streams and buses basically have implementations of connectFrom, `<<`, and `>>` functions. This is also the inspiration behind my design of this trait.

Yes, I think it would be good to standardize the usage of `<<` and `>>` as a pattern.

In reference to Readon's point, what happens if we make IMasterSlave extend IConnectable rather than the other way around? I don't think there's a reason not to make bus connection a feature of IMasterSlave.

andreasWallner:
Maybe a dumb question, but why not use `<>` for the connection? that already works and there is not need to write any code for the Bundle author?

```scala
object BundleWithAssignTraitTest extends App {
  case class TestBundle()
    extends Bundle
      with IMasterSlave {
    val valid = Bool()
    val ready = Bool()

    def asMaster(): Unit = {
      in(ready)
      out(valid)
    }
  }

  case class TestModule() extends Component {
    val io = new Bundle {
      val s = slave(TestBundle())
      val m = master(TestBundle())
    }

    val save = Reg(TestBundle())
    io.s <> save
    io.m <> save
  }

  SpinalConfig().generateVerilog(TestModule())
}
```
generates
```
// Generator : SpinalHDL dev    git head : 303735f207ce73a08d4dbb062e19b5e80425c13d
// Component : TestModule
// Git hash  : 5dab5ea88b968d30abbc1c87929907925133436b

`timescale 1ns/1ps

module TestModule (
  input               io_s_valid,
  output              io_s_ready,
  output              io_m_valid,
  input               io_m_ready,
  input               clk,
  input               reset
);

  reg                 save_valid;
  reg                 save_ready;

  assign io_s_ready = save_ready;
  assign io_m_valid = save_valid;
  always @(posedge clk) begin
    save_valid <= io_s_valid;
    save_ready <= io_m_ready;
  end


endmodule
```

xueweiwujxw:
> > I found that the existing streams and buses basically have implementations of connectFrom, `<<`, and `>>` functions. This is also the inspiration behind my design of this trait.
> 
> Yes, I think it would be good to standardize the usage of `<<` and `>>` as a pattern.
> 
> In reference to Readon's point, what happens if we make IMasterSlave extend IConnectable rather than the other way around? I don't think there's a reason not to make bus connection a feature of IMasterSlave.

I have pushed a commit 7a39e81b202a2d619bf8c253e0f71202e93eaf70 to extend IMasterSlave.

xueweiwujxw:
> Maybe a dumb question, but why not use `<>` for the connection? that already works and there is not need to write any code for the Bundle author?
> 
> ```scala
> object BundleWithAssignTraitTest extends App {
>   case class TestBundle()
>     extends Bundle
>       with IMasterSlave {
>     val valid = Bool()
>     val ready = Bool()
> 
>     def asMaster(): Unit = {
>       in(ready)
>       out(valid)
>     }
>   }
> 
>   case class TestModule() extends Component {
>     val io = new Bundle {
>       val s = slave(TestBundle())
>       val m = master(TestBundle())
>     }
> 
>     val save = Reg(TestBundle())
>     io.s <> save
>     io.m <> save
>   }
> 
>   SpinalConfig().generateVerilog(TestModule())
> }
> ```
> 
> generates
> 
> ```verilog
> // Generator : SpinalHDL dev    git head : 303735f207ce73a08d4dbb062e19b5e80425c13d
> // Component : TestModule
> // Git hash  : 5dab5ea88b968d30abbc1c87929907925133436b
> 
> `timescale 1ns/1ps
> 
> module TestModule (
>   input               io_s_valid,
>   output              io_s_ready,
>   output              io_m_valid,
>   input               io_m_ready,
>   input               clk,
>   input               reset
> );
> 
>   reg                 save_valid;
>   reg                 save_ready;
> 
>   assign io_s_ready = save_ready;
>   assign io_m_valid = save_valid;
>   always @(posedge clk) begin
>     save_valid <= io_s_valid;
>     save_ready <= io_m_ready;
>   end
> 
> 
> endmodule
> ```
> 
> (also works for the updated example w/o the register)

You are right, in most cases, `<>` works fine. However, there are some situations where I need to customize the connection method and autoconnect cannot fully meet my needs.
The Bundle I defined may have different instantiation methods and inconsistent elements, `<>` can not work.

Dolu1990:
@andreasWallner 

> but why not use <> for the connection?

It works only when at least one of the operands are an io of a Component.
If both are internal signal, it can't figure out what to connect to what.

-- 

About IMasterSlaveConnectable[TestBundle] shouldn't it be a class which "extends IMasterSlave with IConnectable" ?

kleinai:
A good question is, are there cases were someone might want `IMasterSlave` without `IConnectable`?

Dolu1990:
legacy code, to not break (maybe ?)
Or if somebody use iMasterSlave for other usage than for a Data / Bundle maybe

andreasWallner:
@xueweiwujxw 

I think there was just one remaining question by @Dolu1990 

> About IMasterSlaveConnectable[TestBundle] shouldn't it be a class which "extends IMasterSlave with IConnectable" ?

this might have been a bit hidden 
would be cool if we could get this in

xueweiwujxw:
@andreasWallner 
Yeah. I committed a trait in `core/src/main/scala/spinal/core/Bundle.scala`
```scala
trait IConnectable[T <: IConnectable[T]] {
  def connectFrom(that: T): T
  def <<(that: T): T = connectFrom(that)
  def >>(into: T): T = {
    into << this.asInstanceOf[T]
    into
  }
}
```

Readon:
> @andreasWallner Yeah. I committed a trait in `core/src/main/scala/spinal/core/Bundle.scala`
> 
> ```scala
> trait IConnectable[T <: IConnectable[T]] {
>   def connectFrom(that: T): T
>   def <<(that: T): T = connectFrom(that)
>   def >>(into: T): T = {
>     into << this.asInstanceOf[T]
>     into
>   }
> }
> ```

I suppose what they mean is not to create a new IMasterSalveConnectable trait, only IConnectable.
If user want to use it, they can define class with `extends IMasterSlave with IConnectable`.

Maybe the next step is to propagate the IConnectable to the whole lib.

xueweiwujxw:
> I suppose what they mean is not to create a new IMasterSalveConnectable trait, only IConnectable. If user want to use it, they can define class with `extends IMasterSlave with IConnectable`.
> 
> Maybe the next step is to propagate the IConnectable to the whole lib.

Oh, maybe  I should remove the IMasterSlaveConnectable?

Dolu1990:
Thanks :)

Dolu1990:
And sorry for how long it toke XD

Readon:
Next step, might be unify existing bus to extend from this trait.

Dolu1990:
yes

Issue #870, opened 2022-09-22
Way too big signal issue


cheerryio:
I have the need to define a signal of large width(which is bigger than 4096), but I get this error ```Way too big signal (toplevel/InWrBufData : in Bits[5120 bits])```. How am I able to disable this phase check "PhaseInferWidth"?

Dolu1990:
Hi, i pushed a open in dev : 

SpinalConfig(bitVectorWidthMax = 5000).generateVerilog( ... )
Should be good ^^

PR #1074, opened 2023-03-31
fix check for wide signals

jonnykl:
Fixes a bug introduced in the commit e352b88bcc9ebbb04e0a80a24597d2fd7814f76e. The check was only updated at one of two locations.

See also #870


Dolu1990:
Nice, thanks :D

PR #1075, opened 2023-04-02
PhysicalUnit formatting

andreasWallner:
Make both HertzNumber and TimeNumber consistent that both have `decompose` / `decomposeString` (missing on HertzNumber before). Also make both formattable via f-interpolation, `String.format`, etc.

Simpler use when printing/reporting/building error messages.

# Impact on code generation

No impact on code generation, but code that printed the classes w/o any formatting will now generate different output. This should not be an issue though because the format was unspecific and not well specified before.

# Checklist

- [x] Unit tests were added


Dolu1990:
<3

Dolu1990:
Fell free to merge

PR #1079, opened 2023-04-08
Generate diagrams based on SpinalHDL code

ZhaokunHu:
This pull request attempts to implement the idea  #635 to automatically generate diagrams based on SpinalHDL code. I have created a new function called GenerateDiagrams that can be used in the following way:
```scala
HDElkDiagramGen(SpinalVerilog(new _))
```
This function will generate a new file named _.html .
If you have an internet connection, you can simply open the HTML file and the dependencies will be loaded automatically from the specified URLs. You can see several diagrams in the pages.
These diagrams are generated by the HDElk tool.(https://github.com/davidthings/hdelk ).
If you do not have an internet connection, you need to download the JS packages from this library and place them in the same folder as your HTML file. Then, according to the readme of this library, you need to add the dependencies to your HTML file. Then you can  open the HTML file to see the diagrams.

Example:
```scala
import spinal.core._
import spinal.lib._

case class FilterConfig(iqWidth: Int,
                        tapNumbers: Int = 33,
                        hwFreq: HertzNumber = 200 MHz,
                        sampleFreq: HertzNumber = 1.92 MHz)

class Filter(fc: FilterConfig) extends Component{
  val din   = slave Flow(Bits(32 bits))
  val dout  = master Flow(Bits(32 bits))
  val flush = in Bool()

  val clockSMP = ClockDomain.external("smp")
  val clockHW = ClockDomain.external("hw")

  val u_fifo_in = StreamFifoCC(
    dataType = Bits(32 bits),
    depth = 8,
    pushClock = clockSMP,
    popClock = clockDomain
  )

  u_fifo_in.io.push << din.toStream
  dout << u_fifo_in.io.pop.toFlow

}

object Filter {
    val fc = FilterConfig(8)
    HDElkDiagramGen(SpinalVerilog(new Filter(fc)))
  }
}
```
It will generate this diagram:
image
I have also tested the GenerateDiagrams class on some more complex SpinalHDL projects to ensure the accuracy of the generated diagrams.

One issue that currently exists is that if you use many complex logic operations in your project, the system will automatically generate many registers to implement these operations, and the generated diagrams may become messy as a result.I have tried various methods to simplify the generated diagrams, but the results have been less than satisfactory.Please let me know if you have any thoughts or suggestions on this approach.

# Impact on code generation
None

# Checklist

- [x] Unit tests were added


Readon:
Good work, Zhaokun.
However, there are still some points could be improved.
1. Would it better to put the file into directory "lib/src/main/scala/spinal/lib/eda/", which represent it is a tool for eda like funciton.
2. About the file name GenerateDiagram is really a board topic, how about HDElkDiagramGen or something more specifical, so that diagram generator using other tech could be named properly.
3. there are still lint errors reported which have been give out at [file list](https://github.com/SpinalHDL/SpinalHDL/pull/1079/files). You can run the check locally on your machine first.  :)
4. The function dealAllSignals is really too huge to review and understand. Simplify it would help to improve the quality of code.
5. On the other hand, the function name dealAllSignals is really vague and it would be better to have a specific name.
6. It would be nice if you could provide some examples of how to use it as a simulation in the tester subpackage.

kleinai:
This is impressive!

I agree this should be under `lib/eda` or `lib/tools` which is where `ModuleAnalyzer` is.

`dealAllSignals` could be several smaller functions to make understanding the flow better. It looks like a candidate for abstracting so that other diagram backends could use it. Not that it has to happen in this PR.

Small suggestion, if the design has a fixed frequency maybe we could put that on the diagram too?

Readon:
> This is impressive!
> 
> I agree this should be under `lib/eda` or `lib/tools` which is where `ModuleAnalyzer` is.
> 
> `dealAllSignals` could be several smaller functions to make understanding the flow better. It looks like a candidate for abstracting so that other diagram backends could use it. Not that it has to happen in this PR.
> 
> Small suggestion, if the design has a fixed frequency maybe we could put that on the diagram too?

lib/tools would be better.

Readon:
one more thing, I think the clk is not necessarily connect visibly. There is already clockdomain color though.

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
```diff
@@ -0,0 +1,510 @@
...
+    val fileName = rtl.toplevelName + ".html"
+    val file = new File(fileName)
+    val pw = new FileWriter(file)
+    val builder = new StringBuilder()
+    builder ++=
+      s"""<!DOCTYPE html>
+         |<html>
+         |<head>
+         |    <meta charset="UTF-8">
+         |    <title>RTL连接图</title>
```
andreasWallner, line 427: Please keep with the remainder of SpinalHDL and use an English title

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
```diff
@@ -0,0 +1,510 @@
...
+         |   }
+         |	}
+         |	  goTop.onclick=function(){
+         |		window.scrollTo({
+         |			top:0,
+         |			behavior:"smooth"
+         |		})
+         |	}
+         |</script>
+         |<script src="/js/elk.bundled.js"></script>
```
andreasWallner, line 486: I had to change these paths to relative ones (w/o the leading `/`), otherwise my browser would not be able to find them when opening the file locally.

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
```diff
@@ -0,0 +1,510 @@
...
+}
+/** Integrating all methods to draw an image*/
+
+class GenerateDiagrams(rtl: SpinalReport[Component]) {
+  def beginGenerate(): Unit = {
+    val module = rtl.toplevel
+    val fileName = rtl.toplevelName + ".html"
+    val file = new File(fileName)
+    val pw = new FileWriter(file)
+    val builder = new StringBuilder()
```
andreasWallner, line 421: You have a `writer` already, why not use that directly (and also pass it to the sub-components for use there?) There is no need to build the string fully in memory?

andreasWallner (commented):
This looks quite promising, I added a few small comments below that I stumbled over when trying it out/skimming over it.

Apart from that:
I played around a bit, and e.g. this code:
```scala
case class Inner() extends Component {
  val io = new Bundle {
    val i = in port Bool()
    val o = out port Bool()
  }
  io.o := RegNext(io.i)
}
```
will generate this image:
image
with the output of the register not being connected.

Also this:
```scala
case class DiagramDemo() extends Component {
  val io = new Bundle {
    val i = in port Bool()
    val i2 = in port Bool()
    val sel = in port Bool()
    val o = out port Bool()

    val i2b = in port Bits(2 bit)
    val muxout = out port Bool()

    val regout = out port Bool()
  }
  io.muxout := io.i2b.mux(
    0 -> (io.i & io.i2),
    1 -> (io.i | io.i2),
    2 -> (io.i ^ io.i2),
    3 -> (io.i)
  )
  io.regout := RegNextWhen(io.i, io.sel)
  io.o := False
  when(io.i) { io.o := True }
}
```
generates 
image
Where we see a similar thing with the outputs not being connected, and the assignments to `io.o` missing completely. 

Nitpick: The blue on blue colorscheme of the examples above is kind of hard to read if you ask me.

PS: please also run scalafmt on the code.

ZhaokunHu:
> This looks quite promising, I added a few small comments below that I stumbled over when trying it out/skimming over it.
> 
> Apart from that: I played around a bit, and e.g. this code:
> 
> ```scala
> case class Inner() extends Component {
>   val io = new Bundle {
>     val i = in port Bool()
>     val o = out port Bool()
>   }
>   io.o := RegNext(io.i)
> }
> ```
> 
> will generate this image: image with the output of the register not being connected.
> 
> Also this:
> 
> ```scala
> case class DiagramDemo() extends Component {
>   val io = new Bundle {
>     val i = in port Bool()
>     val i2 = in port Bool()
>     val sel = in port Bool()
>     val o = out port Bool()
> 
>     val i2b = in port Bits(2 bit)
>     val muxout = out port Bool()
> 
>     val regout = out port Bool()
>   }
>   io.muxout := io.i2b.mux(
>     0 -> (io.i & io.i2),
>     1 -> (io.i | io.i2),
>     2 -> (io.i ^ io.i2),
>     3 -> (io.i)
>   )
>   io.regout := RegNextWhen(io.i, io.sel)
>   io.o := False
>   when(io.i) { io.o := True }
> }
> ```
> 
> generates image Where we see a similar thing with the outputs not being connected, and the assignments to `io.o` missing completely.
> 
> Nitpick: The blue on blue colorscheme of the examples above is kind of hard to read if you ask me.
> 
> PS: please also run scalafmt on the code.
I'm sorry for this silly bug. I accidentally deleted a pair of {} while removing the extra ones, which caused the program to produce incorrect results.I have fixed it and now the result is correct now.
image
image
I am a beginner in using Scala and there are some usages that I am not familiar with, which causes issues in my code writing. Thank you for your advice, I will carefully modify my program according to your suggestions.

ZhaokunHu:
I have made modifications to the program and fixed some problems based on your suggestions .And  I placed it in the spinal.lib.tools package. I have also added a test file which will generate an .html file. You will need to place the js package in https://github.com/davidthings/hdelk 
at the same directory as the HTML file, and then you can open the HTML file using a web browser.

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
ZhaokunHu: I delete the leading" / "and now it is ok.

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
ZhaokunHu: Sorrry,i forgot to change it.I have changed it to "RTL diagrams based on SpinalHDL code"

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
```diff
@@ -0,0 +1,510 @@
...
+    val fileName = topLevelName + ".html"
+    val file = new File(fileName)
+    val pw = new FileWriter(file, true)
+    pw.write(builder.toString())
+    pw.close()
+  }
+}
+/** Integrating all methods to draw an image*/
+
+class GenerateDiagrams(rtl: SpinalReport[Component]) {
```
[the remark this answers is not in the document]
ZhaokunHu: Thanks for your advice.I also got an error when passing a report instance stored in a val but didn't know how to fix it.Now i can fix this problem with your help.

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
ZhaokunHu: I changed it based on your advice.

Readon:
> I have made modifications to the program and fixed some problems based on your suggestions .And I placed it in the spinal.lib.tools package. I have also added a test file which will generate an .html file. You will need to place the js package in https://github.com/davidthings/hdelk at the same directory as the HTML file, and then you can open the HTML file using a web browser.

This is an environment requirement. Just like we should install Quartus before we use Qsysnify.
As a library, SpinalHDL can only passively detect the environment but change it.
So it would be a way to have an environment variable "HDELK_PATH" which save the directory of required js files, thus it can be changed by user. 

It's default value can be set as "https://github.com/davidthings/hdelk", so that if user is online no manual copy is required.
Or if we push the js file into SpinalHDL library, the default link could be "https://rawcontents.github.com/SpinalHDL/SpinalHDL/xxxxxx/hdelk".

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,506 @@
...
+         |   }
+         |	}
+         |	  goTop.onclick=function(){
+         |		window.scrollTo({
+         |			top:0,
+         |			behavior:"smooth"
+         |		})
+         |	}
+         |</script>
+         |<script src="https://cdn.jsdelivr.net/gh/davidthings/hdelk@master/js/elk.bundled.js"></script>
```
Readon, line 481: This can be a default URL base, however, for users do not have network access might want to change this.
So use a variable is better here.:)

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
andreasWallner: That's better, we could also use the component name of the top component in the title e.g. `f"{rtl.toplevelName} RTL diagrams"`

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
andreasWallner: thats cool, should be a good default for most users - and simpler to use as well

lib/src/main/scala/spinal/lib/sim/GenerateDiagram.scala
andreasWallner: I think I explained poorly: What I meant was to use the writer directly, without the StringBuilder. The StringBuilder will piece everything together in RAM, whereas if you use the writer it can be flushed to disk. But you'd have to pass the writer to `GenerateOneDiagram.beginDraw(writer)` to do that.
(and if you want/need the output in a String you can still use `StringWriter` if you use the base `Writer` as the parameter type)

andreasWallner:
> I am a beginner in using Scala and there are some usages that I am not familiar with, which causes issues in my code writing. Thank you for your advice, I will carefully modify my program according to your suggestions.

No worries - stuff like that happens ;-)

I tested the latest version - the CDN links are much more comfortable than the local ones, but I agree with @Readon that it would be nice to have to option of passing an alternative path as a configuration.

I played around a bit more and stumbled over another small thing that's unexpected - it's cool that bundles are shown as a single signal, but accessing a signal from a bundle does not seem to get rendered:
```scala
case class ElkDemo() extends Component {
  val io = new Bundle {
    val axii = slave port Axi4(Axi4Config(addressWidth=32, dataWidth=32, useId=false))
    val axio = master port Axi4(Axi4Config(addressWidth=32, dataWidth=32, useId=false))

    val test = out port Bool()
    val testR = out port Bool() setAsReg()
  }
  io.axii >> io.axio
  io.test := io.axii.r.valid
  io.testR := io.axii.w.valid
}
```
produces
image

ZhaokunHu:
> > I am a beginner in using Scala and there are some usages that I am not familiar with, which causes issues in my code writing. Thank you for your advice, I will carefully modify my program according to your suggestions.
> 
> No worries - stuff like that happens ;-)
> 
> I tested the latest version - the CDN links are much more comfortable than the local ones, but I agree with @Readon that it would be nice to have to option of passing an alternative path as a configuration.
> 
> I played around a bit more and stumbled over another small thing that's unexpected - it's cool that bundles are shown as a single signal, but accessing a signal from a bundle does not seem to get rendered/rendered incorrectly (`test` should be connected directly like `testR` is now, and `testR` should be connected to the FF?):
> 
> ```scala
> case class ElkDemo() extends Component {
>   val io = new Bundle {
>     val axii = slave port Axi4(Axi4Config(addressWidth=32, dataWidth=32, useId=false))
>     val axio = master port Axi4(Axi4Config(addressWidth=32, dataWidth=32, useId=false))
> 
>     val test = out port Bool()
>     val testR = out port Bool() setAsReg()
>   }
>   io.axii >> io.axio
>   io.test := io.axii.r.valid
>   io.testR := io.axii.w.valid
> }
> ```
> 
> produces image


Thank you so much for testing for me! I have solved this bug. 
image

at the latest version:
1. Add a new label for the edges which connect bus to one signal. 
2. Using the writer instead of builder.
3. Change the page name to RTL diagrams of ${rtl.toplevelName}.

ZhaokunHu:
Now the situation where the output port is also used as a register can be displayed correctly. 
image

What's more, the color of the port has been added.
image

Readon:
> Now the situation where the output port is also used as a register can be displayed correctly. image
> 
> What's more, the color of the port has been added. image

Then, it's time to hide the clk & reset port, isn't it? :)

Readon:
Maybe it would be good to use black as default node color, and just distinguish the clock domains by port color only?

Readon:
> Great job! It's really pleasant to see the first circuit diagram generator application building upon `ModuleAnalyzer`/`DataAnalyzer` (my apology for still not having doc for them).
> 
> Things so far look good to me but still, I have a little advice. From an architectural point of view, the generator is tightly coupled with the SpinalHDL internal data model. The process of data transformation is like SpinalHDL data model -> Elk HTML format. A better way to make things more scalable or extensible is to add a middle layer, such as SpinalHDL data model -> JSON -> Elk HTML. The middle JSON format serves as a module interface database that holds all the necessary information for the generator to draw the diagram.
> 
> There will be two benefits if a middle layer is introduced.
> 
> 1. the generated interface database (e.g. a JSON file) can be reused by other tools which also need to elaborate the topology of the module.
> 2. your diagram generator can accept a universal format, not only for SpinalHDL. And this can be a more valuable open-source project.
> 
> Though, it's just a piece of advice and requires time and effort to reconstruct the project (I can see that it's nearly finished and not the early stage of the project). Thus, you can ignore my opinion and keep working. I'm very much looking forward to seeing the final product :)

Good idea! 
Unfortunately, in this stage, it would be hard to archieve that.
1. Hdelk need to be modified to support json load function, whoes difficulty is still not investigated.
2. Zhaokun has a tight schedule on his under graduation report, so that it's not able to support a huge modifications now.
Maybe we could do that later.

There are still problems to be solved, as soon as we tried to draw diagram for Spinal cord or um, whose top level also contains amount of logic inside.

ZhaokunHu:
> Great job! It's really pleasant to see the first circuit diagram generator application building upon `ModuleAnalyzer`/`DataAnalyzer` (my apology for still not having doc for them).
> 
> Things so far look good to me but still, I have a little advice. From an architectural point of view, the generator is tightly coupled with the SpinalHDL internal data model. The process of data transformation is like SpinalHDL data model -> Elk HTML format. A better way to make things more scalable or extensible is to add a middle layer, such as SpinalHDL data model -> JSON -> Elk HTML. The middle JSON format serves as a module interface database that holds all the necessary information for the generator to draw the diagram.
> 
> There will be two benefits if a middle layer is introduced.
> 
> 1. the generated interface database (e.g. a JSON file) can be reused by other tools which also need to elaborate the topology of the module.
> 2. your diagram generator can accept a universal format, not only for SpinalHDL. And this can be a more valuable open-source project.
> 
> Though, it's just a piece of advice and requires time and effort to reconstruct the project (I can see that it's nearly finished and not the early stage of the project). Thus, you can ignore my opinion and keep working. I'm very much looking forward to seeing the final product :)

I appreciate your PR very much. Without the ModuleAnalyzer/DataAnalyzer , this project would have been very difficult. 

However, since this project relies heavily on the internal structure of SpinalHDL, it would be difficult to reconstruct to support a universal format. Additionally, I am not yet familiar enough with JSON and Scala, so I may not be able to complete the reconstruction. I apologize for not being able to accept your suggestion.

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,552 @@
+package spinal.lib.tools
+
+
```
Readon, line 3: I think you should run scalafmt on this code to avoid fmt-lint reports.

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,552 @@
...
+  var outPorts: mutable.Set[ElkPort] = mutable.Set()
+  var children: mutable.Set[ElkNode] = mutable.Set()
+  var highlight = 0
+}
+
+class GenerateOneDiagram(module: Component, topLevelName: String, moduleName: String) {
+  /** Initializing data */
+  private val fileName = topLevelName + ".html"
+  private val file = new File(fileName)
+  private val pw = new FileWriter(file, true)
```
Readon, line 28: This pw could be named as writer?

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,552 @@
...
+    /** Adding the toplevel module and its input/output ports */
+    def findPortHighlight(thisPort:BaseType): Int = {
+      if(clkMap.contains(thisPort.getName())) clkMap(thisPort.getName())
+      else if(clkMap.contains(thisPort.clockDomain.toString())) clkMap(thisPort.clockDomain.toString())
+      else 1
+    }
+    for (topInOut <- topInOuts) {
+      if (haveParent(topInOut)) {
+        val rootParent = findParent(topInOut)
+        if (rootParent.flatten.head.isInput) topNode.inPorts.add(ElkPort(rootParent.getName(),findPortHighlight(topInOut)))
```
Readon, line 130: How about use pattern match here?

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,552 @@
...
+          clkMap.put(thisClk.reset.getName(), clkCounter)
+        if (thisClk.softReset != null)
+          clkMap.put(thisClk.softReset.getName(), clkCounter)
+      }
+    }
+    if (allClk.size > 1) topNode.highlight = 10
+    else if (allClk.size == 1) topNode.highlight = clkMap(allClk.head.toString())
+  }
+
+  private def GenAllNodes(): Unit = {
```
Readon, line 120: Maybe we should use a hierarchy class architecture to help to clean this logic.

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,552 @@
...
+        newNode.highlight = clkMap(topLevelRegister.clockDomain.toString())
+      if (!containedNode.contains(newNode.labelName)) {
+        topNode.children.add(newNode)
+        containedNode.add(newNode.labelName)
+      }
+    }
+  }
+
+  private def GenAllEdges(): Unit = {
+    /** Getting the sonName of net */
```
Readon, line 188: This is really too big to look over.

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,552 @@
...
+      }
+    }
+  }
+
+  /** Generating HDElk language for Node */
+  private def drawNodes(thisNode: ElkNode): Unit = {
+    pw.write(s"""{id:"${thisNode.labelName}",\n""")
+    if (thisNode.typeName != "") pw.write(s"""type:"${thisNode.typeName}",\n""")
+    if (thisNode.highlight != 0) pw.write(s"""highlight:${thisNode.highlight},\n""")
+    if (thisNode.inPorts.nonEmpty) {
```
Readon, line 387: How about moving this logic into port specific class? so all related logic can be put into that.
Above code seems identical while it is in or out.

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,599 @@
...
+    loop.breakable {
+      for (anyParent <- parentList) {
+        anyParent match {
+          case dataParent: Data =>
+            if (dataParent.getClass.getSimpleName != "") {
+              if (dataParent.getName() != "") {
+                returnParent = dataParent
+                loop.break()
+              } else {
+                dataParent.setName(
```
andreasWallner, line 80: The diagrammer should not modify that data model of the HDL code.

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,599 @@
...
+  private def drawClockDomains(): Unit = {
+    writer.write(s"""{id:"ClockDomains",\nchildren:[\n""")
+    for (element <- clkNamesMap) {
+      writer.write(s"""{id:"${element._1}",highlight:${element._2}},\n""")
+    }
+    writer.write(s"""]\n}\n""")
+  }
+
+  /** Integrating all methods to draw an image */
+  def beginDraw(): Unit = {
```
andreasWallner, line 476: You should pass the writer as a parameter instead of opening it for append during construction and closing it here.

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,599 @@
...
+            isContained = true
+        }
+        if (!netIsWrong && !fanOutIsWrong && !isContained && newEdge.source != "" && newEdge.target != "")
+          edges.add(newEdge)
+      }
+    }
+  }
+
+  /** Generating HDElk language for Node */
+  private def drawNodes(thisNode: ElkNode): Unit = {
```
andreasWallner, line 416: I'd move this code into the ElkNode

lib/src/main/scala/spinal/lib/tools/HDElkDiagramGen.scala
```diff
@@ -0,0 +1,599 @@
...
+    val allInnerCells = module.children
+    for (cell <- allInnerCells) {
+      writer.write(
+        s"""<a href="#${cell.getName()}"><button>${
+          cell
+            .getName()
+        }</button></a>&nbsp;\n""")
+    }
+    writer.write(s"""</div><br><br><br><br>\n""")
+    writer.close()
```
andreasWallner, line 589: See above note about not opening/closing the writer multiple times. It should really only be opened once and then passed as a parameter.

andreasWallner (commented):
A few comments of stuff that I observed. Most are suggestions that I think would make the code easier to read, some are small issues.
Keep up the good work!

ZhaokunHu:
I have refactored my code. Please let me know if there are any shortcomings in terms of functionality and which parts of the code still need modification. I will do my best to make the necessary changes.

ZhaokunHu:
> Hi ^^ I think it can be merged.
> 
> overall, maybe the issue currently is for none-small design. maybe having some modes which :
> 
> * omit combinatorial signal draw to keep things smaller (but keep connectivity done across them), so focussing on states
> * Preserve the Area structure / locality, but i guess this maybe not possible using HDElk ?
> * In some modes, we could even imagine that the diagram would "only" represent the connectivity between Area / sub component / io
> 
> One typical example could be recent reimplementation of the StreamFifo, which is now very much structured using Area

image
Thank you for your suggestions.
1. I didn't fully understand what constitutes a combinatorial signal. To ensure the correct connectivity, I can only merge the nodes and edges instead of directly eliminating them. In the latest commit, I attempted some simplifications, but the results were only moderately successful.
2. 2 & 3. Based on your suggestions, representing an Area in HDElk is indeed challenging. Additionally, it's difficult to differentiate signals within an Area as inputs or outputs. Therefore, I cannot add ports to the Area as I do for modules. One possible solution I can think of is to annotate the signal names connecting the Area's internal and external components, as shown in the diagram above.

Dolu1990:
Hi ^^

> I didn't fully understand what constitutes a combinatorial signal

So i would say, in your screenshot above, combinatorial signals (stuff which isn't reg nor io) as for instance when_test12_l64, instead of being a box, could just represent connectivity.

>  it's difficult to differentiate signals within an Area as inputs or outputs

right :) something things are even mixed :/

Readon:
> Hi ^^ I think it can be merged.
> 
> overall, maybe the issue currently is for none-small design. maybe having some modes which :
> 
> * omit combinatorial signal draw to keep things smaller (but keep connectivity done across them), so focussing on states
> * Preserve the Area structure / locality, but i guess this maybe not possible using HDElk ?
> * In some modes, we could even imagine that the diagram would "only" represent the connectivity between Area / sub component / io
> 
> One typical example could be recent reimplementation of the StreamFifo, which is now very much structured using Area

I think we can merge this with a compatible API first. Then we could modify it if it is necessary later.

PR #1081, opened 2023-04-09
[feature] add BusSlaveFactory onWriteBits

### LINKED ISSUE #1068: promote BusSlaveFactory
```scala
def onWrite(address: BigInt, documentation: String = null)(doThat: => Unit): Unit = {
    onWritePrimitive(
      address       = SingleMapping(address),
      haltSensitive = true,
      documentation = documentation
    )(doThat)
  }
```  
it's useful to add writedata param to doThat
it will be like:

```scala
def onWrite(address: BigInt, documentation: String = null)(doThat: Bits => Unit): Unit = {
    onWritePrimitive(
      address       = SingleMapping(address),
      haltSensitive = true,
      documentation = documentation
    )(doThat)
  }
```

chenbo-again:
Closes #1068 

when we talk about bussalvefactory. It's useful for providing a interface to do a callback with a writedata parameter when a write transaction occour.

# Impact on code generation
no old api was changed.

# Checklist

- [ ] Unit tests were added
I do a unit test on my own, I write a simple case like this
```
  factory.onWriteBits(address = 0x100)(bits=>{
    when(bits(0) === True) {
      Pending := True
    }
  })
```
the generated code is like this

```
  assign when_Box_l42 = (bus_HWDATA[0] == 1'b1);
.....

....
        32'h00000100 : begin
          if(when_BusSlaveFactory_l978_regNext) begin
            if(when_Box_l42) begin
              Pending <= 1'b1;
            end
          end
        end
end
```

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
+++ b/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
@@ -167,6 +167,20 @@ trait BusSlaveFactory extends Area{
 
 
   /**
+    * Call doThat(with write data) when a write transaction occurs on address
+    */
+  def onWriteBits(address: BigInt, documentation: String = null)(doThat: Bits => Unit): Unit = {
+    val writeBits = Bits(this.busDataWidth bits)
+    nonStopWrite(writeBits, 0)
+    onWritePrimitive(
+      address       = SingleMapping(address),
+      haltSensitive = true,
+      documentation = documentation
+    )(doThat(writeBits))
+  }
+
+  
+  /**
     * Call doThat when a read transaction occurs on address
     */
   def onRead(address: BigInt, documentation: String = null)(doThat: => Unit): Unit = {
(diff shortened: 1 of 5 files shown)


Readon:
Does this solve #1068 ?

Dolu1990:
Thanks :)

I would say, i doesn't need to be a new primitive itself.

The best would be to add a more fondamental new abstract function in the BusSlaveFactory being : 
def writeBits() : Bits

And then implementing onWriteBits as a combination of onWrite + writeBits

chenbo-again:
It's hard to implement a method writeBits now
```
busCtrl.onWrite(address=0) {
  val bits = busCtrl.writeBits()
}
```

the code is unable to run. because in `onWrite(address)(dothat)`. doThat will only be called in busSlaveFactory build phase. and in this phase, it's hard to add new element like nonStopWrite.

Dolu1990:
> the code is unable to run. because in onWrite(address)(dothat). dothat will only be called in busSlaveFactory build phase. and in this phase, it's hard to add new element like nonStopWrite.

Hoooo you are right.
Maybe instead writeBits need to be implemented like def writeAddress(): UInt, but for data.

PR #1088, opened 2023-04-17
Propagate tags from Tristates when using InOutWrapper

### LINKED ISSUE #1086: InOutWrapper TriState Tag Propagation?
I'm working on a design with an inout/tristate capable pin.  I've written my own set of tags for pin constraints.  I'm noticing tags on the tristate signal don't bubble up when using the `InOutWrapper`.  Is this something that might be useful as a feature?

Loose example:
```scala
case class Rgmii() extends Bundle with IMasterSlave {
  val TX = RgmiiTx()
  val RX = RgmiiRx()
  val MDIO = Mdio() // Provides tristate IO pin

  override def asMaster(): Unit = {
    out(TX)
    in(RX)
    master(MDIO)
  }
}
```
```scala
class TopLevel extends Component {
  val io = new Bundle {
    // ..snip..
    val rgmii = master(Rgmii())
  }

  // ..snip..
  io.rgmii.MDIO.IO.addTag(PinTag("C8")).addTag(PinStandardTag("LVCMOS18"))
}
```
Then in an app object:
```scala
  val report = SpinalVhdl(InOutWrapper(new TopLevel()))
```

I search for tagged IO with this, unless there's a better way.
```scala
  val pinConstraints = report.globalData.toplevel.getAllIo.flatMap { bt =>
    bt.getTag(classOf[PinTag]).map(t => t.getConstraintText(bt))
  }.flatten.mkString(System.lineSeparator())
```

kleinai:
Closes #1086

This is a simple change which copies tags on the tristate signal to the wrapper analog pin.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/Trait.scala
+++ b/core/src/main/scala/spinal/core/Trait.scala
@@ -763,6 +763,7 @@ trait SpinalTag {
   def driverShouldNotChange = false
   def canSymplifyHost       = false
   def allowMultipleInstance = true // Allow multiple instances of the tag on the same object
+  def ioTag                 = false // Propagate tag to IO
 
   def apply[T <: SpinalTagReady](that : T) : T = {
     that.addTag(this)
--- a/lib/src/main/scala/spinal/lib/io/InOutWrapper.scala
+++ b/lib/src/main/scala/spinal/lib/io/InOutWrapper.scala
@@ -32,6 +32,7 @@ object InOutWrapper {
               when(bundle.writeEnable) {
                 newIo := bundle.write
               }
+              newIo.addTags(bundle.getTags())
             }
             case bundle: TriStateOutput[_] if bundle.isOutput || bundle.isMasterInterface => {
               val newIo = inout(Analog(bundle.dataType)).setWeakName(bundle.getName())
@@ -39,6 +40,7 @@ object InOutWrapper {
               when(bundle.writeEnable) {
                 newIo := bundle.write
               }
+              newIo.addTags(bundle.getTags())
             }
             case bundle: ReadableOpenDrain[_] if bundle.write.isOutput && bundle.read.isInput => {
               val newIo = inout(Analog(bundle.dataType)).setWeakName(bundle.getName())
@@ -49,6 +51,7 @@ object InOutWrapper {
                   newIo.assignFromBits(B"0", id, 1 bits)
                 }
               }
+              newIo.addTags(bundle.getTags())
               //            for(bt <- bundle.write.flatten){
               //              for((value, id) <- bt.asBits.asBools.zipWithIndex) {
               //                when(!value){
@@ -66,6 +69,7 @@ object InOutWrapper {
                   newIo(i) := bundle.write(i)
                 }
               }
+              newIo.addTags(bundle.getTags())
             }
             case _ =>
           }
--- a/lib/src/main/scala/spinal/lib/io/InOutWrapper.scala
+++ b/lib/src/main/scala/spinal/lib/io/InOutWrapper.scala
@@ -18,6 +18,10 @@ object InOutWrapper {
         }
       }
 
+      def propagateTags(tristate: Bundle, pin: Data): Unit = {
+        tristate.getTags().filter(_.ioTag).foreach(t => pin.addTag(t))
+      }
+
       for (io <- c.getAllIo) {
         add(io)
       }
@@ -32,7 +36,7 @@ object InOutWrapper {
               when(bundle.writeEnable) {
                 newIo := bundle.write
               }
-              newIo.addTags(bundle.getTags())
+              propagateTags(bundle, newIo)
             }
             case bundle: TriStateOutput[_] if bundle.isOutput || bundle.isMasterInterface => {
               val newIo = inout(Analog(bundle.dataType)).setWeakName(bundle.getName())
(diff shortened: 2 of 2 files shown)


andreasWallner:
I think we at least check the tags properties if it is configured to be propagated. What comes to mind is the `duplicative` property, but maybe it makes sense to have another property for this case?

Dolu1990:
> I think we at least check the tags properties if it is configured to be propagated.

Yes right.

It is often case base case.
I think having a specific trait for tags meant to stay on the component input/output specification would be the best. Then filtering all the tags which implement that trait and adding them to the new input/output

kleinai:
I intended to use this to help propagate tags for pin placement. However now that I think about it more, there might be a better way that doesn't involve tagging signals. Or avoiding the tristate and using a black box driver instead.

andreasWallner:
@kleinai just some thoughts:

I kind of like the idea of tagging, but it has one drawback: it serves the use case were you have a `Bundle` that represents the I/O of a certain devboard (I'm playing around with this for some of the boards that I have, like: https://github.com/andreasWallner/spinalStuff/blob/master/src/main/scala/andreasWallner/iceblink/IrDaLaDemo.scala#L14).
What would be a bit more cumbersome is the use-case that you have a specific project that you want to map to multiple boards. That you could do with a wrapper/method that creates/replaces the tags on the I/Os. (possibly coming from an extension method...)

The second use-case kind of rules out having a trait that extends a `Bundle` with location information (?) since you can't easily map to multiple boards.
Either you have a method that queries the location depending on the board (`def location(platform: String, io: Bundle): Mapping`). That means that you have to have a single class that knows about all the boards - that's shitty to e.g. extend (especially outside of the original codebase).

The other would be to just have an object that provides mapping information (i.e. `class MapMyThingToIceStick(...) { def location(io: Bundle): Mapping}` that you pass to some synthesizer/placer specific serializer together with your toplevel. Also feels kind of cumbersome...

In the end the attribute still feels the most flexible/natural to me... What did you think about as a better way?

PS: I have a PR upcoming to extend InOutWrapper to optionally explicitly instantiate driver blackboxes instead of letting it infer them - it now works since the latest analog fix...

kleinai:
I cobbled together a Vivado build flow which goes from SpinaLHDL to bitstream and pulls certain information out of tags.

For example, RGMII on the XCZU5EV dev board:
```scala
  // 1GbE
  io.rgmii.TX.TXC.addTag(PinTag("E8")).addTag(PinStandardTag("LVCMOS18"))
  io.rgmii.TX.TXD.addTag(PinTag(Seq("E9", "D9", "A9", "A8"))).addTag(PinStandardTag(Seq("LVCMOS18", "LVCMOS18", "LVCMOS18", "LVCMOS18")))
  io.rgmii.TX.TX_CTL.addTag(PinTag("B9")).addTag(PinStandardTag("LVCMOS18"))
  io.rgmii.RX.RXC.addTag(PinTag("E5")).addTag(PinStandardTag("LVCMOS18"))
  io.rgmii.RX.RXD.addTag(PinTag(Seq("A5", "B5", "F8", "C9"))).addTag(PinStandardTag(Seq("LVCMOS18", "LVCMOS18", "LVCMOS18", "LVCMOS18")))
  io.rgmii.RX.RX_CTL.addTag(PinTag("B8")).addTag(PinStandardTag("LVCMOS18"))
  io.rgmii.MDIO.C.addTag(PinTag("A6")).addTag(PinStandardTag("LVCMOS18"))
  io.rgmii.MDIO.IO.addTag(PinTag("C8")).addTag(PinStandardTag("LVCMOS18")) // A single bit tristate type
```
I walk the IO and generate a Tcl of constraints. It's pretty simple but really smooths out the build flow. If I wanted to port to another board I'd probably make a new top level to map the pins differently. Not elegant but at least I don't have to leave Idea to make a bitstream.

Definitely interested in your InOutWrapper change. I'd like to see more explicit Xilinx blackboxes in my designs if I can help it.

I had experimented with a tag trait like `TriStatePropagate` but couldn't get that code to filter and clone the tags properly.

andreasWallner:
@kleinai 
First off, here is the wrapper for Xilinx only: https://github.com/andreasWallner/spinalStuff/blob/master/src/main/scala/andreasWallner/xilinx/XilinxInOutWrapper.scala
I needed that one for designs with a Zynq, Vivado only seems to be happy to infer an IOBUF if it's on the toplevel...
The version that would be able to support more targets that I'm still polishing to upstream you can find there: 
https://github.com/andreasWallner/spinalStuff/blob/master/src/main/scala/andreasWallner/xilinx/InOutWrapper.scala

W.r.t. your example:
this does seem fine to me, you mentioned that you think there might be a better way?

kleinai:
> W.r.t. your example: this does seem fine to me, you mentioned that you think there might be a better way?

After reading your example cases I think tagging ports looks cleaner and flows better in my mind.  Having to map pins somewhere else and match them up could be confusing.

> @kleinai First off, here is the wrapper for Xilinx only: https://github.com/andreasWallner/spinalStuff/blob/master/src/main/scala/andreasWallner/xilinx/XilinxInOutWrapper.scala I needed that one for designs with a Zynq, Vivado only seems to be happy to infer an IOBUF if it's on the toplevel... The version that would be able to support more targets that I'm still polishing to upstream you can find there: https://github.com/andreasWallner/spinalStuff/blob/master/src/main/scala/andreasWallner/xilinx/InOutWrapper.scala

I also had a similar thought, what if InOutWrapper can optionally take a function to map the tristate buffers. The inferred synthesis of tristates is still a good default.

kleinai:
Rebased on dev. Added a tag switch so that only enabled tags will be copied through the wrapper.

kleinai:
A property like `ioTag` to signify that tag should be moved to an IO port?

Dolu1990:
> A property like ioTag to signify that tag should be moved to an IO port?

Yes, to specify that the tag belong to the io port

Dolu1990:
Cool ^^
Feel free to merge

PR #1102, opened 2023-05-01
Compile for all supported versions during CI

andreasWallner:

In our version file we list a few versions that the package claims to be compatible with, we should build for all of them in the CI to prevent possible surprises during a release.

# Impact on code generation

none


Dolu1990:
<3

PR #1107, opened 2023-05-02
UartCtrlRx: bitTimer.reset() to center bit in Majority Voter

dlmiles:

Make implementation follow existing description.
Previously it biased the bit decode towards the leading half of the bit.
VCD and attachments to follow.


dlmiles:
See images at #1108

dlmiles:
Closed as per the #1108 update.

PR #1113, opened 2023-05-09
Regif Chead union struct name add prefix

xiaozhulanshan:
Closes #

Regif Chead union struct name will conflict in different regfile with same field name . So we add prefix in Regif Chead union struct name

# Impact on code generation

# Checklist

- [ ] Unit tests were added


jijingg:
@xiaozhulanshan   thanks

PR #1119, opened 2023-05-13
Fix HDL names of memory ports

andreasWallner:
We have code to name memory access ports with the postfix "_port", but it is not working/used.

A MemPortStatement is a Nameable, the match would never choose the MemPortStatement case.


Dolu1990:
<3<3

PR #1120, opened 2023-05-15
core: RFC object SmtBmcSolver extends Enumeration: TYPOs ?

dlmiles:
Request For Comments:
    
  Boolector definitely looks incorrect spelling to me, has this been tested ?
    
  Just a check Yices is called 'yices' in the configuration for sby and not yices2.

@Readon PING  you can open a commit if any change needed separately, then close this without merge.  Thanks

### DIFF
--- a/core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
+++ b/core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
@@ -29,10 +29,10 @@ sealed trait SbyEngine extends FormalEngin {
 object SmtBmcSolver extends Enumeration {
   type SmtBmcSolver = Value
 
-  val Yices = Value("yices")      // Yices2 ?
-  val Boolector = Value("booleactor")   // TYPO: boolector
+  val Yices = Value("yices")
+  val Boolector = Value("boolector")
   val Z3 = Value("z3")
-  val mathsat = Value("mathsat")
+  // val mathsat = Value("mathsat")
   val cvc4 = Value("cvc4")
 }
 
--- a/core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
+++ b/core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
@@ -29,8 +29,8 @@ sealed trait SbyEngine extends FormalEngin {
 object SmtBmcSolver extends Enumeration {
   type SmtBmcSolver = Value
 
-  val Yices = Value("yices")
-  val Boolector = Value("booleactor")
+  val Yices = Value("yices")      // Yices2 ?
+  val Boolector = Value("booleactor")   // TYPO: boolector
   val Z3 = Value("z3")
   val mathsat = Value("mathsat")
   val cvc4 = Value("cvc4")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
```diff
@@ -29,8 +29,8 @@ sealed trait SbyEngine extends FormalEngin {
 object SmtBmcSolver extends Enumeration {
   type SmtBmcSolver = Value
 
-  val Yices = Value("yices")
-  val Boolector = Value("booleactor")
+  val Yices = Value("yices")      // Yices2 ?
```
Readon, line 32: I have tested the code with yices, it is named without version code. So it is correct.

core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
```diff
@@ -29,8 +29,8 @@ sealed trait SbyEngine extends FormalEngin {
...
+  val Boolector = Value("booleactor")   // TYPO: boolector
```
Readon, line 33: This would be a TYPO, and I did not add testcase for this one. What have been tested are only yices, z3. It would be better to add all support cases. However, I do not use others. :)

Furthermore, I found that I did not even add boolector into the docker image. :(

Readon:
Do you think mathsat is necessary? I did not use it before.

dlmiles:
I have no idea of the relative benefits.  Is there an API provision to add an arbitrary solver name via configuration ?  Rather than being forced to use an Enum from the SpinalHDL code base?

So someone can configure 'mathsat' this way if they need to ?  Then remove the 'mathsat' line from the code ?  Now all use cases are covered for everyone.

dlmiles:
My Thinking
------------------

* KEEP, If the only way for specifying the solver backend is via the Enumeration, so any new solver needs a SpinalHDL code-base change, then leave it in.

* REMOVE, If there is a mechanism to specify any arbitrary backend label via SpinalHDL formal API, then it could be removed.

* COMMENT, It does not make sense to comment it out but leave it in the code-base like that.  This is the worst scenario, it can neither be used, and it teases someone in the situation who can not use `mathsat` because it is commented out for no reason.

It is not upto SpinalHDL to provide the sby/YoSys/solver development environment, these are external projects and an external matter the user is expected to resolve themselves.
All we can do is make that step easier and try to be inclusive of as much eco-system as possible.


Re the docker image concerns:
---------------------------------------------

FWIW even oss-cad-suite does not contain mathsat preloaded, it looks like it is here: https://mathsat.fbk.eu/download.html and GPL3, the contents of oss-ca-suite is based on Ubuntu 20.04 but appears to have everything being built into the Docker image already, libboost 1.74 => 1.71 (I assume due to Ubuntu being older), verilator would goto v5.010+

It looks like oss-cad-suite + Java JDK + sbt/mill + SpinalHDL would base the image on top of Ubuntu 20.04 with the docker image only needing an auto-rebuild on a timed job, no need to maintain the Dockerfile with version change edits to track yosys/verilator until there is a big update to something (like oss-cad-suite moving 22.04)

[MathSAT - <span>Download</span>](https://mathsat.fbk.eu/download.html)
MathSAT 5 An SMT Solver for Formal Verification & More Download Version 5.6.9 (Nov 8, 2022): Linux: 64-bit x86, Reentrant 64-bit x86. Mac OS X: Universal binary. Windows: 64-bit x86. Nightly Builds

dlmiles:
i assume this will be SQUISH, REBASE, FF-MERGE if not possible then MERGE, not sure how you do this from GH webUI, this should only be 1 commit

PR #1124, opened 2023-05-19
Rework PackedBundle packing to avoid width confusion

kleinai:
Original method of translating to and from a packed bitvector results in errors in FIFO, Streams, etc. These classes analyze the bundle type and detect width mismatches.
New change doesn't override asBits or getBitsWidth to avoid problems.

Closes #

# Impact on code generation

# Checklist

- [ ] Unit tests were added


xiaochuang-lxc:
@kleinai  There is a mistake in PackedBundle:
val a=new PackedBundle{
 val b=Bool().pack_from(0)
}
val d=Bits(2 bits)
a.unpack(d)

kleinai:
Follow up in #1187

PR #1134, opened 2023-06-06
AxiMemorySim Random object as parameter

knapheide:

Optionally pass a `scala.util.Random` object into `AxiMemorySim`, so tests can be reproducible.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/AxiMemorySim.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/AxiMemorySim.scala
@@ -288,7 +288,7 @@ case class AxiMemorySim(
   axi : Axi4,
   clockDomain : ClockDomain,
   config : AxiMemorySimConfig,
-  random : Random = Random,
+  random : Random = Random
 ) {
   val memory = SparseMemory()
   val pending_reads = new mutable.Queue[AxiJob]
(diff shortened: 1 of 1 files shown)


Readon:
Maybe a gen function is preferred but a random object?

andreasWallner:
I'm not sure I understand the use-case you'd use this for. To make a test reproducible have a few methods to provide a seed that will ensure that other simulation agents (e.g. the Stream ones) will also have reproducible behavior (e.g.  via the `seed` parameter to `doSim` or via an environment variable `SPINAL_SIM_SEED`).
Do you have a use-case in mind where you just want the AxiMemorySim the behave reproducible but nothing else?

knapheide:
Ah, i didnt realize that `doSim` sets the seed globally.

That wont work though when running multiple simulations in parallel in the same process, right?
I probably should just stop doing that

andreasWallner:
What would be your desired outcome when running multiple simulations in parallel? All our examples enable forking, so a separate jvm process is started for each test - so you can run tests in parallel and still have reproducible tests.

knapheide:
I am iterating over a bunch of configurations i want to test and it just seemed easiest to speed this up using `concurrent.Future.traverse`.
But its not a big problem, i can either look into scalatest, or spawn separate processes myself. I just didnt know that parallel simulations in the same process arent supported.

PR #1140, opened 2023-06-12
Add timeScale support to simulators

### LINKED ISSUE #1138: VSC timeScale and timePrecision
config ``` 1 ps/100 fs ``` , buf simulate actual ``` 1 ns /100fs ```
```scala
SimConfig
    .withTimeScale(1 ps)
    .withTimePrecision(100 fs)

```

because this function use ``` 1ns ``` instead ``` timeScale```
```scala
val timeScale = config.timePrecision match {
    case null => "1ns/1ns"
    case t => "1ns/" + t.replace(" ", "")
  }
```

kleinai:
Closes #1138 

Finishes adding time scale support. This wasn't properly finished with the original set of time precision changes.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -46,6 +46,7 @@ case class SpinalVerilatorBackendConfig[T <: Component](
                                                          optimisationLevel : Int = 2,
                                                          simulatorFlags    : ArrayBuffer[String] = ArrayBuffer[String](),
                                                          withCoverage      : Boolean,
+                                                         timeScale         : TimeNumber = null,
                                                          timePrecision     : TimeNumber = null
 )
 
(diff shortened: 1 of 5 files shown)


andreasWallner:
Would we also need to update https://github.com/SpinalHDL/SpinalHDL/blob/6ed4365caee86508842d305dc6ddf9696e2201ab/core/src/main/scala/spinal/core/sim/package.scala#LL175C1-L175C1 to use the timescale instead of precision?

kleinai:
> Would we also need to update https://github.com/SpinalHDL/SpinalHDL/blob/6ed4365caee86508842d305dc6ddf9696e2201ab/core/src/main/scala/spinal/core/sim/package.scala#LL175C1-L175C1 to use the timescale instead of precision?

Missed your comment.  No, that function converts time into simulation steps.  AFAIK, time scale is primarily used for verilog delays.

andreasWallner:
> No, that function converts time into simulation steps. AFAIK, time scale is primarily used for verilog delays.

That was my thought why we might have to update it.
Quoting from: https://www.chipverify.com/verilog/verilog-timescale

> The time_unit is the measurement of delays and simulation time while the time_precision specifies how delay values are rounded before being used in simulation.

In the old code we used scale == precision, so the calculation code could use precision to calculate the number of simulator cycles needed for a specific timespan, if we have both we have to use the scale (and maybe not round to an integer?) Or do I have some big misundestanding here - verilog is not my native language ;-)

PR #1149, opened 2023-06-28
Fifo rework

Dolu1990:
Merge #1144 StreamFifo / StreamFifoLowLatency reimplementation for better perf

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1318,30 +1318,30 @@ class StreamFifo[T <: Data](val dataType: HardType[T],
     val vec = Vec(check)
   }.vec
 
-  def formalCheckm2sPipe(cond: T => Bool): Bool = this.rework {
+  def formalCheckOutputStage(cond: T => Bool): Bool = this.rework {
     // only with sync RAM read, io.pop is directly connected to the m2sPipe() stage
     Bool(!withAsyncRead) & io.pop.valid & cond(io.pop.payload)
   }
 
   // verify this works, then we can simplify below
   //def formalCheck(cond: T => Bool): Vec[Bool] = this.rework {
-  //  Vec(formalCheckm2sPipe(cond) +: formalCheckRam(cond))
+  //  Vec(formalCheckOutputStage(cond) +: formalCheckRam(cond))
   //}
 
   def formalContains(word: T): Bool = this.rework {
-    formalCheckRam(_ === word.pull()).reduce(_ || _) || formalCheckm2sPipe(_ === word.pull())
+    formalCheckRam(_ === word.pull()).reduce(_ || _) || formalCheckOutputStage(_ === word.pull())
   }
   def formalContains(cond: T => Bool): Bool = this.rework {
-    formalCheckRam(cond).reduce(_ || _) || formalCheckm2sPipe(cond)
+    formalCheckRam(cond).reduce(_ || _) || formalCheckOutputStage(cond)
   }
 
   def formalCount(word: T): UInt = this.rework {
     // occurance count in RAM and in m2sPipe()
-    CountOne(formalCheckRam(_ === word.pull())) +^ U(formalCheckm2sPipe(_ === word.pull()))
+    CountOne(formalCheckRam(_ === word.pull())) +^ U(formalCheckOutputStage(_ === word.pull()))
   }
   def formalCount(cond: T => Bool): UInt = this.rework {
     // occurance count in RAM and in m2sPipe()
-    CountOne(formalCheckRam(cond)) +^ U(formalCheckm2sPipe(cond))
+    CountOne(formalCheckRam(cond)) +^ U(formalCheckOutputStage(cond))
   }
 
   def formalFullToEmpty() = this.rework {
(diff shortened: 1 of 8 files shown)


lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -1085,101 +1090,277 @@ object StreamFifo{
   def apply[T <: Data](dataType: T, depth: Int) = new StreamFifo(dataType,depth)
 }
 
-class StreamFifo[T <: Data](val dataType: HardType[T], val depth: Int) extends Component {
+/**
+  * Improved StreamFifo over the original implementation
```
andreasWallner, line 1094: not sure we need to refer to the original one that does not exist anymore after this merge?

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -1085,101 +1090,277 @@ object StreamFifo{
...
+                            val depth: Int,
+                            val withAsyncRead : Boolean = false,
+                            val withBypass : Boolean = false,
+                            val allowExtraMsb : Boolean = true,
+                            val forFMax : Boolean = false,
+                            val useVec : Boolean = false) extends Component {
   require(depth >= 0)
-  val io = new Bundle {
+  // bypass not supported for sync read case
+  if(!withAsyncRead) require(!withBypass)
```
andreasWallner, line 1112: should we add a note like "use of withAsyncRead=false requires withBypass=false as well"?

tester/src/test/scala/spinal/tester/scalatest/FormalFifoTester.scala
```diff
@@ -44,23 +61,58 @@ class FormalFifoTester extends SpinalFormalFunSuite {
...
         when(!d2_in) { assume(!dut.formalContains(d2)) }
         when(d2_in && !d2_out) { assert(dut.formalCount(d2) === 1) }
 
+        // make sure that d2 does not leave before d1 leaves the StreamFifo
         when(d1_in && d2_in && !d1_out) { assert(!d2_out) }
 
+        // cover the case that FIFO first goes to full, then to empty
+        dut.formalFullToEmpty()
+
+        // TODO for .withProve() induction prove to pass, relationship between push/pop distance and some fill rate
```
andreasWallner, line 83: are the TODOs in here (and the one below on line 124) still a thing?

tester/src/test/scala/spinal/tester/scalatest/SpinalSimStreamFifoTester.scala
```diff
@@ -31,31 +40,68 @@ class SpinalSimStreamFifoTester extends SpinalSimFunSuite {
...
+        assert(dut.io.availability.toInt == depth - queueModel.size)
+        dut.io.flush #= Random.nextFloat() < 0.005
+      }
+
+
       //Push data randomly and fill the queueModel with pushed transactions
       dut.io.push.valid #= false
       dut.clockDomain.onSamplings {
         if (dut.io.push.valid.toBoolean && dut.io.push.ready.toBoolean) {
           queueModel.enqueue(dut.io.push.payload.toLong)
```
andreasWallner, line 64: since we update the testcase, should we use `ScoreboardInOrder` here?

andreasWallner (commented):
Just a few general comments on minor stuff...

One open point in my mind: should some of the parameters also be available from the `Stream.queue` function?

lib/src/main/scala/spinal/lib/Stream.scala
likewise: Yes. Same for `useVec`, as it requires` withAsyncRead`.
We should add this to the code:
```scala
if (useVec) require (withAsyncRead)
```

tester/src/test/scala/spinal/tester/scalatest/FormalFifoTester.scala
likewise: Yes, the TODO's are up-to-date with the code.

I have newer formal code with update TODO's in my `fifo-rework` branch fork, matching better formal coverage.

lib/src/main/scala/spinal/lib/Stream.scala
likewise: Maybe mention: Fully redesigned in release 1.8.2 allowing improved timing closure.

Dolu1990:
> One open point in my mind: should some of the parameters also be available from the `Stream.queue` function?

Yes, i would say, latency and forFMax tunning

lib/src/main/scala/spinal/lib/Stream.scala
Dolu1990: yes, doing it.

lib/src/main/scala/spinal/lib/Stream.scala
Dolu1990: Yes, on it.

tester/src/test/scala/spinal/tester/scalatest/SpinalSimStreamFifoTester.scala
Dolu1990: Baaaa, things are working already, i think it isn't necessary.

PR #1154, opened 2023-07-07
VivadoFlow: Fix issues with getFMax() and getArea(), and improve.

### LINKED ISSUE #1153: VivadoFlow.scala getFMax() does not consider clock pulse slack of non-logic such as BRAMs/PLLs.
During timing (closure) optimizations (https://github.com/SpinalHDL/SpinalHDL/issues/1144) and benchmarking (`Bench.scala`) I noticed we could drive the logic + BRAMs to 925 MHz, which is out of spec (738 MHz) for Alveo U50.

The problem is that `VivadoFlow.bench` does not take pulse width requirements into account when deriving f.max.

A modified version of `getFMax()` I am working on will fix this.

Other vendor tooling support may suffer identical issues.

likewise:
Closes #1153 and #1154.

Make sure pulse timing (for BRAMs, PLLs) is also considered for f.max
Make sure partials are matched with English localization (`0.5 BRAM` vs `0,5 BRAM`) 
URAMs and BRAMs are reported for Ultrascale(+)
`-mode out_of_context` is used during synthesis to prevent the need for I/O pins or XORing outputs.

# Impact on code generation

No impact on code generation.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/eda/xilinx/VivadoFlow.scala
+++ b/lib/src/main/scala/spinal/lib/eda/xilinx/VivadoFlow.scala
@@ -48,7 +48,10 @@ place_design
 route_design
 
 report_utilization
-report_timing"""
+report_timing_summary -warn_on_violation
+report_pulse_width -warn_on_violation -all_violators
+report_design_analysis -logic_level_distribution
+"""
     )
     tcl.flush();
     tcl.close();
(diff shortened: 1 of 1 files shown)


kleinai:
While we're at it, I suggest we synthesize with `-mode out_of_context` to avoid buffer insertions. The synthesis will fail when benchmarking a component with a lot of IO. However, I did not check if OOC noticable effected reported timing.

likewise:
> While we're at it, I suggest we synthesize with `-mode out_of_context` to avoid buffer insertions. The synthesis will fail when benchmarking a component with a lot of IO. However, I did not check if OOC noticable effected reported timing.

Thanks for review, yes, I agree, that prevents issues with many I/Os. It should not matter w.r.t. results, as the module is optimized internally, just not across the I/O interface. In my tests, I get the same results with `-mode out_of_context`.

kleinai:
I also found myself making local changes to read out different utilization metrics. At some point it might be worth it to read the whole table. Assuming AMD doesn't change the format between tool versions of course.

likewise:
> I also found myself making local changes to read out different utilization metrics. At some point it might be worth it to read the whole table. Assuming AMD doesn't change the format between tool versions of course.

The thing I also want to add, is the (histogram or at least the most==worst) amount of logic levels (LL?) in a module. These are the number of LUT/MUX elements between registers. If large, it means this is hard to place and route within timing constraints.

If the histogram is spread out, it means the logic is unbalanced.

It would be great if we can make sure all processing blocks in the library are "decently designed" in terms of LL.

Dolu1990:
Nice thanks :D
Look good to me

About the amount of logic level, yes that could be nice. Either that, or the ratio between routing delay and logic delay ?

likewise:
Thanks for merging. The routing delay mostly becomes influential when the tooling cannot get the different elements close to one another, so is not really useful for single modules (I think -- let me know if you think otherwise).

In other words; It is mostly the number of logic levels that determine the maximum achievable f.max (i.e. logic propagation delays), and then in a bigger integrated design, the f.max is further reduced by routing delays.

PR #1156, opened 2023-07-09
WIP: Attempt to make Tilelink framework generic

kleinai:
Creating this more as an RFC. Playing around with how the Tilelink framework might be adapted to host itself and other buses with less work. More or less a copy and paste of Tilelink with parameters.

Closes #

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/tilelink/meta/Meta.scala
+++ b/lib/src/main/scala/spinal/lib/bus/tilelink/meta/Meta.scala
@@ -1,0 +1,216 @@
+package spinal.lib.bus.tilelink.meta
+
+import spinal.core._
+
+import spinal.core.fiber.{Fiber, Handle, Lock, soon}
+import spinal.lib.bus.misc.{AddressMapping, DefaultMapping, OrMapping, SizeMapping}
+
+import scala.collection.mutable.ArrayBuffer
+
+abstract class ConnectionBase[T <: Bundle, C, DC <: DeferredConfig[C], N <: NodeBase[T, C, DC, N, CB], CB <: ConnectionBase[T, C, DC, N, CB]](val m: N, val s: N) extends Area {
+  val mapping = new Area {
+    var automatic = Option.empty[Any]
+    val value = Handle[AddressMapping]
+  }
+
+  val up, down = new Area {
+    val bus = new Handle[T]
+    val m2s = new Handle[DC]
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^
So, i was kind of thinking about it too. About having some more generic sceleton code using type parameter.

I can see that there is code which could be shared between different bus implementation of the Node class, i'm just mostly not sure of the best way to reuse that code.

In some ways, it seems to me that the following look scary:

```scala
class NodeBase[T <: Bundle, C, DC <: DeferredConfig[C], N <: NodeBase[T, C, DC, N, CB], CB <: ConnectionBase[T, C, DC, N, CB]] 
```

I will take some time to think about it, to see if there would not be another way :)

Dolu1990:
Ahhh interresting, i didn't knew about this syntax XD

kleinai:
Way better to use a trait then build an object.  The classes can be define in an object and share types without needed all the parameter headache.

Usage:
```scala
object Axi4TileLink extends TileLinkApi {
  type T = Axi4
  type C = Axi4Config
  type DC = Config
  type N = Axi4Node
  type CB = Axi4Connection
 
  ... define classes ...
}
```

Example snippet:
```scala
object Example0 extends App {
  SpinalVerilog(new Module {
    import Axi4TileLink._

    val cpuBus = new Axi4Node().setSlaveOnly()
    val debugBus = new Axi4Node().setSlaveOnly()
    val mainBus = new Axi4Node()
    val ramBus = new Axi4Node().setMasterOnly()
    val peripheralBus = new Axi4Node().setMasterOnly()

   ... code ...
```

Dolu1990:
> object Axi4TileLink extends TileLinkApi {

That's a interresting pattern ^^

Hmmmm
I'm still looking around to see if things could be instead more "composed" (to try avoid too much generalization). I pushed the following ( WIP ) : 
https://github.com/SpinalHDL/SpinalHDL/commit/4a33fe066d6a17874c67f3353baf2e573031755f#diff-9c56d14d6a1ec1cfff255c2b4f33e0562ee3f984b60ce88d3767099ee02eeb77

Dolu1990:
I think https://github.com/SpinalHDL/SpinalHDL/pull/1160 would be more or less ok.

Dolu1990:
https://github.com/SpinalHDL/SpinalHDL/pull/1160 merged :)

PR #1160, opened 2023-07-12
spinal.lib.bus.fabric reusable base classes

Dolu1990:
Another attempts of #1156
WIP

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/fabric/NegociateSP.scala
+++ b/lib/src/main/scala/spinal/lib/bus/fabric/NegociateSP.scala
@@ -1,0 +1,32 @@
+package spinal.lib.bus.fabric
+
+import spinal.core._
+import spinal.core.fiber.Handle
+
+import scala.collection.mutable.ArrayBuffer
+
+/**
+  * Negotiation handles for master to slave requests
+  */
+abstract class NegociateSP[S, P] extends Area{
+  val proposed = Handle[S]()
+  val supported = Handle[S]()
+  val parameters = Handle[P]()
+
+  val proposedModifiers = ArrayBuffer[S => S]()
+  val supportedModifiers = ArrayBuffer[S => S]()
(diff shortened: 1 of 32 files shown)


kleinai:
There's a whole thread in TileLink node that's not implemented in `bus.fabric`. Is that assumed to be implementation specific?

Dolu1990:
> There's a whole thread in TileLink node that's not implemented in `bus.fabric`. Is that assumed to be implementation specific?

Yes, the idea is to keep it for the final class.

Also, note that for AXI, there will be no s2m, only m2s.

lib/src/main/scala/spinal/lib/bus/fabric/NegociateSP.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import spinal.core._
+import spinal.core.fiber.{Handle, soon}
+
+import scala.collection.mutable.ArrayBuffer
+
+/**
+  * Negotiation handles for master to slave requests
+  */
+abstract class NegociateSP[S, P] extends Area{
```
kleinai, line 11: Quick fix: this should be `NegotiateSP`

Dolu1990: Sure :D

Dolu1990:
I guess the dev can continue in the bus-fabric branch

@kleinai Were you thinking about implementing a AXI version ?

kleinai:
> I guess the dev can continue in the bus-fabric branch
> 
> @kleinai Were you thinking about implementing a AXI version ?

That was the plan to test out the new changes.

Dolu1990:
> > I guess the dev can continue in the bus-fabric branch
> > @kleinai Were you thinking about implementing a AXI version ?
> 
> That was the plan to test out the new changes.

Cool ^^
Maybe we should do a call then :)

PR #1167, opened 2023-07-21
Fix StreamFifoCC from leaking metastable io.pop.payload when empty

Dolu1990:
So, figured out that StreamFifoCC is reading the memory on every cycle, including when the fifo is empty, potentialy leaking a metastable value on the io.pop.payload.

This PR rework the pop side of the fifo a bit like what was done for the non-CC version.

the PR also fix a similar metastable leak on the RISC-V debug module

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1524,23 +1524,30 @@ class StreamFifoCC[T <: Data](val dataType: HardType[T],
   val finalPopCd = popClock.withOptionalBufferedResetFrom(withPopBufferedReset)(pushClock)
   val popCC = new ClockingArea(finalPopCd) {
     val popPtr      = Reg(UInt(log2Up(2*depth) bits)) init(0)
-    val popPtrPlus  = popPtr + 1
-    val popPtrGray  = RegNextWhen(toGray(popPtrPlus), io.pop.fire) init(0)
+    val popPtrPlus  = KeepAttribute(popPtr + 1)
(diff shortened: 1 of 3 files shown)


Dolu1990:
@Readon It pass SpinalSim fine, but not some of the formal tests.

Seems like some false positive due to bad initial state of the hardware in the formal sim.
For instance, fifo-verify fast pop fail via assert(pushCC.pushPtr - fromGray(pushCC.popPtrGray) <= depth) : 

```
##   0:00:02  Trying induction in step 3..
##   0:00:02  Trying induction in step 2..
##   0:00:03  Trying induction in step 1..
##   0:00:03  Trying induction in step 0..
##   0:00:03  Temporal induction failed!
##   0:00:03  Value for anyconst in unamed (unamed.sv:118.29-118.38): 0
##   0:00:03  Value for anyconst in unamed (unamed.sv:113.28-113.37): 6
##   0:00:03  Assert failed in unamed: unamed.sv:306.67-307.33
--   assert((_zz_32 <= 3'b100)); // Stream.scala:L1559
```

Looking at the vcd, seems like there is no reset + random value on the regs.

Any idea ?

Readon:
This might be caused by the changes of deifnition of popPtrGray.
``` diff
-    val popPtrGray  = RegNextWhen(toGray(popPtrPlus), io.pop.fire) init(0)
+    val popPtrGray  = toGray(popPtr)
```

Dolu1990:
@Readon Thanks ^^

Readon:
> @Readon Thanks ^^

It's my pleasure.

Issue #1165, opened 2023-07-20
Provide some API to speed up SpinalSim hardware signal read


Dolu1990:
I just mesured the time needed to read a signal value using the verilator backend : 

```scala
bt1.toLong   => 40 ns

vs

manager.getLong(signal1) => 5 ns,
```

Redondant code avoided : 

```scala
        val manager = SimManagerContext.current.manager
        val signal1 = manager.raw.userData.asInstanceOf[ArrayBuffer[Signal]](bt1.algoInt)
```

So, adding some API to provide a optimized signal access could realy help speeding up things for performance critical testbench

CPU used to test, AMD 5800X3D

Dolu1990:
Something like : 

```scala
//Once in sim
val proxy = dut.mem.node.bus.a.address.simProxy()
..
//Many time in sim 
val value = proxy.toLong // 5 ns overhead instead of 40
```

SimProxy being in spinal.core.sim package : 

```scala
  implicit class SimBitVectorPimper(bt: BitVector) {
    class SimProxy(bt : BitVector){
      val manager = SimManagerContext.current.manager
      val signal = manager.raw.userData.asInstanceOf[ArrayBuffer[Signal]](bt.algoInt)
      val alwaysZero = bt.getBitsWidth == 0
      def toLong = if(alwaysZero) 0 else manager.getLong(signal)
    }
    def simProxy() = new SimProxy(bt)
}
```

xiaochuang-lxc:
I think the ClockDomain Apis such as waitSampling will speed up more if change like this

Dolu1990:
The threadfull API would not realy get faster, as most of the overhead there is into doing JVM thread pack / unpack and switching threads i guess.

xiaochuang-lxc:
I do a very simple by overwriting the waitSampling(),it can speed up form:
test0:100000 times call 1092ms—>1027ms
test1:1000000 times call 9868ms—>9112ms

Dolu1990:
Ahhh i was expecting less difference ^^

andreasWallner:
I just did the check with modifying `forkStimulus` of a ClockDomain, gives ~10% in a testbench that does not much apart from that clock...

Readon:
Is it possible to modify the logic of toLong to avoid getting things too complicated?

andreasWallner:
> Is it possible to modify the logic of toLong to avoid getting things too complicated?

[...]

PR #1175, opened 2023-08-06
Use simProxy for driving clock

andreasWallner:
As mentioned in #1165, a small change so that a sim clock is driven using a proxy instead of direct access.
On my machine that gave ~10% performance improvement on clock-heavy tests. For more "normal" tests far less of an impact is to be expected.

Go fast ;-)

# Impact on code generation

None

# Checklist

No functionality change ~-  Unit tests were added~
No API change ~-  API changes are or will be documented:~


Dolu1990:
Thanks ^^

PR #1183, opened 2023-08-24
DFI Interface

LurenAA:
Closes #

I would like to add support for the DFI protocol(DDR PHY Interface), which may be useful for the development of certain SDRAM controllers.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/memory/Dfi.scala
+++ b/lib/src/main/scala/spinal/lib/memory/Dfi.scala
@@ -1,0 +1,305 @@
+package spinal.lib.memory
+
+import spinal.core._
+import spinal.lib._
+
+case class DfiTimeConfig(
+    tPhyWrLat: Int,
+    //  the number of cycles between when the write command is driven on the DFI to 
+    // assertion of the dfi_wrdata_en signal
+    tPhyWrData: Int,
+    // The tphy_wrdata parameter works to define the number of cycles from the
+    // assertion of a write command on the DFI control interface 
+    // to when write data is driven on the DFI bus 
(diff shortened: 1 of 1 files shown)


Readon:
So the next step would be 
1. A DDR SDRAM controller use DFI as master port.
2. Write at least one wrapper for DFI to Vivado Serdes connection.

Dolu1990:
> So the next step would be
> 
> 1. A DDR SDRAM controller use DFI as master port.
> 2. Write at least one wrapper for DFI to Vivado Serdes connection.

There is already a DDR2 / DDR3 / SDR controller in the lib : 
https://github.com/SpinalHDL/SpinalHDL/tree/dev/lib/src/main/scala/spinal/lib/memory/sdram/xdr

As phy is has Artix 7 support (serdes)

But the phy interface is different.

Readon:
> > So the next step would be
> > 
> > 1. A DDR SDRAM controller use DFI as master port.
> > 2. Write at least one wrapper for DFI to Vivado Serdes connection.
> 
> There is already a DDR2 / DDR3 / SDR controller in the lib : https://github.com/SpinalHDL/SpinalHDL/tree/dev/lib/src/main/scala/spinal/lib/memory/sdram/xdr
> 
> As phy is has Artix 7 support (serdes)
> 
> But the phy interface is different.

Yes, this work is to fit the legacy controller with more support on DFI, so that the phy can be decoupled from the controller.

Dolu1990:
Hi, sorry for the delay.

I think we can go like this. Thanks :)

PR #1192, opened 2023-09-03
move tests to speed up ci.

### LINKED ISSUE #982: About tests organization
Below is the first part of the path:

```
tester/src/test/scala/spinal/
```

The standard for unit tests would be:

```
PROJECT/src/test/scala/spinal/
```

So let's admit that (most) tests running simulations are integration tests, hence putting them in a separate `tester/` module. It also simplifies dependencies management for the day we want to use features released in `tester/` to test something from `core/` or `lib/`.

IMHO nothing to change here :+1:

---

```
tester/src/test/scala/spinal/tester/
                             ^^^^^^
```

Then there is the `tester/` folder. As it is about integration tests, it forces to import tested things (else using the same package as the test being run would bring in scope the stuff if the same package), making tests look more like examples, to better test the user's API, which seems to be a good thing.

IMHO nothing to change here :+1:

---

Then there are lots of tests with poor organization. Below is a suggestion:

* `drafts/` for worksheets (old things that could be removed IMHO), so that most contributors can forget them
* `path/to/item.scala` where the first element of the path is basically `core`, `lib`, etc. without being repeated.
* `issues/IssueXXX.scala` for bug fixes

For instance:

```
tester/src/test/scala/spinal/tester/scalatest/AhbLite3InterconnectTester.scala
```

tests:

```
lib/src/main/scala/spinal/lib/bus/amba3/ahblite/AhbLite3Interconnect.scala
```

So it would be moved to:

```
tester/src/test/scala/spinal/tester/lib/bus/amba3/ahblite/AhbLite3InterconnectTester.scala
```

Tests about different elements (elements being tested, not elements used to test other elements) would be in the deepest common element in the paths.

I think it is okay to have `App`s to test Scala syntax and types among scalatest-run tests.

This way it would be easier to find and add more tests.

Readon:
Closes #982 continue #1001 

distribute some tests in tester to speed up CI.

# Impact on code generation

# Checklist

- [ ] Unit tests were added


Readon:
now CI is only take 31 minutes for one run.

Dolu1990:
<3
Let's try to make it merged fast, as it move a lot of stuff around.

andreasWallner:
In general I'm all for this, but the unmanaged dependencies make me a bit weary. I understand that it doesn't compile without them as is, and I checked a few resources on how to structure this (i.e. this one: https://users.scala-lang.org/t/sbt-multi-project-dependencies/7948/8), pretty much all advise to have a separate subproject.
I also found https://groups.google.com/g/simple-build-tool/c/RPMxxaRtXn4/m/OX4SLbtdMkEJ which seems to endorse this, but it was for sbt 0.10.

I did a quick check and it seems that even if we have the test in the `test` subproject we can still replicate the package structure and put additional stuff into e.g. the `spinal.core` package which makes the IDE features work. 
We would also be able to get the CI speedup that you are currently seeing, but in that case by running separate jobs with `testOnly spinal.core.*` / `testOnly spinal.lib.*` (and we'd get a similar effect of making it easier to find tests for a certain class)

To be clear: there is no real technical reason I can point to. The main motivation is that I've been on and off debugging why SpinalHDL is so often rebuilt when running sbt from the command line and from googling I often found issues around unmanagedClasspath. It also seems weird to work around the package manager in that way if there is an alternative.

Readon:
> In general I'm all for this, but the unmanaged dependencies make me a bit weary. I understand that it doesn't compile without them as is, and I checked a few resources on how to structure this (i.e. this one: https://users.scala-lang.org/t/sbt-multi-project-dependencies/7948/8), pretty much all advise to have a separate subproject. I also found https://groups.google.com/g/simple-build-tool/c/RPMxxaRtXn4/m/OX4SLbtdMkEJ which seems to endorse this, but it was for sbt 0.10.
> 
> I did a quick check and it seems that even if we have the test in the subproject we can still replicate the package structure and put additional stuff into e.g. the package which makes the IDE features work. We would also be able to get the CI speedup that you are currently seeing, but in that case by running separate jobs with / (and we'd get a similar effect of making it easier to find tests for a certain class). I did the check by keeping all your source changes, but moving e.g. to .`test``spinal.core``testOnly spinal.core.*``testOnly spinal.lib.*``core/test/scala/spinal/core``tester/test/scala/spinal/core`
> 
> To be clear: there is no real technical reason I can point to. The main motivation is that I've been on and off debugging why SpinalHDL is so often rebuilt when running sbt from the command line and from googling I often found issues around unmanagedClasspath. It also seems weird to work around the package manager in that way if there is an alternative.

So do you mean if we move core/src/test/scala/spinal/core directory into tester/src/test/scala/spinal/core would be better than put them into the core lib's test?

Readon:
> > In general I'm all for this, but the unmanaged dependencies make me a bit weary. I understand that it doesn't compile without them as is, and I checked a few resources on how to structure this (i.e. this one: https://users.scala-lang.org/t/sbt-multi-project-dependencies/7948/8), pretty much all advise to have a separate subproject. I also found https://groups.google.com/g/simple-build-tool/c/RPMxxaRtXn4/m/OX4SLbtdMkEJ which seems to endorse this, but it was for sbt 0.10.
> > I did a quick check and it seems that even if we have the test in the subproject we can still replicate the package structure and put additional stuff into e.g. the package which makes the IDE features work. We would also be able to get the CI speedup that you are currently seeing, but in that case by running separate jobs with / (and we'd get a similar effect of making it easier to find tests for a certain class). I did the check by keeping all your source changes, but moving e.g. to .`` test``spinal.core``testOnly spinal.core.*``testOnly spinal.lib.*``core/test/scala/spinal/core``tester/test/scala/spinal/core ``
> > To be clear: there is no real technical reason I can point to. The main motivation is that I've been on and off debugging why SpinalHDL is so often rebuilt when running sbt from the command line and from googling I often found issues around unmanagedClasspath. It also seems weird to work around the package manager in that way if there is an alternative.
> 
> So do you mean if we move core/src/test/scala/spinal/core directory into tester/src/test/scala/spinal/core would be better than put them into the core lib's test?

How about Readon/SpinalHDL@move-tests this one?

Readon:
I have tested on my server that "go to test" function could only jump to files that with the right filename, such as xxxTest.scala for yyy.scala which include xxx class's definition.

Dolu1990:
hmmm kind of like it (stuff still in tester module, but in proper separated packages)
I think for me that's good for merge.

PR #1202, opened 2023-09-14
Bug fix for #1190: VCS run flag not pass to simv

### LINKED ISSUE #1190: VCS runFlags unable to pass into the simulation
There is a `runFlags` in VCSFlags which is used to configure the simulation process. But after the rework of the VCS Flags on PR #911 , the `runFlags` in testbench as following cannot pass into the simulation.
```scala
  val flags = VCSFlags(
    compileFlags = List(
      "-kdb"
    ),
    elaborateFlags = List("-kdb"),
    runFlags = List("-l sim.log")
  )
```

The simulation log:
image

As we can see in the picture, the info of the simulation command is just `[SIMV] ./directionEnc` which did not contain the `-l sim.log` option.

I found it is because the `runFlags` belongs to `VCSFlags` and does not pass to `VpiBackandConfig.runFlags`. Therefore in Line 360:
https://github.com/SpinalHDL/SpinalHDL/blob/09e69e56dd4744abbb284b80a9c7cdd4699cc6f3/sim/src/main/scala/spinal/sim/VCSBackend.scala#L359-L367

the `config.runFlags` contains nothing. So there is nothing after the simulation execute command.

wswslzp:
Closes #1190 

As title.

# Impact on code generation

No impact.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/sim/src/main/scala/spinal/sim/VCSBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VCSBackend.scala
@@ -357,7 +357,7 @@ class VCSBackend(config: VCSBackendConfig) extends VpiBackend(config) {
       val iface = sharedMemIface
       val logger = new LoggerPrintWithTerminationFilter()
       def run(): Unit = {
-        val runCommand = Seq(s"./$toplevelName", config.runFlags).mkString(" ")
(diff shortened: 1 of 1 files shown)


Dolu1990:
<3

PR #1207, opened 2023-09-21
Add init parameter to Timeout

andreasWallner:
Resurrection of an older PR...

Currently `Timeout` is initialized to be `False`.
This can for sure be changed with `init` like this:
```scala
val t = Timeout(5)
t.init(True)
```
(which will produce slightly ugly verilog with two reset statements, but work)

But it feels different from other things, i.e. this doesn't do what one might initially assume: `val t = Timeout(5) init True` (changes type of t).

This change adds a parameter, making it feel more like other library components or functions like e.g. `rise`

RTD change tbd.

# Impact on code generation

None

### DIFF
--- a/lib/src/main/scala/spinal/lib/Utils.scala
+++ b/lib/src/main/scala/spinal/lib/Utils.scala
@@ -600,15 +600,17 @@ class Counter(val start: BigInt,val end: BigInt) extends ImplicitArea[UInt] {
 }
 
 object Timeout {
-  def apply(cycles: BigInt) : Timeout = new Timeout(cycles)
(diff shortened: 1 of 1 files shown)


Dolu1990:
hmm, adding the following to the Timeout class : 

```scala
  def init(value : Bool): this.type = {
    state.removeInitAssignments()
    state.init(value)
    this
  }
```

would work cleanly.
Allowing a : 

```scala
    val t = Timeout(5) init(True)    
```

What about doing things like this ?

andreasWallner:
sure, just saw that some other utilities use the same

Dolu1990:
<3

Issue #278, opened 2020-03-03
Failed to Simulate BlackBox Modules contains $error()


depctg:
When I tries to start simulating a black box module contains `$assert` and `$error`, I got the following error:

```bash
Undefined symbols for architecture x86_64:
  "sc_time_stamp()", referenced from:
      Vunamed::_initial__TOP__1(Vunamed__Syms*) in Vunamed__ALL.a(Vunamed__ALLcls.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Vunamed] Error 1
Exception in thread "main" java.lang.AssertionError: assertion failed: Verilator C++ model compilation failed
```

When I comment out the lines contains $error, the error message is gone. I guess this is caused by Verilator. It wants to attach timestamp to the error message, but it relies on cpp function `sc_time_stamp`. I see some related messages in [verilator repo](https://github.com/grg/verilator/blob/9caffe330b602b630a541207dbb6606e6e3be183/bin/verilator#L3557). I guess theres already java side hooks but external rtls relies on cpp ones.

Zhiyuan

Dolu1990:
Hmm so, i should add in the C++ main generated by SpinalHDL a implementation of sc_time_stamp ?

Dolu1990:
Does sc_time_stamp give the simulation time ? or real time ? Because if that's simulation time, that's quite enoying as it do not give any context, so not realy compatible with multiple simultion running at the same time, unless using thread context variables

depctg:
Thanks for your reply! I guess we need to generate a cpp `sc_time_stamp` function. Currently I do not relies on time information provided by these verilog $errors, so an empty functions to make the compiler happy should be good enough to me.

And, Unfortunately, `sc_time_stamp` returns simulation time. You can check the [example implantation](https://github.com/grg/verilator/blob/9caffe330b602b630a541207dbb6606e6e3be183/bin/verilator#L1580).

Dolu1990:
@depctg I pushed something in dev which should fix it, could you give a try ? Do you know how to try unreleased SpinalHDL ?
https://spinalhdl.github.io/SpinalDoc-RTD/SpinalHDL/About%20SpinalHDL/faq.html#how-to-use-an-unreleased-version-of-spinalhdl-but-commited-on-git

PR #1210, opened 2023-09-26
Update C++ version for verilator (to C++14)

### LINKED ISSUE #1209: Simulation fails when using Verilator of new version which requires C++14 standard
When using Verilator of new version (such as 5.016), simulation fails and complains something like

```
[info] In file included from /usr/local/share/verilator/include/verilated.h:937,
[info]                  from VLinear___024root__DepSet_h572e5853__3.cpp:5:
[info] /usr/local/share/verilator/include/verilated_types.h: In function 'constexpr IData VL_CLOG2_CE_Q(QData)':
[info] /usr/local/share/verilator/include/verilated_types.h:85:1: error: body of 'constexpr' function 'constexpr IData VL_CLOG2_CE_Q(QData)' not a return-statement
[info]    85 | }
[info]       | ^
```

It seems that Verilator requires using C++14 standard to compile the simulation files. However, the compile options are hard coding in `sim/src/main/scala/spinal/sim/VerilatorBackend.scala`. Is it possible to change compile options by user?

andreasWallner:
Closes #1209

With the jump to verilator 5, verilator started using newer C++ syntax. For now it seems that compiling against C++14 is sufficient (when not using the recent timing support that required C++20).

It seems we pinned the C++ version to C++11 to support older compilers (see #278).
This PR removes the pinned version and goes back to using the compiler default that should work in most cases, possibly being faster then if set to C++11. It provides an additional environment variable to add arbitrary setting during compilation, therefore also e.g. allowing for setting the linker on a per-machine basis.

C++11 or larger has been the default for recent compiler for a while (introduced in 2020, default in 2021) - if a user wants to use latest verilator they have to have a matching gcc version (min C++14, available since gcc 6.1, though not as a default).

# Impact on code generation

No impact on code generation.
When updating existing simulation caches may be broken because of C++ version mismatches.

### DIFF
--- a/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
@@ -519,6 +519,10 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
 
     val verilatorBinFilename = if(isWindows) "verilator_bin.exe" else "verilator"
 
+    // allow a user to overwrite/add verilator flags, e.g. C++ version
+    // if the default is too old (see e.g. #278)
+    val envFlags = sys.env.getOrElse("SPINAL_VERILATOR_FLAGS", "")
+
     // when changing the verilator script, the hash generation (below) must also be updated
     val verilatorScript = s""" set -e ;
        | ${verilatorBinFilename}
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks :D

Readon:
It seems that the test time has been reduced slightly, by adopting this PR. On my machine it runs only 272s comparing to previously 296s.

PR #1212, opened 2023-09-28
Fix signal name typos

IanBoyanZhang:
Closes #

Fix signal name typos to improve readability

# Impact on code generation

Should not have any impact on code generation

# Checklist

- [ ] Unit tests were added


andreasWallner:
First off, thanks for the PR!

I'm a bit unsure about this one: in other cases where we corrected spelling mistakes we keep the old spelling in the UI with a deprecation note (to not break other repos outright with a release, e.g. https://github.com/SpinalHDL/SpinalHDL/blob/a9afa0338000d5eb11fdb7786efed93d9698432b/core/src/main/scala/spinal/core/ClockDomain.scala#L390), but we don't have precedent for signal names of components in lib.

@Dolu1990 what do you think, should we do the same here?

Dolu1990:
Hi,


I think for this case that's fine to merge directly, as the SPI stuff is not something which spread everywere in a code base, that's quite specific.

For more general cases, could define a def which provide the old version with wrong typo.

Thanks ^^

PR #1215, opened 2023-10-04
Add option for IO instantitation to InOutWrapper

andreasWallner:

I encountered issues inferring drivers on at least Xilinx (infers only on top level, not for components in a block diagram) and ICE40 w/ yosys (warns on inferred drivers and produced invalid result for me in a case that I just debugged).
To work around those issues, give the InOutWrapper an additional parameter with a callback for generating instantiations of hardmacros.

# Impact on code generation

none

# Checklist

- API changes are or will be documented: RTD tbd


Dolu1990:
Thanks :)

PR #1218, opened 2023-10-06
Update built Scala minor version to latest

andreasWallner:

Continue to build with the supported minor version.
For now I still kept 2.11 as the first in the list, and therefore the one with which all tests are executed, we might want to revisit this in the future.

# Impact on code generation

Checked unittests locally with all Scala versions, no major issues.
I only encountered one issue running the test with 2.13.x, which is that the test for physical number formatting fails - I'll investigate that separately since it also occurs with the current 2.13.6.


Dolu1990:
thanks ^^

PR #1243, opened 2023-11-14
disable scalafmt action

Dolu1990:
So far, that feature kinda adding noise into the github PR, and isn't enforced, so let's just disabled it.


Readon:
Ahhh, I really like the formatted code than not do that at all. :)

Issue #1234, opened 2023-11-03
XSim on Windows : getting it to work as expected through cmd or MSYS2(MINGW64)


oletf:
Greetings, 

I'm currently trying to get into SpinalHDL,
and to use XSim simulation on Windows.

In the test project's `Config.scala` I changed the `def sim = SimConfig.withConfig(spinal).withFstWave`  to `SimConfig.withConfig(spinal).withXSim.withWave.withXilinxDevice("xc7a35tftg256-2")`.

It complains [at this point](https://github.com/SpinalHDL/SpinalHDL/blob/7314a3d140b89c5ddc1f8d46274c28c36b98af46/sim/src/main/scala/spinal/sim/XSimBackend.scala#L212-L216) because I don't have any `sh` in my Path.

I was wondering if it would be possible to make the XSimBackend call the `vivado` command without calling `sh -c [...]` (or at least add a setting to opt-out), 
although I have no idea if it would cause more problems afterwards in the case where other shell based calls or tools are needed in the current state of the simulation backend, 
so feel free to close if it would need too many changes.

When launching simulation from within Windows GUI, Vivado seems to copy an `xsimk.exe` in the project's sim subdir,
and then calls it with `xsim.dir/testbench_archname/xsimk.exe -simmode gui -wdb testbench_archname.wdb -simrunnum 0 -socket 57473`.
 
Thanks forward

Readon:
May be you can use sbt in msys2 environment. And share Vivado related environment variables from windows.

Readon:
This might also be solved through the scala's process helper. It can be used as below:
``` scala
import scala.sys.process._

object Test extends App {
    val cmd = "ls"
    val exitCode = cmd.!
}
```

oletf:
Thanks a lot for the hint,
indeed going through msys2 seems to get the process further !

I still get a problem which seems to be related to Vivado xsi interface path... here's my log:
```
sbt:projectname> runMain projectname.MyTopLevelSim
[info] running (fork) projectname.MyTopLevelSim
[info] [Runtime] SpinalHDL v1.9.4    git head : 270018552577f3bb8e5339ee2583c9c22d324215
[info] [Runtime] JVM max memory : 8152,0MiB
[info] [Runtime] Current date : 2023.11.06 12:57:20
[info] [Progress] at 0,000 : Elaborate components
[info] [Progress] at 0,383 : Checks and transforms
[info] [Progress] at 0,475 : Generate Verilog
[...]

PR #1246, opened 2023-11-16
XSim: better Windows support + add xcix IP import

oletf:
Follow-up to #1234

- Context : from https://github.com/SpinalHDL/SpinalHDL/issues/1234#issuecomment-1812168567
- Description : 
  - calls to Vivado generated scripts are called from `cmd` native shell
  - msys2 still needed but only to build `spinal_xsim.dll` 
  - wdb file moved to xsim subfolder to avoid Vivado cluttering Project root folder when opening wdb viewer
  - added `xcix` import along `xci` IP
# Impact on code generation

None

# Checklist

~~- [ ] Unit tests were added~~
- [ ] XSim Simulation behavior changes started to be documented:
  - on RTD : https://github.com/SpinalHDL/SpinalDoc-RTD/pull/230

### DIFF
--- a/sim/src/main/scala/spinal/sim/XSimBackend.scala
+++ b/sim/src/main/scala/spinal/sim/XSimBackend.scala
@@ -89,48 +89,60 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
   val compilePath        = s"${simulatePath}/compile.${scriptSuffix}"
   val elaboratePath      = s"${simulatePath}/elaborate.${scriptSuffix}"
 
+  val isMsys = sys.env.get("MSYSTEM").isDefined
+
   val msys2Shell = {
     if (isWindows) {
-      val msys2ShellPath =
-        sys.env.getOrElse("MSYS2_ROOT", "C:\\msys64") + "\\msys2_shell.cmd"
-      if (!new File(msys2ShellPath).exists) {
-        throw new Exception(
-          "MSYS2 Shell not found! Please Setup you MSYS2_ROOT Environment Variable."
-        )
+      if (isMsys) { // SBT Called from Msys
+        "sh"
+      } else { // SBT Called from CMD
+        val msys2ShellPath =
+          sys.env.getOrElse("MSYS2_ROOT", "C:\\msys64") + "\\msys2_shell.cmd"
+        if (!new File(msys2ShellPath).exists) {
+          throw new Exception(
+            "MSYS2 Shell not found! Please Setup you MSYS2_ROOT Environment Variable."
+          )
+        }
+        msys2ShellPath + " -defterm -here -no-start -mingw64"
       }
-      msys2ShellPath
     } else {
       ""
     }
   }
+
   val vivadoHome = sys.env.getOrElse(
     "VIVADO_HOME",
     throw new Exception(
       "VIVADO_HOME not found! Setup your vivado environment first."
     )
   )
+  val vivadoBinPath = new File(s"${vivadoHome}/bin").getAbsolutePath
 
-  def doCmd(command: String, cwd: File, message : String) = {
-    val logger = new Logger()
-    if(Process(command, cwd)! (logger) != 0){
-      println(logger.logs)
-      throw new Exception(message)
-    }
-  }
-  def doCmd(command: String, cwd: File, extraEnv: Seq[(String, String)], message : String) = {
+  def doCmd(command: String, cwd: File, message : String, extraEnv: Seq[(String, String)] = Seq.empty) : Unit = {
     val logger = new Logger()
     if(Process(command, cwd, extraEnv :_*)! (logger) != 0){
       println(logger.logs)
       throw new Exception(message)
     }
   }
-  def doCmdMys2(command: String, cwd: File, message : String) = {
-    val cmd_msys2 = s""" $msys2Shell -defterm -here -no-start -mingw64 -c "$command" """.trim
-    doCmd(cmd_msys2, cwd, message)
+  def doCmdVivado(command: String, cwd: File, message: String, extraEnv: Seq[(String, String)] = Seq.empty) = {
+    val cmdVivado =
+      if (isWindows) {
+        if (isMsys) {
+          // msys turns "/c" to "c:\" because of posix path conversion, needs escape
+          s"cmd //c ${vivadoBinPath}/setEnvAndRunCmd.bat ${command}"
+        } else {
+          s"cmd /c ${vivadoBinPath}/setEnvAndRunCmd.bat ${command}"
+        }
+      } else {
+        s"${vivadoBinPath}/setEnvAndRunCmd.sh ${command}"
+      }
+
+    doCmd(cmdVivado, cwd, message, extraEnv)
   }
-  def doCmdMys2(command: String, cwd: File, extraEnv: Seq[(String, String)], message : String) = {
-    val cmd_msys2 = s""" $msys2Shell -defterm -here -no-start -mingw64 -c "$command" """.trim
-    doCmd(cmd_msys2, cwd, extraEnv, message)
+  def doCmdMys2(command: String, cwd: File, message: String, extraEnv: Seq[(String, String)] = Seq.empty) : Unit = {
+    val cmdMsys2 = s""" $msys2Shell -c "$command" """.trim
+    doCmd(cmdMsys2, cwd, message, extraEnv)
   }
 
 
@@ -229,32 +241,25 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
     outFile.close()
 
     val vivadoScriptPath = scriptPath.replace("\\", "/")
-    val command = {
-      val baseCommand = s""" vivado -mode batch -source "$vivadoScriptPath" """.trim
-      if (isWindows) {
-        s"cmd /c $baseCommand"
-      } else {
-        baseCommand
-      }
-    };
-    doCmd(command,
+    val command =  s""" vivado -mode batch -source "$vivadoScriptPath" """.trim
+    doCmdVivado(command,
       new File(workPath),
       "Generation of vivado script failed")
   }
 
   def getScriptCommand(cmd: String) = {
     if (isWindows) {
-      s"cmd /c ${cmd}.bat"
+      s"${cmd}.bat"
     } else {
-      s"bash ${cmd}.sh"
+      s"${cmd}.sh"
     }
   }
 
   def runScript() = {
-    doCmd(getScriptCommand("compile"),
+    doCmdVivado(getScriptCommand("compile"),
       new File(simulatePath),
       "run compile failed")
-    doCmd(getScriptCommand("elaborate"),
+    doCmdVivado(getScriptCommand("elaborate"),
       new File(simulatePath),
       "run elaborate failed")
       val simDesignDir = s"${simulatePath}/xsim.dir"
--- a/sim/src/main/scala/spinal/sim/XSimBackend.scala
+++ b/sim/src/main/scala/spinal/sim/XSimBackend.scala
@@ -84,11 +84,23 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
       ".sh"
     }
   }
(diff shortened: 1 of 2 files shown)


Readon:
Please notice the fact that many people is using SpinalHDL on Windows without MSYS2 environment. So if it is on Windows there might be chance that MSYS2_ROOT is not there at all.

oletf:
The only files I changed are related to XSim usage, 
so if they currently don't use XSim for simulation they won't see any change as those file's code won't run more as they currently do.

But I admit that for any Windows user who wants to use XSim, the `spinal_xsim.dll` needs to be compiled, so it'll be necessary for the user to have at least some `g++` or whatever toolchain which would be able to build it.

I kept using MSYS2 shell in this PR because that's what was used previously already.

I found some mingw in `%VIVADO_HOME%\tps\mingw` which has some `g++` but I haven't checked its limitations.

sim/src/main/scala/spinal/sim/XSimBackend.scala
```diff
@@ -84,18 +84,31 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
...
+  val simulateDir        = s"${workPath}/${toplevelName}_xsim.sim/sim_1/behav/xsim"
+  val simulatePath       = new File(simulateDir).getAbsolutePath
+  val compilePath        = s"${simulatePath}/compile.${scriptSuffix}"
+  val elaboratePath      = s"${simulatePath}/elaborate.${scriptSuffix}"
+
+  val msys2Shell = {
+    if (isWindows) {
+      val msys2ShellPath =
+        sys.env.getOrElse("MSYS2_ROOT", "C:\\msys64") + "\\msys2_shell.cmd"
+      if (!new File(msys2ShellPath).exists) {
```
Readon, line 96: this part is what I concerned.

sim/src/main/scala/spinal/sim/XSimBackend.scala
```diff
@@ -208,9 +230,9 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
 
     val vivadoScriptPath = scriptPath.replace("\\", "/")
     val command = {
-      val baseCommand = s"vivado -mode batch -source $vivadoScriptPath"
+      val baseCommand = s""" vivado -mode batch -source "$vivadoScriptPath" """.trim
       if (isWindows) {
-        s"sh -c \'${baseCommand}\'"
+        s"cmd /c $baseCommand"
```
Readon, line 235: so do you use cmd console by default?

sim/src/main/scala/spinal/sim/XSimBackend.scala
```diff
@@ -111,6 +124,15 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
       throw new Exception(message)
     }
   }
+  def doCmdMys2(command: String, cwd: File, message : String) = {
+    val cmd_msys2 = s""" $msys2Shell -defterm -here -no-start -mingw64 -c "$command" """.trim
```
Readon, line 128: this command would not use env variable in msys if you start it through cmd shell on Windows.

Readon:
there are normally 3 ways to use.
1. on linux
2. on Windows, use cmd shell
3. on Windows, use bash of msys2

sim/src/main/scala/spinal/sim/XSimBackend.scala
oletf: yes, as mentioned in the issue a few times : 
my goal was using native cmd console by default to call everything that can be : 
`vivado.bat`, and its generated `compile.bat` and `elaborate.bat` : 
so whatever if user starts sbt from cmd or from msys shell :
- all `.bat` will be run by cmd
- the `g++` command to build `spinal_xsim.dll`

sim/src/main/scala/spinal/sim/XSimBackend.scala
oletf: yes, because on windows the build of `spinal_xsim.dll` depends only on the `VIVADO_HOME` env var to copy locally needed include files.

This behavior was already there before those commits, and I kept it,
as it allows to have this only step which actually doesn't depends on any PATH (cmd or msys) env var.

sim/src/main/scala/spinal/sim/XSimBackend.scala
oletf: indeed I completely missed that this would fail on msys2 shell, thanks for spotting it!

`msys2Shell` would get :
- `sh` if sbt run by msys (check with `sys.env.get("MSYSTEM")` ?)
- the full `%MSYS2_ROOT%\\msys2_shell.cmd -defterm -here -no-start -mingw64` if ran by cmd

would that be good enough ?

pretty unsure about checking msys vs cmd with only `MSYSTEM` envvar, 
thought about checking if `sh` exists, but there can be some other `sh` in cmd PATH

andreasWallner:
> there are normally 3 ways to use.
> 
>     1. on linux 
>     2. on Windows, use cmd shell 
>     3. on Windows, use bash of msys2

For the windows cases: aren't they the same in the sense that g++ must always run in the msys environment (since the compiler is installed there), whereas vivado "must" always run in the cmd environment (since there are no environment settings files for bash, only `settings64.bat`)?

I think from that point it makes sense to call g++ via msys and vivado via cmd.

Readon:
> > there are normally 3 ways to use.
> > ```
> > 1. on linux 
> > 2. on Windows, use cmd shell 
> > 3. on Windows, use bash of msys2
> > ```
> 
> For the windows cases: aren't they the same in the sense that g++ must always run in the msys environment (since the compiler is installed there), whereas vivado "must" always run in the cmd environment (since there are no environment settings files for bash, only `settings64.bat`)?
> 
> I think from that point it makes sense to call g++ via msys and vivado via cmd.

you are using method 2.

I am currently use the MSYS2 shell to start Vivado, the description in SpinalHDL/SpinalDoc-RTD#230 has been tested without cmd at all. and also with the all-in-one msys2 bundle.

A easy and unify method is prefered, because the environment of Windows norrmally a mess while the number of programs increasing.

andreasWallner:
@Readon anything else from your side, from my POV we can merge this

Readon:
> @Readon anything else from your side, from my POV we can merge this (I tried it on Linux)

I still could not run this under msys2 environment.

Readon:
> > @Readon anything else from your side, from my POV we can merge this (I tried it on Linux)
> 
> I still could not run this under msys2 environment.

What I am using is the portable environment installed by MSYS2-installer. I think in MSYS2 there sould be no assumption that there would be a cmd.exe exist.

Readon:
I also tested in the cmd console it still prompt that "Please Setup you MSYS2_ROOT Environment Variable." after I have set it by "set MSYS2_ROOT="xxxxxxx"".  I have to define all environment variable in Windows System's advance settings. It might be not good if I just want to run it once, but pollute my whole environment variables.

After that, I still get " Generation of vivado script failed" error with backtrace like below.
``` bash
[error]         at spinal.sim.XSimBackend.doCmd(XSimBackend.scala:125)
[error]         at spinal.sim.XSimBackend.doCmdVivado(XSimBackend.scala:141)
[error]         at spinal.sim.XSimBackend.generateScript(XSimBackend.scala:245)
```

Readon:
One more thing, if the xcix is supported after this PR, how about add a minimal test on it? So that we could test this each time the CI runs.

oletf:
> What I am using is the portable environment installed by MSYS2-installer. I think in MSYS2 there sould be no assumption that there would be a cmd.exe exist.

For any MSYS2 installation, I assumed that we're on Windows, and that the `cmd.exe` native windows shell existing since the very beginnings of the OS is always present.
Also the fact that launching a `.bat` is not possible without a minimal `cmd /c` call, and that even calling a `.bat` from msys2 will call it through a cmd subprocess.
Or did you refer to the `msys2_shell.cmd` script which maybe could not have been yet added in your current msys2 version ?

> "set MSYS2_ROOT="xxxxxxx""

Windows cmd syntax is very misleading on many things, 
if you got spaces in the path you want to set I think it's better to do it by `set "MSYS2_ROOT=xxxxxxx"` ([see this for details](https://stackoverflow.com/questions/34353504/how-to-set-environment-variables-with-spaces)).
I always forget this it's terrible...

>  I have to define all environment variable in Windows System's advance settings.

Which other env var than the `MSYS2_ROOT` and `VIVADO_HOME` did you have to set to reach to the errors you got ?
Now I'm actually unsure whether java's `sys.env.get` gets Windows or MSYS2 env var when the java process in running as child of msys2 shell.

 Writing this paragraph above I now realize that the failing at `XSimBackend.scala:245` might actually be either : 
- caused by either msys2 env being different from mine
- Vivado path resolution failing at some point
- both

So yeah I think I cannot properly test running this by sbt from msys2, since my sbt install is in the windows side,
and msys2 is just used for g++ in my case.
I might have overlooked another thing for it to work in your specific case, 
please let me know how I can try to reproduce your case on my side !

oletf:
> One more thing, if the xcix is supported after this PR, how about add a minimal test on it? So that we could test this each time the CI runs.

I admit I just added xcix because I needed it in my project, and just adding it to the IP import list didn't cause any problem from Vivado's pov.

Not sure about how to add a proper test for this (still lacking experience in scala testing workflow), 
is there an `xci` test which already exists I could refer to ?

Readon:
> For any MSYS2 installation, I assumed that we're on Windows, and that the `cmd.exe` native windows shell existing since the very beginnings of the OS is always present. Also the fact that launching a `.bat` is not possible without a minimal `cmd /c` call, and that even calling a `.bat` from msys2 will call it through a cmd subprocess. Or did you refer to the `msys2_shell.cmd` script which maybe could not have been yet added in your current msys2 version ?

I suppose it is not a good assumption. I run the TestXSim in a standalone MSYS2 environment, there is no cmd. However, I pushed a PR on your repos, may be that helps. And it can call the `compile.bat` and `elaborate.bat` files, while you have used `setEnvAndRunCmd.sh` first by `msys2_shell.cmd` script.

> Windows cmd syntax is very misleading on many things, if you got spaces in the path you want to set I think it's better to do it by `set "MSYS2_ROOT=xxxxxxx"` ([see this for details](https://stackoverflow.com/questions/34353504/how-to-set-environment-variables-with-spaces)). I always forget this it's terrible...

Yes, this might be the problem.

> Which other env var than the `MSYS2_ROOT` and `VIVADO_HOME` did you have to set to reach to the errors you got ? Now I'm actually unsure whether java's `sys.env.get` gets Windows or MSYS2 env var when the java process in running as child of msys2 shell.

It's weird that it can get VIVADO_HOME when I define it in console but in the system. And fails to get for `MSYS2_ROOT`.

> So yeah I think I cannot properly test running this by sbt from msys2, since my sbt install is in the windows side, and msys2 is just used for g++ in my case. I might have overlooked another thing for it to work in your specific case, please let me know how I can try to reproduce your case on my side !

If you want you can install the [MSYS2-Installer](https://github.com/Readon/msys2-installer/releases/tag/all-in-one-20231108131831), and define the environment variable in that. It can work either on current upstream code, and on the PR I have pushed on your repo.


The only problem remains is that I could not get the Windows starter, by cmd, to work now.

Readon:
> > One more thing, if the xcix is supported after this PR, how about add a minimal test on it? So that we could test this each time the CI runs.
> 
> I admit I just added xcix because I needed it in my project, and just adding it to the IP import list didn't cause any problem from Vivado's pov.
> 
> Not sure about how to add a proper test for this (still lacking experience in scala testing workflow), is there an `xci` test which already exists I could refer to ?

I think there is no, and the xcix is normally a commercially used one, I am not sure if you can upload it here.

Readon:
Sorry for the late response. 
I have tested the latest version on my machine. It works on MSYS2 standalone environment alright.

However, on Windows cmd & MSYS2 mixed environment it still report "UnsatisfiedLinkError", even when I defined VIVADO_HOME and add %VIVADO_HOME%\lib\win64.o to my PATH.
Almost the same as #1234 states.

Readon:
Sorry for the late response. Thanks for the contribution.

PR #1247, opened 2023-11-17
Commenting (and scaladocing) idslXXX

ronan-lashermes:

To understand the sources better, I documented idslplugin and idslpayload with comments and scaladoc.
Added a bit of cleanup, removing old commented code.

Comments are semi-automatic with GPT-4 and a manual refinement phase.

# Impact on code generation

No code has been modified, this should not cause any issue.

# Checklist

- Unit tests were added -> Not applicable
- API changes are or will be documented -> Not applicable

### DIFF
--- a/idslpayload/src/main/scala/spinal/idslplugin/Annotation.scala
+++ b/idslpayload/src/main/scala/spinal/idslplugin/Annotation.scala
@@ -1,9 +1,43 @@
 package spinal.idslplugin
 
+
+/**
+ * A trait representing a callback mechanism for post-initialization operations.
+ * 
+ * This trait is used to define an action that should be taken after an object has been initialized.
+ * Classes or traits implementing `PostInitCallback` can define custom behavior that is executed
+ * after the initial construction and setup of the instance.
+ */
 trait PostInitCallback {
+  /**
+   * The method to be called after the initialization of the object.
+   *
+   * Implementing classes should override this method to define custom post-initialization behavior.
+   * This method is expected to return the current instance (`this`) of the class.
+   *
+   * @return the current instance of the class implementing this trait.
+   */
   def postInitCallback(): this.type
 }
 
+/**
+ * A trait for callback handling during value assignments.
+ * 
+ * This trait provides a mechanism to define custom behavior whenever a value is assigned to a variable.
+ * It is particularly useful in scenarios where additional actions need to be taken or side-effects need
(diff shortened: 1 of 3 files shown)


andreasWallner:
@ronan-lashermes 

I appreciate the effort that you put into this, but to me much of this may help if you try to get into it but feels like overkill/disrupting, e.g.:
   
          // Check if the function call (Apply node) is a constructor.
          if (a.fun.symbol.isConstructor) {

my 0.02$ - I'll leave a decision for @Dolu1990

ronan-lashermes:
> I appreciate the effort that you put into this, but to me much of this may help if you try to get into it but feels like overkill/disrupting, e.g.:

Thanks for the feedback. If I'm honest, I was also testing a workflow for semiautomatic doc generation. I was quite positively surprised by the results. This pull request is around 30 min work.

Since I am not familiar with scala reflection API, there may be "trivial" comments.
I agree that it may not be necessary.
Let's see what @Dolu1990 think.

Dolu1990:
Hi,

So, i don't feel so confortable with it.

Here is an example : 

```scala
/**
 * A trait representing a callback mechanism for post-initialization operations.
 * 
 * This trait is used to define an action that should be taken after an object has been initialized.
 * Classes or traits implementing `PostInitCallback` can define custom behavior that is executed
 * after the initial construction and setup of the instance.
 */
trait PostInitCallback {
```

There are the issues : 

- The comment is very long for what it say. It could have been : 
"PostInitCallback allows the instances which implement it to have an automatic callback after their construction. This callback is managed by the idslplugin (a custom scala compiler plugin)"

- "post-initialization operations"
post-initialization isn't specific enough, as it could be used in much wider context than what the PostInitCallback is about.

- "after an object has been"
Scala has a object keyword, which refer to static / singleton stuff, better using "instance"

- "has been initialized"
initialization isn't specific enough, construction should be used i would say.

- The same concept is repeated 3 times "post-initialization operations" "an object has been initialized" " the initial construction"

Main thing i think is that a comment should add more value than its own length.

Don't get me wrong, i think that writing good comments is hard, and i'm so bad at it i'm kinda never writting any XD

Dolu1990:
> // Check if the function call (Apply node) is a constructor.

Agree that doesn't add much value for the disruption it add

andreasWallner:
Closing this because of lack of feedback

PR #1272, opened 2024-01-08
VerilatorBackend now cache improvement

Dolu1990:
VerilatorBackend now cache the whole shared object instead of .a + headers the avoiding linking

On the VexiiRiscv CPU simulation (no SoC) i got the following "Verilator compilation done in" : 
- Uncached : ~2.7s
- Cached : ~1.8s
- Cached + this PR : ~0.4s

### DIFF
--- a/sim/src/main/scala/spinal/sim/Misc.scala
+++ b/sim/src/main/scala/spinal/sim/Misc.scala
@@ -50,7 +50,7 @@ sealed class WaveFormat(val ext : String = "???"){
 
 
 trait Backend{
-  var uniqueId       = Backend.allocateUniqueId()
+  var uniqueId       = Backend.allocateUniqueId().toLong
   def isBufferedWrite : Boolean
 }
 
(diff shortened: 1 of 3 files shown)


Dolu1990:
@dlmiles :D

dlmiles:
good to see improvements

maybe the next thing is to cause a change (forcing at least one item to need to be recompiled) and note how many things are actually compiled, this might be an 'single file change incremental compile and link test' (maybe adding / removing  `/*verilator public */` across builds is such a scenario.)
also make a configuration change (that affects at least one verilator command line option) and confirm this causes a full rebuild, this might be a 'full recompile and link test' (maybe changing the -Dverilog_define=1 across builds is such a scenario)

also ensure ccache is not being used :D  (I think other peoples perceptions of the matter are mitigated by their use of ccache, I have never used ccache and worked on the source of the problem)

Dolu1990:
Thanks, i will try ^^

PR #1275, opened 2024-01-11
Update build.sc dependencies

KireinaHoro:
Fixes build issues with mill and synchronizes dependencies.


KireinaHoro:
This also includes changes to `build.sc` to allow using `spinalhdl` as a submodule (see an example [here](https://github.com/KireinaHoro/enzian-pio-nic/blob/075e38b392d256089a825f08bbf08b0e8ff5a9cc/build.sc#L20-L22)), by making all build modules cross-build traits, so another project could extend them (and override `millSourcePath`, for example).

Dolu1990:
Thanks :D

PR #1281, opened 2024-01-14
Fix comments for trait BitwiseOp

g0t00:

just minor change to comments which showed wrong info when hovering operator with metals before

# Impact on code generation
none

# Checklist

no code changes...


Dolu1990:
Thanks ^^

PR #1289, opened 2024-01-18
driveStream on BusSlaveFactory

### LINKED ISSUE #1265: BusSlaveFactory driveFlow does not honor byteEnable
It seems like the `driveFlow` function in `BusSlaveFactory` does not honour `byteEnable`:

https://github.com/SpinalHDL/SpinalHDL/blob/68b6158700fc2440ea7980406f927262c004faca/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala#L512-L516

I had the impression that the `valid` bit in the flow should not be on when `byteEnable` was all zero for the width of the flow. This scenario is tricky however, since it's not clear how to properly handle partial `byteEnable`s on a `Flow`:
- do we trigger the Flow but with false data on those that the byte is not enabled (similar to current situation), or
- do we drop the write completely?

A real-scenario where this would matter:  I am using a Axi4SlaveFactory after an AXI width converter.  For a wide write request with some bytes disabled with `strb`, the AXI width converter issues requests that contains beats with zero `strb`.  This then falsely triggers the downstream `Flow`.

KireinaHoro:
Closes #1265

The bus slave factory `driveFlow` does not currently honour `byteEnable`.  As discussed, this PR adds an optional mode where it checks for all-zero `byteEnable` and does not trigger valid on the flow.  For backwards compatibility we keep the old behaviour as default.

It also adds `driveStream` and `readStreamBlockCycles` as two blocking-mode operations on the bus control class.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
+++ b/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
@@ -406,12 +406,16 @@ trait BusSlaveFactory extends Area{
 
   /**
     * Create a writable Flow register of type dataType at address and placed at bitOffset in the word
+    *
+    * @param checkByteEnable do not trigger flow if byte enable is all zero.  See [[https://github.com/SpinalHDL/SpinalHDL/issues/1265]]
+    *                        for the discussion about this behaviour.
     */
-  def createAndDriveFlow[T <: Data](dataType  : T,
-                                    address   : BigInt,
-                                    bitOffset : Int = 0): Flow[T] = {
+  def createAndDriveFlow[T <: Data](dataType       : T,
+                                    address        : BigInt,
+                                    bitOffset      : Int = 0,
+                                    checkByteEnable: Boolean = false): Flow[T] = {
     val flow = Flow(dataType)
-    driveFlow(flow, address, bitOffset)
+    driveFlow(flow, address, bitOffset, checkByteEnable)
     flow
   }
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
thanks ^^

lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
```diff
@@ -502,28 +506,56 @@ trait BusSlaveFactory extends Area{
...
+  def driveStream[T <: Data](that: Stream[T], address: BigInt, bitOffset: Int = 0): Unit = {
+    val wordCount = (bitOffset + widthOf(that.payload) - 1) / busDataWidth + 1
+    onWritePrimitive(SizeMapping(address, wordCount * wordAddressInc), haltSensitive = false, null) {
+      when(!that.ready) {
+        writeHalt()
+      }
+    }
+    val flow = Flow(that.payloadType())
+    driveFlow(flow, address, bitOffset, checkByteEnable = true)
+    that << flow.toStream
```
Readon, line 556: Here the flow.toStream might not wait the ready signal.
[This code](https://github.com/SpinalHDL/SpinalHDL/blob/2879ece341aa3d09c297c38c13270fe2d29786b8/lib/src/main/scala/spinal/lib/Flow.scala#L70) demostrate that.

KireinaHoro:
The idea here is that the onWritePrimitive will block the bus until the stream is ready; in a word, Flow.toStream should never overflow.Am 28.01.2024 um 16:06 schrieb Readon ***@***.***>:﻿
@Readon commented on this pull request.



In lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala:
>        driveMultiWord(that.payload, address)
       that.valid := regValid
     }
   }
 
+  /**
+   * Emit on that a transaction when a write happen at address, by using data placed at bitOffset in the word.
+   * Block the write transaction until the transaction succeeds (stream becomes ready).
+   */
+  def driveStream[T <: Data](that: Stream[T], address: BigInt, bitOffset: Int = 0): Unit = {
+    val wordCount = (bitOffset + widthOf(that.payload) - 1) / busDataWidth + 1
+    onWritePrimitive(SizeMapping(address, wordCount * wordAddressInc), haltSensitive = false, null) {
+      when(!that.ready) {
+        writeHalt()
+      }
+    }
+    val flow = Flow(that.payloadType())
+    driveFlow(flow, address, bitOffset, checkByteEnable = true)
+    that << flow.toStream

Here the flow.toStream might not wait the ready signal.
This code demostrate that.

Readon:
> The idea here is that the onWritePrimitive will block the bus until the stream is ready; in a word, Flow.toStream should never overflow.Am 28.01.2024 um 16:06 schrieb Readon ***@***.***>:﻿ @Readon commented on this pull request. In lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala:
> driveMultiWord(that.payload, address)
> that.valid := regValid } } + /** + * Emit on that a transaction when a write happen at address, by using data placed at bitOffset in the word. + * Block the write transaction until the transaction succeeds (stream becomes ready). + */ + def driveStream[T <: Data](that: Stream[T], address: BigInt, bitOffset: Int = 0): Unit = { + val wordCount = (bitOffset + widthOf(that.payload) - 1) / busDataWidth + 1 + onWritePrimitive(SizeMapping(address, wordCount * wordAddressInc), haltSensitive = false, null) { + when(!that.ready) { + writeHalt() + } + } + val flow = Flow(that.payloadType()) + driveFlow(flow, address, bitOffset, checkByteEnable = true) + that << flow.toStream Here the flow.toStream might not wait the ready signal. This code demostrate that. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

ohhh, what if the bus have no backpressure at all? I think where are some there.

KireinaHoro:
> ohhh, what if the bus have no backpressure at all? I think where are some there.

Yeah then this obviously won't work (same for the `readStreamBlockCycles`).  I guess we need some detection of the bus (for example `readHalt` should throw an exception [instead of silently ignoring](https://github.com/SpinalHDL/SpinalHDL/blob/b5a3010ffa91eb33706e426749815d61d881d10f/lib/src/main/scala/spinal/lib/bus/bram/BRAMSlaveFactory.scala#L48)) to prevent these functions from being used on such a `busSlaveFactory`?

Readon:
> > ohhh, what if the bus have no backpressure at all? I think where are some there.
> 
> Yeah then this obviously won't work (same for the `readStreamBlockCycles`). I guess we need some detection of the bus (for example `readHalt` should throw an exception [instead of silently ignoring](https://github.com/SpinalHDL/SpinalHDL/blob/b5a3010ffa91eb33706e426749815d61d881d10f/lib/src/main/scala/spinal/lib/bus/bram/BRAMSlaveFactory.scala#L48)) to prevent these functions from being used on such a `busSlaveFactory`?

That might be good.

PR #1292, opened 2024-01-22
allow CounterFreeRun to take bitCount

KireinaHoro:
Self-explanatory small change to make it easier to use.


Dolu1990:
Thanks :)

PR #1293, opened 2024-01-23
add setOutputAsReg

KireinaHoro:
Similar to `Data.setAsReg`, but aware of the port direction that it only sets output ports as register.  This allows a better shorthand like the following:

```scala
val io = new Bundle {
  val masterStream = master(Stream(Bits(4 bits))) setOutputAsReg()
}
```

Using `setAsReg` here would lead to the complaint of `REGISTER DEFINED AS COMPONENT INPUT`.

@Dolu1990 before merging, could you check if `MultiData` is the right place to put this function?  Alternatively, should this be the default behaviour (if direction present, set as register only when is output) for `setAsReg`?

### DIFF
--- a/core/src/main/scala/spinal/core/Data.scala
+++ b/core/src/main/scala/spinal/core/Data.scala
@@ -347,6 +347,11 @@ trait Data extends ContextUser with NameableByComponent with Assignable with Spi
 
   /** Set baseType to reg */
   def setAsReg(): this.type
+  /** Recursively set baseType to reg only for output */
+  def setOutputAsReg(): this.type = {
+    flatten.filter(_.dir == out).foreach(_.setAsReg())
+    this
+  }
   /** Set baseType to Combinatorial */
   def setAsComb(): this.type
 


lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala
```diff
@@ -36,7 +36,7 @@ object Axi4Stream {
     val keep = (config.useKeep) generate Bits(config.dataWidth bit)
     val last = (config.useLast) generate Bool()
     val dest = (config.useDest) generate UInt(config.destWidth bit)
-    val user = (config.useUser) generate Bits(config.userWidth*config.dataWidth bit)
+    val user = (config.useUser) generate Bits(config.userWidth bit)
```
Dolu1990, line 39: Ahh i guess this is an accident ^^

core/src/main/scala/spinal/core/HardMap.scala
```diff
@@ -0,0 +1,63 @@
+package spinal.core
+
+import spinal.idslplugin.Location
+
+import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
+
+class HardMap extends MultiData {
+  val storage = mutable.LinkedHashMap[NamedType[Data], Data]()
```
Dolu1990, line 9: should be in another PR

KireinaHoro: Ah yes, it's an accident that I pushed some other stuff... will do a repush

core/src/main/scala/spinal/core/MultiData.scala
```diff
@@ -103,6 +103,11 @@ abstract class MultiData extends Data {
     this
   }
 
+  def setOutputAsReg(): this.type = {
```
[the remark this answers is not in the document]
KireinaHoro: Implemented this idea in Data now

KireinaHoro:
@Dolu1990 reimplemented `setOutputAsReg` in Data.  So you think it's a good idea to keep this separate?  (I don't know how to implement it in `Data` and merge it with `setAsReg`)

Dolu1990:
> So you think it's a good idea to keep this separate

Yes :)

All good, thanks !

PR #1297, opened 2024-01-29
Composable exp

Dolu1990:
Sync VexiiRiscv related improvments


Readon:
It seems there are a lot of changes there.

Dolu1990:
Yes, quite a fire little fix there and there of a few line each.
The biggest changes are : 
- Verilator waves being saved into a folder which is named with the test name itself.
- making the vivado flow generating a proper vivado project instead of TCL only stuff.

If you are unconfortable with any of those change, i can isolate it into its dedicated PR.

Readon:
> Yes, quite a fire little fix there and there of a few line each. The biggest changes are :
> 
> * Verilator waves being saved into a folder which is named with the test name itself.
> * making the vivado flow generating a proper vivado project instead of TCL only stuff.
> 
> If you are unconfortable with any of those change, i can isolate it into its dedicated PR.

It's OK. It would be perfect if the doc can get updated.

Dolu1990:
Added the dac in https://github.com/SpinalHDL/SpinalDoc-RTD/commit/01659ba9afa3989a7c780ef55edbc24927d28299

PR #1300, opened 2024-02-04
Fix RegIf SpinalEnum

### LINKED ISSUE #1286: RegIf doesn't support fields of SpinalEnum type
When trying to create a field of for a SpinalEnum type, the creation fails at [RegInst.scala#142](https://github.com/SpinalHDL/SpinalHDL/blob/eb5a3a705ab25bfc6e236a51352f02934c22572b/lib/src/main/scala/spinal/lib/bus/regif/RegInst.scala#L142)

I'm not familiar enough (yet) with Spinal's internals to estimate the amount of work to implement this, but it would make my code much more beautiful if SpinalEnums are supported :-)

yportne13:
Fix #1286
add a match case for SpinalEnum when init the reg


jijingg:
good, @yportne13 , could you update [SpinalDOC](https://github.com/SpinalHDL/SpinalDoc-RTD)，Especially the mapping relationship between resetvalue and enum values

PR #1301, opened 2024-02-06
spinal.lib.Stream fix isFree

Angelic47:

Based on the comments, the `isFree` property within `spinal.lib.Stream` is intended for use in scenarios where:  
 - The `valid` signal is low and the `ready` signal is high, indicating that the receiver is ready but the sender has not transmitted data.  
Under these conditions, `isFree` should be high.  

If my understanding is correct, there appears to be a logical error with `isFree`, 
where an accidental use of `||` (logical OR) instead of `&&` (logical AND) has led to its improper functioning.  

This pull request addresses and corrects this issue.  

# Impact on code generation

`!io.valid || io.ready` => `!io.valid && io.ready`

# Checklist

- Unit tests were added -> Not applicable
- API changes are or will be documented -> Not applicable

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -289,7 +289,7 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
 
 /** Return True when the bus is ready, but no data is present
   */
-  def isFree: Bool = signalCache(this ->"isFree")((!valid || ready).setCompositeName(this, "isFree", true))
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

I would say, the comment was wrong, and the name of the function may not have been the best.

So overall, the idea of this function is to return True if the stream isn't stuck.
So, isFree meaning "Nothing is stuck here"

I think we should not change the functionality of the function. (backward compatibility / usecase)

Angelic47:
Alright, since this isn't an logical error, we can just close this pull request. However, perhaps we need to modify the comments, as they indeed are quite misleading.

Angelic47:
:D Good job👍

PR #1308, opened 2024-02-08
StreamMux with stream select use join

### LINKED ISSUE #1307: StreamMux with streamed select no correct join
I have a problem with using Stream Mux with the `createSelector` helper when using this factoy
```scala
  def apply[T <: Data](select: Stream[UInt], inputs: Vec[Stream[T]]): Stream[T] = {
```
this is using this to generate the selector:
```scala
def createSelector(): Stream[UInt] = new Composite(this, "selector") {
      val stream = Stream(cloneOf(select))
      val reg = stream.haltWhen(input.isStall).toReg(U(0))
      select := reg
    }.stream
```
This construction breaks when the inputs stream gets valid before the select stream.
This does not get correctly tested for in mux_with_selector formal test, as there is no assert validating the select stream with the output. (Only the value after createSelector register)
I am currently using a workaround where I join the select stream with the output of the mux to fix the handshaking.
But I do not know if this is the cleanest solution, also this breaks the createSelector API. (But I am unsure if this is fixable without breaking that):
```scala
def apply[T <: Data](select: Stream[UInt], inputs: Seq[Stream[T]]): Stream[T] = {
    val c = new StreamMux(inputs(0).payload, inputs.length)
    (c.io.inputs, inputs).zipped.foreach(_ << _)
    c.io.select := select.payload
    StreamJoin(c.io.output, select).map(_._1)
  }
```
I will provide draft PR with this included, but unsure how to best procede...

g0t00:
Closes #1307 

Proposed fix for #1307.
Not sure how to handle createSelector. AFAIK it can not be fixed as the input stream has to be halted when there is no valid selection. Should we just remove?

# Impact on code generation

# Checklist

- [X] Unit tests were modified

The part of the API currently described on the RTD is not changed.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -867,7 +867,7 @@ object StreamMux {
     c.io.output
   }
 
-  /** syncSel joins the selection stream with the the output stream. <
+  /** syncSel joins the selection stream with the the output stream.
     * Making sure the selection stream is fully synchronized with the selected stream
     */
   def syncSel[T <: Data](select: Stream[UInt], inputs: Seq[Stream[T]]): Stream[T] = {
@@ -878,7 +878,7 @@ object StreamMux {
   }
 
   /** regSel select uses haltWhen on the selection stream, thus making sure it is only consumed when data is selected.
-    * Caution: the other direction is not synchronized. (input valid without selection stream valid). See StreamMux.syncSel for a fully synchronied version.
+    * Caution: the other direction is not synchronized. (input valid without selection stream valid). See StreamMux.syncSel for a fully synchronized version.
     */
   def regSel[T <: Data](select: Stream[UInt], inputs: Seq[Stream[T]]): Stream[T] = {
     val c = new StreamMux(inputs(0).payload, inputs.length)
@@ -918,16 +918,25 @@ object StreamDemux {
     c.io.outputs
   }
 
+  /** regSel select uses haltWhen on the selection stream, thus making sure it is only consumed when data is selected.
+    * Caution: the other direction is not synchronized. (input valid without selection stream valid). See StreamDemux.syncSel for a fully synchronized version.
+    */
   def regSel[T <: Data](input: Stream[T], select: Stream[UInt], portCount: Int): Vec[Stream[T]] = {
     val c = new StreamDemux(input.payload, portCount)
     c.io.input << input
     select >> c.io.createSelector()
     c.io.outputs
   }
+
+  /** syncSel joins the selection stream with the the input stream.
+    * Making sure the selection stream is synchronized with the input stream
+    * If the select stream payload is out of range for the port count it is ignored.
+    */
   def syncSel[T <: Data](input: Stream[T], select: Stream[UInt], portCount: Int): Vec[Stream[T]] = {
     val c = new StreamDemux(input.payload, portCount)
-    c.io.input << StreamJoin(input, select).map(_._1)
-    select >> c.io.createSelector()
+    val joined = StreamJoin(input, select).takeWhen(select.payload < portCount)
+    c.io.input << joined.map(_._1)
+    c.io.select := joined._2
     c.io.outputs
   }
 
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -248,7 +248,7 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
     this.ready := ret.ready && counter.willOverflowIfInc
     (ret, counter)
   }
-  
+
   /**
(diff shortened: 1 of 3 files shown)


Dolu1990:
Hi ^^

Thanks for the PR :D

> Not sure how to handle createSelector. AFAIK it can not be fixed as the input stream has to be halted when there is no valid selection. Should we just remove?

So, i agree i'm realy not sure about createSelector, if it was designed like it on purpose, or by mistake.
Overall, i agree with you, we need to join the two stream.

I would say, do not remove it, because some may use it with that functionality on purpose.
The most we could do i think would be to rename createSelector into something like : createStreamRegSelect (something very explicity about what it does).

And on the side, we can add a function which does things in the way "we would expect". named like : createStreamSelect

Sound good ?

Dolu1990:
@Readon

g0t00:
> Hi ^^
> 
> Thanks for the PR :D
> 
> > Not sure how to handle createSelector. AFAIK it can not be fixed as the input stream has to be halted when there is no valid selection. Should we just remove?
> 
> So, i agree i'm realy not sure about createSelector, if it was designed like it on purpose, or by mistake. Overall, i agree with you, we need to join the two stream.
> 
> I would say, do not remove it, because some may use it with that functionality on purpose. The most we could do i think would be to rename createSelector into something like : createStreamRegSelect (something very explicity about what it does).
> 
> And on the side, we can add a function which does things in the way "we would expect". named like : createStreamSelect
> 
> Sound good ?

Make sense to me. So I renamed the original function. And made it possible to still use this in companion object.
Also the companion object defaults now to the "correct way". I am not sure If you can create a `createStreamSelect` In the class like the already existing create function, as it has to manipulate the streams...
But I think it is way more clean to use the Wrapper from the companion object anyway.

Dolu1990:
Hi,

Thanks for the patch ^^
Seems ok to me, the PR was in a "draft" state, do you feel it's ok to merge ?

Readon:
please only commit changes of logic, but not the auto format ones.

g0t00:
> please only commit changes of logic, but not the auto format ones.

Okay i can roll back the Scala format. But Is there a reason why the files are not formatted? Seems weird to me...

Dolu1990:
> Okay i can roll back the Scala format. But Is there a reason why the files are not formatted? Seems weird to me...

Autoformat can't comprehend the beauty of manual alignment <3
More generaly, i would say, one rule doesn't fit them all. Mostly because SpinalHDL is a internal DSL (lot of tricks) and hardware description has its own pattern

Readon:
> Okay, thanks for the explanation. I added syncSelc und regSel for StreamDemux. Should be ready for review now.

Good job. The only thing I concern about is the function name which seems like a encoded one. Would you please use names as @Dolu1990 described? 

> The most we could do i think would be to rename createSelector into something like : createStreamRegSelect (something very explicity about what it does).
like this?

g0t00:
> > Okay, thanks for the explanation. I added syncSelc und regSel for StreamDemux. Should be ready for review now.
> 
> Good job. The only thing I concern about is the function name which seems like a encoded one. Would you please use names as @Dolu1990 described?
> 
> > The most we could do i think would be to rename createSelector into something like : createStreamRegSelect (something very explicity about what it does).
> > like this?

Ah yeah for some reason I reverted the change while rolling back auto-format

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -866,10 +866,23 @@ object StreamMux {
     c.io.output
   }
 
-  def apply[T <: Data](select: Stream[UInt], inputs: Vec[Stream[T]]): Stream[T] = {
+  /** syncSel joins the selection stream with the the output stream.
+    * Making sure the selection stream is fully synchronized with the selected stream
+    */
+  def syncSel[T <: Data](select: Stream[UInt], inputs: Seq[Stream[T]]): Stream[T] = {
```
Readon, line 872: I mean this function name and the regSel also.

g0t00: Now I am confused. Those are the names proposed by @Dolu1990 in the issue (#1307)
Can you propose alternative names, which make it more clear?

Dolu1990: @g0t00 Yes right.
But now i'm seeing it, i'm thinking maybe the best would be to have it named : 
joinSel

So, we would have joinSel and regSel

I think the issue was that syncSel was too abstract, sync can mean many things.
But joinSel is very explicit :)
Sorry for the change.

g0t00: Seems good to me actually prefer that.

Dolu1990:
Seems good to me.
@Readon All good for you ? Or i missed something ?

Readon:
> Seems good to me. @Readon All good for you ? Or i missed something ?

Ok, it's good to me.
Thanks.

Readon:
If this is good enough, may be documented them is a good idea.

Issue #201, opened 2019-02-14
Wishbone intercon will generate spurius logic


wifasoi:
It wold be cool if the wishbone Intercon will generate only a Decoder when you have one master and multiple slave or only a Arbiter when you have multiple master and one slave.

Now it generate a decoder and arbiter when is not necessary (one decoder for each master and one arbiter for each slave)

wifasoi:
After some though i think this is a non issue, for this case is simpler to use a wishbone decoder/arbiter than instantiate an intercon. So I'll close this issue and add some more docstring to clarify this case

Issue #1310, opened 2024-02-13
WishboneSlaveFactory - doWrite is never called in pipelined mode


janschiefer:
It may be totally possible that I am completely and utterly wrong...

but I have been playing with WishboneSlaveFactory and and noticed some odd behaviour:

- When driving WishboneSlaveFactory from a master in "classic" mode everything works fine.

- However, when I try a "single write operation" in Wishbone B4 **pipelined mode** with a cycle according to specification, the data is never written to the register. In fact the internal signal "askWrite" is fired, but the signal "doWrite" is never activated.

![image](https://github.com/SpinalHDL/SpinalHDL/assets/29553827/7e1591fe-bc44-40da-91a4-47e417923672)

Wishbone specification B4:
![image](https://github.com/SpinalHDL/SpinalHDL/assets/29553827/0245f49a-18f1-473f-a2f0-eb1fc5ef0894)

If I extend the STB-Pulse one more cycle - which is not according to the specification - "doWrite" is called and the register is successfully written:

![image](https://github.com/SpinalHDL/SpinalHDL/assets/29553827/a2d78835-1035-4ce7-930b-d74f81de8027)

The problem seems to be, that SpinalHDL Lib's wishbone implementation also expects STB to be high i**n the cycle when ACK is received**. However, the specification requires STB only to be high in the cycle before ACK when in pipelined mode:

![image](https://github.com/SpinalHDL/SpinalHDL/assets/29553827/f060c87d-ec2b-4243-bed5-e54542ba6fdd)

My WB Slave test code:
```
package tangBusTest

import spinal.core._
import spinal.lib._
import spinal.lib.bus.wishbone._

// Hardware definition
case class WishboneLed(wb_config: WishboneConfig) extends Component {
  val io = new Bundle {
    val leds_out = out Bits (8 bits)
    val wb_bus = slave(Wishbone(wb_config))
  }

  val slave_factory = WishboneSlaveFactory(io.wb_bus)

  if (wb_config.useERR) io.wb_bus.ERR := False

  val leds_reg = slave_factory.createReadWrite(Bits(8 bits), address = 0x08, bitOffset = 0)

  val test_read_reg = slave_factory.read(U(value = 112, 8 bits), address = 0x16, bitOffset = 0)

  io.leds_out := leds_reg

}

object WishboneLed {}
```

I hope, I have found a valid issue and not wasted your time.

janschiefer:
Versions:
[...]

PR #1312, opened 2024-02-15
Wishbone bus bugfixes

### LINKED ISSUE #1311: Wishbone BusIf SEL usage
I think there is a bug [here](https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/WishboneBusInterface.scala#L37) but it's possible I'm reading the wishbone spec wrong.

[From here](https://cdn.opencores.org/downloads/wbspec_b3.pdf#page=66) it seems like the SEL line is a byte select / partial word mechanism not a chip select like it looks like in the code. 

It doesn't look like BusIf really supports partial word writes, so perhaps the easiest thing is to just remove the use of SEL in that class altogether. I can make this change but I wanted to make sure I wasn't misunderstanding anything before doing so.

jdavidberger:
Closes #1311 
Addresses #201 too

There were a few issues in WishBone related library code that are fixed by this PR. I can split this into multiple PRs if that makes it easier to go through.

## [Remove selMatch logic; buffer out the ACK signal to match the spec](https://github.com/SpinalHDL/SpinalHDL/commit/0da1c222342538e4bf2d747f961134191d84f14c)

Fixes #1311. Also only asserts ACK after a valid cycle.

## [Make wb arbiter a pass through when there is only one input](https://github.com/SpinalHDL/SpinalHDL/commit/17eda435eac98885fa6ba6f6726e54397ea81d4a)

Makes an instantiated arbiter behave as a passthrough when it is 1 to 1

I did this to make it easier to read the generated code and it seemed worth including.

## [Add config option to Wishbone bus to specify if it's byte addressable…](https://github.com/SpinalHDL/SpinalHDL/commit/19e8051c6a23bded12c5c893ca74530b7b5aa75e) 

There is an issue in the current code base where `WishboneSlaveFactory` assumed the bus was word addressable, but the `BusIf` subsystem assumed it was byte addressable. This makes it a configuration option on the bus config, defaulting to word addressable. I have no strong opinions on whether or not the default should be byte or word addressable, but AFAICT it's not mandated either way by the spec and I've seen both used. 

## [Update wishbone interconnect to use AddressMapping instead of just Si…](https://github.com/SpinalHDL/SpinalHDL/commit/32958ae7d32c4d533268ab1101a64fa3c49c434f) 

This lets you use DefaultMapping and friends with the wishbone interconnect. 

## [Only insert an arbiter in a wishbone interconnect when necessary](https://github.com/SpinalHDL/SpinalHDL/commit/7fefcec799f22a925e62ae216b3245bb67eb7a36)

Related to [Make wb arbiter a pass through when there is only one input](https://github.com/SpinalHDL/SpinalHDL/commit/17eda435eac98885fa6ba6f6726e54397ea81d4a), but this removes the arbiter component altogether. This generates simpler code but should have no behavioral impact. 

Note that I did not do this for the decoders because that could change behavior.

## [Bring WishboneDecoder more in line with other decoders. This commit c…](https://github.com/SpinalHDL/SpinalHDL/commit/b442c0441be11fea140f4145a029c084f3146dbe)

`MuxOH` has an oddity where the index which wins in the event of a tie changes from the end of the vector for n > 2 to the beginning of the vector when n == 2. This meant that if you had a DefaultMapping (or equivalent SizeMapping) entry in the decoder, it needed to be at the end of list with two slaves and at the beginning of the list for N > 2. This was confusing behavior. The AXI decoder doesn't have this issue because it doesn't use `MuxOH`; and instead does something more like this commit. 

# Impact on code generation

There is no impact on code generation directly; but these should make the generated code somewhat easier to read since it indirectly simplifies some of the arbiter constructs. 

# Checklist

- [x] Unit tests were added: 
- [X] API changes are or will be documented: The only user facing API change has a docstring for it

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
+++ b/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
@@ -5,20 +5,20 @@ import spinal.lib._
 
 object AddressGranularity extends Enumeration {
   type AddressGranularity = Value
-  @deprecated("This option will be REMOVED in future versions, and the default will be Word granular. It exists " +
+  @scala.deprecated("This option will be REMOVED in future versions, and the default will be Word granular. It exists " +
     "to prevent changing library behavior for existing code.")
-  val Unspecified = Value
+  val UNSPECIFIED = Value
 
   /**
    * Word granular addressing increments by one per word. This is the standard defined behavior of the bus.
    */
-  val Word = Value
+  val WORD = Value
 
   /**
    * Byte granular addressing increments by one per byte. This is not in adherence with the wishbone standard but is
    * a commonly encountered configuration with external IP.
    */
-  val Byte = Value
+  val BYTE = Value
 }
 
 
@@ -54,7 +54,7 @@ case class WishboneConfig(
   val tgcWidth : Int = 0,
   val tgdWidth : Int = 0,
   val useBTE : Boolean = false,
-  private val addressGranularity : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified
+  private val addressGranularity : AddressGranularity.AddressGranularity = AddressGranularity.UNSPECIFIED
 ){
   def useTGA = tgaWidth > 0
   def useTGC = tgcWidth > 0
@@ -65,11 +65,11 @@ case class WishboneConfig(
 
   def pipelined : WishboneConfig = this.copy(useSTALL = true)
 
-  def wordAddressInc(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified): Int = {
-    val effectiveAddressGranularity = if (addressGranularity == AddressGranularity.Unspecified) addressGranularityIfUnspecified else addressGranularity
-    require(effectiveAddressGranularity != AddressGranularity.Unspecified, "If the bus has not been configured with an address granularity, one must be provided to wordAddressInc")
+  def wordAddressInc(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity = AddressGranularity.UNSPECIFIED): Int = {
+    val effectiveAddressGranularity = if (addressGranularity == AddressGranularity.UNSPECIFIED) addressGranularityIfUnspecified else addressGranularity
+    require(effectiveAddressGranularity != AddressGranularity.UNSPECIFIED, "If the bus has not been configured with an address granularity, one must be provided to wordAddressInc")
     effectiveAddressGranularity match {
-      case AddressGranularity.Byte => dataWidth / 8
+      case AddressGranularity.BYTE => dataWidth / 8
       case _ => 1
     }
   }
@@ -268,7 +268,7 @@ case class Wishbone(config: WishboneConfig) extends Bundle with IMasterSlave {
   def doWrite : Bool    = doSend &&  WE
   def doRead  : Bool    = doSend && !WE
 
-  def byteAddress(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified) : UInt = {
+  def byteAddress(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity = AddressGranularity.UNSPECIFIED) : UInt = {
     ADR << log2Up((config.dataWidth / 8) / config.wordAddressInc(addressGranularityIfUnspecified))
   }
 }
--- a/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
+++ b/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
@@ -52,8 +52,9 @@ case class WishboneConfig(
 
   def pipelined : WishboneConfig = this.copy(useSTALL = true)
 
-  def wordAddressInc(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity): Int = {
+  def wordAddressInc(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified): Int = {
     val effectiveAddressGranularity = if (addressGranularity == AddressGranularity.Unspecified) addressGranularityIfUnspecified else addressGranularity
+    require(effectiveAddressGranularity != AddressGranularity.Unspecified, "If the bus has not been configured with an address granularity, one must be provided to wordAddressInc")
     effectiveAddressGranularity match {
       case AddressGranularity.Byte => dataWidth / 8
       case _ => 1
--- a/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
+++ b/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
@@ -5,7 +5,20 @@ import spinal.lib._
 
 object AddressGranularity extends Enumeration {
   type AddressGranularity = Value
-  val Unspecified, Byte, Word = Value
+  @deprecated("This option will be REMOVED in future versions, and the default will be Word granular. It exists " +
+    "to prevent changing library behavior for existing code.")
+  val Unspecified = Value
+
+  /**
+   * Word granular addressing increments by one per word. This is the standard defined behavior of the bus.
+   */
+  val Word = Value
+
+  /**
+   * Byte granular addressing increments by one per byte. This is not in adherence with the wishbone standard but is
+   * a commonly encountered configuration with external IP.
+   */
+  val Byte = Value
 }
 
 
@@ -41,7 +54,7 @@ case class WishboneConfig(
   val tgcWidth : Int = 0,
   val tgdWidth : Int = 0,
   val useBTE : Boolean = false,
-  val addressGranularity : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified
+  private val addressGranularity : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified
 ){
   def useTGA = tgaWidth > 0
   def useTGC = tgcWidth > 0
--- a/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
+++ b/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
@@ -3,6 +3,12 @@ package spinal.lib.bus.wishbone
 import spinal.core._
 import spinal.lib._
 
+object AddressGranularity extends Enumeration {
+  type AddressGranularity = Value
+  val Unspecified, Byte, Word = Value
+}
+
+
 /** This class is used for configuring the Wishbone class
   * @param addressWidth size in bits of the address line
   * @param dataWidth size in bits of the data line
@@ -16,6 +22,7 @@ import spinal.lib._
   * @param tgcWidth size in bits of the tag cycle line, deafult to 0 (disabled)
   * @param tgdWidth size in bits of the tag data line, deafult to 0 (disabled)
   * @param useBTE activate the Burst Type Extension, default to false (disabled)
+  * @param addressGranularity This specifies the address granularity for the bus.
   * @example {{{
   * val wishboneBusConf = new WishboneConfig(32,8).withCycleTag(8).withDataTag(8)
   * val wishboneBus = new Wishbone(wishboneBusConf)
@@ -33,7 +40,8 @@ case class WishboneConfig(
   val tgaWidth : Int = 0,
   val tgcWidth : Int = 0,
   val tgdWidth : Int = 0,
-  val useBTE : Boolean = false
+  val useBTE : Boolean = false,
+  val addressGranularity : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified
 ){
   def useTGA = tgaWidth > 0
   def useTGC = tgcWidth > 0
@@ -43,6 +51,14 @@ case class WishboneConfig(
   def isPipelined = useSTALL
 
   def pipelined : WishboneConfig = this.copy(useSTALL = true)
+
+  def wordAddressInc(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity): Int = {
+    val effectiveAddressGranularity = if (addressGranularity == AddressGranularity.Unspecified) addressGranularityIfUnspecified else addressGranularity
+    effectiveAddressGranularity match {
+      case AddressGranularity.Byte => dataWidth / 8
+      case _ => 1
+    }
+  }
 
   def withDataTag(size : Int)    : WishboneConfig = this.copy(tgdWidth = size)
   def withAddressTag(size : Int) : WishboneConfig = this.copy(tgaWidth = size)
--- a/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
+++ b/lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
@@ -255,6 +255,9 @@ case class Wishbone(config: WishboneConfig) extends Bundle with IMasterSlave {
   def doWrite : Bool    = doSend &&  WE
   def doRead  : Bool    = doSend && !WE
 
+  def byteAddress(addressGranularityIfUnspecified : AddressGranularity.AddressGranularity = AddressGranularity.Unspecified) : UInt = {
+    ADR << log2Up((config.dataWidth / 8) / config.wordAddressInc(addressGranularityIfUnspecified))
+  }
 }
 
 object Wishbone{
--- a/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneBusInterfaceTester.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneBusInterfaceTester.scala
@@ -46,8 +46,16 @@ class SpinalSimWishboneBusInterfaceTester extends SpinalAnyFunSuite{
       dut.clockDomain.waitSampling()
 
       val sco = ScoreboardInOrder[WishboneTransaction]()
-      val monitor = WishboneMonitor(dut.io.bus, dut.clockDomain){ bus =>
-        sco.pushRef(WishboneTransaction.sampleAsMaster(bus))
+      for(ref_output <- List(
+        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 1),
+        WishboneTransaction(0, 0), WishboneTransaction(4, 0),
+        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 2)
+      )) {
+        sco.pushRef(ref_output)
+      }
+
+      WishboneMonitor(dut.io.bus, dut.clockDomain){ bus =>
+        sco.pushDut(WishboneTransaction.sampleAsMaster(bus))
       }
 
       val dri = new WishboneDriver(dut.io.bus, dut.clockDomain)
@@ -55,13 +63,7 @@ class SpinalSimWishboneBusInterfaceTester extends SpinalAnyFunSuite{
       dri.drive(scala.collection.immutable.Seq(WishboneTransaction(0, 100), WishboneTransaction(4, 2)), we = true)
       dri.drive(scala.collection.immutable.Seq(WishboneTransaction(0), WishboneTransaction(4)), we = false)
 
-      val expectedOutput = List(
-        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 1),
-        WishboneTransaction(0, 0), WishboneTransaction(4, 0),
-        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 2)
-      )
-      print(sco.ref)
-      assert(sco.ref.toList == expectedOutput)
+      sco.checkEmptyness()
     }
   }
 
--- a/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneBusInterfaceTester.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneBusInterfaceTester.scala
@@ -45,7 +45,7 @@ class SpinalSimWishboneBusInterfaceTester extends SpinalAnyFunSuite{
 
       dut.clockDomain.waitSampling()
 
-      val wordInc = conf.wordAddressInc(AddressGranularity.Byte)
+      val wordInc = conf.wordAddressInc(AddressGranularity.BYTE)
 
       val sco = ScoreboardInOrder[WishboneTransaction]()
       for(ref_output <- List(
@@ -76,12 +76,12 @@ class SpinalSimWishboneBusInterfaceTester extends SpinalAnyFunSuite{
   }
 
   test("sync_byte"){
-    val conf = WishboneConfig(32,32, addressGranularity = AddressGranularity.Byte)
+    val conf = WishboneConfig(32,32, addressGranularity = AddressGranularity.BYTE)
     testBus(conf, description="sync_byte")
   }
 
   test("sync_word"){
-    val conf = WishboneConfig(32,32, addressGranularity = AddressGranularity.Word)
+    val conf = WishboneConfig(32,32, addressGranularity = AddressGranularity.WORD)
     testBus(conf, description="sync_word")
   }
 
--- a/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneBusInterfaceTester.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneBusInterfaceTester.scala
@@ -45,11 +45,13 @@ class SpinalSimWishboneBusInterfaceTester extends SpinalAnyFunSuite{
 
       dut.clockDomain.waitSampling()
 
+      val wordInc = conf.wordAddressInc(AddressGranularity.Byte)
+
       val sco = ScoreboardInOrder[WishboneTransaction]()
       for(ref_output <- List(
-        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 1),
-        WishboneTransaction(0, 0), WishboneTransaction(4, 0),
-        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 2)
+        WishboneTransaction(0, 0x12345678L), WishboneTransaction(1 * wordInc, 1),
+        WishboneTransaction(0, 0), WishboneTransaction(1 * wordInc, 0),
+        WishboneTransaction(0, 0x12345678L), WishboneTransaction(1 * wordInc, 2)
       )) {
         sco.pushRef(ref_output)
       }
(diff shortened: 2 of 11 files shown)


andreasWallner:
First off - thanks for the PR!

Most of the stuff looks good to me, the thing I'm not convinced by is the byteAddressable change. 

First off I'm not sure this should be a bus property itself. If we want to have this then the BusSlaveFactory looks like a better place to me, also so that behavior between different buses (AHB, APB, ...) stays consistent.

Second it's a difference in the APIs that BusSlaveFactory and RegIf provide: The current contract of the BusSlaveFactory is that the address that you provide as a parameter needs to be on the bus to address the register, you as the designer have to make that choice explicitly. The BusIf API has a different contract. Also from this POV I'd rather put that as a parameter in the BusSlaveFactory.

If we want to keep it this way then I'd have some minor comments:
- wording wise it might be misleading since we don't make bytes addressable, but we change what the "unit" of the address is
- isn't the default in your current code byte-addressable, not word (parameter description)?

@Dolu1990 what's you opinion on this?

andreasWallner:
FYI: In case you've not see it yet there is #1310, just as an info when you are using Wishbone pipelined - having a look at that now.

jdavidberger:
> First off I'm not sure this should be a bus property itself. If we want to have this then the BusSlaveFactory looks like a better place to me, also so that behavior between different buses (AHB, APB, ...) stays consistent.

I actually had it in there at first but a few things made me move it: It would need to be specified in `WishboneSlaveFactory` and `WishboneBusInterface` separately. It made my code more fragile since I had the `WishboneConfig` defined once in my code but had a half a dozen interfaces in my design around it. I eventually came around to think of it as a part of the wishbone configuration itself since you can't do much generically with the bus without knowing that setting.

Part of the thing I ran into was that I wanted to attach both RegIf and BusSlaveFactory in different components on the same bus interconnect and they were incompatible without remapping the address.  

> wording wise it might be misleading since we don't make bytes addressable, but we change what the "unit" of the address is

Agree, I will change the verbiage to be address granularity, that seems more accurate?

> isn't the default in your current code byte-addressable, not word (parameter description)?

This is probably a verbiage issue too; it's coded up as the number of increments going one word forward takes. So default of 1 is word/addressable granular.

andreasWallner:
> Part of the thing I ran into was that I wanted to attach both RegIf and BusSlaveFactory in different components on the same bus interconnect and they were incompatible without remapping the address.

This is currently entirely possible, just not by passing the same register parameters to RegIf and BusSlaveFactory (which are two mostly unrelated interfaces ATM)? It looks to me like the issue here is an assumption that both take the same parameters. 
While I understand that this can be misleading, we can't change the defaults here without breaking a bunch of existing code that uses them currently (we'd either change the behavior of RegIf or BusSlaveFactory).

I'd stick with not having this as a property on the bus, but add it as a parameter to the factory.

It would be interesting to have a more general solution, but then we should have a plan which impact this has also on interconnect generators, etc. The recent fabric negotiation code might be a place for this to go.

> This is probably a verbiage issue too; it's coded up as the number of increments going one word forward takes. So default of 1 is word/addressable granular.

Looking at it again yes... this being a divider seems kind of weird to me (but it makes sense since you wanted to change the behavior to match RegIf), I think it would be more logical to pass the increment.

andreasWallner:
Another point: your last commit removes to ability to have an early ACK which is explicitly allowed by 3.10 in the spec.
We should keep that possibility - if stalling of the bus is needed (or for timing closure) this should be a parameter - otherwise we have throughput compared to the current implementation?

jdavidberger:
> While I understand that this can be misleading, we can't change the defaults here without breaking a bunch of existing code that uses them currently (we'd either change the behavior of RegIf or BusSlaveFactory).

I agree that barring a major version change coming up soon breaking existing code is a red line.

So two options on this:
- Replace the poorly named addrInc in the WishboneConfig field with an enum that is something like:
```
enum AddressGranularity:
    case Unspecified, Byte, Word
```
with the default being unspecified. When unspecified it preserves existing behavior of RegIf / WishboneSlaveFactory. 
- Keep it as a config option on RegIf and WishboneSlaveFactory separately. Probably use the above enum definition though since it's very clear what it means.

With how I'm using wishbone in my designs, the first option would keep the code base cleaner since I have half a dozen of these things sprinkled around on the same bus, but I'll do make the second one happen if you think thats better.

 > Another point: your last commit removes the ability to have a comb ACK which is explicitly allowed by 3.10, 3.30 & 3.40 in the spec.

This is an issue, I'll fix it and ref the spec in the code there

andreasWallner:
There is one other thing to think about here that I overlooked before: both options present an issue for generic peripherals: when building bus independent peripherals (e.g. like here: https://github.com/andreasWallner/spinalStuff/blob/master/src/main/scala/andreasWallner/io/spi/SpiSlave.scala#L95) we pretty much depend on the generators behaving the same - otherwise they are not replaceable.

I'm not sure it's worth it to get slightly more consistent behaviour between two quite different APIs.

Quick sidenote: looking at the other wishbone bug I linked above I noticed that the current test case for the generator does not really verify anything, it looks like it wasn't finished back in 2018 - so we shouldn't currently depend on it/trust it.

jdavidberger:
> There is one other thing to think about here that I overlooked before: both options present an issue for generic peripherals: when building bus independent peripherals (e.g. like here: https://github.com/andreasWallner/spinalStuff/blob/master/src/main/scala/andreasWallner/io/spi/SpiSlave.scala#L95) we pretty much depend on the generators behaving the same - otherwise they are not replaceable.

This would work with option 1 pretty easily I think. Since the bus has a reference to it's config, the generated BusSlaveFactory doesn't need any special configuration options to make it work whether the users bus is word granular or byte granular. Or am I misunderstanding this?

jdavidberger:
I made the following changes:

- Moved typos and simple bug fix changes to the top of the commits
- Added testing logic for `SpinalSimWishboneSlaveFactoryTester`. This was to ensure I didn't change default behavior in following commits. I did not fix the Pipelined operations; I'm not super familiar with that operational mode and figured functional testing in classic mode was a net win. 
- Add unit tests for WishboneBusInterface
- Made the modification discussed above wrt address granularity field. 

I put that one at the end since it's got unresolved feedback. The main demonstration here is that the existing unit tests did not need modifications in order to pass so the changes shouldn't break existing code.

vr2045:
Hello, this is a show stopper bug for me too :( Wishbone gets taken over as the ACK is the wrong polarity. Please merge to baseline when possible so we can all get this much needed patch for Wishbone.

tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneBusInterfaceTester.scala
```diff
@@ -0,0 +1,78 @@
...
+      dri.drive(scala.collection.immutable.Seq(WishboneTransaction(0, 100), WishboneTransaction(4, 2)), we = true)
+      dri.drive(scala.collection.immutable.Seq(WishboneTransaction(0), WishboneTransaction(4)), we = false)
+
+      val expectedOutput = List(
+        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 1),
+        WishboneTransaction(0, 0), WishboneTransaction(4, 0),
+        WishboneTransaction(0, 0x12345678L), WishboneTransaction(4, 2)
+      )
+      print(sco.ref)
+      assert(sco.ref.toList == expectedOutput)
```
andreasWallner, line 64: No need to do the comparison like this, just push the expected transactions via `sco.pushDut(...)` and do a `sco.checkEmptyness()` at the end of the test.

tester/src/test/scala/spinal/tester/scalatest/SpinalSimWishboneSlaveFactoryTester.scala
```diff
@@ -33,25 +35,34 @@ class SpinalSimWishboneSlaveFactoryTester extends SpinalAnyFunSuite{
...
-      //   else
-      //     scoreboard.pushDut(WishboneTransaction.sampleAsMaster(bus))
-      // }
-
-      for(repeat <- 0 until 1000){
-        driver.drive(for(x <- 1 to 10) yield WishboneTransaction(address=10).randomizeData(dut.io.bus.config.dataWidth),true)
-        dut.clockDomain.waitSampling(10)
-        driver.drive(for(x <- 1 to 10) yield WishboneTransaction(address=10),false)
-        dut.clockDomain.waitSampling(10)
+      scoreboard.ref ++= List(
```
andreasWallner, line 41: same as above, please use `scoreboard.pushRef(...)` and add a `scoreboard.checkEmptiness()` at the end of the test

andreasWallner (changes requested):
Thanks for adding some tests here!

It's totally ok to comment out the pipeline tests for now - they anyhow didn't do anything and looking at the other bug pipeline mode is broken anyhow.

You have one comment in the code w.r.t. retry: not sure it's really needed the usual use-cases of the BusSlaveFactory are register interface that respond quickly and don't stall for long - but let's leave that for later.

Regarding the AddressGranularity:
- The current implementation will not directly break generic peripherals, but a peripheral originally written with that feature in mind can't be switched over to any other BusSlaveFactory.
- You introduce a requirement on the bus being passed to a component: if the component was written with one setting on the bus the caller always has to provide the bus with that setting. The code in the peripheral will depend on it since the addresses will be written as such (1,2,3 vs. 1,4,8 for a 32bit bus).
- Adding the three options here only increases possible confusion from my POV: you could now have a BusSlaveFactory that needs to be passed byte addresses, or possibly some addresses aligned to a granularity set elsewhere.

The only advantage that I see so far is that the constants that you use when using two distinct APIs are the same (if you even provide addresses with the BusIfAdapter).
I think we should stick with the the BusSlaveFactory using/expecting byte addresses consistently. (you can still build systems that e.g. drop to lower two bits of the address in any case)

jdavidberger:
> I think we should stick with the the BusSlaveFactory using/expecting byte addresses consistently. (you can still build systems that e.g. drop to lower two bits of the address in any case)

I think we are on the same page on this. BusSlaveFactory by contract uses byte granularity. 

I think the confusion is that I thought `wordAddressInc` was in reference to the raw ADR on the bus. Looking through the code more I realized it was applied to the internal addressing which is byte granular. So overriding that was the wrong thing to do.

Look at the latest commit; I think that does the right thing from what you are saying.  I added testing which makes sure this is mapped out right.

andreasWallner:
I'm sorry, but my points from last comment mostly remain:
- if we were to use the same code with an AXI-Lite or APB BusSlaveFactory the result would be different
- the 3 options still remain, modifying the behavior of the BusSlaveFactory

I'd drop `AddressGranularity` completely. In the current framework this should not be a property on the Bus.

After looking at it again I see now that the old Wishbone factory was not consistent with the other BusSlaveFactories. I'd be for making them consistent (and use the bus address instead of the currently used `byteAddress`) but I'd like to discuss that with @Dolu1990 since that would break some user code (but make other user code work as expected)...
(the `val byteAddress = bus.ADR << log2Up(bus.config.dataWidth/8)` used in the WishboneSlaveFactory` should IMO not be there and `bus.ADR` should be used instead)

jdavidberger:
Is the idea then that if you have a word granular wishbone bus you should just map it to byte granular before invoking any of the bus interface stuff? 

I'm not sure I really understand the point about AXI / APB having different results, but I can just drop all the granularity commits to get the PR through? This would reintroduce the issue in WishboneSlaveFactory but that could be solved separately

Jeff-Ciesielski:
Howdy! Just wanted to toss in my 0.02 since we use spinal's wishbone infra quite a lot at LeafLabs.  Great conversation in this thread so far,  it's pretty rough to have two similar subsystems in the spinal ecosystem that behave differently and ultimately make mixing and matching impossible :/. 

My coworker @RiceShelley just hit this same issue re: RegIF vs WishboneSlaveFactory today and had a long talk about it before realizing there was a fix underway. Big thanks @jdavidberger for tackling this.   

> Is the idea then that if you have a word granular wishbone bus you should just map it to byte granular before invoking any of the bus interface stuff?

This was Rice's workaround:
```
  val busIf =
    WishboneBusInterface(
      bus = WishboneToByteAddr(io.wb),
      sizeMap = SizeMapping(0x0000, 100 Byte)
    )
```

It works, but IMO is a little odd / error prone / un...magical? It puts the onus on the device to shift the address but really, this only works because we have control over both ends of the connection and know that the master has byte granularity.

IMO making it a parameter of the bus config (what was done here) makes a lot of sense.  Fundamentally, granularity IS a part of the wishbone config for a given port just as much as the data or address width, so while I agree that it creates a little bit of friction, at the very least it would allow spinal to check when attaching two busses that they're compatible and throw an error if they aren't at elaboration time (or, more ideally, make it happen automatically if possible).  

>You introduce a requirement on the bus being passed to a component: if the component was written with one setting on the bus the caller always has to provide the bus with that setting. The code in the peripheral will depend on it since the addresses will be written as such (1,2,3 vs. 1,4,8 for a 32bit bus).

This is already the case with SEL, no? Same goes for data width. In any event, the wishbone spec defines granularity at a port level (on both controllers and peripherals), and given that peripherals are intended to be reused/shared/etc it seems to me that the infrastructure should support (at a type/configuration level) the concept of granularity precisely to prevent the mistakes you're talking about.  The alternative is requiring the complete decoupling of the register addressing scheme and having different wrappers for your peripheral if it's reused with different controllers.

andreasWallner:
First off: Thanks for getting me to read the stuff again, I was somehow focusing on BusSlaveFactory

I've spent some time thinking, reading the spec and discussing with @Dolu1990 about it (some of this might be a recap):
- the thing that should be consistent across all implementations of `BusSlaveFactory` is the addresses are passed as byte addresses. that ensures that code can be generic w.r.t. the bus interface and is the one thing common across the different implementations
- we should not introduce a parameter on the bus config that changes the behavior of the `BusSlaveFactory`, that makes this even more complicated

what the spec says (please correct me if I'm wrong I don't deal with Wishbone often, both from Wishbone B4, 2.2.3)
- The spec seems to mandate that ADR_O is addressing a word of the width of the data port, no byte addresses there
- the granularity of access is then given by the width of the SEL signal (which is currently implemented/used incorrectly)

It also looks to me like the current `WishboneSlaveFactory` and `BusIf` are able to generate valid & interoperable peripherals (given the bugfixes here and upcoming fix to the pipeline stuff, albeit ignoring the SEL lines currently in the generators)

Also: From how I understand it, as long as ones keeps with the wishbone spec when assigning data width and SEL you should implictly get a check for compatibility (since the assignment checks whether the buses have the same address width, data width and sel width).

I see two remaining points:
- using the `WishboneSlaveFactory` one can currently create a register that is not addressable - we should add an assert on that
- The current wishbone `BusIf` generates something that is not spec compliant as far as I can tell (since it expects a byte address on the bus). I think we should fix that and make it spec compliant. 
(in the long term it would make sense to me to use `BusIf` as a wrapper around `BusSlaveFactory`, but that's for another day...)

How to fix this:
- I don't think that we should implement a parameter so that the `WishboneSlaveFactory` is able to generate the same invalid peripheral.
- Should we just bite the bullet and fix the wishbone `BusIf`, which is currently not spec compliant (if either of you can confirm our reading of the spec)

jdavidberger:
> what the spec says (please correct me if I'm wrong I don't deal with Wishbone often, both from Wishbone B4, 2.2.3)

> The spec seems to mandate that ADR_O is addressing a word of the width of the data port, no byte addresses there
the granularity of access is then given by the width of the SEL signal (which is currently implemented/used incorrectly)

I think this is wrong and I wanted to respond quickly to the point:

https://wishbone-interconnect.readthedocs.io/en/latest/02_interface.html#master-signals

> ADR_O()
> The address output array [ADR_O()] is used to pass a binary address. The higher array boundary is specific to the address width of the core, and the lower array boundary is determined by the data port size and granularity. For example the array size on a 32-bit data port with BYTE granularity is [ADR_O(n..2)]. In some cases (such as FIFO interfaces) the array may not be present on the interface.

[They also give a concrete example of a 16 bit word bus with 8 bit granularity ](https://cdn.opencores.org/downloads/wbspec_b4.pdf#page=105)

In terms of granularity as a configuration value, the spec also has in 2.15:

> The WISHBONE DATASHEET MUST indicate the port granularity. The granularity MUST be indicated as: 8-bit, 16-bit, 32-bit or 64-bit.

This is put in the same context as mandating what information must be available to be a wishbone bus, which really make me think it's reasonable to have it as part of the bus config

Jeff-Ciesielski:
(Wishbone gives me such a headache sometimes :laughing: )

> The spec seems to mandate that ADR_O is addressing a word of the width of the data port, no byte addresses there

After re-re-reading I think it *implies* this, but it leaves it a little bit ambiguous due to some somewhat unclear/contradictory examples. 

> [They also give a concrete example of a 16 bit word bus with 8 bit granularity](https://cdn.opencores.org/downloads/wbspec_b4.pdf#page=105)

They do, but, crucially...they leave out ADR handling :|

I think that Wishbone might actually be slightly underspecified in this regard.  (among others) The spec gives three examples pertaining to this:
- A 32 bit peripheral with byte granularity showing ADR_O clipped by two bits (DATA-word addressed) with a description of SEL
- A 16 bit peripheral with WORD Granularity showing no SEL (Granularity == DATA-word size)
- A 16 bit peripheral with BYTE granularity showing SEL[1..0] (Granularity < DATA-word size)

Examples 2/3 DO NOT mention the ADR lines in any way, only how they handle SEL.

As @jdavidberger mentioned, the first example is spelled out as:

> The higher array boundary is specific to the address width of the core, and the lower array boundary is determined by the data port size and granularity. For example the array size on a 32-bit data port with BYTE granularity is [ADR_O(n..2)]

I think there is a bit open to interpretation here, or at the very least, there is some condition implied by the existence of SEL that is left to the reader to deduce.

Assuming 32 bit ADR and 32 bit DATA :
- My read is that one could have a port with DWORD granularity, giving you 2^32 DATA-WORDS / 2^34 BYTES of addressable memory, meaning you'd use ADR_O(n..0) and no SEL
- OR you could have BYTE granularity with ADR_O(n..2), and you select your byte(s) with SEL[3..0]
- OR you could have WORD granularity with....?

That last case has NO example, and in fact there is never a call out to the actual direct relationship of SEL and ADR, however SEL's presence and width is defined by port granularity, so it stands to reason (at least to me) that if this were something like a DSP with 16 bit granularity on a 32 bit bus, there is no concept of a byte.  SEL would be present, so you'd use that to select between 16 bit values, and you'd send addresses that corresponded to 32 bit boundaries (so ADR[n..1]).

Or, in other words: "...ADR_O is addressing a word of the width of the data port..."

> The current wishbone BusIf generates something that is not spec compliant as far as I can tell ...
> ...
> Should we just bite the bullet and fix the wishbone BusIf, which is currently not spec compliant (if either of you can confirm our reading of the spec)

I fully agree that they should be consistent, and I think that it's reasonable for them to only support byte addressing as long as the higher level wishbone infrastructure (interconnect/etc) aren't tied to that.

I also think that, in an ideal world where code didn't exist out in the wild and this were getting built from scratch, defining wishbone in the library in terms of DATA_WIDTH, ADDRESSABLE_WORDS (or WORD_ADDRESS_WIDTH), and GRANULARITY explicitly would be better than the implicit check based on DATA_WIDTH and SEL_WIDTH that we currently have.   Given that we don't live in that sort of utopia I'm  walk back my support of the explicit granularity unless it was done in such a way as to wrap the generation of SEL/ADR widths (so an abstraction on top of instead of an inherent change to the config itself).

jdavidberger:
I agree the spec is sort of hard to decipher with granularity (and with a bunch of other stuff to be honest). The crux of it is for me is when they say ADR_O(n..2) do they mean the core has N ADR lines or N-2? [the sel stuff](https://cdn.opencores.org/downloads/wbspec_b4.pdf#page=62) being a bit flag for the bytes and _not_ a count of valid bytes still seems odd to me. 

It's also unclear to me whether or not SEL is a spec required signal if the granularity is less than the word size. It does seem clear if one exists it is logically sized based on the granularity and data width. I've been thinking of it as optional but maybe it's not. 

For what it's worth litex handles it [like so](https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/wishbone.py#L164). This doesn't mirror my understanding of how the ADR lines should be as it informed them -- my application is using the litex wishbone bus so it was my first exposure to wishbone.

Jeff-Ciesielski:
> I agree the spec is sort of hard to decipher with granularity (and with a bunch of other stuff to be honest)

:100: 


> The crux of it is for me is when they say ADR_O(n..2) do they mean the core has N ADR lines or N-2?

Practically, a lot of hardware exposes the latter (see Vex as an example, it exposes `PC>>2` on the ADR_O port and is used in a LOT of different hardware ecosystems).  Whether or not this is really actually correct still unclear to me as well, but If you squint and say "ADR_I/Os value reflects the DATA word size in individual increments) and adjust accordingly depending on your application, it makes sense enough.

To be even more confusing though, when you read the section on burst transfers and address handling...it says this:
```
An incrementing burst is defined as multiple accesses to consecutive addresses. Each
transfer the address is incremented. The increment is dependent on the data array
[DAT_O()], [DAT_I()] size; for an 8bit data array the increment is 1, for a 16bit data array
the increment is 2, for a 32bit data array the increment is 4, etc
```

Which IMO sort of implies that ports should always expose byte addresses even if they ignore the LSBs due to their granularity :|.

Thanks, I hate it :laughing: 

>  [the sel stuff](https://cdn.opencores.org/downloads/wbspec_b4.pdf#page=62) being a bit flag for the bytes and not a count of valid bytes still seems odd to me.

It's actually really useful.  As a (somewhat contrived) example: I have a peripheral in one of my projects that exposes a 2x4 character LED display that's driven by a really funky IC.  The devices has 2 32 bit registers with BYTE granularity (one byte per character).  When I want to update, I can issue a write to that port with the SEL line(s) for any individual LED in the group, or for any combination of them without having to do a read/modify/write.

Lots of other ways to skin that cat (each character gets its own 32 bit word with only the bottom 8 bits / I keep them all registered somewhere else and do the updates there and always do 32 bit updates), but the flexibility in implementation is nice.

andreasWallner:
@Jeff-Ciesielski Would be great to get some authoritative answer here. I'll ask Dolu as well, I think he has some contacts. (One datapoint I stumbled over was the Zip-CPU blog, there it's also described with the ADDR_O always addressing the a data width wide signal.

@jdavidberger I'd like to hold off until we can clarify what the behavior of the ADDR lines should be, but the remaining fixes would be greatly appreciated. Should we go ahead and merge the other stuff that you fixed?

jdavidberger:
Sure, do you want me to pop the top few commits off or do you want to just do it on your end? Might be better if you just pull those in to dev and I can merge that into this branch so the PR with the comment history is still active

andreasWallner:
Ok, @Dolu1990 got confirmation that the reading of ADDR_O as always being a word address is correct.
So we should make the fix in the RegIf generator. Only thing is that we don't want to break backwards compatiblity so at least for the time being we'd like to add a mandatory parameter to it's constructor to decide between word address (standard compliant) and bytes address (non standard compliant, but for people that are using it that way ATM).

We should not support the non-standard mode in the other generator or explicitiy in the bus. The reason to make it a mandatory parameter for now is to call attention to the change for all that update SpinalHDL.

Would you be up to change this in the PR?

PS: another reason for SEL being enable lines are RAMs, those often have enables for bits/bytes and those can be used directly

jdavidberger:
Is there a reference or something to point to to resolve the granularity thing? It'd be nice to have a solid thing to point to in the scala doc for this; if for no other reason than to prevent the inevitable issues that will be opened. 

Adding a mandatory parameter would introduce a build-time breakage on backwards compatibility. I'm guessing it breaks binary compatibility if someone is linking into the library. I don't see a great way around this -- the best option would be to obsolete the entire `WishboneBusInterface` and introduce `WishboneBusStandardInterface` or something but cedeing the proper namespace long term isn't great. Worth a big call out in the changelog though.

I can make this modification; here is roughly what I'd do:

- *Move* the granularity enum to the WishboneBusInterface as a non optional parameter. Documentation calls out this thread and supporting documents. 
- *Keep* `byteAddress` helper member of the bus; *Remove* any reference to granularity in the config -- the helper function just does the correct shifting (Rationale is we might as well have the helper function; it needs to be invoked in a few places anyway).
- *Add* a helper function to `Wishbone` named `fixByteGranularAddressing : Wishbone`. This takes what is logically a byte-granular address schemes and returns a standard word granular bus. This is what will be used on the WishboneBusInterface; and will be useful when people interface with other IPs. 

If this is in line with what you were thinking, I'll go ahead and make those changes

vr2045:
A related reference is the core of the address incrementer for the FuseSoC Wishbone system [here](https://github.com/fusesoc/wb_common/blob/f17eeb1db43e78e5cde8e41506967ccb2d38c4cb/wb_common.v#L41). 

[This line](https://github.com/fusesoc/wb_common/blob/f17eeb1db43e78e5cde8e41506967ccb2d38c4cb/wb_common.v#L62) to me implies that the Wishbone address is byte oriented in the FuseSoC wishbone cores that are part of this system.

PS: this wb_common file and the address incrementer is used in various FuseSoC Wishbone BFM's as well as synthesizable components so think this is at least one de-facto standard due to its wide usage.

I think that due to lack of clarity in the spec, there can be an option to allow the user to specify byte oriented granularity of the Wishbone.

Dolu1990:
> I think that due to lack of clarity in the spec, there can be an option to allow the user to specify byte oriented granularity of the Wishbone.

Sound fair to me. An option in the wishbone config itself to specify the kind of address. With its default set to "word".

jdavidberger:
> Sound fair to me. An option in the wishbone config itself to specify the kind of address. With its default set to "word".

This is the existing branch but with `Word` as a default instead of `Unspecified`. I can remove the 'unspecified' value altogether and make it default to word. This WILL cause breaking backwards compatibility changes though in the bus interface code.

I can also mark unspecified as `deprecated` and leave it as the default. On the next version breaking rev, we can just remove unspecified and make the default `word`. 

I'll do whichever; just wanted to present an option that didn't change behavior of existing code.

Dolu1990:
Ahhh i didn't thinked about "unspecified" XD
Hmm i'm a bit scared that Unspecified will just endup being a temporary solution which endup just breaking over time, as many component realy need to have it specified.
It will also creates some blind spot.

No good solution, only less worst ones XD

jdavidberger:
I'm less worried about that; only `class Wishbone` should ever actually look at the granularity. It exposes `byteAddress` which is what components should look at. I can make `addressGranularity` itself private.

jdavidberger:
I went ahead and added the deprecation and private access modifier since this was the easiest option. I can still just remove unspecified outright if that is what we want to do here.

lib/src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala
```diff
@@ -3,6 +3,25 @@ package spinal.lib.bus.wishbone
...
+  /**
+   * Word granular addressing increments by one per word. This is the standard defined behavior of the bus.
+   */
+  val Word = Value
+
+  /**
+   * Byte granular addressing increments by one per byte. This is not in adherence with the wishbone standard but is
+   * a commonly encountered configuration with external IP.
+   */
+  val Byte = Value
```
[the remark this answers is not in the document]
Dolu1990: But else, seems all fine for me to merge.
Should somebody have a say about it speak now, else stay silent forever <3 (kiding)

Dolu1990:
So, how do we feel about the PR ?
Do we have a consensus ?

Dolu1990:
I guess that one consensus we can have is that bad spec should be a crime :D

andreasWallner:
We should just remove all of the Wishbone stuff from the library 😲 (only joking here)

The thing that I see coming up downstream is the request to prevent connection of a word and a byte bus...

Since that seems to be the only solution to also generate the broken code needed for some systems - LGTM.

andreasWallner (approved):

jdavidberger:
I can add something for tying together busses of unlike granularities in a later PR. I need pipeline WB stuff now so I'm looking at that right now. Is there anything else needed for this PR? I don't want to stack more onto it but I want to base anything I do for pipeline mode on top of the changes here.

Dolu1990:
Thanks :D

Issue #1314, opened 2024-02-19
Formal verification:  Verilog generated when using synchronous resets is not understood by SymbiYosys


janschiefer:
SpinalHDL: 1.10.1
Scala version: 2.12.18
sbt version: 1.9.8
SymbiYosys version: Giit 19.02.204

Problem:
- SymbiYosys chokes on "assert(xxx) else begin end"- statements ( maybe not supported? )
- also, a ";" is missing after the assert-statement

Fix:
- remove  generation of this block... at least for now:
```
else begin
        $fatal(""); // MyTopLevelFormal.scala:L23
        $finish;
      end
```
- generate  ";" behind assert

Full demonstration: https://github.com/janschiefer/SpinalTemplateSbt/tree/code-generation-sync-reset

Let's try to verify code, that geneate a sychronous reset:

Config.scala:
```
package projectname

import spinal.core._
import spinal.core.sim._

object Config {
  def spinal = SpinalConfig(
    targetDirectory = "hw/gen",
    defaultConfigForClockDomains = ClockDomainConfig(
      resetActiveLevel = HIGH,
      resetKind = SYNC
    ),
    onlyStdLogicVectorAtTopLevelIo = true
  )

  def sim = SimConfig.withConfig(spinal).withFstWave
}
```

MyTopLevelFormal.scala:
```
package projectname

import spinal.core._
import spinal.core.formal._

// You need SymbiYosys to be installed.
// See https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Formal%20verification/index.html#installing-requirements
object MyTopLevelFormal extends App {
  FormalConfig
    .withConfig( Config.spinal )
    .withBMC(10)
    .doVerify(new Component {
      val dut = FormalDut(MyTopLevel())

      // Ensure the formal test start with a reset
      assumeInitial(clockDomain.isResetActive)

      // Provide some stimulus
      anyseq(dut.io.cond0)
      anyseq(dut.io.cond1)

      // Check the state initial value and increment
      assert(dut.io.state === past(dut.io.state + U(dut.io.cond0)).init(0))
    })
}
```

SymbiYosys logfile:
```
SBY  2:03:42 [/home/jschiefer/Code/SpinalTemplateSbt/./simWorkspace/unamed/unamed_bmc] Removing directory '/home/jschiefer/Code/SpinalTemplateSbt/simWorkspace/unamed/unamed_bmc'.
[...]

PR #1315, opened 2024-02-19
Verilog generated when using synchronous resets is not understood by SymbiYosys - remove "assert() else begin end"-block generation for now

janschiefer:
See issue https://github.com/SpinalHDL/SpinalHDL/issues/1314

Remove "assert() else begin end"-block generation for now.


Readon:
This "else begin" strings is used for assert otherthan formal verification.
There two use case of assert now.
1. assert in simulation.
2. assert in formal verification.

PR #1319, opened 2024-02-20
Some little change on inout signal

yportne13:
1. remove `{` `}`
2. add size `[1:0]`
3. align brackets
4. add comment `//~`
5. remove `analog_wrap`

1.10.1:
```verilog
MyBlack b (
  .io_ttttttttttttttttttttttttt (b_io_ttttttttttttttttttttttttt[1:0]), //o
  .io_x                         ({analog_wrap_0}) 
);
assign ret[1 : 0] = analog_wrap_0[1 : 0];
```

this pr:
```verilog
MyBlack b (
  .io_ttttttttttttttttttttttttt (b_io_ttttttttttttttttttttttttt[1:0]), //o
  .io_x                         (ret[1:0]                           )  //~
);
```

```scala
import spinal.core._
import spinal.lib._
import spinal.lib.io._

class InOut extends Component {
  val b = new MyBlack
  val a = new MyBlack
  val c = new MyBlack
  val ret = Analog(Bits(2 bit)) 
  b.io.x := ret
  c.io.x := ret
  a.io.x := ret

}
class MyBlack extends BlackBox {
  val io = new Bundle {
    val ttttttttttttttttttttttttt = out Bits(2 bits)
    val x = inout(Analog(Bits(2 bits)))
  }
}

object testInOutTop extends App {
  SpinalVerilog(new InOut)
}
```


yportne13:
`assign` an analog signal might lead to the insertion of digital buffers in some eda tools. To avoid generating such code, two modifications [phase.scala#389](https://github.com/SpinalHDL/SpinalHDL/pull/1319/files#diff-0ae53b8431a1d28d269ed1e47ce1e97ad2bddcb8ccbcb3eb656fe5474d2e5bddL389) and [phase.scala#419](https://github.com/SpinalHDL/SpinalHDL/pull/1319/files#diff-0ae53b8431a1d28d269ed1e47ce1e97ad2bddcb8ccbcb3eb656fe5474d2e5bddL419) were made but it seems that these changes have introduced other bugs which are difficult for me to resolve

yportne13:
maybe something like this?
```scala
        val count = filtred.count(e => e.isInOut && e.component == c)
        val countA = filtred.count(e => e.isAnalog && e.component == c)
        count match {
          case 0 => { //No analog inout, will need to create a connection seed later on
            if(countA == 1)
              seeds += island.elements.find(e => e.bt.isAnalog && e.bt.component == c).get.bt
            else {
            val groups = island.elements.map(b => anlogToGroup.get(b.bt)).filter(_.nonEmpty).map(_.get).toArray.distinct
            val finalGroup = groups.length match {
              case 0 => {
                val group = new Group()
                analogGroups += group
                group
              }
              case 1 => groups.head
              case _ =>{
                val ghead = groups.head
                for(group <- groups.tail){
                  groups.head.islands ++= group.islands
                  analogGroups -= group
                  for(i <- group.islands){
                    for(b <- i.elements){
                      anlogToGroup(b.bt) = ghead
                    }
                  }
                }
                ghead
              }
            }
            finalGroup.islands += island
            for(b <- island.elements){
              anlogToGroup(b.bt) = finalGroup
            }
            }
          }
          case 1 => seeds += island.elements.find(e => e.bt.isInOut && e.bt.component == c).get.bt //Got a analog inout to host the connection
          case _ => PendingError("MULTIPLE INOUT interconnected in the same component"); null
        }
```

Dolu1990:
Hi,

> assign an analog signal might lead to the insertion of digital buffers in some eda tools

what do you mean by digital buffers ? Which kind of assign (example) ? I'm trying to understand what exactly you are trying to avoid. Where would be the digital buffer inserted in the example ? (ret is never used in a digital manner, always analog)

jijingg:
> Hi,
> 
> > assign an analog signal might lead to the insertion of digital buffers in some eda tools
> 
> what do you mean by digital buffers ? Which kind of assign (example) ? I'm trying to understand what exactly you are trying to avoid. Where would be the digital buffer inserted in the example ? (ret is never used in a digital manner, always analog)

hi @Dolu1990 , 

first question , in case1, what the `ret`  signal used for, which declared and assigned but never used ?  
second,  Our SYN-team requires us to prohibit the use of assign statements in analog signal and avoid inserting buffers
```verilog
assign   a =  b ; 
```
for some syn-tools , may be syn like this , For analog signals, it is unacceptable
```verilog
BUFFD1BWP169H3PDSVT G3404(.I(b), .Z(a));
```

Dolu1990:
Hi @jijingg 

> first question , in case1, what the ret signal used for, which declared and assigned but never used ?

So, mostly because tristate stuff in verilog / vhdl are a pain, when there is no inout going out in the current component to "bake" the analog island, SpinalHDL will "bake" the analog island into a new signal (analog_wrap_0). all other analog signal connected to the island (ret in our case) become regular digital read only signals (for simulation wave trace and for digital read of the analog island)

> Our SYN-team requires us to prohibit the use of assign statements in analog signal and avoid inserting buffers

As far as i understand this requirement, it make sense to avoid somebody connecting two analog stuff together, as it will not be infered as a direct analog connection, but with a buffer. Make sense to have that catched in lint.
But here, the intended design from SpinalHDL isn't in the same situation. In the generated netlist, the ret signal will never be used in a analog manner anymore. It is just a probe.

Does this make sense or i missed something ?
How can i assign the value of an analog signal into an digital thing in verilog ?

jijingg:
tristate signal 

> How can i assign the value of an analog signal into an digital thing in verilog ?


As far as I understand, it's not possible to connect an analog signal to a digital signal directly. If you need to pull out an analog signal, you cannot use "assign." The probe point should be "analog_wrap_0" instead of the "ret" signal, and external pads should use analog pads rather than digital pads

Dolu1990:
I pushed an alternative proposal for the Phase.scala changes : 
https://github.com/SpinalHDL/SpinalHDL/commit/c93422d034988f168334d52c729928f6c6b08d1f

So, the idea is : 
- If there is an single inout, use it as seed
- else if there is an single analog, use it as a seed
- else create seed

The idea is to keep flexibility.
while allowing to use single analog signal to avoid potential synthesis issues.

Dolu1990:
I just tested, the
[error] 	spinal.core.ChecksTester
[error] 	spinal.core.InOutTesterCocotbBoot
[error] 	spinal.core.AnalogConnectionCocotbBoot

are passing on the analog1319 branch.
It seems the issure remining now is more related to the ComponentEmitterVerilog.scala changes

Dolu1990:
Thanks :)

PR #1320, opened 2024-02-20
SpinalFormalConfig: Set _hasAsync accodring to clock config; set includeFormal as default in SpinalFormalConfig

janschiefer:
Closes #

https://github.com/SpinalHDL/SpinalTemplateSbt/issues/39
https://github.com/SpinalHDL/SpinalHDL/issues/1314

- SymbiYosys multiclock mode has to be used when using ResetKind == ASYNC, so enable multiclock mode when using this reset mode in the default clock domain
- always use ".includeFormal" in FormalConfig, to generate the "right" kind of "assert" depending on mode ( simulation vs. verification )

# Impact on code generation

RTL:
- use .includeFormal as default and therefore do not generate "assert(...) else ... begin" blocks in formal verification mode

SymbiYosis configuration file:
- use "multiclock on" when an using an ASYNC reset in the default clock domain

# Checklist
- [X] API changes are or will be documented: No changes on outward facing API

### DIFF
--- a/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
+++ b/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
@@ -62,7 +62,7 @@ case class SpinalFormalConfig(
     var _backend: SpinalFormalBackendSel = SpinalFormalBackendSel.SYMBIYOSYS,
     var _keepDebugInfo: Boolean = false,
     var _skipWireReduce: Boolean = false,
-    var _hasAsync: Boolean = false,
+    var _hasAsync: Boolean = true,
     var _timeout: Option[Int] = None,
     var _engines: ArrayBuffer[FormalEngin] = ArrayBuffer()
 ) {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
```diff
@@ -118,7 +119,8 @@ case class SpinalFormalConfig(
   }
 
   def withConfig(config: SpinalConfig): this.type = {
-    _spinalConfig = config
+    _spinalConfig = config.includeFormal
```
Readon, line 122: How about only keep this line to help user easily work around the config problem?

Readon: Multiclock features of formal tools only good for the design which is really have several clocks. It might work slower.

janschiefer: You have a actually working alternative with ASYNC reset without multiclock?

Readon: > You have a actually working alternative with ASYNC reset without multiclock?

The tester in located in tester/ subproject are all use ASYNC reset as default. The SYNC reset has not been tested.
However, although the design itself use ASYNC reset, the verification actually run is still in SYNC manner.
So I think it doesn't matter.

Readon:
according to #1435 this can be closed.

PR #1324, opened 2024-02-22
Inject valCallback for traits

### LINKED ISSUE #1322: NamedType returns null when called outside of a Component
I'm trying to restructure some of my common code between components with Scala `trait`s, but it seems like we cannot create `NamedType`s for use with `HardMap` properly inside traits:

```scala
package pionic

import spinal.core._

case class Good() extends Component {
  val KeyA = NamedType(Bool())
  val hm = HardMap(Seq(KeyA))
}

trait BadTrait {
  this: Component =>

  val KeyB = NamedType(Bool())
  val hm = HardMap(Seq(KeyB))
}
case class Bad() extends Component with BadTrait

object NamedTypeNull extends App {
  SpinalVerilog(Good())
  SpinalVerilog(Bad())
}
```

Generates the following NPE:

```
Exception in thread "main" scala.MatchError: null (of class spinal.core.NamedType)
        at spinal.core.HardMap.keyToName(HardMap.scala:20)
        at spinal.core.HardMap.update(HardMap.scala:28)
        at spinal.core.HardMap.add(HardMap.scala:32)
        at spinal.core.HardMap$.$anonfun$apply$1(HardMap.scala:11)
        at spinal.core.HardMap$.$anonfun$apply$1$adapted(HardMap.scala:11)
        at scala.collection.immutable.List.foreach(List.scala:333)
        at spinal.core.HardMap$.apply(HardMap.scala:11)
        at pionic.BadTrait.$init$(NamedTypeNull.scala:14)
        at pionic.Bad.<init>(NamedTypeNull.scala:16)
        at pionic.NamedTypeNull$.$anonfun$new$2(NamedTypeNull.scala:20)
        at spinal.core.internals.PhaseCreateComponent.$anonfun$impl$273(Phase.scala:2556)
        at spinal.core.fiber.Engine$.$anonfun$create$1(AsyncCtrl.scala:147)
        at spinal.core.fiber.AsyncThread.$anonfun$jvmThread$1(AsyncThread.scala:59)
        at spinal.core.fiber.EngineContext.$anonfun$newJvmThread$1(AsyncCtrl.scala:39)
        at spinal.sim.JvmThread.run(SimManager.scala:51)
```

Is there a way to make this pattern work?

KireinaHoro:
Closes #1322.  Refactored the compiler plugin to recognize traits with a `Component` self-type (actually with `ValCallback`), and inject calls to `valCallback` automatically.

### DIFF
--- a/idslplugin/src/main/scala/spinal/idslplugin/components/MainTransformer.scala
+++ b/idslplugin/src/main/scala/spinal/idslplugin/components/MainTransformer.scala
@@ -46,7 +46,7 @@ class MainTransformer(val global: Global) extends PluginComponent with Transform
       val clazz = cd.impl.symbol.owner
       val valCallbackClazz = if (cd.mods.hasFlag(Flag.TRAIT)) {
         val selfType = cd.impl.self.tpt.tpe
-        selfType.parents.find(_.toString == "spinal.core.Component").get.typeSymbol
+        selfType.parents.find(typeHasTrait(_, "spinal.idslplugin.ValCallback")).get.typeSymbol
       } else clazz
       val func = valCallbackClazz.tpe.members.find(_.name.toString == "valCallback").get
       val body = cd.impl.body.map {
--- a/idslplugin/src/main/scala/spinal/idslplugin/components/MainTransformer.scala
+++ b/idslplugin/src/main/scala/spinal/idslplugin/components/MainTransformer.scala
@@ -37,6 +37,40 @@ class MainTransformer(val global: Global) extends PluginComponent with Transform
       }
     }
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

The missmatch come from https://github.com/SpinalHDL/SpinalHDL/blob/d04f565baf76fbbeef745cc8d47cb613b1aa4909/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala#L105

This PR make the vallcalback not working on them, so, those two signals doesn't get any name, and that is picked up by that regression, which check the graph size.

Expected number is 251 :)

KireinaHoro:
Dropping this since I started working with the Fiber/Plugin system in #1322.

PR #1325, opened 2024-02-22
Compile tester with compiler plugin in Mill

KireinaHoro:
To run the test cases in Mill correctly, `tester` should be compiled with the IDSL plugin.

The equivalence of `sbt tester/testOnly` is `mill "tester[2.13.12].test.testOnly"` (replace `2.13.12` with `_` to test for all cross targets).

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -104,8 +104,9 @@ trait Core extends SpinalModule with SpinalPublishModule with CrossSbtModule {
   }
 }
 
-object tester extends Cross[Tester](Version.SpinalVersion.compilers)
+object crossTester extends Cross[Tester](Version.SpinalVersion.compilers)
 trait Tester extends SpinalModule with SpinalPublishModule with CrossSbtModule {
+  override def millSourcePath = os.pwd / "tester"
   def mainClass = Some("spinal.tester")
   def moduleDeps = Seq(core(crossScalaVersion), sim(crossScalaVersion), lib(crossScalaVersion))
   def scalacOptions = super.scalacOptions() ++ idslplugin(crossScalaVersion).pluginOptions()
@@ -114,4 +115,10 @@ trait Tester extends SpinalModule with SpinalPublishModule with CrossSbtModule {
   object test extends CrossSbtModuleTests with TestModule.ScalaTest {
     def ivyDeps = Agg(ivy"org.scalatest::scalatest::${scalatestVersion}")
   }
+}
+
+object tester extends Module {
+  val defaultTester = crossTester(Version.SpinalVersion.compilers.head)
+  def testOnly(args: String*) = T.command { defaultTester.test.testOnly(args: _*) }
+  def runMain(mainClass: String) = T.command { defaultTester.runMain(mainClass) }
 }
--- a/build.sc
+++ b/build.sc
@@ -104,8 +104,8 @@ trait Core extends SpinalModule with SpinalPublishModule with CrossSbtModule {
   }
 }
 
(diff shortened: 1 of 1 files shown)


Readon:
Hi, thanks for the contribution.
One more thing, the command have no default scala version used, so that the legacy command could not work now.
``` bash
mill tester.test.testOnly xxxx
```

Can you assign the default version? So that the compatibility is maintained.
I suppose now the 2.12.18 is the default version of scala of sbt.

KireinaHoro:
> Hi, thanks for the contribution. One more thing, the command have no default scala version used, so that the legacy command could not work now.
> 
> ```shell
> mill tester.test.testOnly xxxx
> ```
> 
> Can you assign the default version? So that the compatibility is maintained. I suppose now the 2.12.18 is the default version of scala of sbt.

This behaviour is [supported by mill](https://mill-build.com/mill/Cross_Builds.html#_default_cross_modules) through writing:

```shell
$ mill "tester[].test.testOnly"
[300/300] tester[2.12.18].test.testOnly
...
```

Quotes is needed only for zsh due to the squared brackets; they're not needed on bash.  I guess we could document this somewhere?

Readon:
> 
> This behaviour is [supported by mill](https://mill-build.com/mill/Cross_Builds.html#_default_cross_modules) through writing:
> 
> ```shell
> $ mill "tester[].test.testOnly"

This still incompatible. is there other solution?

> [300/300] tester[2.12.18].test.testOnly
> ...
> ```
> 
> Quotes is needed only for zsh due to the squared brackets; they're not needed on bash.
> 
> The version selection for mill is kept in sync with sbt through [Version.sc](https://github.com/SpinalHDL/SpinalHDL/blob/dev/project/Version.sc) symlinked to [Version.scala]
(https://github.com/SpinalHDL/SpinalHDL/blob/dev/project/Version.scala) for sbt. I guess we could document this somewhere?

This could not be worked on Windows easily.

KireinaHoro:
> This still incompatible. is there other solution?

I'm not sure about what you mean by "compatible"; with whom?  Do you mean the sbt syntax?

Readon:
> > This still incompatible. is there other solution?
> 
> I'm not sure about what you mean by "compatible"; with whom? Do you mean the sbt syntax?

imcompatible with the documented command line [here](https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Developers%20area/mill%20support.html#run-a-specified-test-suite)

KireinaHoro:
Ok I pushed a fix.  So the current API does not require the brackets anymore:

```shell
$ mill tester.test.testOnly \*RepeatabilityTester
$ mill tester.runMain <main class>
...
```

To be honest the `test` submodule in `tester.test.testOnly` is there only to keep the API the same as the docs and IMO we should update the docs instead.

Readon:
> Ok I pushed a fix. So the current API does not require the brackets anymore:
> 
> ```shell
> $ mill tester.test.testOnly \*RepeatabilityTester
> $ mill tester.runMain <main class>
> ...
> ```
> 
> To be honest the `test` submodule in `tester.test.testOnly` is there only to keep the API the same as the docs and IMO we should update the docs instead.

The thing is to keep it running simple and consistent. There might be some shell is based on the old command line.

Also the mill version only run test classes under tester subproject. We have split the tester subproject into several subprojects now. If interested PR is welcome.

Readon:
Thanks.

Dolu1990:
Thanks :D

PR #1332, opened 2024-02-29
Add exit() to StateMachine class

AtaraxiaZ:
Closes #

The two styles of writing state machine should better be consistent. 

For example, the first style of using a StateMachine is

```scala
val stateA: State = new State {
    whenIsActive {
        exit()
    }
}
```

The second style is 
```scala
val stateA = new State

StateA
    .whenIsActive {
        exitFsm()
   }
```
The inconsistency happen when using `exit()`, it can only be use in the first style, and If you want to use it in the
second style, you need to use exitFsm() which is not documented. So from a user friendly perspective, I think add
a new `exit()` method in the StateMachine class would be useful. And the second style would also be written as

```scala
val stateA = new State

StateA
    .whenIsActive {
        exit()
   }
```

# Impact on code generation
No impact.

# Checklist
Maybe it's not required for this small modification?

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/fsm/StateMachine.scala
+++ b/lib/src/main/scala/spinal/lib/fsm/StateMachine.scala
@@ -362,6 +362,10 @@ class StateMachine extends Area with StateMachineAccessor with ScalaLocated {
     wantStart := True
   }
 
+  def exit(): Unit = {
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks :)

PR #1333, opened 2024-02-29
Handle Quotes in Attribute values

g0t00:

Attribute values currently do not escape double quote characters, thus invalid vhdl/verilog is emitted when there are quotes in string attribute values.

# Impact on code generation

# Checklist

- [x] Unit tests were added

### DIFF
--- a/tester/src/test/scala/spinal/tester/scalatest/AttributeEmitTests.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/AttributeEmitTests.scala
@@ -1,0 +1,17 @@
+package spinal.tester.scalatest
+
+import spinal.tester.SpinalAnyFunSuite
+import spinal.core._
+
+class AttributeEmitTests extends SpinalAnyFunSuite {
+  test("check escaping of quotes in attributes") {
+    class Test extends Component {
+      val a = RegInit(False)
+      a.addAttribute("testname", "test \"value\" ")
+    }
+    val verilog = SpinalVerilog(new Test)
+    assert(verilog.getRtlString().contains("test \\\"value\\\""))
+    val vhdl = SpinalVhdl(new Test)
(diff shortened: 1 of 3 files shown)


g0t00:
I do not know if there are other ways to emit strings into the rtl. If so they probably have a similar problem.

tester/src/test/scala/spinal/tester/scalatest/AttributeEmitTests.scala
```diff
@@ -0,0 +1,17 @@
...
+class AttributeEmitTests extends SpinalAnyFunSuite {
+  test("check escaping of quotes in attributes") {
+    class Test extends Component {
+      val a = RegInit(False)
+      a.addAttribute("testname", "test \"value\" ")
+    }
+    val verilog = SpinalVerilog(new Test)
+    assert(verilog.getRtlString().contains("test \\\"value\\\""))
+    val vhdl = SpinalVhdl(new Test)
+    assert(vhdl.getRtlString().contains("test \"\"value\"\""))
```
andreasWallner, line 15:
```suggestion
    assert(vhdl.getRtlString().contains(raw"""test ""value"""""))
```

andreasWallner (approved):
Thanks!

I'd only make this minor change to raw literals (which don't need escaping) for readability

g0t00:
> Thanks!
> 
> I'd only make this minor change to raw literals (which don't need escaping) for readability

Ah great. Did not know those exist in scala.

andreasWallner:
I see a bunch of tests failing, they seem connected to escaping - would you have a look?

g0t00:
> I see a bunch of tests failing, they seem connected to verilog escaping/literals - would you have a look?

Should be fixed.

andreasWallner:
Thanks!

PR #1338, opened 2024-03-05
make cutLongExpressions optional

yportne13:
make cutLongExpressions optional


Dolu1990:
Thanks :)

Just to understand, what where the reasons for this need ?

PR #1339, opened 2024-03-05
If the output is not connected, set it to be an empty pair of parentheses

yportne13:
If the child component output connect to nothing, the origin way is:
```verilog
  wire                a_io_y;
  wire                a_io_z;
  wire                a_io_p;
  wire                useless;

  Sub a (
    .io_x (io_x  ), //i
    .io_y (a_io_y), //o
    .io_z (a_io_z), //o
    .io_p (a_io_p)  //o
  );
  assign useless = (a_io_y && 1'b1);
  assign io_t = a_io_z;
```
this pr will change this to:
```verilog
  wire                a_io_y;
  wire                a_io_z;
  wire                useless;

  Sub a (
    .io_x (io_x  ), //i
    .io_y (a_io_y), //o
    .io_z (a_io_z), //o
    .io_p (      )  //o
  );
  assign useless = (a_io_y && 1'b1);
  assign io_t = a_io_z;
```
the testcase above is generate from this:
```scala
class TestOutput extends Component {
  val io = new Bundle {
    val x = in Bool()
    val t = out Bool()
  }

  val a = new Sub
  a.io.x := io.x
  val useless = Bool()
  useless := a.io.y && True
  io.t := a.io.z

}

class Sub extends BlackBox {
  val io = new Bundle {
    val x = in Bool()
    val y = out Bool()
    val z = out Bool()
    val p = out Bool()
  }

}
```

### DIFF
--- a/core/src/main/scala/spinal/core/Spinal.scala
+++ b/core/src/main/scala/spinal/core/Spinal.scala
@@ -175,7 +175,7 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         var singleTopLevel             : Boolean = true,
                         var noAssertAtTimeZero         : Boolean = false,
                         var cutLongExpressions         : Boolean = true,
-                        var cleanOutput                : Boolean = false
+                        var emitFullComponentBindings  : Boolean = true
 ){
   def generate       [T <: Component](gen: => T): SpinalReport[T] = Spinal(this)(gen)
   def generateVhdl   [T <: Component](gen: => T): SpinalReport[T] = Spinal(this.copy(mode = VHDL))(gen)
--- a/core/src/main/scala/spinal/core/Spinal.scala
+++ b/core/src/main/scala/spinal/core/Spinal.scala
@@ -174,7 +174,8 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         bitVectorWidthMax              : Int = 4096,
                         var singleTopLevel             : Boolean = true,
                         var noAssertAtTimeZero         : Boolean = false,
-                        var cutLongExpressions         : Boolean = true
+                        var cutLongExpressions         : Boolean = true,
+                        var cleanOutput                : Boolean = false
 ){
   def generate       [T <: Component](gen: => T): SpinalReport[T] = Spinal(this)(gen)
   def generateVhdl   [T <: Component](gen: => T): SpinalReport[T] = Spinal(this.copy(mode = VHDL))(gen)
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -165,7 +165,7 @@ class ComponentEmitterVerilog(
         val componentSignalName = (sub.getNameElseThrow + "_" + io.getNameElseThrow)
         val name = component.localNamingScope.allocateName(componentSignalName)
         val noUse = signalNoUse(io)
-        if (!io.isSuffix && (io.isVital || !noUse))
+        if (!io.isSuffix && ((io.isVital || !noUse) || !spinalConfig.cleanOutput))
           declarations ++= emitExpressionWrap(io, name)
         referencesOverrides(io) = name
       }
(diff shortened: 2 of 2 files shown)


Dolu1990:
Hi,

I'm a bit split with keeping the old behaviour.

The reason is for simulation / debugability purposes.
If we keep the .io_p (a_io_p)  //o, it has the good thing that it provide visibility to a.io.p value from the parent component.
Meaning in my wave, i can select the parent component, and probe all the IO of all the sub component with the sub component name as a prefix.

What are the advantages of .io_p (      )  //o ? (lint ?)

yportne13:
> Hi,
> 
> I'm a bit split with keeping the old behaviour.
> 
> The reason is for simulation / debugability purposes. If we keep the .io_p (a_io_p) //o, it has the good thing that it provide visibility to a.io.p value from the parent component. Meaning in my wave, i can select the parent component, and probe all the IO of all the sub component with the sub component name as a prefix.
> 
> What are the advantages of .io_p ( ) //o ? (lint ?)

In our project, only a small part of code is writen in SpinalHDL. Most are verilog/sv. All code reviews are conducted on Verilog. Most reviewers do not understand SpinalHDL. In the review, identifying which module outputs have not been used is crucial. Currently, the style in which SpinalHDL generates is not as obvious as `io_p()` style. Therefore, this change is made to ensure consistency with our other Verilog code style. This is also the reason for that `cutLongExpression` pr. Maybe also change this pr to optional?

Dolu1990:
Hi,

Yes, i think it should be optional (off by default, to privilege debugability)

core/src/main/scala/spinal/core/Spinal.scala
```diff
@@ -174,7 +174,8 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         bitVectorWidthMax              : Int = 4096,
                         var singleTopLevel             : Boolean = true,
                         var noAssertAtTimeZero         : Boolean = false,
-                        var cutLongExpressions         : Boolean = true
+                        var cutLongExpressions         : Boolean = true,
+                        var cleanOutput                : Boolean = false
```
Dolu1990, line 178: cleanOutput as a name isn't explicit enough i think.

Finding a good short name is quite hard for that one XD
hmmm, maybe a parameter which specify the reverse of the current clean output, something like : 
emitFullComponentBindings, meaning, always connect all the pins of a sub component into its parent ?

Dolu1990: Default bing true

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -1732,4 +1740,21 @@ end
   fillExpressionToWrap()
   emitEntity()
   emitArchitecture()
+
+  def signalNoUse(sig: BaseType): Boolean = {
+    var noUse = true
+    component.dslBody.walkStatements {
```
Dolu1990, line 1746: Ahh i would say this doesn't scale well. for each request it will check the whole component dslBody. 
I would say, the best would be to generate a noUse linkedhashmap that specify for every output of the current component if it is used once. And then use that linkedhashmap for

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -1728,8 +1736,36 @@ end
     }
   }
 
+  var outputSignalNoUse: Set[BaseType] = Set()
```
Dolu1990, line 1739: Scala immultable Set do not scale.
You can use the mutable.LinkedHashSet instead

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -1728,8 +1736,36 @@ end
...
+  if(!spinalConfig.emitFullComponentBindings) {
+    component.children.foreach(sub =>
+      sub.getAllIo
+      .foreach(io => if(io.isOutput) {
+        outputSignalNoUse = outputSignalNoUse + io
+      }
+    ))
+  }
+  component.dslBody.walkStatements {
```
Dolu1990, line 1748: Shoulden't it be in the above if statement ? to avoid doing it when it isn't necessary.

yportne13:
Is there any issue with this PR that needs to be addressed before it can be merged as soon as possible? I also have another PR related to SystemVerilog interface changes that is waiting to be submitted, and there is a slight overlap in the changes with this PR.

Dolu1990:
Thanks ^^
All good :)

PR #1344, opened 2024-03-11
UnionElement simGet

### LINKED ISSUE #1342: `Union` elements cannot be accessed inside SpinalSim properly
```scala
import spinal.core._
import spinal.core.sim._
import spinal.lib._
import spinal.lib.sim._

case class MyUnion() extends Union {
  val a = newElement(UInt(64 bits))
  val b = newElement(Bits(64 bits))
}

object TestUnionSim extends App {
  val sim = SimConfig.withConfig(SpinalConfig()).withVerilator.allOptimisation

  val dut = sim.compile(new Component {
    val o = out(MyUnion())

    o.a := U(32, 64 bits)
  })
  dut.doSim("access-union-member") { dut =>
    println(s"Output port as int: ${dut.o.a.get.toBigInt}")
  }
}
```

This results in the following error with SpinalSim:

```
[Progress] Verilator compilation done in 3051.907 ms
[Progress] Start unamed access-union-member simulation with seed 172859342
[Error] Simulation failed at time=0
It seems like you used some SpinalHDL hardware elaboration API in the simulation. If you did, you shouldn't.
Exception in thread "main" java.lang.NullPointerException
        at spinal.core.signalCache$.apply(Misc.scala:169)
        at spinal.core.signalCache$.apply(Misc.scala:172)
        at spinal.core.UnionElement.get(Union.scala:12)
        at pionic.TestUnionSim$.$anonfun$new$1(TestUnionSim.scala:22)
        at pionic.TestUnionSim$.$anonfun$new$1$adapted(TestUnionSim.scala:21)
        at spinal.core.sim.SimCompiled.$anonfun$doSimApi$2(SimBootstraps.scala:610)
        at spinal.sim.SimManager.threadBody$1(SimManager.scala:230)
        at spinal.sim.SimManager.$anonfun$run$1(SimManager.scala:233)
        at spinal.sim.SimThread.$anonfun$jvmThread$1(SimThread.scala:93)
        at spinal.sim.JvmThread.run(SimManager.scala:51)
```

Also, it's a bit unexpected that I have to use `.get` on the `UnionElement`; seems like the implicit cast didn't work without it:

```
[error] /local/home/pengxu/work-local/enzian-pio-nic/hw/spinal/pionic/TestUnionSim.scala:22:45: value toBigInt is not a member of spinal.core.UnionElement[spinal.core.UInt]
[error]     println(s"Output port as int: ${dut.o.a.toBigInt}")
```

KireinaHoro:
Closes #1342.

### DIFF
--- a/core/src/main/scala/spinal/core/Union.scala
+++ b/core/src/main/scala/spinal/core/Union.scala
@@ -8,7 +8,7 @@ object UnionElement {
   implicit def wrapped[T <: Data](ue: UnionElement[T]) = ue.get()
 }
 
-class UnionElement[T <: Data](val t: HardType[T], host: Union) {
+class UnionElement[T <: Data](val t: HardType[T], val host: Union) {
   def get(): T = signalCache(this, "get") (host.raw.aliasAs(t))
 }
 
--- a/core/src/main/scala/spinal/core/core.scala
+++ b/core/src/main/scala/spinal/core/core.scala
@@ -473,6 +473,7 @@ package object core extends BaseTypeFactory with BaseTypeCast {
       }
     }
 
+    def mask = ((BigInt(1) << (high - low + 1)) - 1) << low
   }
 
 
--- a/core/src/main/scala/spinal/core/sim/package.scala
+++ b/core/src/main/scala/spinal/core/sim/package.scala
@@ -30,6 +30,7 @@ import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.collection.Seq
 import scala.util.Random
+import scala.util.control.Breaks._
 
 /**
   * Simulation package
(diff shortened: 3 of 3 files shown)


KireinaHoro:
The idea is similar to `aliasAs` during elaboration: find the offset by looking at all elements.  This is probably not general enough yet, not sure if `flatten` works for non-bundle elements.  I tested locally for a simple Bundle as Union element usecase.

KireinaHoro:
Seems like `manager.getBigInt` does not change after setBigInt... I'm debugging

Dolu1990:
Thanks ^^

> Seems like manager.getBigInt does not change after setBigInt... I'm debugging

Hmm did you waited a bit of time in between the two ? else it is normal SpinalSim will buffer write until next delta cycle

KireinaHoro:
Good point, but I didn't find the wait delta cycle API, is this inside SimManager?

Dolu1990:
sleep(0)  ^^
 
```scala
object TestSimDeltaCycle extends App{
  SimConfig.withWave.compile(new Component {
    val x = in UInt(8 bits)
  }).doSim{dut =>
    dut.x #= 0
    sleep(10)
    println(dut.x.toInt) // => 0
    dut.x #= 42
    println(dut.x.toInt) // => 0
    sleep(0) //Wait next delta cycle
    println(dut.x.toInt) // => 42
  }
}
```

KireinaHoro:
This does not work if we are not inside a thread, right?  I got a null pointer exception...

Dolu1990:
> This does not work if we are not inside a thread, right? I got a null pointer exception...

right.

> I'm not sure if sleep(0) should be inside the #= of the SimProxy,

Should not

> Forcing the user to call it feels weird since there's no such requirement for bundles.

?

I don't understand. It should always be the case that you can't read the result of a #= until the next delta cycle. The reason is to avoid race conditions in testbench (that's the default for VHDL, Verilog is cursed :D)

I never realy had the need to read back the value i just wrote in a testbench. The idea is more to generate a value in scala time, store it to some reference data structure and assign it to hardware.

In which scenario are you doing what you are trying to  do ?

KireinaHoro:
So the user would not be trying to read back the #= value in the same delta cycle, but you see here that how I implemented `UnionElement.SimProxy.#=`:

- read out the full union
- wipe out part of it that corresponds to the `UnionElement`
- OR the actual value
- write the new full union back

We need to read the same union multiple times in multiple `#=`.  Either we need a way to reliably increment the delta cycle (for both inside and outside of a thread), or we need a different API for unions:

- readback the whole union (with a UnionRicher?)
- stage multiple changes (on the UnionElementRicher)
- commit once

Dolu1990:
> So the user would not be trying to read back the #= value in the same delta cycle, but you see here that how I implemented UnionElement.SimProxy.#=:

Ahhhh i see, i understand now, yes this is problematic.

Hmm now we are going into the tricky tricky lands XD We should not rely on sleep(0) for that.

> commit once

Right So yes, would need the user to instanciate a class to build a BigInt value from multiple "set" request, and then use that BigInt to assign the raw of the union.
Not so great. but i dont' realy see a nicer way.

KireinaHoro:
Pushed a solution that would allow the same `simGet` API, with the requirement to do a `commit()` on the `UnionElement` (actually the `SimUnionElementPimper`).  @Dolu1990 does this look good?

Dolu1990:
Nice ^^

yes :)

KireinaHoro:
Good to merge then :)

Dolu1990:
thanks ^^

PR #1350, opened 2024-03-12
jtagvpi: Don't use blocking IO, to avoid blocking the simulation

### LINKED ISSUE #1349: JTAG VPI Simulation gets blocked by IO
I recently switched my VexRiscv simulation to the official RISC-V JTAG and JTAG VPI as a network interface and was wondering why the simulation suddenly is excruciatingly slow. Turns out, that the JTAG VPI implementation has a blocking IO call in its main loop. Line 81, i.e., the `readFully` call is blocking. Since it's executed in a `SimThread` the simulation gets blocked. In practice, its not fully blocked, since a connected OpenOCD on the VPI port will send regular packages, so the simulation just becomes really slow.

https://github.com/SpinalHDL/SpinalHDL/blob/7c6b13792617daef52f6b706632925f71ebde99e/lib/src/main/scala/spinal/lib/com/jtag/sim/JtagVpi.scala#L79-L83

A trivial fix would be to do a `while` loop that waits until enough bytes are present in the input buffer, the more elegant fix would probably to do some proper asynchronous IO (e.g., incremental non-blocking reading of `VpiCmd` packages).

rpls:
Closes #1349 

A first trivial fix for the blocking IO problem in the JTAG VPI implementation. Just adds a `while` loop with a `sleep`, which is similar to the behaviour of the `JtagTCP` implementation.

# Impact on code generation

Shouldn't have any impact on code generation, just the simulation. 

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/com/jtag/sim/JtagVpi.scala
+++ b/lib/src/main/scala/spinal/lib/com/jtag/sim/JtagVpi.scala
@@ -78,6 +78,11 @@ object JtagVpi {
       // main receive and processing loop
       while (this.synchronized(connection) != null) {
         try {
+          // Wait for input data, otherwise readFully would block the simulation
+          while(inputStream.available() < MaxSizeOfVpiCmd) {
+            sleep(jtagClkPeriod * 200)
+          }
+
           inputStream.readFully(buffer)
 
           val vpiCmd = deserializeVpiCmd(buffer)


Dolu1990:
Seems good to me ^^

rpls:
I raised the number of bytes to wait for a bit more.

andreasWallner:
Why wait for the max size of a VPI command, not min size? (in the end the whole code is a bit sketchy since we might only have received half of a command until now, but that's not your fault)

rpls:
> Why wait for the max size of a VPI command, not min size? 

Is there a minimum size? I was looking at that code and the corresponding JTAG VPI code I found around the net as well (OpenOCD, other projects that use VPI, etc.). Exactly because of the name `MaxSize...`. And as far as I can tell, each package consists of the full size. And the code following that also *always* loads the full size (`readFully` will always wait for enough data to fill the entire buffer array... which is sized at `MaxSizeOfVpiCmd` bytes).

andreasWallner:
You are right, misremembered the code when the last review, sry

Dolu1990:
What if openocd send less than MaxSizeOfVpiCmd ? seems it is a dead lock.

```scala
          while(inputStream.available() < MaxSizeOfVpiCmd) {
            sleep(jtagClkPeriod * 200)
          }
```

Seems like the whole implementation (include before this patch) assume that once there is something in inputStream, there is enough for a whole packet. Meaning the following would be fine : 

```scala
          while(inputStream.available() > 0) {
            sleep(jtagClkPeriod * 200)
          }
```

?

andreasWallner:
@Dolu1990 "max size" is a bit of a misnomer it seems - from what I can tell (and what @rpls posted above) all VPI commands have the same length.
Since it's a TCP stream you can't rely on all bytes arriving at the same time, so there is a need to check for `> MaxSizeOfVpiCmd`

Dolu1990:
Ahhh ok ^^

Thanks !

PR #1351, opened 2024-03-12
Report when a timeout has happened for `BusSlaveFactory.readStreamBlockCycles`

KireinaHoro:
Improvement for `readStreamBlockCycles` to report to caller that a timeout has happened.  This is useful to trigger some action for a timeout (e.g. invalidate a cacheline, etc.)

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
+++ b/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
@@ -596,17 +596,22 @@ trait BusSlaveFactory extends Area{
    * @param that data to read over bus
    * @param address address to map at
    * @param blockCycles cycles to block read transaction before returning NACK
+   * @param timeout whether the read transaction timed out (returned NACK)
    * @tparam T type of stream payload
    */
-  def readStreamBlockCycles[T <: Data](that: Stream[T], address: BigInt, blockCycles: UInt): Unit = {
+  def readStreamBlockCycles[T <: Data](that: Stream[T], address: BigInt, blockCycles: UInt, timeout: Bool = null): Unit = {
     val counter = Counter(blockCycles.getWidth bits)
     val wordCount = (1 + that.payload.getBitsWidth - 1) / busDataWidth + 1
+    timeout != null generate { timeout := False }
     onReadPrimitive(SizeMapping(address, wordCount * wordAddressInc), haltSensitive = false, null) {
       counter.increment()
       when(counter.value < blockCycles && !that.valid) {
         readHalt()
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

What is the usage of readStreamBlockCycles in general, and also what is the usage of timeout ? I mean practical use case.

(I general i avoid this kind of interfaces)

KireinaHoro:
The practical use case is for a polling interface to reduce bus traffic when no data is present.  If a CPU core would busy poll a register when there's no data present (NACK), the interconnect will be spammed with useless NACK traffic.  `readStreamBlockCycles` would allow the device to block the transaction to a specified upper bound before returning the NACK, thus allowing for lower overhead for NACKs on the bus.

The timeout could be used for event counting (i.e. counting how many times a NACK was sent) in the simplest case.  In my project this is used to trigger a cacheline clean over the coherency protocol, so that the CPU can read again without hitting the NACK case that's cached.

Dolu1990:
> The practical use case is for a polling interface to reduce bus traffic when no data is present. If a CPU core would busy poll a register when there's no data present (NACK), the interconnect will be spammed with useless NACK traffic.

I would say the best would be the CPU doing pulling, with a few "nop" in between. The reason is : 
- some interconnect decoder will only allow a single slave to have inflight transactions at a given time. The reason is to preserve ordering between requests. So if you have a slave which hold something, it will block the decoder, even if you have some buffers in between
- A CPU which is waiting on some IO memory transaction to complet will not be capable to handle any interrupts for all that duration.

> In my project this is used to trigger a cacheline clean over the coherency protocol

hmm suggesting multiple masters. What kind of memory ordering do you have in your memory interconnect ?

KireinaHoro:
I would say that if this is suitable or not depends very much on the actual system.  The downsides you describe apply for maybe embedded micro-controllers, but does not form a concern in data-centers:

- server-class CPUs have stronger interconnects at the SoC level that certainly handles multiple in-flight transactions
- interrupts do not matter since CPU cores are dedicated to handling requests (e.g. RPC) and do not have other meaningful work to do

We are currently doing research about integrating scheduling decisions on the CPU into the NIC itself, so blocking the CPU core is certainly not a problem.

Our current experiment platform is the Enzian with a ThunderX-1 CPU.  They run aarch64 cores, so a weak memory ordering (not sure if this is what you're asking).

KireinaHoro:
Btw the PR also has a fix for a corner case (`that.valid` going high right when the counter expires).

Dolu1990:
Ahh right, i see, for this it make sense.
Thanks ^^

PR #1352, opened 2024-03-13
Proof of concept for isunknown usage

### LINKED ISSUE #1347: Checking for unknown values in simulation
Is there any current support for this? I couldn't find it. I was hoping for something that essentially generated `$isunknown`. 

The use case is wanting to assert in certain places in my simulation that such and such signal is fully defined.

If support for this does not exist, but would be accepted as a PR, I can make one. Would appreciate guidance in terms of where such an operator would live -- "Operator.formal.IsUnknown" makes sense to me but I'm not really clear if it fits in there and also don't know how to map it for VHDL.

jdavidberger:
Closes #1347 

This isn't ready to merge in this form it's mostly just for discussion. I didn't realize initially that it would require adding command line flags to enable SV functionality. 

Having to add in `.addSimulatorFlag("-g2012").addSimulatorFlag("-D__HAS_SYSTEM_VERILOG__")` is maybe bearable but makes this a pretty niche feature. IVerilog at least on my setup requires the -g2012 flag to support $isunknown. 

AFAICT verilog has no standard macro definitions for which version of the standard is being used -- which coming from a C/C++ background is bewildering and I'd love to be wrong on this. 

Here is the task list I imagine would need to be completed to make this mergable:
- The sim config would need some high level option for which standard of verilog to adhere to which dutifully enables the correct simulator flags based on the backend, and also some notion of whether that version supports system verilog or not would generate the other macro. 
- The IS_UNKNOWN macro would have to be moved somewhere sensible; and probably only emitted whenever the macro was used. 
- When the backend doesn't support the $isunknown directive -- non SV verilog and VHDL -- a warning should be emitted that says 'this will always be 0'. 

I still think it's worth doing but would understand if that wasn't the overall consensus. Being able to add in verification into components to assert they don't leak 'X' out in simulations is useful; particularly since all Mem's come up with that by default. But it is a lot of hoops to jump through to get there. 

# Checklist

- [ x] Unit tests were added
- [x ] API changes are or will be documented:

### DIFF
--- a/core/src/main/scala/spinal/core/BitVector.scala
+++ b/core/src/main/scala/spinal/core/BitVector.scala
@@ -515,5 +515,5 @@ abstract class BitVector extends BaseType with Widthable {
     cloneOf(this).setWidth(w).asInstanceOf[T]
   }
 
-  def isUnknown: Bool = wrapUnaryWithBool(new Operator.Formal.IsUnknown)
+  def isUnknown: Bool = wrapUnaryWithBool(new Operator.BitVector.IsUnknown)
 }
--- a/core/src/main/scala/spinal/core/BitVector.scala
+++ b/core/src/main/scala/spinal/core/BitVector.scala
@@ -514,4 +514,6 @@ abstract class BitVector extends BaseType with Widthable {
     val w = list.filter(!_.hasTag(tagAutoResize)).map(e => widthOf(e)).max
     cloneOf(this).setWidth(w).asInstanceOf[T]
   }
+
+  def isUnknown: Bool = wrapUnaryWithBool(new Operator.Formal.IsUnknown)
 }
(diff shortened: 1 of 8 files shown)


jdavidberger:
When I went looking to see where system verilog options were, I had missed it on the config mode enum. I have it set to key off of that now.

This should be fully backwards compatible with all existing uses of the API except for one case: If someone had something like:

```
SimConfig.withIVerilog.
      .withConfig(SpinalConfig(mode = SystemVerilog))
```

before these commits it would effectively ignore the `mode = SystemVerilog`. After these commits, it configures IVerilog for system verilog and generates sv files. I don't think this can actually break any existing code, and it would be a weird thing for peope to have had prior to this commit so I think the slight behavior change should be fine.

Dolu1990:
Thanks :)

PR #1356, opened 2024-03-14
BlackBox Generic update

jijingg:
1. BlackBox Generic add RawExpression for som special use  
   for example ,  we want pass an verilog expression ` {256{1'b1}}` or  `{2*4{1'b1}, 32'h323, 16'b001110}`  directly,  current string can't meet our requirement
```scala
addGenric("XXX_TABLE", RawExpression("{2*32*4{1‘b1}}")
```
```verilog
    .RT_PORT_NUM        (2             ),
    .XXX_TABLE          ({2*32*4{1'b1}}),
    .TGTID_W            (8             ),
```
   Of course, it also carries risks and does not check whether the expression is a valid and legal Verilog expression. Therefore, when you want to use it, you must be clear about what you are doing. I suggest keeping `RawExpression`  for those special needs

2. Optimize generate code style for paramter instance ， Especially alignment

image

4. show path when generate verilog/vhdl log
Convenient for users to see the location of generated code
```SHELL
[Progress] at 6.046 : Checks and transforms
[Progress] at 6.191 : Generate Verilog to ./out/xxgen/xxxgen20240314/
```

### DIFF
--- a/core/src/main/scala/spinal/core/BlackBox.scala
+++ b/core/src/main/scala/spinal/core/BlackBox.scala
@@ -44,7 +44,7 @@ class Generic {
 }
 
 case class GenericValue(e : Expression) extends SpinalTag
-case class RawExpression(e: String)
+case class VerilogValues(v: String)
 
 class BlackBoxImpl{
 //  def getRtl(mode : SpinalMode): Unit = {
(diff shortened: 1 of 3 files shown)


jijingg:
@Dolu1990  do you have any comment on `RawExpression`  ，This is not a very elegant option and lacks security verification. As a small-scale workround way, i think  it's  acceptable

Dolu1990:
Hi,

> This is not a very elegant option and lacks security verification. As a small-scale workround way, i think it's acceptable
i think that's ok

Maybe instead of RawExpression, it should be VerilogExpression ? that way it would at least catch that this isn't for VHDL generation

All good.

jijingg:
> Hi,
> 
> > This is not a very elegant option and lacks security verification. As a small-scale workround way, i think it's acceptable
> > i think that's ok
> 
> Maybe instead of RawExpression, it should be VerilogExpression ? that way it would at least catch that this isn't for VHDL generation

OK, rename `RawExpression` to `VerilogValues`,  Values may more accurency than expression,  if it's ok , i will merge after ci pass

PR #1359, opened 2024-03-17
Add CountLeadingZeroes utility function

fayalalebrun:
This adds a utility function used for counting the number of consecutive zero bits starting from the MSB.

Such a function is very useful for floating point and posit arithmetic. In fact, several ISA's have a dedicated count leading zeroes instruction, like `lzcnt` in x86.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/Utils.scala
+++ b/lib/src/main/scala/spinal/lib/Utils.scala
@@ -1488,3 +1488,26 @@ object Shift{
     logic | scrap.asBits.resized
   }
 }
+
+/** Counts the number of consecutive zero bits starting from the MSB. */
+object CountLeadingZeroes {
+  def apply(in: Bits): UInt = {
+    val amountWidth = log2Up(in.getWidth) + 1
+    // MSB to LSB
+    in.asBools.reverse
+      // Map to number of leading zeroes in segment and whether all bits in segment are zero
+      .map(x => ((~x).asUInt, ~x))
+      .reduceBalancedTree((left, right) => {
+        val sum = U(0, amountWidth bits)
+        // When left segment is all zero, we can add up the right segment's zeroes
+        when(left._2) {
+          sum := left._1.resize(amountWidth) + right._1.resize(amountWidth)
+        } otherwise {
+          sum := left._1.resize(amountWidth)
+        }
+
+        (sum, left._2 && right._2)
+      })
+      ._1
+  }
+}
--- a/tester/src/test/scala/spinal/tester/scalatest/SpinalSimLibTester.scala
+++ b/tester/src/test/scala/spinal/tester/scalatest/SpinalSimLibTester.scala
@@ -76,4 +76,24 @@ class SpinalSimLibTester extends SpinalAnyFunSuite {
         }
     }
   }
+
+  for (bitCount <- 1 to 5) {
+    test("CountLeadingZeroes " + bitCount) {
+      SimConfig.noOptimisation
+        .compile(new Component {
+          val input = in(Bits(bitCount bits))
+          val output = out(UInt(log2Up(bitCount) + 1 bits))
+          output := CountLeadingZeroes(input)
+        })
+        .doSim(seed = 42) { dut =>
+          for (i <- 0 until 1 << bitCount) {
+            val expected = Integer.numberOfLeadingZeros(i) - 32 + bitCount
+            dut.input #= i
+            sleep(1)
+            val out = dut.output.toInt
+            assert(out == expected, f"in: ${i.toBinaryString} bits: ${bitCount}")
+          }
+        }
+    }
+  }
 }


KireinaHoro:
+1 on this.  I have a implementation that I'm using internally here, maybe for comparison: https://github.com/KireinaHoro/enzian-pio-nic/blob/3c881a09a16f48d740908a341a41264adc30bc89/hw/spinal/pionic/package.scala#L17-L46 The implementation I have make use of recursive instantiation instead of the `reduceBalancedTree` function, which I have seen for the first time.  Would you mind taking a look and see if there's any fundamental difference?

On a different note, it would be useful to include counting trailing zeroes as well, implemented as a very simple wrapper around CLZ.

fayalalebrun:
@KireinaHoro I think both implementations will result in a balanced tree. However, your implementation is more efficient because it does not require addition and only uses the necessary amount of bits at every level. The only advantage of the function I wrote is the ability to accept odd widths. That shouldn't be too difficult to add to yours though.

Do you mind if I rewrite this based on your implementation? I will also add a count trailing zeroes function.

KireinaHoro:
Sure!  I have a comment in there citing the original stackoverflow answer; I simply transcribed the Verilog solution.  If you want to attribute, that should be the source.

fayalalebrun:
I have added my implementation based on yours. For supporting non-power of two's the best solution I could think of was to pad the input with zeroes and subtract the difference. This may sound bad at first, but due to the constant inputs, the synthesizer should be able to collapse all of the redundant multiplexers.

KireinaHoro:
Sounds legit, that'd be how I would do it...

Dolu1990:
Nice thanks ^^

Dolu1990:
I would be curious to see how it compare to something in the kind of : 

OHToUInt(OHMasking.firstV2(input.reverse))

dreamflyings:
> 我很好奇它与以下东西相比如何：
> 
> OHToUInt（OHMasking.firstV2（输入.反转））

Thank you 🙏 for providing this, it is a common practice.

Readon:
> I would be curious to see how it compare to something in the kind of :
> 
> OHToUInt(OHMasking.firstV2(input.reverse))

So should we revert this?

fayalalebrun:
@Readon I did the test and synthesized both with yosys and a 512 bit input. 

> I would be curious to see how it compare to something in the kind of :
> 
> OHToUInt(OHMasking.firstV2(input.reverse))

This approach resulted in:
```
ABC RESULTS:               NOT cells:      806
ABC RESULTS:               NOR cells:     1412
ABC RESULTS:              NAND cells:     1461
ABC RESULTS:        internal signals:     4287
```

While the approach in this MR resulted in:
```
ABC RESULTS:               NOT cells:      494
ABC RESULTS:              NAND cells:      749
ABC RESULTS:               NOR cells:     1079
ABC RESULTS:        internal signals:     1514
```

So roughly half resource usage using the approach here.

Looking at the code, I think each one of those two commands has to create a separate tree on 512 bit input. So that would explain the doubled resource usage.

Readon:
> @Readon I did the test and synthesized both with yosys and a 512 bit input.
> 
> > I would be curious to see how it compare to something in the kind of :
> > OHToUInt(OHMasking.firstV2(input.reverse))
> 
> This approach resulted in:
> 
> ```
> ABC RESULTS:               NOT cells:      806
> ABC RESULTS:               NOR cells:     1412
> ABC RESULTS:              NAND cells:     1461
> ABC RESULTS:        internal signals:     4287
> ```
> 
> While the approach in this MR resulted in:
> 
> ```
> ABC RESULTS:               NOT cells:      494
> ABC RESULTS:              NAND cells:      749
> ABC RESULTS:               NOR cells:     1079
> ABC RESULTS:        internal signals:     1514
> ```
> 
> So roughly half resource usage using the approach here.
> 
> Looking at the code, I think each one of those two commands has to create a separate tree on 512 bit input. So that would explain the doubled resource usage.

oh, nice, that is reasonable.

Readon:
This also reminds me of something, this is actually an encoder here, so shouldn't we provide the common encoder and decoder circuits first?

At least, explicit specified their function by name.

Readon:
So this CountLeadingZeroes ==== Decoder, and UIntToOh ==== Encoder. 
The common usage might be 3-8 encoder and 8-3 decoder in textbook.

Dolu1990:
@Readon Do you mean changing the name of things ? xxxDecoder instead of CountLeadingZeroes ?

Readon:
> @Readon Do you mean changing the name of things ? xxxDecoder instead of CountLeadingZeroes ?

emm， I mean add the xxxDecoder and xxxEncoder is good. The CountLeadingZero like function can reference on that.😀

Dolu1990:
Ahhh sure ^^
As long as the xxx name are good.

Readon:
They only use different naming policy, one based on technology, one based on usage.

likewise:
I think I copied a solution from Gitter once, not sure how it compares.

https://github.com/likewise/SpinalCorundum/blob/main/src/main/scala/corundum/LeadingZeroes.scala

PR #1360, opened 2024-03-17
Rename FlowByCCToggle, deprecating the old name

andreasWallner:
FlowByCCToggle is different than other CDC primitives in that it's unsafe to use in a general use-case. It must be used as a component in a CDC structure, or only in certain circumstances.

Rename to indicate this, add a comment to explain.

Checking the current uses in lib:
- Apb3CCToggle is fine, since both uses are part of a handshake, limiting datarate
- Jtag uses both are limited by the protocol to wait for a response to the instruction

# Checklist

No documentation update needed since FlowCCByToggle is undocumented atm.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Flow.scala
+++ b/lib/src/main/scala/spinal/lib/Flow.scala
@@ -189,19 +189,54 @@ object RegFlow{
   }
 }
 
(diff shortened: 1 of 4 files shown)


Dolu1990:
Hi,

What's about FlowCCUsafeByToggle ? (to preserve the FlowCC prefix)

andreasWallner:
Also fine for me

PR #1361, opened 2024-03-17
Add messages to asserts in Utils

andreasWallner:
Prevents to user from seeing just an "assertion failed" message and having to dig into Utils


Dolu1990:
Thanks ^^

PR #1363, opened 2024-03-18
Fix simulator flags not passed to elaboration

fayalalebrun:
For the RTL sources I was using, ghdl required the `-frelaxed` option for analysis and elaboration. However, setting this flag via `addSimulatorFlag` only resulted in the flag being used for analysis.

Not sure if this was done on purpose, so I will ping those I saw made related contributions: @likewise @wswslzp @jiegec

### DIFF
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -277,9 +277,12 @@ object SpinalGhdlBackend {
 
   def apply[T <: Component](config: SpinalGhdlBackendConfig[T]) : Backend = {
     val vconfig = new GhdlBackendConfig()
-    vconfig.analyzeFlags = config.simulatorFlags.mkString(" ")
-    if (config.timePrecision != null) {
-      vconfig.elaborationFlags = s"--time-resolution=${config.timePrecision.decompose._2}"
+    val flagsConcat = config.simulatorFlags.mkString(" ")
+    vconfig.analyzeFlags = flagsConcat
+    vconfig.elaborationFlags = flagsConcat + {
+      if (config.timePrecision != null) {
+        s" --time-resolution=${config.timePrecision.decompose._2}"
+      } else ""
     }
     vconfig.runFlags = config.runFlags.mkString(" ")
     vconfig.logSimProcess = config.enableLogging


Dolu1990:
Hi,

hmm maybe we would need a more precise way to define flags, as different simulators may have different stage which needs different stages.

Maybe, the simulator flags would need to be specify more as : 

```scala
simConfig.addSimulatorFlag(simAnalyse -> "--miaou")
simConfig.addSimulatorFlag(simElaborate -> "--wuff")
```
Which could then be shortened via some sugger as :
```scala
simConfig.addAnalyseFlag("--miaou")
simConfig.addElaborateFlag("--wuff")
```

?

fayalalebrun:
Sounds good, but then shouldn't the flag structure be specific to each simulator, since for example I imagine elaborate flags don't have much meaning outside of ghdl? An idea could be that the flags are optional arguments to the `withGhdl` call.

Dolu1990:
> Sounds good, but then shouldn't the flag structure be specific to each simulator, since for example I imagine elaborate flags don't have much meaning outside of ghdl? An idea could be that the flags are optional arguments to the withGhdl call.

Yes, i think you are right. That would be better.

fayalalebrun:
Looking at the code, it seems like VCS also makes use of a special object to holds its flags, and from what I can tell does not use the flags given to `SimConfig` via `addSimulatorFlag` and `addRunFlag`.

Is it okay to do it in the same way, breaking compatibility?

Dolu1990:
ahhh i see. I think what should have been done overall, is to have in the SpinalSimConfig an : 

```scala
var _ghdlConfig : GhdlConfig = new GhdlConfig

def ghdlConfig = _backendConfig.asInstanceOf[GhdlConfig]
```

```scala
class GhdlConfig{
   var elaborationFlags ....
   def setElaborationFlag ...
}
```

Then this can be used freely


```scala
val cfg = SimConfig()
cfg.withGhdl()
cfg.ghdlConfig.setElaborationFlag( ... )
cfg.doSim(...)
```


> Is it okay to do it in the same way, breaking compatibility?

Hmmm no it isn't, it's fine if we have two way to define overlapping things.
Not breaking user space is important.

Dolu1990:
Thanks ^^

Let's me know when you are ready for merge. (look good to me)

fayalalebrun:
Yes it's ready to go. I think it's a bit ugly but at least it is consistent with the current interface.

Just out of curiosity, what is the difference between GhdlBackendConfig and SpinalGhdlBackendConfig? Why are two needed?

Dolu1990:
> GhdlBackendConfig and SpinalGhdlBackendConfig?

GhdlBackendConfig is is from spinal.sim, and doesn't know anything about SpinalHDL.core itself, very abstract.
While SpinalGhdlBackendConfig is a bigger set of configs, which knows about SpinalHDL.core

Which make quite a bit of redondencies, but ensure decoupeling.

> Yes it's ready to go. I think it's a bit ugly but at least it is consistent with the current interface.

Could add a layer of paint on the top :)

PR #1364, opened 2024-03-19
add system verilog interface support

yportne13:
This pr is still work in progress. related discussion: #1323 . This PR enhances the Interface type based on Bundle, and when the code generation setting is set to System Verilog and the switch for generating interface is turned on, this type of signal will be generated as an interface form. If the switch is not turned on, it is equivalent to a Bundle. The remaining issues are as follows:

* ~~Interface declaration width: If a signal in an Interface is not used, its width will be set to -1 in the PhaseInferWidth stage, which will result in a `[-2:0]` signal being generated when the Interface declaration is created. In the current SpinalHDL code generation logic, if a signal in a Bundle is not used, this signal will eventually be pruned.~~
* name: Currently, in order to ensure the quality of generated code, it is required that an Interface can correctly execute the `getName` method. However, there are some unexpected situations where it becomes impossible to execute `getName`, such as `Vec[Interface].apply(UInt)`, even if all Interfaces inside `Vec` have been manually named, the result after applying `apply` will still lose its name.
* setDefinitionName: different from Component, an Interface can be Cloned. The setDefinitionName method should occur before all Clone behaviors; otherwise, the signals generated in the code from the Clone will be a different Interface, which is not expected. However, currently, there is no way to enforce this constraint.
* ~~assignment: Currently, all assignment behaviors are performed in a spread manner. In many cases, the generated code will be quite ugly, such as connecting the same interface signal of two sub-modules together. If done manually, you would only need to instantiate an interface in the current module and then fill in the two sub-modules. However, the currently generated code is that for each sub-module's interface signal, a separate wrapper is generated, and then all signals of the wrapper are unfolded and connected one by one.~~
* ~~parameter: should be nearly same as `addGeneric` in `BlackBox`~~
* ~~definitionName: if some Interfaces have same definitionName, but there is some difference in those Interfaces like port width, they will be automatically rename to `${origin_name}_1`, `${origin_name}_1_1` and so on. But it should be `${origin_name}_{idx}`~~
* ~~by now, in interface, only BaseType is allowed. Consider add the support for Interface in Interface~~


yportne13:
a small demo:
```scala

import spinal.core._
import spinal.lib._

case class MyInt(width: Int) extends Interface with IMasterSlave {
  val a = Bits(width bits)
  val b = Bool()
  val c = SInt(8 bits)

  override def asMaster = mst
  @modport
  def mst = {
    out(a, b, c)
  }

  @modport
  def slv = {
    in(a, b, c)
  }

}

case class StreamIF[T <: Data](payloadType :  HardType[T]) extends Interface with IMasterSlave with DataCarrier[T] {
  val valid = Bool()
  val payload = payloadType()
  val ready = Bool()

  override def asMaster = mst
  @modport
  def mst = {
    out(valid, payload)
    in(ready)
  }

  @modport
  def slv = {
    in(valid, payload)
    out(ready)
  }

  def map[U <: Data](f: T => U): StreamIF[U] = {
    val new_payload = f(this.payload)
    val ht: HardType[U] = HardType(new_payload)
    val that: StreamIF[U] = StreamIF(ht)
    that.setName(s"${this.getName()}_map")
    that.valid := this.valid
    that.payload := new_payload
    this.ready := that.ready
    that
  }

  def <<(that: StreamIF[T]) = {
    this.valid := that.valid
    this.payload := that.payload
    that.ready := this.ready
  }

  def >>(that: StreamIF[T]) = {
    that << this
  }

  def fire = valid && ready
  def freeRun() = {
    ready := True
    this
  }
}

class TestInterface extends Component {
  val io = new Bundle {
    val x = slave(MyInt(8))
    val y = master(MyInt(9))
    val axi_s = slave(StreamIF(UInt(8 bits)))
    val axi_m = master(StreamIF(UInt(8 bits)))
    val t = out Bits(1 bits)
  }

  val i1 = new TestInterfaceSub
  i1.io.x := io.x
  io.t := B(i1.io.t)
  val i2 = new TestInterfaceSub
  i2.io.x.a := io.x.a
  i2.io.x.b := !io.x.b
  i2.io.x.c := io.x.c + 1
  val i3 = new TestAxi
  i3.io.axi_s << io.axi_s.map(_ + 3)
  io.axi_m << i3.io.axi_m.map(x => x + 1)

  val r = Reg(MyInt(9))

  r.a := io.x.a.resized
  r.b := !io.x.b
  r.c := io.x.c + 2

  io.y := r

}

class TestInterfaceSub extends Component {
  val io = new Bundle {
    val x = in(MyInt(8))
    val t = out Bool()
  }

  val a = new ISub
  a.io.x := io.x
  val useless = Bool()
  useless := a.io.y && True
  io.t := a.io.z

}

class TestAxi extends Component {
  val io = new Bundle {
    val axi_s = slave(StreamIF(UInt(8 bits)))
    val axi_m = master(StreamIF(UInt(8 bits)))
  }

  io.axi_s >> io.axi_m

}

class ISub extends BlackBox {
  val io = new Bundle {
    val x = in(MyInt(8))
    val y = out Bool()
    val z = out Bool()
    val p = out Bool()
  }

  //io.y := io.x

}

object TestInterfaceTop extends App {
  SpinalConfig().generateSystemVerilog(new TestInterface)
  SpinalConfig(svInterface = true).generateSystemVerilog((new TestInterface).setDefinitionName("TestInterfaceActive"))
  SpinalConfig().generateVerilog(new TestInterface)
}

```

Dolu1990:
Hi,

Thanks for the PR.
I tried the demo. I think there is one fondamental things we should think about is : 

Do we want SpinalHDL to emit systemVerilog interface definition. i'm talking only about the definition, not the usage. For instance : 

```systemverilog
interface MyInt_1 () ;

  logic  [8:0]    a ;
  logic           b ;
  logic  [7:0]    c ;

  modport mst (
    output  a,
    output  b,
    output  c
  );

  modport slv (
    input   a,
    input   b,
    input   c
  );

endinterface
```

There is a few reason i can see against it : 
- Defining new interface definition in the generated code mean it isn't realy interpolable with other .sv (human written or generated). If for instance i want to integrate some generated SV in a handwritten toplevel, it create the issue that SpinalHDL will have its own interface definitions. If for instance i want to integrate some handwritten SV in SpinalHDL as black box there is a similar issue aswell

That is why i think the best would be that SpinalHDL should not generate interface definition but instead be able to refer to human defined interface definition from some external .sv. In other words, interface definition would a bit like blackboxes.  

Or maybe i missed something ?

Maybe we can organize a call to talk about all of this ?

Dolu1990:
@jijingg 

> no no. Let's forget about the scenario of mixed development with traditional SV teams. If I only want to develop using SpinalHDL, the purpose of using Interface is only to reduce code volume

Ahhhhhh, got it, i wasn't looking at it from that perspective ^^

hmmm i have to think about it.

jijingg:
> In other words, interface definition would a bit like blackboxes.

This statement is half right， if we want'  use  human defined interface definition from some external .sv.  right  this interface is like BlackBox


> * Defining new interface definition in the generated code mean it isn't realy interpolable with other .sv (human written or generated). If for instance i want to integrate some generated SV in a handwritten toplevel, it create the issue that SpinalHDL will have its own interface definitions. If for instance i want to integrate some handwritten SV in SpinalHDL as black box there is a similar issue aswell

Recall that if we wanted to use an existing Verilog code, we would declare a BlackBox to replicate its signal name one-to-one
So now if we also want to use an existing SV code, we should declare a Bundle with Inteface one by one and then link it with our internal Interface， If the external interface is the same as the interface I use in Spinal, then the Sv interface we generate is essentially the same as the external handwritten interface. You can use either.  

We can refer to this example https://github.com/SpinalHDL/SpinalHDL/discussions/1323#discussioncomment-8768749

`MY_APB_IF`  is exists handwritten SV,  and  the `Apb4` is spinalhdl's Bundle , If I want to use it, then I need to declare it and hand over the interconnection between these two to the user. This way, Blackbox's Inteface can also be easily resolved。

Of course, SpinalHDL generate `MY_APB_IF` is not necessary, its name is exactly the same as the existing one
```scala
class MyBlackBox extends BlackBox{
  val x   = in Bits(32 bit)
  val y   = out Bits(32 bit)
  val cfg = master(MY_APB_IF(32, 32))
}

class Top extends Component{
  val a = in Bits(32 bit)
  val b = out Bits(32 bit)
  val spinal_apb = slave(Apb4(Apb4Config(32, 32)))

  val ut = new MyBlackBox

  ut.x    := a
  b       := ut.y
  ut.cfg  >< spinal_apb                          //way1: ut.cfg implement a method connect with spinalhdl's Apb4
//  twoDiffrenceAPBconnect(ut.cfg,  spinal_apb)  //way2: implement a function connect two different APB
}
```

core/src/main/scala/spinal/core/Interface.scala
```diff
@@ -0,0 +1,148 @@
...
+      case i: Int       => genericElements += Tuple3(name, i, default)
+      case i: BigInt if i <= Integer.MAX_VALUE && i >= Integer.MIN_VALUE => genericElements += Tuple3(name, i.toInt, default)
+      case d: Double        => genericElements += Tuple3(name, d, default)
+      case boolean: Boolean => genericElements += Tuple3(name, boolean, default)
+      case t: TimeNumber    => genericElements += Tuple3(name, t, default)
+    }
+    name
+  }
+  def addParameter(name: String, that: Any, default: String = null): String = addGeneric(name, that, default)
+  val widthGeneric = scala.collection.mutable.LinkedHashMap[Data, String]()
```
Dolu1990, line 56: Could be BitVector instead of Data right ?

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -54,6 +54,7 @@ class ComponentEmitterVerilog(
   val definitionAttributes  = new StringBuilder()
   val beginModule = new StringBuilder()
   val declarations = new StringBuilder()
+  var declaredInterface: Set[SVIF] = Set()
```
Dolu1990, line 57: I would say to avoid imutable Set for performance reasons, and instead go for either mutable.HashSet if you know you will never iterate on the set.
But if you iterate on the set it is very important to use mutable.LinkedHashSet in order to preserve the insertion order, allowing to remain deterministic.

Dolu1990:
I was looking how the verilog generation find the name for signals inside interface, seems like currently, each BaseType get its name set will the full path. Ex . "io_axi_s.valid"

I'm not saying it is a good or bad solution, just for you to know, in the ComponentEmitterVerilog, there is the "wrappedExpressionToName : mutable.HashMap[Expression, String]" (from ComponentEmitter) which allows to create aliases for the generation.

So "wrappedExpressionToName(myBaseType) = "io_axi_s.valid" would also work :)

Dolu1990:
Hi ^^

One question would be : 
Would it be faisable that we would not have to use to have a dedicated SVIF class, and instead, directly generate all bundles as interfaces ?
There may be ways do make do it.

Or you see some blocking things with it ?
Just asking :)

yportne13:
> Hi ^^
> 
> One question would be : Would it be faisable that we would not have to use to have a dedicated SVIF class, and instead, directly generate all bundles as interfaces ? There may be ways do make do it.
> 
> Or you see some blocking things with it ? Just asking :)

1. for me, I don't want to make all the bundle into interface. In SpinalHDL, I use much more bundle than I use interface in sv.
2. to support interface in interface, trace all the way up to find the root interface is needed. If regard bundle as interface, then you may get a lot of `io` interface.
3. by now there is still a little issues about Vec and Bundle in interface, there is some special opera on those types
4. a lot of extra things like paremeter, definitionName, modport, may not suitable to add into Bundle

jijingg:
> Would it be faisable that we would not have to use to have a dedicated SVIF class, and instead, directly generate all bundles as interfaces ?

> for me, I don't want to make all the bundle into interface. In SpinalHDL, I use much more bundle than I use interface in sv.

I understand what you two mean, but the name of `SVIF` may need to be discussed.
Because it inherits from `Bundle`, it needs a name that is similar to `Bundle` but carries the meaning of `SV interface`
maybe  `IntfBundle`  or `Interface`  is this ok?

Readon:
How about specify the Bundle need to be initiated to SV interface in config like
```
SpinalSystemVerilog(SVInterface = [ABundle, BBundle])
```
so that class ABundle and BBundle and their subclass would be implemented by interface. 

if SVInterface is empty than all Bundle remains unchanged, the same with previous generation.
also if all Bundle is necessary specify SVInterface = [ Bundle ] ?

Dolu1990:
So maybe for now keeping SVIF as its own class instead of merging it into Bundle allows to keep things clearly decoupled.
Eventualy later in the future it can be merge in X or Y ways, but this shouldn't be a priority for now. (step by step)

So, in order to move this forward with the issues.
Could you provide a self contained code example with each issues you know about which remains ?
That would provide a clear target / view on what we could help :)
Maybe adding them directly in the Scalatest regressions.

yportne13:
Currently, the remaining issues are not very important(at least for my use case) and beyond my ability to resolve. The interface is already sufficient for my needs at the moment

Dolu1990:
Hi,

In this issue description, there is : 

```
definitionName: if some Interfaces have same definitionName, but there is some difference in those Interfaces like port width, they will be automatically rename to ${origin_name}_1, ${origin_name}_1_1 and so on. But it should be ${origin_name}_{idx}
```

Maybe [improve interface definition name](https://github.com/SpinalHDL/SpinalHDL/pull/1364/commits/f539637a8f0a08e261c1ca6cfeee8ab5080cde5b)
[f539637](https://github.com/SpinalHDL/SpinalHDL/pull/1364/commits/f539637a8f0a08e261c1ca6cfeee8ab5080cde5b) solved it ?

yportne13:
> Hi,
> 
> In this issue description, there is :
> 
> ```
> definitionName: if some Interfaces have same definitionName, but there is some difference in those Interfaces like port width, they will be automatically rename to ${origin_name}_1, ${origin_name}_1_1 and so on. But it should be ${origin_name}_{idx}
> ```
> 
> Maybe [improve interface definition name](https://github.com/SpinalHDL/SpinalHDL/pull/1364/commits/f539637a8f0a08e261c1ca6cfeee8ab5080cde5b) [f539637](https://github.com/SpinalHDL/SpinalHDL/pull/1364/commits/f539637a8f0a08e261c1ca6cfeee8ab5080cde5b) solved it ?

sorry for late reply, yes! it solved it! I'll mark this line as solved

jijingg:
hi @Dolu1990 ,  I changed the name from `SVIF` to `Interface`. Currently, I believe the functionality of Interface that we have developed has achieved the goal of supporting SystemVerilog Interfaces. Furthermore, Interface is independent of Bundle and does not introduce any contamination to existing code. Even if `Interface `replaces `Bundle`, the generated code remains exactly the same in Verilog mode. Additionally, we have conducted some validation and trials in my project, and I think it meets the conditions for integration at present.

Dolu1990:
Hi @jijingg @yportne13 

Thanks :)

> Furthermore, Interface is independent of Bundle and does not introduce any contamination to existing code

That is a nice thing. Keep in mind that from my perspective, getting big PR is always scary ^^
So yes, the less "contamination" the better from a maintenability perspective.

jijingg:
> How about specify the Bundle need to be initiated to SV interface in config like
> 
> ```
> SpinalSystemVerilog(SVInterface = [ABundle, BBundle])
> ```
> 
> so that class ABundle and BBundle and their subclass would be implemented by interface.
> 
> if SVInterface is empty than all Bundle remains unchanged, the same with previous generation. also if all Bundle is necessary specify SVInterface = [ Bundle ] ?

hi @Readon ，This is also a good approach. However, when there are a lot of Interfaces within a subsystem, it may appear too lengthy. Additionally  The way to print the same Interface in different locations may vary. Therefore, we added a method called `notSVIF `to define when to print the Interface and when not to.  
So I still hope that users can define which Bundles need to be instantiated as Interfaces, and use "notSVIF" to mark where they don't need to be expanded.

PR #1367, opened 2024-03-19
Fix corner case of `readStreamBlockCycles`

KireinaHoro:
We should only set `respReady` when we have issued the read; otherwise, we could issue a spurious NACK due to the counter overflowing in the previous iteration.


andreasWallner:
Thanks!

PR #1369, opened 2024-03-20
Constraint writer for Xilinx Vivado (reopened: CDC rework)

KireinaHoro:
Redoing #1306.  I think I accidentally closed this due to some branch dancing errors, now the PR head branch should be safe.

I also dropped from the last state of the PR stuff related to `Flow`: `m2sPipe` and `FlowCCByToggle` (cf. #1360), since these are not inherently safe CDC primitives.

@andreasWallner sorry to bother you again but can you review it (or just a brief look, since most of the stuff has been fixed already)?  Thanks a lot!

### DIFF
--- a/core/src/main/scala/spinal/core/Trait.scala
+++ b/core/src/main/scala/spinal/core/Trait.scala
@@ -810,7 +810,15 @@ object noLatchCheck                  extends SpinalTag
 object noBackendCombMerge            extends SpinalTag
 object crossClockDomain              extends SpinalTag{ override def moveToSyncNode = true }
 object crossClockBuffer              extends SpinalTag{ override def moveToSyncNode = true }
-case class crossClockFalsePath(source: Option[BaseType] = None, destIsReset: Boolean = false) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
+
+sealed trait TimingEndpointType
+object TimingEndpointType {
+  case object DATA extends TimingEndpointType
+  case object RESET extends TimingEndpointType
+  case object CLOCK_EN extends TimingEndpointType
+}
+
+case class crossClockFalsePath(source: Option[BaseType] = None, destType: TimingEndpointType = TimingEndpointType.DATA) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
 case class crossClockMaxDelay(cycles: Int, useTargetClock: Boolean) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
 object randomBoot                    extends SpinalTag{ override def moveToSyncNode = true }
 object tagAutoResize                 extends SpinalTag{ override def duplicative = true }
--- a/core/src/main/scala/spinal/core/Trait.scala
+++ b/core/src/main/scala/spinal/core/Trait.scala
@@ -810,6 +810,8 @@ object noLatchCheck                  extends SpinalTag
 object noBackendCombMerge            extends SpinalTag
 object crossClockDomain              extends SpinalTag{ override def moveToSyncNode = true }
 object crossClockBuffer              extends SpinalTag{ override def moveToSyncNode = true }
+case class crossClockFalsePath(source: Option[BaseType] = None, destIsReset: Boolean = false) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
+case class crossClockMaxDelay(cycles: Int, useTargetClock: Boolean) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
 object randomBoot                    extends SpinalTag{ override def moveToSyncNode = true }
 object tagAutoResize                 extends SpinalTag{ override def duplicative = true }
 object tagTruncated                  extends SpinalTag{
--- a/lib/src/main/scala/spinal/lib/CrossClock.scala
+++ b/lib/src/main/scala/spinal/lib/CrossClock.scala
@@ -135,7 +135,7 @@ object ResetCtrl{
     )
 
     val solvedOutputPolarity = if(outputPolarity == null) clockDomain.config.resetActiveLevel else outputPolarity
-    val falsePathAttrs = List(crossClockFalsePath(Some(input), destIsReset = true))
+    val falsePathAttrs = List(crossClockFalsePath(Some(input), destType = TimingEndpointType.RESET))
     samplerCD(BufferCC(
       input       = (if(solvedOutputPolarity == HIGH) False else True) ^ inputSync,
       init        = if(solvedOutputPolarity == HIGH) True  else False,
--- a/lib/src/main/scala/spinal/lib/CrossClock.scala
+++ b/lib/src/main/scala/spinal/lib/CrossClock.scala
@@ -6,9 +6,10 @@ import scala.collection.Seq
 import scala.collection.mutable.ArrayBuffer
 
 object BufferCC {
-  def apply[T <: Data](input: T, init: => T = null, bufferDepth: Option[Int] = None, randBoot : Boolean = false): T = {
-    val c = new BufferCC(input, init, bufferDepth, randBoot)
+  def apply[T <: Data](input: T, init: => T = null, bufferDepth: Option[Int] = None, randBoot : Boolean = false, inputAttributes: Seq[SpinalTag] = List(), allBufAttributes: Seq[SpinalTag] = List()): T = {
+    val c = new BufferCC(input, init, bufferDepth, randBoot, inputAttributes, allBufAttributes)
     c.setCompositeName(input, "buffercc", true)
+    // keep hierarchy for timing constraint generation
     c.io.dataIn := input
 
     val ret = cloneOf(c.io.dataOut)
@@ -49,7 +50,7 @@ object BufferCC {
   }
 }
 
-class BufferCC[T <: Data](val dataType: T, init :  => T, val bufferDepth: Option[Int], val  randBoot : Boolean = false) extends Component {
+class BufferCC[T <: Data](val dataType: T, init :  => T, val bufferDepth: Option[Int], val  randBoot : Boolean = false, inputAttributes: Seq[SpinalTag] = List(), allBufAttributes: Seq[SpinalTag] = List()) extends Component {
   def getInit() : T = init
   val finalBufferDepth = BufferCC.defaultDepthOptioned(ClockDomain.current, bufferDepth)
   assert(finalBufferDepth >= 1)
@@ -64,12 +65,16 @@ class BufferCC[T <: Data](val dataType: T, init :  => T, val bufferDepth: Option
 
   buffers(0) := io.dataIn
   buffers(0).addTag(crossClockDomain)
+  inputAttributes.foreach(buffers(0).addTag)
   for (i <- 1 until finalBufferDepth) {
     buffers(i) := buffers(i - 1)
     buffers(i).addTag(crossClockBuffer)
   }
+  buffers.map(b => allBufAttributes.foreach(b.addTag))
 
   io.dataOut := buffers.last
+
+  addAttribute("keep_hierarchy", "TRUE")
 }
 
 
@@ -97,7 +102,7 @@ class PulseCCByToggle(clockIn: ClockDomain, clockOut: ClockDomain, withOutputBuf
 
   val finalOutputClock = clockOut.withOptionalBufferedResetFrom(withOutputBufferedReset)(clockIn)
   val outArea = finalOutputClock on new Area {
-    val target = BufferCC(inArea.target, False)
+    val target = BufferCC(inArea.target, False, inputAttributes = List(crossClockFalsePath()))
 
     io.pulseOut := target.edge(False)
   }
@@ -130,10 +135,12 @@ object ResetCtrl{
     )
 
     val solvedOutputPolarity = if(outputPolarity == null) clockDomain.config.resetActiveLevel else outputPolarity
+    val falsePathAttrs = List(crossClockFalsePath(Some(input), destIsReset = true))
     samplerCD(BufferCC(
       input       = (if(solvedOutputPolarity == HIGH) False else True) ^ inputSync,
-      init        = (if(solvedOutputPolarity == HIGH) True  else False),
-      bufferDepth = bufferDepth)
+      init        = if(solvedOutputPolarity == HIGH) True  else False,
+      bufferDepth = bufferDepth,
+      allBufAttributes = falsePathAttrs)
     )
   }
 
@@ -164,7 +171,7 @@ object ResetCtrl{
         inputPolarity = resetCd.config.resetActiveLevel,
         outputPolarity = clockCd.config.resetActiveLevel,
         bufferDepth = bufferDepth
-      ).setCompositeName(resetCd.reset, "syncronized", true)
+      ).setCompositeName(resetCd.reset, "synchronized", true)
     )
   }
 }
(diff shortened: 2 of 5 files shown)


lib/src/main/scala/spinal/lib/CrossClock.scala
```diff
@@ -6,9 +6,10 @@ import scala.collection.Seq
 import scala.collection.mutable.ArrayBuffer
 
 object BufferCC {
-  def apply[T <: Data](input: T, init: => T = null, bufferDepth: Option[Int] = None, randBoot : Boolean = false): T = {
-    val c = new BufferCC(input, init, bufferDepth, randBoot)
+  def apply[T <: Data](input: T, init: => T = null, bufferDepth: Option[Int] = None, randBoot : Boolean = false, inputAttributes: Seq[SpinalTag] = List(), allBufAttributes: Seq[SpinalTag] = List()): T = {
```
andreasWallner, line 9: Do we even need `allBufAttributes`? The only use-case I see is for false paths - and for those a single one is sufficient to break the timing arc that we want to prevent.

lib/src/main/scala/spinal/lib/eda/xilinx/VivadoConstraintWriter.scala
```diff
@@ -0,0 +1,171 @@
...
+
+import java.io.{File, PrintWriter, Writer}
+import scala.collection.mutable
+import scala.language.postfixOps
+
+object VivadoConstraintWriter {
+  def apply[T <: Component](
+                             report: SpinalReport[T],
+                             filename: String = null,
+                             trustSpinalTimings: Boolean = false
```
andreasWallner, line 16: remove this since it's currently not implemented?

andreasWallner (changes requested):
Looks quite good to me

lib/src/main/scala/spinal/lib/eda/xilinx/VivadoConstraintWriter.scala
KireinaHoro: done

lib/src/main/scala/spinal/lib/CrossClock.scala
KireinaHoro: Location used: https://github.com/SpinalHDL/SpinalHDL/pull/1369/files#diff-5e6bbb8e52385971285b669e6d58869c6b0c3f3e9fb488b8917e2a30931879d3R138-R143

andreasWallner: You are right, I was forgetting that we don't create any `-through` constraints, but `-from`/`-to`.

andreasWallner (approved):

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -362,7 +362,10 @@ class Stream[T <: Data](val payloadType :  HardType[T]) extends Bundle with IMas
     val rValid = RegNextWhen(self.valid, self.ready) init(False)
     val rData = RegNextWhen(self.payload, if(holdPayload) self.fire else self.ready)
 
-    if (crossClockData) rData.addTag(crossClockDomain)
+    if (crossClockData) {
+      rData.addTag(crossClockDomain)
+      rData.addTag(new crossClockMaxDelay(2, true))
```
Dolu1990, line 367: So m2sPipe crossClockData=true is used in the context where you already have some "arbitration" to tell you when some data cross clock data are stable, and you want to sample them at a safe moment. (small area cross clock)

I would set it to 1 to be safe, as the design around it may assume so. as it cost "nothing"

core/src/main/scala/spinal/core/Trait.scala
```diff
@@ -810,6 +810,8 @@ object noLatchCheck                  extends SpinalTag
 object noBackendCombMerge            extends SpinalTag
 object crossClockDomain              extends SpinalTag{ override def moveToSyncNode = true }
 object crossClockBuffer              extends SpinalTag{ override def moveToSyncNode = true }
+case class crossClockFalsePath(source: Option[BaseType] = None, destIsReset: Boolean = false) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
+case class crossClockMaxDelay(cycles: Int, useTargetClock: Boolean) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
```
Dolu1990, line 814: I see 3 cases here : 
- Using target clock (useTargetClock=true)
- Using source clock (?)
- Using the fastest of both clock to avoid race condition (?)


Also, maybe we should see it more named as a "multiCyclePath" ?

andreasWallner: I don't really see the "multiCyclePath" here: that's usually used for a signal path staying in its own clock domain, but allowing more then 1 cycle propagation delay. 
Here we are talking about a max. propagation delay (name is taken from the STA tools)

Dolu1990: Ahh good then :)

core/src/main/scala/spinal/core/Trait.scala
```diff
@@ -810,6 +810,8 @@ object noLatchCheck                  extends SpinalTag
 object noBackendCombMerge            extends SpinalTag
 object crossClockDomain              extends SpinalTag{ override def moveToSyncNode = true }
 object crossClockBuffer              extends SpinalTag{ override def moveToSyncNode = true }
+case class crossClockFalsePath(source: Option[BaseType] = None, destIsReset: Boolean = false) extends SpinalTag { override def allowMultipleInstance: Boolean = false }
```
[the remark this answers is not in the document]
KireinaHoro: Or should we have a sealed trait for the destination type instead?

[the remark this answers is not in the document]
Dolu1990: Ahhhh yes, that is right, like a enum, yes yes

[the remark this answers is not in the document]
KireinaHoro: fixed

lib/src/main/scala/spinal/lib/Stream.scala
KireinaHoro: fixed

core/src/main/scala/spinal/core/Trait.scala
KireinaHoro: Yeah as of now we can specify to use either the source or dest clock, but not the fastest.  Should this be done in Spinal (look at the CD and find out which one is faster), or in Tcl?

Dolu1990: Hmmm, i would say, keeping it in SpinalHDL is fine

KireinaHoro: Then we can simply ask the caller to determine the appropriate clock domain to use.  Do we actually have a case where we need to use the fastest?

Dolu1990: As long as it isn't to the user to figure it out. I think that is ok.
But the one thing is that not everybody provide clock informations to SpinalHDL, so if the TCL generation isn't used, it should be fine not have clock domain frequancy specified

KireinaHoro: Yeah I think as is it's good then?  It's an attribute pushed by the CDC library

Dolu1990: What is good ? 

> Do we actually have a case where we need to use the fastest?

Hmmm i'm not sure anymore.
- For a gray counter, you want the source clock for constraint.
- For sampling some data when you know it is stable, you want the destination clock fo constraint

But i may miss some special cases ?

andreasWallner: I'd somewhat argue against binding this to SpinalHDL timings here:
- as you already said, often we want source/destination clock for the constraint - independent of their speed
- in such cases we shouldn't require timings to be available - especially for the use-case of building reusable components
- if we really come across another use-case where we need "fastest" we can still add it

core/src/main/scala/spinal/core/Trait.scala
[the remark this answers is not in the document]
andreasWallner: if we build this with an enum and keep the current prohibition for multiple tags then we can' have a path were both data and clock get a false path constraint - I have to admit that I can't think of a use-case for that one ATM, but reset & clock does make some sense...
Not sure having separate booleans isn't the preferable solution here?

andreasWallner:
@KireinaHoro @Dolu1990 what should we do with the two open threads? Would be great if we could merge this^^

core/src/main/scala/spinal/core/Trait.scala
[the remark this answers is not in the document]
KireinaHoro: So IMO it doesn't prevent a path having both data and reset/clock enable?  A path is defined by the start point to endpoint and this tag is added to the first buffer (i.e. destination) of the timing path, so we shouldn't have any conflicts, no?

core/src/main/scala/spinal/core/Trait.scala
KireinaHoro: Yeah the safest way is to do it in Tcl IMO, and that's how we implement it right now (apart from the clock defs, but that's of less concern).  Do we need any changes here?

core/src/main/scala/spinal/core/Trait.scala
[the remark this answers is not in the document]
andreasWallner: Currently the tag has `allowMultipleInstances` set to `false`. Combine that with the current enum approach - you can only have a single tag on a given path, meaning you can't e.g. have a false path on both clock and reset.
(since you'd have to place two tags in the same path, which `allowMultipleInstances` should not allow - and the constraint generator would also have to be able to deal with).

That's why I thought the approach with separate booleans was the better one in this case.

core/src/main/scala/spinal/core/Trait.scala
andreasWallner: I think we are in agreement to keep it as-is, what about you @Dolu1990 ?

core/src/main/scala/spinal/core/Trait.scala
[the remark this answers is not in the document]
KireinaHoro: I still don't understand.  When you say _clock_, do you mean the clock enable signal in a CD?  My understanding is that even if the start point is different, the end point is always distinct between CE and reset, no?

core/src/main/scala/spinal/core/Trait.scala
Dolu1990: > But i may miss some special cases ?

I think the reason why we used "fastest of both side" as constraints, was to be very safe by using a "design blind"  aproache. Meanwhile, if more care was used, we could have used the correct side of the CC as constraint.

=> Ok for me as-is.

core/src/main/scala/spinal/core/Trait.scala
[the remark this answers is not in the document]
andreasWallner: Lets say I wanted to do have the false path applied to both reset & clock enable, like so:
```scala
def doSmthg(x: Bool) = {
  x.addTag(crossClockFalsePath(destType=TimingEndpointType.RESET)
  x.addTag(crossClockFalsePath(destType=TimingEndpointType.CLOCK_EN)
}
```
this would currently not work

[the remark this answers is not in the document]
andreasWallner: I checked with
```scala
  val io = new Bundle {
    val i = in port Bool()
    val o = out port Bool()
  }
  io.o := BufferCC(io.i, inputAttributes = List(crossClockFalsePath(destType=TimingEndpointType.RESET), crossClockFalsePath(destType=TimingEndpointType.CLOCK_EN)))
}
```
which gives the error
```
[error]  Conflicting tags added to the same item! crossClockFalsePath(None,RESET) ; crossClockFalsePath(None,CLOCK_EN)
```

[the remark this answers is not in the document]
KireinaHoro: Uh but what you're doing doesn't make sense: you're not actually using it to drive the clock enable/reset signal.  If you look at this on the netlist level, clock enable (CE) and reset (PRE, etc.) are always _separate_ pins and thus create separate paths; you can always add the attribute on the non-shared parts of the paths and not having to worry about tag conflict.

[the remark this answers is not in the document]
andreasWallner: I think the example is actually not that esoteric - Altera recommends a synchronizer using the CE input of a FF, depending on where the reset for that is coming from you need the constraint on both of them.

Looks like this: https://www.intel.com/content/www/us/en/programmable/quartushelp/current/index.htm#da_rules/cdc_50102.htm
Personally I think these should then be delay constraints instead of false path constraints, but I've seen the latter used for that.

Instead of allowing multiple instances I'd go for multiple arguments, much shorter, just as extendible and if used with explicit parameter names just as readable.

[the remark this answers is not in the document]
andreasWallner: @KireinaHoro
How should we continue here? Modify it or leave as it is and drop the uniqueness requirement once we really need it?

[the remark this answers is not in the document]
KireinaHoro: I'm with leaving it as is.  We can always come back when such a use case comes up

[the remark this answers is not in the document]
andreasWallner: I'll merge this then, thanks!

andreasWallner:
@KireinaHoro Thanks!

Dolu1990:
@KireinaHoro  Hi ^^

I'm currently looking into getting a more generalized aproach to get timings constraints.
Idea being to have some tool analyse the SpinalHDL netlist, and have a  abstract software interface  to listen at it.
(see https://github.com/SpinalHDL/SpinalHDL/blob/8c1fd9f75c27d97c288d46710c8a695a283ed7d6/lib/src/main/scala/spinal/lib/eda/TimingExtractor.scala)

On the way, i found some issues with the way the netlist is analiysed in the Xilinx version from this PR.
Let me know if you are interrested ^^
trying to get this right :D

KireinaHoro:
@Dolu1990 hi!

Sounds like a sensible approach, since the logic walking the DSL to derive constraints should be shared across all tool vendors, even though the actual constraint syntax might differ.  I'd be happy to test the new approach on my local project :)

Tell me about what's wrong with the current VivadoConstraintWriter :)

Dolu1990:
Cool ^^

> Tell me about what's wrong with the current VivadoConstraintWriter :)

Mostly the def doWalkStatements which match on BaseType instead of Assignement statements
Aswell as the 

```scala
    var source = s.source.asInstanceOf[BaseType]
    if (!resetIsDriver) {
      AnalysisUtils.seekNonCombDrivers(source) { case b: BaseType =>
        source = b
      }
    }
```

which assumes the source is a basetype, aswell as assuming only one final source can be reached.

Another missing thing is that crossclock memories are ignored

I'm trying to get those things fixed in https://github.com/SpinalHDL/SpinalHDL/blob/8c1fd9f75c27d97c288d46710c8a695a283ed7d6/lib/src/main/scala/spinal/lib/eda/TimingExtractor.scala + a few fixes in the AnalysisUtils :)

KireinaHoro:
> Mostly the def doWalkStatements which match on BaseType instead of Assignement statements
and
> assuming only one final source can be reached

You're right, I didn't (and maybe still don't) have a very clear mental model of how a SpinalHDL netlist looks like.

Excited to see the new design!

Dolu1990:
no worries ^^

PR #1373, opened 2024-03-23
Add tag to keep I/O unchanged in InOutWrapper

andreasWallner:
Sometimes we may want to have InOutWrapper convert some of our drivers to TriState (etc.) drivers, but not others. This PR adds a tag to mark signals that we don't want to be wrapped.


Dolu1990:
Thanks :D

PR #1374, opened 2024-03-25
Inline output signal

yportne13:
keep on this pr #1339
if a submodule output signal directly connect to a combine signal, do not create a wrapper. Use this signal.
may also fix this discussion #1366


jijingg:
@yportne13  good , could you add those relevant options and descriptions to the SpinalHDL-DOC

PR #1377, opened 2024-03-29
Priority mux

### LINKED ISSUE #1376: Performance of PriorityMux
While rethink on the #1359 under ultra wide input scenario, the PriorityEncoder is required, which request an efficient PriorityMux first. 

The existing PriorityMux generate the code in daisy chain mode, as below.
``` Verilog
output_1 <= (sel[0] ? inputs_0 : (sel[1] ? inputs_1 : (sel[2] ? inputs_2 : inputs_3)));
```
This heavily depends on the capability of the synthesizer to reduce it's critical path.

Does anyone have investigated on this pattern with various tools? If this is necessary to be refactored?

Readon:
Closes #1376

# Impact on code generation

# Checklist

- [ ] Unit tests were added


Readon:
test code should be added before merge.

andreasWallner:
No qualms in general with this, just a small comment: I did have to think what `highPreced` means, maybe we can find something that is "more direct" (not sure how to say it, I guess it's because I thought of preceding, not precedence...).
`msbPriority`? `msbFirst`? (or keep it simple and let the user call reverse, that's actually shorter?)

Dolu1990:
In terms of API,

One possibility is to have : 

```scala
object PriorityMux{
   def apply(manyParameters) ...
   def first(...) = apply(..., highPreced= false)
   def last(...) = apply(..., highPreced= true)
}
```

A bit like for OHMasking

Readon:
> In terms of API,
> 
> One possibility is to have :
> 
> ```scala
> object PriorityMux{
>    def apply(manyParameters) ...
>    def first(...) = apply(..., highPreced= false)
>    def last(...) = apply(..., highPreced= true)
> }
> ```
> 
> A bit like for OHMasking

The first and last names here are a bit weird.
For OHMasking, the use of first to indicate that the first 1 was found is understandable, but for mux, first doesn't represent its choice very well. IMO.

Dolu1990:
> > In terms of API,
> > One possibility is to have :
> > ```scala
> > object PriorityMux{
> >    def apply(manyParameters) ...
> >    def first(...) = apply(..., highPreced= false)
> >    def last(...) = apply(..., highPreced= true)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > A bit like for OHMasking
> 
> The first and last names here are a bit weird. For OHMasking, the use of first to indicate that the first 1 was found is understandable, but for mux, first doesn't represent its choice very well. IMO.

Sure better name can be found.

Readon:
So do we have an agreements on this PR? Should it be the next step after implementing Device related thing as it is in #1381 ?

Dolu1990:
> So do we have an agreements on this PR?

Yes

> Should it be the next step after implementing Device related thing as it is in https://github.com/SpinalHDL/SpinalHDL/discussions/1381 ?

Yes i think so.

PR #1386, opened 2024-04-14
Add BRAM RegIf support; Add BRAMDriver

fayalalebrun:

This adds support for RegIf to the BRAM bus. In order to implement tests, I also added a BRAMDriver.

# Checklist

- [ x ] Unit tests were added
- [ x ] API changes are or will be documented:
  -  I believe the docs for RegIf apply, but let me know if more are needed.

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
+++ b/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
@@ -33,9 +33,9 @@ case class BRAMBusInterface(bus: BRAM, sizeMap: SizeMapping, regPre: String = ""
 
   override def writeAddress(): UInt = bus.addr << underbitWidth
 
-  override def readHalt(): Unit = {}
+  override def readHalt(): Unit = assert(false, "BRAM bus does not support halting")
 
-  override def writeHalt(): Unit = {}
+  override def writeHalt(): Unit = assert(false, "BRAM bus does not support halting")
 
   override def busDataWidth: Int = bus.config.dataWidth
 
--- a/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
+++ b/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
@@ -7,10 +7,10 @@ import spinal.lib.Delay
 
 case class BRAMBusInterface(bus: BRAM, sizeMap: SizeMapping, regPre: String = "", readDelay: Int = 1)(implicit moduleName: ClassName) extends BusIf {
   override val withStrb: Boolean = true
-  override val wstrb: Bits = withStrb generate(Bits(strbWidth bit))
-  override val wmask: Bits = withStrb generate(Bits(busDataWidth bit))
-  override val wmaskn: Bits = withStrb generate(Bits(busDataWidth bit))
-  withStrb generate(wstrb := bus.we)
+  override val wstrb: Bits = Bits(strbWidth bit)
+  override val wmask: Bits = Bits(busDataWidth bit)
+  override val wmaskn: Bits = Bits(busDataWidth bit)
+  wstrb := bus.we
   initStrbMasks()
 
   override val askWrite: Bool = bus.we.orR.allowPruning()
--- a/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
+++ b/lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
@@ -5,7 +5,7 @@ import spinal.lib.bus.misc.SizeMapping
 import spinal.lib.bus.bram.BRAM
 import spinal.lib.Delay
 
-case class BRAMBusInterface(bus: BRAM, sizeMap: SizeMapping, regPre: String = "", readDelay: Int = 1)(implicit moduleName: ClassName) extends BusIf {
+case class BRAMBusInterface(bus: BRAM, sizeMap: SizeMapping, regPre: String = "")(implicit moduleName: ClassName) extends BusIf {
   override val withStrb: Boolean = true
   override val wstrb: Bits = Bits(strbWidth bit)
   override val wmask: Bits = Bits(busDataWidth bit)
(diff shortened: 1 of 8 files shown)


lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
```diff
@@ -0,0 +1,43 @@
+package spinal.lib.bus.regif
+
+import spinal.core._
+import spinal.lib.bus.misc.SizeMapping
+import spinal.lib.bus.bram.BRAM
+import spinal.lib.Delay
+
+case class BRAMBusInterface(bus: BRAM, sizeMap: SizeMapping, regPre: String = "", readDelay: Int = 1)(implicit moduleName: ClassName) extends BusIf {
+  override val withStrb: Boolean = true
+  override val wstrb: Bits = withStrb generate(Bits(strbWidth bit))
```
andreasWallner, line 10: as far as I can see `we` is `wstrb` and it should be treated unconditionally

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  override val askRead: Bool = (!askWrite).allowPruning()
+
+  override val doWrite: Bool = (askWrite && bus.en).allowPruning()
+
+  override val doRead: Bool = (askRead && bus.en).allowPruning()
+
+  override val readData: Bits = Bits(busDataWidth bits)
+
+  bus.rddata := Delay(readData, readDelay) init B(0)
```
andreasWallner, line 26: I may have to check again, but looking at the BRAM bundle I don't see how arbitrary delay can work with the bus - since there is no response signal in the bus to indicate when a response is available?

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  override val writeData: Bits = bus.wrdata
+
+  override val readError: Bool = False
+
+  override def readAddress(): UInt = bus.addr << underbitWidth // BRAM uses word-aligned addresses
+
+  override def writeAddress(): UInt = bus.addr << underbitWidth
+
+  override def readHalt(): Unit = {}
```
andreasWallner, line 36: wondering whether there should be an assert here and below - it's currently unused, but if used for a read with a delay this would break

fayalalebrun: Good point, I think an assert to prevent this from being called with BRAMBus is the best approach.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
fayalalebrun: I wouldn't call it an arbitrary delay, it is more so that with BRAM you always get a response after a constant number of cycles, depending on how it is configured. So the register interface must match the constant read delay of the BRAM.

Perhaps I should make this a configuration option in the bus itself?

andreasWallner: The other components that deal with the BRAM bus also can't deal with arbitrary read delay from the BRAM (Decoder, BRAM BusSlaveFactory and Axi4SharedToBram depend on a 1 cycle delay). That's why I'm hesitant w.r.t. adding this.

fayalalebrun: I have moved the setting to BRAMConfig, as I think it is the cleanest solution. Then, to make sure the rest of the components are compatible:
- An assert when connecting two BRAM buses to make sure they have the same read latency.
- Decoder, RegIf and BRAMDriver support any read latency.
- BusSlave has an assert preventing read latency other than 1.
- Axis4SharedToBram creates its own BRAM config, so it will always have an implicit read latency of 1.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  override val askRead: Bool = (!askWrite).allowPruning()
+
+  override val doWrite: Bool = (askWrite && bus.en).allowPruning()
+
+  override val doRead: Bool = (askRead && bus.en).allowPruning()
+
+  override val readData: Bits = Bits(busDataWidth bits)
+
+  bus.rddata := readData   //Delay(readData, bus.config.readLatency) init B(0),  readData already reg-out
```
fayalalebrun, line 26: @jijingg Why did you make this change? readData is not already registered by default, see the other regif interfaces. Even if it was, this would break the case when readLatency > 1.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
andreasWallner: Looks good for most of this, but I don't see how the BRAMDecoder would support any latency? If you are accessing different RAM blocks every cycle, then the `selIndex` needs to be delayed for the number of read cycles, otherwise you'd mux the data from the wrong RAM to the `rddata`?

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
fayalalebrun: @jijingg Is it okay if I remove your commit? Happy to discuss any concerns you may have.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/BRAMBusInterface.scala
fayalalebrun: Yes you are right. I remember making that change, but perhaps I only thought about it.

I will make the edit after jijingg responds and we decide what to do with his commit.

andreasWallner:
@fayalalebrun Thanks!

PR #1388, opened 2024-04-18
add an option to remove timescale

yportne13:
add a `withTimescale` option in SpinalConfig so that people can choose wheather or not generate this:
```verilog
`timescale 1ns/1ps
```
the first commit default to `with timescale`, and the second commit change the default to `without timescale`

### DIFF
--- a/core/src/main/scala/spinal/core/Spinal.scala
+++ b/core/src/main/scala/spinal/core/Spinal.scala
@@ -175,7 +175,7 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         var singleTopLevel             : Boolean = true,
                         var noAssertAtTimeZero         : Boolean = false,
                         var cutLongExpressions         : Boolean = true,
-                        var withTimescale              : Boolean = true,
+                        var withTimescale              : Boolean = false,
                         var emitFullComponentBindings  : Boolean = true
 ){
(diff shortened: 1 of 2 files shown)


Dolu1990:
Personnaly i'm good with this change. But i don't know this can mess up with other people flow to sudently not have the timescale generated by default ?

yportne13:
> Personnaly i'm good with this change. But i don't know this can mess up with other people flow to sudently not have the timescale generated by default ?

then remove the last commit is ok. but i don't know how to remove the last commit

Dolu1990:
So far, i think the full PR is good, let's just wait a few days to see if there is any objection.

Dolu1990:
Thanks :)

KireinaHoro:
This does break Verilator by default... I'm confused why the default behaviour is changed here :/

Dolu1990:
hmm kinda same, maybe the right thing would have been to have by default the timescale enabled.

Verilator has this behaviour that once there is one timescale, he want it everywere XD

@KireinaHoro @yportne13 Should we change the default to "with time scale" ?

yportne13:
Either option works for me, I don't have a strong preference

Dolu1990:
SpinalHDL 1.10.2 released with timescale back by default

PR #1389, opened 2024-04-20
Fix bin files passed to .sby read statement

fayalalebrun:

Currently, generating `.sby` files for formal verification fails if the design contains initialized memory. This is because the `.bin` file is currently being passed to the `read` statement, which expects only Verilog and SystemVerilog files.

This MR adds a simple filter to not include `.bin` files for the purposes of the `read` statement. I think this solution is not as robust as it could be, but `SpinalReport` currently does not seems to convey the distinction between generated RTL sources and bin files.

# Impact on code generation

# Checklist

- [ ] Unit tests were added


andreasWallner:
Thanks!
Merging this to get it fixed, could I ask you for a PR that adds a testcase for this?

PR #1401, opened 2024-04-27
Fix verilator version parsing

fayalalebrun:

On my system, verilator parsing fails since the regex in question does not seem to match the output of `verilator --version`. This is the output:
```
Verilator 5.022 2024-02-24 rev UNKNOWN.REV
```

I also tested on a Ubuntu system and it seems the output does not seem to match there either:
```
Verilator 4.038 2020-07-11 rev v4.036-114-g0cd4a57ad
```

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
@@ -566,7 +566,7 @@ JNIEXPORT void API JNICALL ${jniPrefix}disableWave_1${uniqueId}
 
     val verilatorVersionProcess = Process(Seq(verilatorBinFilename, "--version"), new File(workspacePath))
     val verilatorVersion = verilatorVersionProcess.lineStream.mkString("\n") // blocks and throws an exception if exit status != 0
-    val verilatorVersionDeci = BigDecimal("v([0-9]*\\.[0-9]*)".r.findFirstIn(verilatorVersion).get.tail)
+    val verilatorVersionDeci = BigDecimal("([0-9]+\\.[0-9]+)".r.findFirstIn(verilatorVersion).get)
 
     if (cacheEnabled) {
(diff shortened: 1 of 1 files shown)


andreasWallner:
The regex is looking for the second mention of the version (which has the `v` prefix), so it does in fact match what you pasted from your Ubuntu system.

I'm kind of asking myself which version to trust here, both in you example from Ubuntu, as well as from the version included in oss-cad-suite I see different version numbers reported in both places - the ones that are consistent are the self-built ones... (in my case in the system path and the local one)

```
1 ~/git/SpinalHDL[dev]% verilator --version
Verilator 4.228 2022-10-01 rev v4.228
1 ~/git/SpinalHDL[dev]% /opt/oss-cad-suite-20230105/bin/verilator --version
Verilator 5.005 devel rev v5.004-46-g816cff009
1 ~/git/SpinalHDL[dev]% /opt/oss-cad-suite-20230926/bin/verilator --version
Verilator 5.017 devel rev v5.014-134-gd2c72a7f2
1 ~/git/SpinalHDL[wavefile_prefix]% ../verilator/bin/verilator --version
Verilator 4.228 2022-10-01 rev v4.228
```

In your case, where is the 5.x from that you are using?

fayalalebrun:
The one with UNKNOWN comes from Manjaro. Perhaps the second version is set by the distro.

I also tried Fedora, and this one is missing the "v":
```
Verilator 5.022 2024-02-24 rev fedora-5.022
```

fayalalebrun:
OpenSUSE:
```
Verilator 4.228 2022-10-01 rev UNKNOWN.REV
```

Alpine:
```
Verilator 5.018 2023-10-30 rev UNKNOWN.REV
```

Dolu1990:
Thanks ^^

PR #1402, opened 2024-04-28
Add wave prefix for verilator & VCS

andreasWallner:
The prefix was already somewhat implement in the backends, expose it to the user (at least for verilator and VCS).
IVerilog and GHDL will follow, but I saw that they are currently also missing support for the `_testFile` parameter which I´d add at the same time.

`_testFile` does allow us to specify a separate folder for just the waveform file, but we can't put the results from multiple tests into the same directory due to the name collision. This is nice for two reasons:

- workspace folder may be renamed automatically, making it a chore to pick the right one
- if your waveform viewer supports switching between files, keeping signals the same they are easier to find.

# Impact on code generation

None


Dolu1990:
Thanks ^^

PR #1409, opened 2024-05-08
[regif] feature upgrade refactoring

jijingg:
This update is based on practical feedback from real projects, and the changes are quite significant. In addition to adding some important features, many factory functions have been added, and common register file reuse scenarios have been solved. suggestions from software usage have been adopted to optimize.`.h` file. 
   Despite such significant refactoring modifications, backward compatibility should be maintained in principle, and it will not affect old code. 

- [x] 1. remove BusIfVistor mode instead directly access for simplify
- [x] 2. suport fifo[wrfifo/rdfifo]/ram interface 
- [x] 3. readLogic support replace
- [x] 4. add RegInstGrp layer
- [x] 5. refactor BusIf Document, but keep old api
- [x] 6. add usefull Factory method(Interrupt, SCR etc..)
- [x] 7. localbus add 
- [x] 8. demo example update 
- [x] 9. add reuse Block layer 
- [x] 10. optmize CHeader /HTML format 
 
# Checklist

- [x] Unit tests were added  done
- [ ] API changes are or will be documented:  TODO


andreasWallner:
I know this is still a draft, but a few thoughts:
- what is lbus/mbus - I couldn't find any spec; is this some company internal bus that you are using? Should it be in spinal.lib?
- removing the visitor breaks current users of the library - we should not do that
- in general many of these changes seem to break existing users (ones with custom output generators, by moving classes, etc.)
- this seems to only support FIFO writes, not reads, when adding FIFO access it should allow for both; the current design can't be easily extended for read

possibly more to follow, have to look at it some more

jijingg:
> * removing the visitor breaks current users of the library - we should not do that
@andreasWallner 
1. Although there isn't a standard, it's basically an unwritten but widely used bus approach, which may have some naming differences, but the logic is the same. Since every team will use it, I suggest providing an implementation in Spinal to avoid everyone reinventing the wheel. （Of course, we can discuss naming further. I also don't think "lbus" or "mbus" is a good idea. That includes their signal names as well.）
2. Secondly, maintaining backward compatibility as the first priority for open-source projects, which is not violated here. The original method can still be used, and API conversion can be used to accomplish this seamlessly, without impacting users.
3. Same as 2, it is insensitive to existing users and cannot cause crashes of existing code
4. Regarding the logic behind `bus.readata`, there is still work to be done, which is why it's currently in draft status. However, the FIFO reading you mentioned is indeed a good idea, even though I haven't encountered such a scenario yet. I believe FIFO reading could have potential use, and I will consider implementing it.

Readon:
FIFO is really a special case, when writing to a FIFO it should be tested if it is writable first. So the data and status of FIFO might need to be added separately, which is a little bit different from the existing regif interface.

On the other hand, before read, the host should check if there is data in the FIFO.

jijingg:
> FIFO is really a special case, when writing to a FIFO it should be tested if it is writable first. So the data and status of FIFO might need to be added separately, which is a little bit different from the existing regif interface.
> 


yes  the FIFO port cannot strictly guarantee no data loss,
if the user scenario is sensitive to operation loss, there are two ways for users to solve it
1. Calculate the maximum throughput per unit time and choose the issuance operation to ensure that the FIFO does not overflow per unit time.
2. The status information of FIFO can be hung on registers for software to read, and the software's issuance operation needs to refer to the FIFO's status as handshake

> On the other hand, before read, the host should check if there is data in the FIFO.

right, If it is an effective read operation, the software does need to first determine whether there is data. But to be honest,
FIFO reading may have use-case ,   but I haven't encountered them yet.  to be honest,   I'm still on the fence about whether it's necessary to implement the read functionality of a FIFO.
 So the functionality of reading from a FIFO seems to have too many limitations and is somewhat redundant. It has to worry about HW-write overflows and also consider the possibility of SW-read miss

Readon:
> n the fence about whether it's necessary to implement the read functionality of a FIFO.
> So the functionality of reading from a FIFO seems to have too many limitations and is somewhat redundant. It has to worry about HW-write overflows and also consider the possibility of SW-read miss

Ahhh，We use a lot while designing an instrument.
On common case is that a streamlized SPI or UART controller might want to use FIFO to cache the data to be transmitted.

jijingg:
> Ahhh，We use a lot while designing an instrument.
> On common case is that a streamlized SPI or UART controller might want to use FIFO to cache the data to be transmitted.

i got it,  so now If we  sensitive to packet loss, we may need some peripheral auxiliary circuits or handshake to ensure data. Since that's the case, it's necessary to add this feature , but the specific data caching correctness needs to be guaranteed by the user themselves, is it OK?

Readon:
> > Ahhh，We use a lot while designing an instrument.
> > On common case is that a streamlized SPI or UART controller might want to use FIFO to cache the data to be transmitted.
> 
> i got it, so now If we sensitive to packet loss, we may need some peripheral auxiliary circuits or handshake to ensure data. Since that's the case, it's necessary to add this feature , but the specific data caching correctness needs to be guaranteed by the user themselves, is it OK?

Yes, I think they should guarantee the correctness by software. We can only provide the option.

lib/src/main/scala/spinal/lib/bus/regif/AccessType.scala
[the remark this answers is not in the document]
jijingg: done  XD, thanks

jijingg:
At present, I think it ready for merge The update of the document is still in progress it will completed within this week.  do  you guys have any comments @andreasWallner @Readon @mrberman87 @Dolu1990

Dolu1990:
For me, you can merge you feel it.

andreasWallner:
@jijingg 
comment are coming up, just takes a while to get through 66 files ...

jijingg:
> @jijingg comment are coming up, just takes a while to get through 66 files ...

hi, Let me merge first. If there are any issue  need improvement, we will continue to initiate the PR

lib/src/main/scala/spinal/lib/bus/localbus/MemBus.scala
```diff
@@ -0,0 +1,39 @@
+package spinal.lib.bus.localbus
+
+import spinal.core._
+import spinal.lib._
+
+case class MemBusConfig(aw: Int, dw: Int)
+
+case class MemBus(c: MemBusConfig) extends Interface with IMasterSlave {
```
andreasWallner, line 8: MemBus is a near duplicate of the BRAM bus (slight difference in WE, which would actually be a reasonable future feature request to support) and the AsyncMemoryBus (which can stall) - I think we should extend one of those with a parameter instead of adding yet another new Bus that is not reasonably different.

lib/src/main/scala/spinal/lib/bus/localbus/MinBus.scala
```diff
@@ -0,0 +1,35 @@
...
+
+case class MinBusConfig(aw: Int,
+                        dw: Int,
+                        pw: Int = 4,
+                        withStrb: Boolean = false
+                       ){
+  val sw  = dw >> 3
+}
+
+case class MinBus(c: MinBusConfig) extends Interface with IMasterSlave {
```
andreasWallner, line 14: I hope you'll describe the meaning (and timing) of these signals somewhere - its really hard to know what some of them are supposed to be (e.g. the difference between rvld and rdy, what `prot` does).
Apart from that this is a slight extension of AsyncMemoryBus and would maybe be better as some parameters there?

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemVIP.scala
```diff
@@ -0,0 +1,108 @@
+package spinal.lib.bus.localbus
+
+import spinal.core._
+import spinal.core.sim._
+class MemVIP{
+  private val mem = scala.collection.mutable.HashMap[Long, BigInt]()
```
andreasWallner, line 6: it would make sense if this would reuse `SparseMemory`

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemBusDriver.scala
```diff
@@ -0,0 +1,43 @@
+package spinal.lib.bus.localbus
+
+import spinal.core._
+import spinal.core.sim._
+
+case class MemBusDriver(bus : MemBus, clockdomain : ClockDomain) {
```
andreasWallner, line 6: All other bus drivers initialize the bus when constructed - this one should do this as well.

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemBusDriver.scala
```diff
@@ -0,0 +1,43 @@
...
+    bus.wr.randomize()
+    bus.addr.randomize()
+    bus.wdat.randomize()
+  }
+
+  def read(address : BigInt) : BigInt = {
+    bus.ce #= true
+    bus.wr #= false
+    bus.addr #= address
+    bus.wdat #= 1234 //.randomize()
```
andreasWallner, line 25: this should use randomize like in the comment?

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemVIP.scala
```diff
@@ -0,0 +1,108 @@
...
+      while(true){
+        cd.waitSampling()
+        if(bus.ce.toBoolean && bus.wr.toBoolean){
+          add(bus.addr.toLong, bus.wdat.toBigInt)
+        }
+      }
+    }
+  }
+
+  def hang(bus: MinBus, delay: Int = 10)(cd: ClockDomain): MemVIP = {
```
andreasWallner, line 43: what does `hang` mean - the only thing that comes to my mind is that it gets stuck which doesn't appear to be the case - looking at this and the one above shouldn't these just be `apply` methods?

lib/src/main/scala/spinal/lib/bus/regif/Block/FifoInst.scala
```diff
@@ -0,0 +1,40 @@
...
+import spinal.core._
+import spinal.lib.Stream
+
+abstract class FifoInst(name: String, addr: BigInt, doc: String, grp: GrpTag = null)(bi: BusIf) extends RegSlice(name, addr, doc, size = bi.wordAddressInc, grp)(bi){
+  override val regType: String = "FIFO"
+
+  val hitDoRead: Bool
+  val hitDoWrite : Bool
+
+  val bus = Stream(Bits(bi.busDataWidth bit))
```
andreasWallner, line 12: Having a `Stream` here creates an issue for `WrFifoInst` where it's lying to the user: Stream implies that backpressure can be applied, which is not the case there. So for write this should be a `Flow` instead. This would correctly indicate that it can't stall the bus, and the user can use `toStream(overflow: Bool)` to convert to a stream with an overflow check (or use `toStream` if they don't care).

In short: this should be moved to `WrFifoInst` (as a `Flow`) and `RdFifoInst` (kept as a `Stream`)

lib/src/main/scala/spinal/lib/bus/regif/BusIf.scala
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  def newRAMAt(address: BigInt, size: BigInt, doc: String, grp: GrpTag = null)(implicit symbol: SymbolName) = {
+    assert(address % wordAddressInc == 0, s"located Position not align by wordAddressInc: ${wordAddressInc}")
+    val res = creatRAM(symbol.name, address, size, doc, grp)
+    regPtr = address + scala.math.ceil(size.toDouble/wordAddressInc).toLong * wordAddressInc
+    res
+  }
+
+  def creatRAM(name: String, addr: BigInt, size: BigInt, doc: String, grp: GrpTag = null) = {
```
andreasWallner, line 171: why leave out a single character in `create`? typing this I always have to remind myself to make this mistake.
I see that `creatReg` existed for the old API - we should deprecate that one and fix the spelling.

lib/src/main/scala/spinal/lib/bus/regif/BusIfAdapter/AhbLite3BusInterface.scala
```diff
@@ -6,36 +6,38 @@ import spinal.lib.bus.misc.SizeMapping
...
+  val reg_rderr: Bool = Reg(Bool(), init = False)
+  val reg_rdata: Bits = Reg(Bits(busDataWidth bits), init = defualtReadBits)
+
   val wstrb: Bits = withStrb generate (Bits(strbWidth bit))
   val wmask: Bits = withStrb generate (Bits(busDataWidth bit))
   val wmaskn: Bits = withStrb generate (Bits(busDataWidth bit))
   initStrbMasks()
   override def getModuleName = moduleName.name
 
-  val readError = Bool()
```
andreasWallner, line 15: these were one of the changes that I was referring to when I said that the PR can't be backwards compatible: since you are changing the name of a public field every code referring to it is broken after this

lib/src/main/scala/spinal/lib/bus/regif/BusIfBase.scala
```diff
@@ -1,37 +1,50 @@
...
   val doRead: Bool
+  val cg_en: Bool = True
+  val NS: Bool = False //NS(Non-Secure Access flag bit)
+
+  val bus_rderr: Bool
+  val bus_rdata: Bits
+  val reg_rderr: Bool
+  val reg_rdata: Bits
 
-  val readData: Bits
```
andreasWallner, line 17: This was one of the changes I was referring to when I said this PR is not backwards compatible. This change breaks ever custom bus implementation a user of the library has

andreasWallner (commented):
Sorry for the long time, but since this is nearly a rewrite I had to copy code to an external editor to have useful diffs - a PR of this size is not reviewable.

jijingg:
> Sorry for the long time, but since this is nearly a rewrite I had to copy code to an external editor to have useful diffs - a PR of this size is not reviewable.

XD , yes  it's realy a significant refactoring,  stemming from some feedback on our real project,  i think it's necessary to do this upgrade. 

thank you for taking the time to conduct such a careful review and providing multiple valuable feedback. I will make some revisions to the PR again

lib/src/main/scala/spinal/lib/bus/regif/Block/FifoInst.scala
jijingg: make sense ,  `Flow` is better than `Stream` for WrFfifo

lib/src/main/scala/spinal/lib/bus/regif/BusIf.scala
jijingg: XD，  fixed https://github.com/SpinalHDL/SpinalHDL/pull/1434/commits/e153944f85917791e882c727a25df809b9621a15

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemBusDriver.scala
jijingg: done

lib/src/main/scala/spinal/lib/bus/localbus/MinBus.scala
jijingg: `prot `is widely used in the AMBA protocol, integrating the habits of the AMBA protocol to represent secure bits.  and yes , adding some comments would be better

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemVIP.scala
jijingg: As a Mem VIP (Verification IP), it must be Sparse, and here we emphasize that VIP ， so did the MemVIP be better

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemBusDriver.scala
jijingg: done

lib/src/main/scala/spinal/lib/bus/localbus/MinBus.scala
andreasWallner: I understand that `prot` is used by AMBA - but nobody can know that you are using the same names here compared to AMBA. This bus does in no way show a connection to the AMBA protocols. That why I think it needs at least a description of the intent of the signals + their timing.

lib/src/main/scala/spinal/lib/bus/localbus/sim/MemVIP.scala
andreasWallner: I'm sorry - I don't understand what you are saying.

PR #1412, opened 2024-05-10
Fix spaces in file paths for sby generation

### LINKED ISSUE #1408: Linux: Spaces in path break formal verification
I installed SpinalHDL following the guide on the documentation. The Windows section mentions to avoid spaces in the project path so I assumed that on Linux it would be fine. Everything I've tested so far works other than formal verification. In `Project/./simWorkspace/unamed/unamed_bmc/ERROR` I am seeing
```
ERROR: sby file syntax error: '[files]' section entry expects up to 2 arguments, 9 specified
```
I am guessing that this is because in the `config.sby`, there are spaces in the [files] section. The fragments of the path that are separated by spaces are then treated as separate arguments, which is why there are 9 arguments. I tried adding \ before all the spaces but when I used runMain again, the modified config just got overridden.

fayalalebrun:
Closes #1408 

The `sby` tool does not support spaces in paths. Since absolute paths are used in the "files" section of the .sby file, users whose projects are in paths with spaces cannot currently use SpinalHDL formal verification.

This fixes the issue by using relative paths instead of absolute paths. That is, as long as the component does not have a space in its name, but this shouldn't be support by VHDL/Verilog anyways.

# Impact on code generation

Files in the .sby "files" section use relative paths instead of absolute paths.

# Checklist

- [ ] Unit tests were added

@TheImunityGamer it would be great if you could confirm this fixes your issue.

### DIFF
--- a/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
+++ b/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
@@ -225,7 +225,7 @@ case class SpinalFormalConfig(
 
       val dst = rtlDir.resolve(src.getName)
       FileUtils.copyFileToDirectory(src, rtlDir.toFile())
-      rtlFiles.append(dst.toString)
+      rtlFiles.append(workingWorksplace.relativize(dst).toString)
     }
 
     _backend match {
--- a/core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
+++ b/core/src/main/scala/spinal/core/formal/SymbiYosysBackend.scala
@@ -117,7 +117,7 @@ class SymbiYosysBackend(val config: SymbiYosysBackendConfig) extends FormalBacke
   }
 
   def genSby(): Unit = {
-    val localSources = config.rtlSourcesPaths.map(f => new File(f).getAbsolutePath).mkString("\n")
+    val localSources = config.rtlSourcesPaths.map(f => new File(f)).mkString("\n")
     val read = config.rtlSourcesPaths
       .map(f => {
         Paths.get(f).getFileName


Readon:
The main reason on putting file name as absolute path is that relative path is sensitive to the path that runs sby. So that not only test cases in lib should be tested but also with user project.

Readon:
> > The main reason on putting file name as absolute path is that relative path is sensitive to the path that runs sby. So that not only test cases in lib should be tested but also with user project.
> 
> I tested it with my current project and it works there.

I have tested it on our internal projects it works too.

Readon:
Thanks.^^

Readon:
> I haven't been able to make sbt use the dev branch yet. I hoped it would be as simple as setting the spinalVersion value from the template project to "dev", but sbt keeps complaining about this.

I have tested it.
You need to modify scala version also.
The default dev's scala version have been changed to 2.12

TheImunityGamer:
I'm using the right Scala. I'm using version 2.12.18. I just keep getting told that `com.github.spinalhdl:spinalhdl-core_2.12:dev` was not found.

Readon:
> I'm using the right Scala. I'm using version 2.12.18. I just keep getting told that `com.github.spinalhdl:spinalhdl-core_2.12:dev` was not found.

I mean the scala version specified in your SpinalTemplateSbt's build.sbt file.

fayalalebrun:
@TheImunityGamer Here is an example: https://github.com/andreasWallner/spinalStuff/blob/master/extras/develop.sbt

andreasWallner:
Since you are pointing to my example just a quick note: this only works correctly if the scala version of your project matches the default (currently "2.12.18", see https://github.com/SpinalHDL/SpinalHDL/blob/dev/project/Version.scala - but you could change that for your local usage).
This setup is nice for ongoing co-development of your codebase and SpinalHDL (though many things are easier tried using extensions via implicit classes). If you just want to use the current dev it would be easier to publish SpinalHDL to your local cache, see https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Introduction/faq.html#how-to-use-an-unreleased-version-of-spinalhdl-but-committed-on-git

@TheImunityGamer if you continue to have problems with that please open a new issue containing your build.sbt or ask in the gitter chat.

TheImunityGamer:
I found out that `sbt publishLocal` really wants to be run from within a directory without spaces. And I'm guessing using `ProjectRef(file("relative path"))` in your build.sbt also wants the project to have no spaces in its path.

So sby got further this time but is now saying:
```
Removing directory '<path_to_project>/simWorkspace/unamed_1/unamed_1_bmc'.                                                                                                                                                
Copy '<path_to_project>/simWorkspace/unamed_1/rtl/unamed_1.sv' to '<path_to_project>/simWorkspace/unamed_1/unamed_1_bmc/src/unamed_1.sv'.
engine_0: smtbmc --progress yices                                                                                                                                                                                                                                                                          
base: starting process "cd <path_to_project>/./simWorkspace/unamed_1/unamed_1_bmc/src; yosys -ql ../model/design.log ../model/design.ys"                                                                                  
base: bash: line 1: cd: too many arguments                                                                                                                                                                                                                                                                 
base: Can't open log file `../model/design.log' for writing!                                                                                                                                                                                                                                               
base: finished (returncode=1)                                                                                                                                                                                                                                                                              
base: task failed. ERROR.                                                                                                                                                                                                                                                                                  
summary: Elapsed clock time [H:MM:SS (secs)]: 0:00:00 (0)                                                                                                                                                                                                                                                  
summary: Elapsed process time [H:MM:SS (secs)]: 0:00:00 (0)                                                                                                                                                                                                                                                
summary: engine_0 (smtbmc --progress yices) did not return a status                                                                                                                                                                                                                                        
summary: engine_0 did not produce any traces                                                                                                                                                                                                                                                               
DONE (ERROR, rc=16)                                                                                                                                                                                                                                                                                        
```
That line with the `cd` is trying to cd to a path with spaces.

TheImunityGamer:
tldr at the bottom

You're right, the second issue was with the file at `oss-cad-suite/share/yosys/python3/sby_core.py`. I needed to change these SbyProc statements and make them use `\"` to get the space escaping in the cd commands:
```python
proc = SbyProc(
                self,
                model_name,
                [],
                "cd \"{}/src\"; {} -ql ../model/design.log ../model/design.ys".format(self.workdir, self.exe_paths["yosys"])
            )
```
It seemed to me that this workaround could be avoided by cd-ing to a directory without spaces and then running sby. It seems sby will use paths relative to the current working directory.

Anyways, after that, there were some other issues, like yosys complaining that it needed clk2fflogic, so I forced it to be enabled (putting True in an if statement. I think it used to rely on a command line option). Then there was another issue with spaces in the path and cd not being able to get there. I then decided to run sby manually after cd-ing into `simWorkspace/unamed_1` and it worked.

tldr: made a bunch of changes to some files. Now I need to run normal SpinalHDL formal one time, it fails, then I do it manually, and it works.

PR #1414, opened 2024-05-10
Streamhistory

Readon:
Closes #

StreamHistory is a kind of history which would have a valid for each element to demonstrate if the data is valid for using.
It acts like FIFO, also follow first in first out policy, but with internal data visibility.

# Impact on code generation

# Checklist

- [ x ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1627,6 +1627,51 @@ class StreamFifoCC[T <: Data](val dataType: HardType[T],
   }
 }
 
+object StreamHistory {
+  def apply[T <: Data](that: Stream[T], length: Int): Area = new Area{
+    val ret = new StreamHistory(that.payloadType, length)
+    ret.io.push << that
+    val output = cloneOf(that)
+    output << ret.io.pop
+    val states = cloneOf(ret.io.states)
+    (states, ret.io.states).zipped.map(_ << _)
+  }
+}
+
+class StreamHistory[T <: Data](dataType: HardType[T], length: Int) extends Component {
+  val io = new Bundle {
+    val push          = slave  Stream(dataType)
+    val pop           = master Stream(dataType)
+    val states        = Vec(master Flow(dataType), length)
+  }
+
+  def builder(prev: Stream[T], left: Int): List[Stream[T]] = {
+    left match {
+      case 0 => Nil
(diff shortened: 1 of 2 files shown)


lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -1627,6 +1627,50 @@ class StreamFifoCC[T <: Data](val dataType: HardType[T],
...
+  def builder(prev: Stream[T], left: Int): List[Stream[T]] = {
+    left match {
+      case 0 => Nil
+      case 1 => prev :: Nil
+      case _ => prev :: builder(
+        {
+          val stream = Stream(dataType)
+          stream <-< prev
+
+          val next = Stream(dataType)
```
Dolu1990, line 1657: I'm not sure to understand why "next" stream is required, why not returning directly "stream" ?

Readon: the stream's ready should be also asserted when it's valid is not true.
next is used for interface only.

lib/src/main/scala/spinal/lib/Stream.scala
```diff
@@ -1627,6 +1627,50 @@ class StreamFifoCC[T <: Data](val dataType: HardType[T],
...
+          next.valid := stream.valid
+          next.payload := stream.payload
+          stream.ready := next.ready | !stream.valid
```
Readon, line 1660: @Dolu1990 Here is the point.

lib/src/main/scala/spinal/lib/Stream.scala
Readon: that means elements pushed at input would automatically move forward to pop end.

Dolu1990: > the stream's ready should be also asserted when it's valid is not true.

The <-< operator does that already by default i think (it collapse bubbles) : 
https://github.com/SpinalHDL/SpinalHDL/blob/c150e90db33e389503821077caa942194745b271/lib/src/main/scala/spinal/lib/Stream.scala#L369

Also, as by default it collapse bubbles, the ready signal has no meaning when the valid is low.

or i missed something ?

Readon: You are right, I missed the collapse bubbles function previously.
Now it has been fixed.

PR #1415, opened 2024-05-13
VexiiRiscv branch merge

Dolu1990:
Lot of little fixes / improvments.
The main change is related to the fiber framework to provide a better error reporting.

Can be merged if it passes the regressions i would say.


core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -955,9 +970,16 @@ package object sim {
       } else {
         throw new Exception("???")
       }
-
     }
 
+    def forkStimulus() : Unit = {
+      val hz = cd.frequency match {
+        case ClockDomain.FixedFrequency(value) => value.toBigDecimal
```
andreasWallner, line 977:
```suggestion
        case ClockDomain.FixedFrequency(value) => value.toBigDecimal,
        case _ => throw new Exception(s"Can't forkStimulus() w/o explicit frequency since frequency of ClockDomain $cd is not known")
```
?

lib/src/main/scala/spinal/lib/com/uart/UartCtrl.scala
```diff
@@ -191,10 +191,10 @@ object UartCtrl {
 }
 
 case class UartCtrlInitConfig(
-  baudrate : Int = 0,
-  dataLength : Int = 0,
-  parity : UartParityType.E = null,
-  stop : UartStopType.E = null
+  var baudrate : Int = 0,
```
andreasWallner, line 194: Should these really be vars?
The user can never make those vals again so that they stay consistent, but if you want/need to be able to change them you can always have a `var config: UartCtrlInitConfig` and reassign to that from `config.copy(...)` (as with pretty much all other configs in lib)

Dolu1990: Overall, i'm more and more "pissed" by the all functional paradigm / val ^^
config.copy is kinda too dogmatic i think. While mutability is just easy ?
Often it work just great to have a default config and mutate it a bit.

> (as with pretty much all other configs in lib)

Maybe that is wrong aswell

lib/src/main/scala/spinal/lib/bus/tilelink/Ram.scala
```diff
@@ -1,15 +1,17 @@
...
 
-class Ram (p : NodeParameters, bytes : Int) extends Component {
+class Ram (p : NodeParameters,
+           bytes : Int) extends Component {
   val io = new Bundle{
     val up = slave port Bus(p)
   }
 
-  val mem = Mem.fill(bytes/p.m.dataBytes)(Bits(p.m.dataWidth bits))
+  val mem = Mem.fill(bytes/p.m.dataBytes)(Bits(p.m.dataWidth bits)).simPublic()
```
andreasWallner, line 14: should we make the memory simPublic by default?

Dolu1990: hmmm, probably not, need to revert this.

lib/src/main/scala/spinal/lib/com/uart/UartCtrl.scala
andreasWallner: Not sure - If you pass one config object into multiple constructors (forgetting to copy them) and one ends up changing it, it's suddenly modified for all places. Kind of the curse of Java reference semantics.

Dolu1990: > Not sure - If you pass one config object into multiple constructors (forgetting to copy them) and one ends up changing it, it's suddenly modified for all places. Kind of the curse of Java reference semantics.

Yes, right.

On the other hand, imutability prevent "natural" patterns as : 

```scala
      val uart = new TilelinkUartFiber()
      uart.config.initConfig.baudrate = 9600
```

Forcing you to get everything ahead of time through parameters, preventing preset default values.

In other words, imutability feel a bit like (in another context) "everything private in class, use only getters/setters"

PR #1418, opened 2024-05-15
Change the usage interface of StreamHistory the same as Stream Adapter and Extender.

Readon:
Closes #

The interface by creating new Area might lead states accessing problem, this might solve it.

# Impact on code generation

# Checklist

- [x] Unit tests were added


Readon:
> I'm not sure anymore now.
> 
> What kind of history do we want ? I guess the history should move when we push something in it.
> 
> So, i'm realy not sure we should have an stream output to the StreamHistory.
> 
> Basicaly it would become just a shift register which shift on input stream fire More like a FlowHistory

The problem is that in some scenario there are requirement that user want to proceed several valid datas then output data. after processing the output data can be poped. shift registers as FlowHistory is not able to control the pace.

Dolu1990:
Ahhh i see the usage.
One issue seems to me that  : 
prev.pipelined(true, true)
can store up to 2 transactions, which will make the history "hide" stuff.

How long of an history are you using ?

Readon:
> Ahhh i see the usage. One issue seems to me that : prev.pipelined(true, true) can store up to 2 transactions, which will make the history "hide" stuff.
> 
> How long of an history are you using ?

that would not be a good solution.
I myself only use cases that length equals 3.

According my experiments that there are cases we need to add cut off to ready to meet timing constraints. So I just want to design ahead of time.

Dolu1990:
Hmmm
> According to my experience that there are cases we need to add cut off to ready to meet timing constraints. So I just want to design ahead of time.

Could it be done before the stream history ?

StreamHistory(myStream.s2mPipe(), ...) ?

Readon:
> Hmmm
> 
> > According to my experience that there are cases we need to add cut off to ready to meet timing constraints. So I just want to design ahead of time.
> 
> Could it be done before the stream history ?
> 
> StreamHistory(myStream.s2mPipe(), ...) ?

Yes, of course, user can cascade it by themself also. so I changed the title and update it. Now the main modification is on its interface to user.

Dolu1990:
hmm
I'm think trying to use s2m / m2s in general is a bad idea for an history functionality, whats about reworking the design with : 
image

The idea would be to :  
- Have zero latency
- Implementd as a shift register
- Push new transaction are pushed the the first free register on the right
- The oldest stored transaction would always be on the most right register

Readon:
> hmm I'm think trying to use s2m / m2s in general is a bad idea for an history functionality, whats about reworking the design with : image
> 
> The idea would be to :
> 
> * Have zero latency
> * Implementd as a shift register
> * Push new transaction are pushed the the first free register on the right
> * The oldest stored transaction would always be on the most right register

Though, the logic would bloat a little by the insert location determine logic, which would be a first non-valid signal finding logic.

Dolu1990:
Yes right, it would mostly cost (2input mux * data width) for each register stage

Readon:
> The good thing about that alternative design compared to m2sPipe, is that the oldest transaction will always be present at the same history index, independently of the lenght of the history, and with no latency.

That sounds a little bit like a partial functional HistoryModifiable.

Readon:
> The good thing about that alternative design compared to m2sPipe, is that the oldest transaction will always be present at the same history index, independently of the lenght of the history, and with no latency.

That sounds a little bit like a partial functional HistoryModifiable.

Readon:
> Yes right, it would mostly cost (2input mux * data width) for each register stage

There should be mux control signal generator though.

Readon:
> The good thing about that alternative design compared to m2sPipe, is that the oldest transaction will always be present at the same history index, independently of the lenght of the history, and with no latency.

I think that has to be renamed to StreamHistoryFifo due to its working policy. 
StreamHistory <==> History, cache chain move one by one.
StreamHistoryFifo <==> StreamFifo, data can be move out ASAP.

Readon:
> > The good thing about that alternative design compared to m2sPipe, is that the oldest transaction will always be present at the same history index, independently of the lenght of the history, and with no latency.
> 
> I think that has to be renamed to StreamHistoryFifo due to its working policy. StreamHistory <==> History, cache chain move one by one. StreamHistoryFifo <==> StreamFifo, data can be move out ASAP.

The name that StreamAccessibleFifo might be much better. Actually, it works like a FIFO but with direct internal data access capability.
I think that could makes another PR. 
I already have a initial version. However, I need to test its resource usage and timing first.

Readon:
Now, under the configuration that has 32 bits width and 12 length.
StreamAccessibleFifo use 272 LUTs which is 258 more than StreamHistory's 14 LUTs.
The data is tested in Xilinx Vivado as an independent project.

Readon:
Also, StreamFifo use the least resources.

Dolu1990:
I think the current "StreamHistory" is too generic, as it take some very specific design/behaviour decisions.
What about renaming the current "StreamHistory" into StreamStageChain ?

> The name that StreamAccessibleFifo might be much better

I Agree

Readon:
Stage seems not visible then shift, it is actually is a shift register groups.
^^

Dolu1990:
> StreamShiftChain

Good for me ^^
Feel free to merge

PR #1423, opened 2024-05-23
Correct frequency calculation in newSlowedClockDomain

### LINKED ISSUE #1422: ClockDomain.newSlowedClockDomain uses current clock domain rather than clockdomain on which it is called
The current frequency used for calculating the factor is taken from the current clockdomain, rather than the clock domain on which the method is called.

https://github.com/SpinalHDL/SpinalHDL/blob/a4baf0df6ca1a2706ad79f6f460bd5483421968b/core/src/main/scala/spinal/core/ClockDomain.scala#L445

louiecaulfield:
`ClockDomain.newSlowedClockDomain()` was using the current clockdomain's frequency rather than the clock domain on which the method was called to calculate the resulting domain's factor.

Closes #1422

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/ClockDomain.scala
+++ b/core/src/main/scala/spinal/core/ClockDomain.scala
@@ -442,7 +442,7 @@ def renamePulledWires(clock     : String = null,
   }
 
   def newSlowedClockDomain(freq: HertzNumber): ClockDomain = {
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks :)

Issue #1426, opened 2024-05-24
Formal: support for temporal sequences


NikLeberg:
Hi there, not an issue more of a question.

Currently for formal verification SymbiYosys and Verilog/SystemVerilog are used. As documented [here](https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Formal%20verification/index.html#formal-primitives) this is somewhat _limited_ to boolean `assert` `cover` and `assume` and some variations of `past` as this is what the underlying Yosys supports.

One really great thing for formal verification would be temporal sequences. Something like (stolen from [tomverbeure](https://tomverbeure.github.io/rtl/2019/01/04/Under-the-Hood-of-Formal-Verification.html#systemverilog-assertions)):
```systemverilog
sequence a_seq_b12_seq_c
    a ##1 b[*1:2] ## c
endsequence
```
This sequence results in a match when the following happens:
 - ‘a’ evaluates to 1
 - after that ‘b’ evaluates to 1 for one or for two consecutive clock cycles
 - after that, ‘c’ evaluates to 1

One can then as example do `assert(a_seq_b12_seq_c)` or `cover(a_seq_b12_seq_c)`.

Now, similar functionality is supported in the PSL language together with VHDL. GHDL supports a large range of PSL. GHDL can then be run as [Yosys plugin](https://github.com/ghdl/ghdl-yosys-plugin). And ontop formal verification with SymbiYosys is possible see: [`tmeissner/psl_with_ghdl`](https://github.com/tmeissner/psl_with_ghdl/tree/master).

For me this sounds like a possible solution to bring temporal sequences to SpinalHDL. Has somebody tested something in this regards? Maybe even worked on anything? I'd love to help in any way. But alone I don't think I could come up with a good solution.

Happy for any comments.
Kind regards, Nik

Readon:
The existing Formal facilities is based on VerilogHDL with SymbiYosys only.
There might be bunch of things to be added and tested.

NikLeberg:
> There might be bunch of things

What do you mean by that?
 - Are there existing things that are actively being worked on? Where could I tap in to support it?
 - Or is there nothing and SpinalHDL is simply open for contributions in this regards?

Readon:
> > There might be bunch of things
> 
> What do you mean by that?
> 
>     * Are there existing things that are actively being worked on? Where could I tap in to support it?

no, as I know.

you're welcome to commit PR.

> 
[...]

PR #1429, opened 2024-05-30
formal: add yosys+ghdl backend

NikLeberg:
As discussed in #1426 I see great benefit in supporting GHDL as formal backend (actually SymbiYosys+Yosys+GHDL). This very simple PR introduces a new selectable backend `FormalConfig.withSymbiYosysAndGhdl`. By itself this is non functional. No formal verification can be done as the required VHDL/PSL generation is yet missing. It is merely a simple stepping stone in this general direction.

This is marked as WIP as it's my first PR here and I'd love for feedback on:
 - [ ] code comments?
 - [ ] function/variable naming
 - [ ] code style

related #1426
see also https://github.com/SpinalHDL/docker/pull/11 that installs the required [`ghdl-yosys-plugin`](https://github.com/ghdl/ghdl-yosys-plugin)

# Impact on code generation

If a current formal testbench uses  `FormalConfig.withSymbiYosysAndGhdl` it will break the build. SpinalHDL will try to emit VHDL for the various formal operators and the `match` statements will produce an exception on the obviously non implemented operators.

# Checklist

- [ ] Unit tests were added


core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
```diff
@@ -48,6 +48,7 @@ object FormalWorkspace {
 class SpinalFormalBackendSel
 object SpinalFormalBackendSel {
   val SYMBIYOSYS = new SpinalFormalBackendSel
+  val SYMBIYOSYS_AND_GHDL = new SpinalFormalBackendSel
```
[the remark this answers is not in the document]
NikLeberg: 👍 renamed to `SYMBIYOSYS_GHDL`

NikLeberg:
I have added some new commits that allow the use of `anyseq`, `anyconst`, `allseq` and `allconst` as supported [by GHDL and Yosys](https://symbiyosys.readthedocs.io/en/latest/verilog.html#unconstrained-variables) and also (non initial)  `assert`, `assume` and `cover` statement generation.

My implementation works with the following simple test:

```scala
import spinal.core._
import spinal.core.formal._
import spinal.tester._

case class Dut() extends Component {
  val io = new Bundle {
    val a = in port Bool()
  }
}

class FormalTest extends SpinalFormalFunSuite {
  def formalConfig = FormalConfig.withAsync

  test("FormalTest.Verilog") {
    formalConfig
      .withProve()
      .doVerify(new Component {
        setDefinitionName("FormalTestVerilog")
        val dut = FormalDut(Dut())
        anyconst(dut.io.a)
        assume(dut.io.a === True) // comment me, and the formal prove fails
        assert(dut.io.a === True)
      })
  }

  test("FormalTest.VHDL") {
    formalConfig.withSymbiYosysAndGhdl
      .withProve()
      .doVerify(new Component {
        setDefinitionName("FormalTestVHDL")
        val dut = FormalDut(Dut())
        anyconst(dut.io.a)
        assume(dut.io.a === True) // comment me, and the formal prove fails
        assert(dut.io.a === True)
      })
  }
}
```

In the above example, if the `assume` line is removed, then yosys correctly fails the test as the value 0/False fails the assert:
```
##   0:00:00  Value for anyconst in FormalTestVHDL (/28): 0
##   0:00:00  Assert failed in FormalTestVHDL: /23
```

The above code could be added as testcase.. to where? I'm not sure I understand the test folder structure and organisation. Could you @Readon maybe tell me where this could be appropriate, if it makes sense?

The implementation does not feel too _hacky_ for me ^^. But please feel free to comment on style/functionality!

Readon:
There is a shouldFail function which pass while verification failed.

You can add the vhdl verification cases in the same file as verilog do.

Readon:
your dut have not driven by clock. it's just combinational

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
```diff
@@ -23,8 +23,10 @@ package spinal.core.formal
 import java.io.{File, PrintWriter}
 import org.apache.commons.io.FileUtils
 import spinal.core.internals.{PhaseContext, PhaseNetlist}
+import spinal.core.internals.{Operator, DataAssignmentStatement}
 import spinal.core.sim.SimWorkspace
-import spinal.core.{BlackBox, Component, GlobalData, SpinalConfig, SpinalReport}
+import spinal.core.{BlackBox, Component, GlobalData, SpinalConfig, SpinalReport, SpinalError, SpinalProgress}
+import spinal.core.{noUndrivenCheck, cloneOf}
```
Readon: I think noUndrivenCheck is not necessary, if you met this, you might miss to declare the input and output as in/out or master/slave.

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
```diff
@@ -71,6 +74,11 @@ case class SpinalFormalConfig(
     this
   }
 
+  def withSymbiYosysAndGhdl: this.type = {
```
Readon, line 77: How about just use withGhdl?

core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
```diff
@@ -452,10 +453,10 @@ class ComponentEmitterVhdl(
       inc
       b ++= initialStatlementsGeneration
       dec
-      b ++= s"${tabStr}elsif ${emitClockEdge(emitReference(clock, false), clockDomain.config.clockEdge)}"
+      b ++= s"${tabStr}elsif ${emitClockEdge(emitReference(clock, false), clockDomain.config.clockEdge)} then\n"
```
Readon, line 456: Why it is necessary to change this `then`'s position?

Readon:
Maybe you could include the tester of yours, then. Or just pick up some existing formal verification checker like FormalMuxTester/FormalDemuxTester as an testcase first. Keep the consistency with the existing solution is a good start point.

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
NikLeberg: I don't see how input/output master/slave could impact this?
The functions `anyseq` `anyconst` `allseq` and `allconst` in VHDL are litterally only a signal with an attribute. No given value whatsoever and as such of course completely undriven.

Example:
```vhdl
architecture arch of FormalTestVHDL is
  attribute anyconst : boolean;
  signal toplevel_dut_1_io_a_anyconst : std_logic;
  attribute anyconst of toplevel_dut_1_io_a_anyconst : signal is true;
begin
  dut_1 : entity work.Dut
    port map ( 
      io_a => toplevel_dut_1_io_a_anyconst 
    );
  end process;
end arch;
```

core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
NikLeberg: To re-use the `emitClockEdge` function for the formal clocking statement and the usual process clock statement.
The formal clock statement looks like `@rising_edge(clk)`.
Previously I had a second almost identical function that also did
```scala
edgeKind match {
  case RISING  => "rising_edge"
  case FALLING => "falling_edge"
}
```
But instead of having two functions that do basically the same thing, I opted to slightly modify the existing one.
The `then` is now a the two places that also have the corresponding `if` which IMO is better to read anyway.

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
NikLeberg: 👍 renamed to `withGhdl`

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
Readon: so does this equal to 
``` Verilog
assign dut_delay = $anyconst(delay_width);
```
If so the current syntax is working, then the check should be pass, so that the only modification might be in Emitter of VHDL, as I understand.

NikLeberg: Hmm I see. The problem is, that in VHDL the unconstrained variables are not an assignment at all and instead a simple attribute on an otherwise undriven signal. To convert the SpinalHDL assignment to the attribute I implemented a _pre processing_ phase `PhaseFormalVhdlFixups` in `FormalBootstraps.scala` which turns the assignment to an attribute.

Initially this conversion also removed the assignment itself from the statement graph. It could be left intact, and the implementation of VHDL could just ignore the statement. Something like so:

```diff
diff --git a/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala b/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
index 22b719fba..a3bcddc94 100644
--- a/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
+++ b/core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
@@ -307,10 +307,11 @@ override def impl(pc: PhaseContext): Unit = {
           val sig = cloneOf(target)
             .addAttribute(kind)
             .setCompositeName(target, kind)
-            .addTag(noUndrivenCheck)
+            // .addTag(noUndrivenCheck)
+          target.allowOverride()
           target := sig
         }
-        s.removeStatement()
+        // s.removeStatement()
       case _ =>
     }
   }
diff --git a/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala b/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
index cad692584..3f8d8f42d 100644
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
@@ -816,6 +816,7 @@ class ComponentEmitterVhdl(
 
   def emitAssignment(assignment: AssignmentStatement, tab: String, assignmentKind: String): String = {
     assignment match {
+      case r if r.source.isInstanceOf[Operator.Formal.RandomExp] => "" // handled in FormalBootstraps.PhaseFormalVhdlFixups
       case _ =>
         if (!assignment.target.isInstanceOf[SpinalStruct])
           s"$tab${emitAssignedExpression(assignment.target)} ${assignmentKind} ${emitExpression(assignment.source)};${emitLocation(assignment)}\n"
```

But now the `allowOverride` is required as Spinal thinks that the signal is assigned twice...

core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
```diff
@@ -624,19 +677,21 @@ class ComponentEmitterVhdl(
         statement match {
           case assignment: AssignmentStatement  => b ++= emitAssignment(assignment,tab, assignmentKind)
           case assertStatement: AssertStatement =>
-            val cond = emitExpression(assertStatement.cond)
+            if (!spinalConfig.formalAsserts) {
+              val cond = emitExpression(assertStatement.cond)
 
-            require(assertStatement.message.isEmpty || (assertStatement.message.size == 1 && assertStatement.message.head.isInstanceOf[String]))
+              require(assertStatement.message.isEmpty || (assertStatement.message.size == 1 && assertStatement.message.head.isInstanceOf[String]))
```
Readon, line 683: line 683 and 685 would better not be committed.

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
Readon: So, if the expression can be identified as Formal.RandomExp then how about generate the attributes string for all Formal.RandomExp (Maybe it is in emitAttributesDef, I am not very familiar with the backends.) and skip it while emitAssignment, just like you do.

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
```diff
@@ -272,3 +290,28 @@ case class SpinalFormalConfig(
     }
   }
 }
+
+class PhaseFormalVhdlFixups extends PhaseNetlist {
```
Readon, line 294: I understand the usage of this part, how ever IMO this class is best been put into core\src\main\scala\spinal\core\internals\PhaseVhdl.scala not in FormalBootstraps.

core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
NikLeberg: What do you mean by that? The lines 683/685/687 were already existing, I simply wrapped them in the `if (!spinalConfig.formalAsserts)` conditional. They are perfectly valid, and even required, to generate the appropriate assert statements in an non-formal context (usually simulation) where report messages and severities are allowed.

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
NikLeberg: What a nice idea!
With the new commits, I moved the code to the `ComponentEmitterVhdl.scala` to the various other _preprocessing_ steps in `emitArchitecture()`. It converts tha assignments to tags/attributes which the rest of the component emitter correctly emits, and then goes on to ignore/skip the assignment in `emitAssignment()`.

Readon: So may be the steps remains should be add test cases in the existing verification tester?
you can search them by using FormalConfig keyword.

NikLeberg: I already have many test cases that test various conditions and functionalities for correctness i.e. equivalence between SystemVerilog and VHDL generation... uploading them (as I'm heavily editing them) is not good right now, especially since the CI runner will fail anyway because https://github.com/SpinalHDL/docker/pull/11 is not merged yet.

The biggest issue I currently have, is that the PSL asserts cannot live within a conditional/process. I.e. when I write:
```scala
when(False) {
  assert(False)
}
```

The assert should obviously never trigger. In Verilog it generates:
```systemverilog
  assign when_FormalApi_l379 = 1'b0;
    if(reset) begin
    end else begin
      if(when_FormalApi_l379) begin
        assert(1'b0); // FormalApi.scala:L381
      end
    end
  end
```

but for VHDL it currently only generates (manually trimmed):

```vhd
architecture arch of AssertWithinCondTest is
begin
  FormalApi_L381: assert always (pkg_toStdLogic(false) = '1') abort (reset = '1') @rising_edge(clk); -- FormalApi.scala:L381
end arch;
```

So right now I'm trying to figure out how I can walk backwards the data model. I'm at the `AssertStatement` and need to figure out in which conditional scopes this statement is, up to the root level. With this info I could then generate the necessary conditionals and AND it to the conditional of the assert itself.

Maybe @Dolu1990 can help how to walk the conditional scopes backwards/up?

Readon: ChatGPT tell me that it is
``` VHDL
property when_FormalApi_l379_implication is always (
    when_FormalApi_l379 -> next (FALSE)
);
```
If so, is it possible to convert it to be
``` VHDL
FormalApi_L381: assert always (when_FormalApi_l379 -> next (FALSE) abort (reset = '1') @rising_edge(clk);
```

Readon: > What a nice idea! With the new commits, I moved the code to the `ComponentEmitterVhdl.scala` to the various other _preprocessing_ steps in `emitArchitecture()`. It converts tha assignments to tags/attributes which the rest of the component emitter correctly emits, and then goes on to ignore/skip the assignment in `emitAssignment()`.

After having a conversation with @Dolu1990 , he thinks editing emit for formal verification code would not be a good practice then using a standalone phase.

NikLeberg: > After having a conversation with @Dolu1990 , he thinks editing emit for formal verification code would not be a good practice then using a standalone phase.

I don't know if I understand this sentence correctly: Is he saying that it would be better to not handle formal stuff in `ComponentEmitterVhdl.scala` at all? So better would be to handle in in a new phase?

Changes to `ComponentEmitterVhdl.scala` are required anyway as it does not know how to handle `Operator.Formal.<x>` yet. The Verilog/SystemVerilog emitter does know.

IMHO we could later always move out the formal specific stuff to a separate phase i.e. if it becomes too big. Currently it is only that small part that dies the transformation of random assignments to attributes.

NikLeberg: > ```vhdl
> FormalApi_L381: assert always (when_FormalApi_l379 -> next (FALSE) abort (reset = '1') @rising_edge(clk);
> ```
Yes, with properties and temporal sequences is how this can be done. A naive implementation could also just do:
```vhdl
FormalApi_L381: assert always ((not when_FormalApi_l379) or FALSE) abort (reset = '1') @rising_edge(clk);
```
But handling it with properties i.e. breaking it up in multiple lines is better for readability.

Readon: > > After having a conversation with @Dolu1990 , he thinks editing emit for formal verification code would not be a good practice then using a standalone phase.
> 
> I don't know if I understand this sentence correctly: Is he saying that it would be better to not handle formal stuff in `ComponentEmitterVhdl.scala` at all? So better would be to handle in in a new phase?
> 
> Changes to `ComponentEmitterVhdl.scala` are required anyway as it does not know how to handle `Operator.Formal.<x>` yet. The Verilog/SystemVerilog emitter does know.
> 
> IMHO we could later always move out the formal specific stuff to a separate phase i.e. if it becomes too big. Currently it is only that small part that does the transformation of random assignments to attributes.

You can have a look at #1435 as reference.

Readon: > 
> But handling it with properties i.e. breaking it up in multiple lines is better for readability.

agree，but we can implement it step by step.

maybe after define a FormalProperty traits first.

tester/src/test/scala/spinal/tester/code/FormalApi.scala
```diff
@@ -0,0 +1,418 @@
...
+    * hence the failing assert should not trigger. The condition of the failing
+    * case is true and the assert should trigger.
+    *
+    * @param backend formal backend to use, i.e. use SymbiYosys or GHDL
+    * @param doFail  false = condition is always false; true = always true
+    */
+  def runAssertWithinCondTest(backend: SpinalFormalBackendSel, doFail: Boolean = false): Unit = {
+    val config = getConfig(backend).withAsync.withBMC(3)
+
+    config.doVerify(new Component {
```
Readon, line 393: In verilog verification mode, the FormalDut is preferred to wrap around the dut.
Where the sync reset maner is guaranteed for formal verification code, but remains the design unchanged.

tester/src/test/scala/spinal/tester/code/FormalApi.scala
```diff
@@ -0,0 +1,418 @@
...
+  *  - [x] assume
+  *  - [x] cover
+  *  - [x] assert
+  *  - [ ] QOL show backtrace
+  *  - [ ] assume/assert inside conditional scope
+  *  - [x] assert during reset
+  *  - [x] assert outside of reset
+  *  - [ ] multiple clock domains
+  */
+class FormalApiTest extends SpinalFormalFunSuite {
```
Readon, line 35: How about keep the xxxxxxFunSuite naming pattern for test suite?

NikLeberg: What do you mean? This is not a class meant to be used as `FunSuite`. It is just a testing class where multiple `test()` calls are made. The current naming I have borrowed from [`BitVectorTests.scala#L31`](https://github.com/SpinalHDL/SpinalHDL/blob/499d6dca89c7c3b38cd9de1d5f4412d8166a7252/tester/src/test/scala/spinal/core/BitVectorTests.scala#L31) where it is `xxxxxxTest`.

tester/src/test/scala/spinal/tester/code/FormalApi.scala
NikLeberg: Ah, I see now how your changes in #1435 come together. But still, wrapping it in a `FormalDut` is a bit hideous in this instance. As `FormalDut` and the new `FormalPhase` only work with a hierarchy of components I'd have to write something hideous like: 
```scala
config.doVerify(new Component { FormalDut( new Component {
...
```
This is of course only because the components here are really small and simple. I think I'll just call `setFormalTester()` inside the component and then it's fine.

Readon: Ok, one thing need to notify is that the way to design the FormalDut is to split the formal verification code a part from the DUT explicitly.

tester/src/test/scala/spinal/tester/code/FormalApi.scala
Readon: Ok, I missunderstand it a API that wrap for verifications which could inherit from.

tester/src/test/scala/spinal/tester/code/FormalApi.scala
NikLeberg: Agreed. If the test (i.e. component) would be more complicated I'd also use the FormalDut to split the parts explicitly. For these very simple cases I don't.

core/src/main/scala/spinal/core/formal/FormalBootstraps.scala
NikLeberg: The newest commits did move the formal phase fixups to the new `FormalPhase` from #1435. I also managed to unroll the scope a statement is in to build the conditional for it. This is then used as a preface in the PSL property (with `|->`) such that conditional assume and asserts are possible now.

Readon: Do you think we should use the same symbol with SVA and PSL? 
The property has antecedent and consequent which act like `when() othewise` block.
Should we create a new syntax sugar for that?

NikLeberg: Which symbol do you refer to? Are you suggesting to also use the `|->` operator for SVA? I don't think that would be good as the open version of Yosys does not support that.

My current goal is simply to reach feature equivalence of verilog and vhdl backends. 

To extend that with the advanced features (temporal statements, properties etc.) I plan on doing in a future but separate PR. Step by step. :)

Readon: > To extend that with the advanced features (temporal statements, properties etc.) I plan on doing in a future but separate PR. Step by step. :)

Nice plan. the test failed now is because of the CI environment. I think we need to upgrade it first.
How about upgrade the docker image with ghdl and related plugin only? First.

Readon:
What is missing while CI? Does it require ghdl 4.1? Planning on upgrade the docker image for CI.

NikLeberg:
> What is missing while CI?

Oh, I haven't noticed that the action for pr https://github.com/SpinalHDL/docker/pull/11 did fail and you had to revert. I'll loot into it.

> Does it require ghdl 4.1?

But yes. GHDL as formal backend requires not only a recent version of GHDL (we discovered a bug in how `abort` in PSL was handled) but also requires the mentioned `yosys-ghdl-plugin`. Without that, GHDL and Yosys can't work together.

> Planning on upgrade the docker image for CI.

According to the pr [workflow run](https://github.com/SpinalHDL/docker/actions/runs/9525081205/job/26258793211) it failed because the `libgnat-13` apt package was not found. It is available for ubuntu 24... oh, I totaly havent noticed your https://github.com/NikLeberg/spinalhdl-docker/pull/1 ... sorry about that. I'll check if the lower version works

NikLeberg:
Why is `pastValid()` implemented with `initial()` instead of `init()` ?

See [`formal/package.scala#L20`](https://github.com/SpinalHDL/SpinalHDL/blob/42b22b3eea7db0d9570ba3d17c2828673c47dd3a/core/src/main/scala/spinal/core/formal/package.scala#L20)
```scala
  def pastValid() : Bool = signalCache(ClockDomain.current -> "formal.pastValid")(RegNext(True) initial(False)  setWeakName("formal_with_past"))
```

This fails in VHDL, where the initial assignment is not made. I.e. the `initial` call produces the declaration:
`signal formal_with_past : std_logic;`

Whereas using `init` this correctly procudes the initial assignment in VHDL (Verilog is unchanged, works both ways?):
`signal formal_with_past : std_logic := pkg_toStdLogic(false);`

Is there a specific reason for that?

Readon:
> Why is `pastValid()` implemented with `initial()` instead of `init()` ?
> 
> See [`formal/package.scala#L20`](https://github.com/SpinalHDL/SpinalHDL/blob/42b22b3eea7db0d9570ba3d17c2828673c47dd3a/core/src/main/scala/spinal/core/formal/package.scala#L20)
> 
> ```scala
>   def pastValid() : Bool = signalCache(ClockDomain.current -> "formal.pastValid")(RegNext(True) initial(False)  setWeakName("formal_with_past"))
> ```
> 
> This fails in VHDL, where the initial assignment is not made. I.e. the `initial` call produces the declaration: `signal formal_with_past : std_logic;`
> 
> Whereas using `init` this correctly procudes the initial assignment in VHDL (Verilog is unchanged, works both ways?): `signal formal_with_past : std_logic := pkg_toStdLogic(false);`
> 
> Is there a specific reason for that?

As my understanding, $initial is a SVA statement for init condition of time 0. init is a function set the register value after reset.

NikLeberg:
In both cases, the following verilog is emitted: 
```verilog
  initial begin
    formal_with_past = 1'b0;
  end

  always @(posedge clk) begin
    formal_with_past <= 1'b1;
  end
```

NikLeberg:
I added commit #[54e65b5](https://github.com/SpinalHDL/SpinalHDL/pull/1429/commits/54e65b59b92315a45146fede40c15e4e13010083) that implements the discussed change from `initial` to `init` for the formal `pastValid`. Maybe @Dolu1990 can help us out and explain if we would break something with it? Is seems as if he added the `initial` code infrastructure in the first place.

NikLeberg:
About CI.. I don't think your pr #1455 did what you think it did? It is now using the `latest` tag that was last pushed/released to one year ago.

Readon:
> About CI.. I don't think your pr #1455 did what you think it did? It is now using the `latest` tag that was last pushed/released to one year ago.



> About CI.. I don't think your pr #1455 did what you think it did? It is now using the `latest` tag that was last pushed/released to one year ago.



> About CI.. I don't think your pr #1455 did what you think it did? It is now using the `latest` tag that was last pushed/released to one year ago.

The lastest tag point to the latest released version. 
After fully test the docker image we can release the image and it will work though.
Now, it's not necessary. IMO.

Readon:
you can use the master version to test first.

NikLeberg:
Ah, so I have to change the image tag that is used in the action yaml? Can do... that commit shall never enter dev though. I can also test locally. What do you prefer?

Readon:
> Ah, so I have to change the image tag that is used in the action yaml? Can do... that commit shall never enter dev though. I can also test under the same conditions as in CI locally. What do you prefer?

It's ok, before merge we should check all modified files first.

NikLeberg:
Hi @Readon I think implementation wise this is done. I'm passing all tests locally with latest upstream GHDL version #[3b05281](https://github.com/ghdl/ghdl/commit/3b05281987f194478fea538a843411c775931a09). This latest GHDL version is required as there was a problem with failing prove tasks that was fixed. I'll open a PR in Spinal/docker.

About this PR: Where and how much documentation shall be added?

NikLeberg:
The failures in `tester-formal` were because:
 - I forgot to change an assert in `FormalApiTest` -> fixed
 - CI did not yet use the updated (https://github.com/SpinalHDL/docker/pull/15) docker image -> a re-run should pick up the new image and then pass

Readon:
Sorry for the late reply. I just want to release the docker image first then merge this before.
However, it seems that ghdl would not release frequently. So we should do it in our steps.

Now, I all there is a patch for nfa corruption at ghdl/ghdl#2683 .
We can update the docker image to that version, and then with this PR.

Readon:
Now all environment has been prepared, and I tried it on my own machine.
One problem we need to concern about is that all ghdl supported formal verification runs in formal related jobs, so that the running time of CI would increase a lot due to long verification time.
Is it possible to split the ghdl related test case into standalone CI jobs? @NikLeberg 
Thanks.

NikLeberg:
> Is it possible to split the ghdl related test case into standalone CI jobs?

Currently there is no difference made between default symbiyosys or ghdl formal tests. They just run sequentially one by one.

I'm no expert in sbt and frequently have to fight with it to do the expected thing. So is there a general way to filter tests? Other testing frameworks I know from e.g. C have the possibility to _tag_ testcases and then later exclude / include special tagged cases from running. Has sbt something similar?

PR #1435, opened 2024-06-03
Verify component without withAsync option with synchronous reset as default.

### LINKED ISSUE #1430: Formal proofs started to fail with newer versions of sby
After updating to a newer version of yosys (Latest oss-cad-suite version), many of my proofs started to fail with the following error in the logfile:
`ERROR: $check cell $cover$anon.sv:42$9 with TRG_WIDTH > 1 is not support by async2sync, use clk2fflogic.`

Here is a small example which reproduces the issue:
```
object Formal extends App {
  FormalConfig.withDebug
    .withBMC(40)
    .withProve(40)
    .withCover(40)
    .doVerify(
      new Component {
        val a = Stream(UInt(1 bit))
        val r = Stream(UInt(1 bit))

        r << a
        cover(
          a.fire && r.fire && past(
            a.fire && r.fire && past(a.fire && r.fire && pastValid())
          )
        )
      }
    )
}

```

Readon:
Closes #1430

# Impact on code generation

# Checklist

- [ ] Unit tests were added


Readon:
This might lead to a very slow formal verification in the lib itself.

Readon:
The checker fails because of "multiclock on" statement would lead to an erron initial value while prove.
In theory, the initial value in prove should follow the assert and assume condition. However, when multiclock on the prove tools would not follow the initial condition.
This is tested on existing implementation and current version of yosys.

Readon:
@Dolu1990 How can I change the default clockDomain's reset kind in Component?

Readon:
Now the existing verification are mostly in synchronous mode, which is faster.
I would test if it works in latest yosys.

Readon:
As tested in yosys 0.41 [here](https://github.com/Readon/SpinalHDL/actions/runs/9352794782), the synchronous design have been passed. 
However, it failed on FifoCC tester, which turned multiclock on in a strange mode.
The waveform shows that it fails during the second reset time, the assertion does not follow the reset check priority.
It's not a proper time to upgrade to the latest version, IMO.

Readon:
I have also moved the ModuleAnalyzer and DataAnalyzer from lib to core directory, because there might be some analysis could be easily done by those tools, also in core.

Readon:
OK, all the verification testers have been run through under the Yosys 0.41.

PR #1437, opened 2024-06-04
inline literal input for some case

yportne13:
commit 075e654c1 makes all the literal input not inline. this commit inline literal input for some case

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitter.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitter.scala
@@ -99,7 +99,20 @@ abstract class ComponentEmitter {
   }
 
   def isSubComponentInputBinded(data: BaseType) = {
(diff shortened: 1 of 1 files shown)


jijingg:
@yportne13 could you add some demo codes so that others can clearly handle what cases

yportne13:
```scala
object TestInline extends App {
  SpinalVerilog(new Component{
    val sub = new Component {
      val x = in Bits(1 bits)
    }
    sub.x.setAll()
  })
}
```
origin:
```verilog
module unamed_1 (
);

  wire       [0:0]    sub_x;

  unamed sub (
    .x (sub_x)  //i
  );
  assign sub_x = 1'b1;

endmodule
```
this commit:
```verilog
module unamed_1 (
);


  unamed sub (
    .x (1'b1)  //i
  );

endmodule
```

PR #1443, opened 2024-06-07
a little improve on interface

yportne13:
add a basic support for bundle in interface.
add a basic sim for interface
If there's no hurry to release a new version, this PR can be kept open for now and will include a little more improvements and test cases.


Dolu1990:
Let's me know when you want a review ^^

yportne13:
one last thing, for Bundle, I can call `parent` on the element to get that Bundle. But for Vec, what should I do?

Dolu1990:
The push just above will now set parent of vec aswell

yportne13:
done @Dolu1990

PR #1444, opened 2024-06-10
add missing match case for ElabOrderId.getName

KireinaHoro:
`ElabOrderId.getName` does not assign a name to the patch phase, which would cause a match error during elaboration.


andreasWallner:
thanks!

PR #1448, opened 2024-06-11
Constraint writer: use regex to match for net names

KireinaHoro:
Noticed a commit that I have locally and forgot to push to the PR branch for the constraint writer.  We need to use hierarchical names to find nets, when the generated SpinalHDL module is not the top-level module.


Dolu1990:
Thanks ^^

Issue #1445, opened 2024-06-10
Support for AMBA ACE4


KireinaHoro:
I'm looking at adding some support for ACE 4 on top of the `amba4.axi` classes we already have in `spinal.lib`, but I'm not super sure on what's the best way to proceed here.  ACE adds two extra channels `AC` and `CR`, but also extra signals to `AR`, `AW` and `W` signals on the original AXI definition.  Should I

- derive `Axi4Ax` and `Axi4R` channels and add the ACE signals, or
- add flags like `useSnoop` on the `Axi4Ax` object?

Dolu1990:
I would say, adding withAce in the Axi4Config, then, hmm, are the signal added to AR and AW exactly the sames ?

KireinaHoro:
![grafik](https://github.com/SpinalHDL/SpinalHDL/assets/7804730/388aaf96-3e24-4c67-a440-f4552de6422b)

In a word, the snoop signals specify the read and write transactions as cache transactions that would also enforce the cache state of a memory location.  For example, a master can issue ReadShared in a read transaction via indicating it through `ARSNOOP`; the interconnect will broadcast ReadShared on the `AC` channel to masters in the same shareable domain to fulfill this requirement.

Dolu1990:
Ahhh sad XD
So, adding withAce to the Axi4Config, and adding all the related channels to Axi4, aswell as additional signals to Axi4Ar / Axi4Aw should be fine i think.  ?


Did you considered Tilelink (<3<3) ?

KireinaHoro:
Ah so basically not having a `amba4.ace`, but just use `Axi4` directly?

Yeah I wanted to, but I'm interfacing with the PS on a Zynq UltraScale+, which has an ACE port :/

Dolu1990:
> but just use Axi4 directly?

Yes
I mean in some way, you can see AXI ace as an extentions of AXI so, why not ^^

> which has an ACE port :/

Ahhhh damned XD
i didn't knew they were providing ace support to the fpga fabric XD
Do they specify which subset of AXI ACE they use / you should support ? 
Because that's a biiiiiiiiiiiiig spec with a loooooooooooooooot of variants.

KireinaHoro:
> i didn't knew they were providing ace support to the fpga fabric XD Do they specify which subset of AXI ACE they use / you should support ?

[...]

PR #1450, opened 2024-06-12
ACE & ACE-Lite signals

KireinaHoro:
As discussed in #1445, this is still a draft.

- [x] basic signals for ACE4 and ACE4-Lite
- [x] encodings for enums
- [x] basic test for signals (checked locally, test not pushed)
- [ ] `>>` operator, upgrade & downgrade between ACE and ACE-Lite
- [ ] test if it connects to the Zynq ACE port
- [ ] sample app to see if it actually works
- [ ] better abstraction?
- [ ] stretch goal: a bridge to/from TL-C?

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
@@ -29,7 +29,15 @@ package spinal.lib.bus.amba4.axi
 import spinal.core._
 import spinal.lib._
 
-
+case class Ace4Config(isAceLite       : Boolean = true,
+                      useSnoopData    : Boolean = false,
+                      useAwUnique     : Boolean = false,
+                      snoopDataWidth  : Int = -1,
+                     ) {
+  if (useSnoopData)
+    require(List(32, 64, 128, 256, 512, 1024) contains snoopDataWidth,
+      "Valid snoop data width: 32, 64, 128, 256, 512 or 1024 bit")
+}
 
 /**
  * Configuration class for the Axi4 bus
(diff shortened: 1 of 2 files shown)


lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
```diff
@@ -298,6 +326,91 @@ object Axi4{
     def DECERR = B"11" // Decode error
   }
 
+  // ARM IHI 0022E, Table C3-7
+  object arsnoop {
```
Dolu1990, line 330: Arsnoop ?

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
```diff
@@ -70,6 +80,10 @@ case class Axi4Config(addressWidth : Int,
   def useArwUser = arwUserWidth >= 0 //Shared AR/AW channel
   def arwUserWidth = Math.max(arUserWidth, awUserWidth)
 
+  def isAce = aceConfig.nonEmpty
```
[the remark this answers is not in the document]
KireinaHoro: done

lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4.scala
KireinaHoro: Apart from the AC/CR/CD snoop channels for a **snooped** master, ACE and ACE-Lite also adds ARSNOOP and AWSNOOP signals for a **snooping** master.  This (along with the `awsnoop` object) define encodings for these fields.

PR #1456, opened 2024-06-16
add driveTo method return stream directly.

### LINKED ISSUE #1446: About convenient anonymous bundle.
I am trying to practice on some full stream design, where need a bunch of interface data type between stream, almost each of the stream stage.
So that I want to define some internal data type in anonymous bundle as below, but not `case class` for all cases.
``` Scala
val chainIn = input.translateWith(new Bundle { val recipe = inRecipe; val totalLen = recipeTotalLen })
```
However, this code would lead to error "Spinal can't clone class XXXXX datatype".
According to some issues, it is caused by the clone is not clone able which is required by Stream.

Is there any other convenient way to define this?

Readon:
Closes #1446
return stream directly by driveTo which simplify usage.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/misc/pipeline/Node.scala
+++ b/lib/src/main/scala/spinal/lib/misc/pipeline/Node.scala
@@ -150,20 +150,20 @@ trait NodeApi extends NodeBaseApi {
     con(that.payload, getNode)
   }
 
-  def driveTo[T <: Data](con: (Node) => T): Stream[T] = {
+  def driveTo[T <: Data](that: Flow[T])(con: (T, Node) => Unit): Unit = {
+    arbitrateTo(that)
+    con(that.payload, getNode)
+  }
+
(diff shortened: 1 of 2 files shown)


Dolu1990:
Hi ^^

Hmm i think the name would need to change.

instead of (for instance)
```scala
def driveTo[T <: Data](con: (Node) => T): Stream[T] = {

We could have : 

def toStream[T <: Data](con: (Node) => T): Stream[T] = {
```

Readon:
> Hi ^^
> 
> Hmm i think the name would need to change.
> 
> instead of (for instance)
> 
> ```scala
> def driveTo[T <: Data](con: (Node) => T): Stream[T] = {
> 
> We could have : 
> 
> def toStream[T <: Data](con: (Node) => T): Stream[T] = {
> ```

Nice, better than driveTo.

Issue #1441, opened 2024-06-06
PackedBundle EndiannessSwap


oletf:
Hi, 

I'm currently trying to have a `PackedWordBundle` which contains an `UInt`, 
but I'd need this word's bytes endianness to be swapped in its packed form.

I've tried to have assignments done on a `signalCache` without much success...
```scala
case class TestPack() extends PackedWordBundle(32 bits) {
  private val word1BE = UInt(32 bits).packFrom(0).inWord(0)
  def word1 = signalCache(this, "word1")(EndiannessSwap(word1BE))
  //...other packed things
}
```
tried here to only use assignments on `word1` so the actual UInt word1BE gets the swapped signal, which is used once `packed` is called.
(getting `ASSIGNMENT OVERLAP` on the first assignment)

I probably missed an easier way to get what I need in Spinal, 
also not sure about how I should use signalCache in this situation if needed...

Any idea on how I could get this ?
 
thanks forward

andreasWallner:
As far as I can tell `EndianessSwap` is not from the SpinalHDL library so I have to guess a bit.

The most likely reason is that EndianessSwap returns a new signal that has the bits assigned with swapped endianess. So if you try to drive `myPack.word1 := ...` somewhere in your code you are not describing a driver for `word1BE` but a non-conditional driver for `word1` alone which conflicts with the driver inside `EndianessSwap`.

Many of the functions on UInt don't work "in both directions" (pretty much only the "Bit extraction" operations from the docs do). The `signalCache` does not help here as the issue already occurs on the first usage.

What's your use-case why you need the swapped endianess?
Do you just want to bit-order to change? `pack` has an optional parameter for this. If you want to byte-order to change this is currently not supported OOTB by PackedBundle as far as I know.

oletf:
> As far as I can tell `EndianessSwap` is not from the SpinalHDL library so I have to guess a bit.

it's from lib.Utils : [`EndianessSwap`](https://github.com/SpinalHDL/SpinalHDL/blob/987cc87b8eedf8a833373ca13b8f62ecc8ce1df6/lib/src/main/scala/spinal/lib/Utils.scala#L461-L477).

> Many of the functions on UInt don't work "in both directions" (pretty much only the "Bit extraction" operations from the docs do). The signalCache does not help here as the issue already occurs on the first usage.

Thanks for the insight, I'm still lacking understanding of internals of Spinal.

> What's your use-case why you need the swapped endianess?

[...]

PR #1457, opened 2024-06-16
Packed bundle assignment bug

dokleina:
Updated PackedBundle's `packed` method to avoid Assignment Overlap errors when the Bundle contains only a single field.

As demonstrated in #1441, a `PackedBundle` with a single field can cause an `ASSIGNMENT OVERLAP` error. This is likely because assignment completely re-assigns an internal Bits buffer object, whereas in the case of multiple fields, no one single field overwrites the entire buffer. Adding `allowOverride` to the buffer clears up the issue.

# Impact on code generation

None anticipated or expected.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/PackedBundle.scala
+++ b/lib/src/main/scala/spinal/lib/PackedBundle.scala
@@ -77,7 +77,7 @@ class PackedBundle extends Bundle {
 
   def packed: Bits = {
     val maxWidth = mappings.map(_._1.high).max + 1
-    val packed = B(0, maxWidth bit)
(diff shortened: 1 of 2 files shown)


dokleina:
I don't understand why `lib-test` failed, it almost looks like it might have timed out or something. Nothing I added in this PR should have broken it.

Dolu1990:
It seems that the CI was canceled by some ways i rerun it, your changes seems good to me aswell.

Dolu1990:
Thanks ^^

Issue #1463, opened 2024-06-21
formal: GlobalClock() produces invalid VHDL


NikLeberg:
When the `spinal.core.formal.GlobalClock()` is used, an invalid name is given to the signal in VHDL.
Which generates with VHDL to:
```vhdl
signal _global_clk : std_logic;
attribute gclk of _global_clk : signal is true;
signal _global_reset : std_logic;
```
In VHDL, leading `_` are not allowed.

This was not previously surfaced as VHDL was not a formal target/backend so far. Soon it could become one though.
Maybe drop the `_` in the internal clock definition?

Readon:
> When the `spinal.core.formal.GlobalClock()` is used, an invalid name is given to the signal in VHDL. Which generates with VHDL to:
> 
> ```vhdl
> signal _global_clk : std_logic;
> attribute gclk of _global_clk : signal is true;
> signal _global_reset : std_logic;
> ```
> 
> In VHDL, leading `_` are not allowed.
> 
> This was not previously surfaced as VHDL was not a formal target/backend so far. Soon it could become one though. Maybe drop the `_` in the internal clock definition?

if I understand right, after _global_clk is changed to global_clk it would work?
However, how would it be to sperate the global clock from others? Would there be any conflict on the global clock with normal clock names?

NikLeberg:
> if I understand right, after _global_clk is changed to global_clk it would work?

Yes

> However, how would it be to sperate the global clock from others? Would there be any conflict on the global clock with normal clock names?

I don't know... the default ones from spinal are named `clk` and `reset`. So I don't think there is any conflict?

Readon:
> > if I understand right, after _global_clk is changed to global_clk it would work?
> 
> Yes
> 
> > However, how would it be to sperate the global clock from others? Would there be any conflict on the global clock with normal clock names?
> 
> I don't know... the default ones from spinal are named `clk` and `reset`. So I don't think there is any conflict?

'clk' and 'reset' are only default name. There could be custom names there.

NikLeberg:
I added commit # [2391508](https://github.com/SpinalHDL/SpinalHDL/pull/1429/commits/2391508638fe9283acddb7543189804e68a9c0ba) to PR #1429 that removes the underline.

Local testing showed no tests to fail. I think only if a user previously named their clocks `global` would this break anything.

NikLeberg:
[...]

PR #1468, opened 2024-06-28
validate component and declaration naming

NikLeberg:
related #1463

Spinal now checks the names of components and declarations for validity in the currently active language.
If an invalid name is found, it fails the generation.
If an invalid name in another than the currently active language is found, it reports a warning.

# Impact on code generation

Spinal will now refuse to generate code that has invalid identifiers. Previously Spinal did just generate it and at a later step the user had to find out (through GHDL, Quartus, etc.) that the generated code was invalid.

# Checklist

- [x] Unit tests were added
- [x] ~API changes are or will be documented:~ -> no changes to API

### DIFF
--- a/core/src/main/scala/spinal/core/internals/Phase.scala
+++ b/core/src/main/scala/spinal/core/internals/Phase.scala
@@ -2481,6 +2481,40 @@ class PhaseAllocateNames(pc: PhaseContext) extends PhaseMisc{
     for (c <- sortedComponents.reverse) {
       c.allocateNames(pc.globalScope)
     }
+
+    val invalidVhdlIdentifier = List (
+      ("[^\\w]".r, "contains non-alphanumeric characters"),
+      ("_$".r, "ends with an underscore"),
+      ("__".r, "contains a double underscore"),
+      ("^[^a-zA-Z]".r, "doesn't start with a letter")
+    )
+    val invalidVerilogIdentifier = List (
+      ("[^\\w$]".r, "contains non-alphanumeric or dollar characters"),
+      ("^[^a-zA-Z_]".r, "doesn't start with a letter or underscore")
+    )
+    def checkName(namedObj: Nameable): Unit = {
+      var name = namedObj.getName()
+      if (!name.startsWith(pc.globalData.anonymSignalPrefix)) {
+        for ((regex, reason) <- invalidVhdlIdentifier) {
+          if (regex.findFirstIn(name).exists(_ => true)) {
+            val msg = s"Name of $namedObj is invalid in VHDL because it $reason."
+            if (pc.config.mode == VHDL) SpinalError(msg)
+            else SpinalWarning(msg)
+          }
+        }
+        for ((regex, reason) <- invalidVerilogIdentifier) {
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks :D

The one thing i'm a bit worrying is the execution overhead of regex.
I realy have no idea how much processing it cost.
Do you have any idea ?

Readon:
A rough comparison can be carried by testing time of SpinalHDL.
This is normal one.


Here is the one that test time of this PR.


Slightly increase but not that much.

NikLeberg:
Generally the overhead of the RegEx(es) should be low. It does not use any repeats (i.e. `+` or `*` in the pattern) and as such does not have to do backtracking. And backtracking is the _costly_ thing in regex. So it should be fine I guess?

I tried to give a quantitative answer and ran multiple iterations of a Spinal generation, with and without the proposed changes. But the JIT compilation of Scala optimized it so well that it resulted in roughly identical runtimes for both cases. 🤷‍♂️

core/src/main/scala/spinal/core/internals/Phase.scala
```diff
@@ -2481,6 +2481,40 @@ class PhaseAllocateNames(pc: PhaseContext) extends PhaseMisc{
...
+
+    val invalidVhdlIdentifier = List (
+      ("[^\\w]".r, "contains non-alphanumeric characters"),
+      ("_$".r, "ends with an underscore"),
+      ("__".r, "contains a double underscore"),
+      ("^[^a-zA-Z]".r, "doesn't start with a letter")
+    )
+    val invalidVerilogIdentifier = List (
+      ("[^\\w$]".r, "contains non-alphanumeric or dollar characters"),
+      ("^[^a-zA-Z_]".r, "doesn't start with a letter or underscore")
```
Dolu1990, line 2493: > or underscore

Isn't this a VHDL rule ? Starting with underscore is often used in verilog, this for instance allow to make a signal invisible in verilator.

Dolu1990: Also, if this is moved to the VHDL rullset, a new issue appear that all the _zz_???? generated in the Verilog mode will trigger it as warning :)

NikLeberg: It tests the _negative_. I.e. the pattern `^[^a-zA-Z_]` matches with any string/name that does not start with either an underscore or an letter. If there is a match, then an error is generated. So the checks pass if there is _no_ match.

So it is at the correct place with the verilog rules.

NikLeberg: The line `if (!name.startsWith(pc.globalData.anonymSignalPrefix)) {` should prevent any of the automatically generated anonymous signals to trigger warnings or errors... Does it not? 😅

Dolu1990: Hoooooooooooo right XD
my bad. thanks ^^

Dolu1990:
Thanks ^^

NikLeberg:
It's a pleasure :)

PR #1475, opened 2024-07-11
RFC: Add blackboxing for simple 1-port synchronous ROMs

bunnie:

The motivation for this is an issue I experienced in a recent tape-out. We added the AES extensions to a Vex configuration. The simulation tools indicated no problems with the `$readmemb` style ROM, but for complicated reasons this was a lie (the actual ROM was all-`X` but due to the way the compare loop ran in the test bench, it would not fail on `X`). 

In the end, the synthesis tool really either needed a black boxed ROM, or the ROM presented as a `case` statement in RTL.

This PR attempts to generate exactly that.

# Impact on code generation

In the case that no black boxing is requested, the generated code should be unaffected. However, a black box model of the ROM will be generated even if it is not used.

In the case that black boxing is requested, the ROM is replaced with a module that instantiates the black boxed ROM file, and a .v file that is an RTL model of the ROM. The name of the module and the ROM file I think is unique per ROM -- unlike the RAM black boxes where a single model can be re-used so long as the capacity is the same, identical capacity ROMs are still different models since they should have different contents.

Unfortunately, I could not figure out how to get `ComponentEmitterVerilog.scala`'s `emitMem()` routine to detect if the memory is a black box or not. I think the blackbox attribute is attached to a `Component` but `Mem` does not inherit `Component` properties, so I *think* it's not actually possible to tell between the two cases.

So, what ends up happening is you end up with both a `.bin` and a `.v` description of the ROM, and in the case that you are using a black box, there is still a dummy line that tries to `$readmemb` the `.bin` file, into a variable that isn't used. I ran this through verilator and it seemed harmless.

Note that in the case that a blackbox is requested, you actually do want both the `.v` blackbox and the `.bin` file, because the `.bin` file is used to generate the hard-macro for the ASIC, so both artifacts are in fact useful in that situation. The problem is more that the code is also generated to read in the .bin file to a useless variable, and when no black box is requested, you get this extra .v artifact that's not used by the design.

# Checklist

I'm happy to try and figure out how to add unit tests and API docs if the PR seems useful. However, I'm not even sure if this is the approach you'd really want to take to generate the black box model. I'm also happy to just keep this in my branch and pull it forward over time. Note that the block box model has at least been verified on the AES use case!

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/internals/Phase.scala
+++ b/core/src/main/scala/spinal/core/internals/Phase.scala
@@ -898,7 +898,26 @@ class PhaseMemBlackBoxingDefault(policy: MemBlackboxingPolicy) extends PhaseMemB
     }
 
     if (mem.initialContent != null) {
-      return "Can't blackbox ROM"  //TODO
+      if (topo.portCount == 1 && topo.readsSync.size == 1 && topo.readWriteSync.isEmpty) {
+        mem.component.rework {
+          val port = topo.readsSync.head
+
+          val rom = new Rom_1rs(
+            wordWidth = port.width,
+            wordCount = mem.wordCount*mem.width/port.width,
+            technology = mem.technology
+          )
+
+          rom.io.addr.assignFrom(port.address)
+          rom.io.en.assignFrom(port.clockDomain.isClockEnableActive)
+
+          wrapConsumers(port, rom.io.data)
+
+          rom.setName(mem.getName())
+        }
+      } else {
+        return "Can't blackbox ROM" // TODO
+      }
       //      } else if (topo.writes.size == 1 && topo.readsAsync.size == 1 && topo.portCount == 2) {
     } else if (topo.writes.size == 1 && (topo.readsAsync.nonEmpty || topo.readsSync.nonEmpty) && topo.writeReadSameAddressSync.isEmpty && topo.readWriteSync.isEmpty) {
       mem.component.rework {
--- a/core/src/main/scala/spinal/core/MemBlackBox.scala
+++ b/core/src/main/scala/spinal/core/MemBlackBox.scala
@@ -259,6 +259,32 @@ class Ram_1wrs(
   noIoPrefix()
 }
 
+/**
+ * Rom 1rs
+ */
+class Rom_1rs(
+                val wordWidth      : Int,
+                val wordCount      : Int,
+                val technology     : MemTechnologyKind
+              ) extends BlackBox {
+
+  addGenerics(
+    "wordCount"      -> Rom_1rs.this.wordCount,
+    "wordWidth"      -> Rom_1rs.this.wordWidth,
+    "technology"     -> Rom_1rs.this.technology.technologyKind
+  )
+
+  val io = new Bundle {
+    val clk    =  in Bool()
+    val en     =  in Bool()
+    val addr   =  in UInt(log2Up(wordCount) bit)
+    val data   = out Bits(wordWidth bit)
+  }
+
+  mapCurrentClockDomain(io.clk)
+  noIoPrefix()
+}
+
 
 /**
   * Ram 2wrs
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -369,7 +369,15 @@ class ComponentEmitterVerilog(
     for (child <- component.children) {
       val isBB             = child.isInstanceOf[BlackBox] && child.asInstanceOf[BlackBox].isBlackBox
       val isBBUsingULogic  = isBB && child.asInstanceOf[BlackBox].isUsingULogic
-      val definitionString =  if (isBB) child.definitionName else getOrDefault(emitedComponentRef, child, child).definitionName
+      val definitionString =  if (isBB)
+        {
+          if (child.definitionName == "Rom_1rs") {
+            // each ROM blackbox needs a unique name since it contains the ROM data
+            s"${child.name}_${child.definitionName}"
+          } else {
+            child.definitionName
+          }
+        } else getOrDefault(emitedComponentRef, child, child).definitionName
 
       val instanceAttributes = emitSyntaxAttributes(child.instanceAttributes)
 
@@ -1248,13 +1256,13 @@ class ComponentEmitterVerilog(
             logics ++= s"    ${emitReference(mem, false)}[$index] = ${filledValue.length}'b$filledValue;\n"
           }
         }
-      }else {
-
-
+      } else {
         val withSymbols = memBitsMaskKind == MULTIPLE_RAM && symbolCount != 1
         for (i <- 0 until symbolCount) {
+          // TODO: fix for multiple symbols
           val symbolPostfix = if(withSymbols) s"_symbol$i" else ""
           val builder = new mutable.StringBuilder()
+          val v_builder = new mutable.StringBuilder()
           for ((value, index) <- mem.initialContent.zipWithIndex) {
             val unfilledValue = value.toString(2)
             val filledValue = "0" * (mem.getWidth - unfilledValue.length) + unfilledValue
@@ -1262,9 +1270,11 @@ class ComponentEmitterVerilog(
               builder ++=  s"${filledValue.substring(symbolWidth * (symbolCount - i - 1), symbolWidth * (symbolCount - i))}\n"
             } else {
               builder ++= s"$filledValue\n"
+              v_builder ++= s"            'd$index: data <= 32'b$filledValue;\n"
             }
           }
 
+          // Emit a .bin file template for the ROM
           val romStr = builder.toString
           val relativePath = romCache.get(romStr) match {
             case None =>
@@ -1279,8 +1289,58 @@ class ComponentEmitterVerilog(
               file.getName
             case Some(x) => x
           }
-
           logics ++= s"""    $$readmemb("${relativePath}",${emitReference(mem, false)}${symbolPostfix});\n"""
+
+          // Emit a verilog blackbox template for the ROM
+          val template =
+            """
+              |`resetall
+              |`timescale 1ns / 1ps
+              |`default_nettype none
+              |
+              |module {prefix}_Rom_1rs #(
+              |    parameter wordCount = 512,
+              |    parameter wordWidth = 32,
+              |    parameter addrWidth = $clog2(wordCount)
+              |)
+              |(
+              |    input  wire                             clk,
+              |    input  wire                             en,
+              |    input  wire [addrWidth - 1:0]           addr,
+              |    output reg  [wordWidth - 1:0]           data,
+              |);
+              |
+              |always @(posedge clk) begin
+              |    if (en) begin
+              |        case (addr)
+              |{romvals}
+              |
+              |            default: data <= 32'h0;
+              |        endcase
+              |    end else begin
+              |        data <= data;
+              |    end
+              |end
+              |
+              |endmodule
+              |`resetall
+              |""".stripMargin
+
+          val refFullName = emitReference(mem, false)
+          val lastUnderscoreIndex = refFullName.lastIndexOf('_')
+          val baseName = if (lastUnderscoreIndex != -1) refFullName.substring(0, lastUnderscoreIndex) else refFullName
+
+          val content = template
+            .replace("{romvals}", v_builder.toString)
+            .replace("{prefix}", s"${baseName}")
+
+          val filePath = s"${pc.config.targetDirectory}/${baseName}_Rom_1rs.v"
+          val file = new File(filePath)
+          emitedRtlSourcesPath += filePath
+          val writer = new java.io.FileWriter(file)
+          writer.write(content)
+          writer.flush()
+          writer.close()
         }
       }
 


Dolu1990:
Hi,

Maybe the PR has 2 things which should be decoupled (blackboxing / rom generation using case statement) ?
Meaning that emitting rom using case statement could be a SpinalConfig parameter. 

Did you tried the SpinalConfig(inlineRom = true).generateVerilog(...) ? It will avoid using readmemb and instead do things as : 

```verilog
  reg [7:0] rom [0:15];

  initial begin
    rom[0] = 8'b00000000;
    rom[1] = 8'b00000001;
    rom[2] = 8'b00000010;
    rom[3] = 8'b00000011;
    rom[4] = 8'b00000100;
    rom[5] = 8'b00000101;
    rom[6] = 8'b00000110;
    rom[7] = 8'b00000111;
    rom[8] = 8'b00001000;
    rom[9] = 8'b00001001;
    rom[10] = 8'b00001010;
    rom[11] = 8'b00001011;
    rom[12] = 8'b00001100;
    rom[13] = 8'b00001101;
    rom[14] = 8'b00001110;
    rom[15] = 8'b00001111;
  end
```

I'm not sure how well synthesis tools eat this kind of patterns.

Anyway, i'm fine with the idea of having rom blackbox, aswell as having an option to generate rom as case statement.
Would it be fine to have those 2 things decoupled or there is a reason behind it ? (just asking, i can modify the PR :))

bunnie:
Synthesis tools for this ASIC flow won't respect any "initial begin" statements, unfortunately, so the use of `initial begin` to setup the ROM won't work. Has to be a "pure RTL" with no initial state for the tools to build it.

Even worse, the tools will happily take the `initial begin` statements without warning or error, and just give you X in simulation and no components in the output, which is a big difference between verilator behavior and synopsis flow behavior.

No reason to couple them. I don't have a clear concept of the SpinalHDL internal modular abstraction so in my head the two halves (black boxing and code gen) would be linked. But now that you mention it, I see that there is no actual link between those two files.

I'm happy to rework the PR into two PRs, but, before I do that, do you know if there is a way I can detect if a ROM is blackboxed at the codegen stage? I think a `Mem` object doesn't know if it is blackboxed, so I emit both kinds of code, regardless of whether blackboxing is on or not.

Dolu1990:
So between the two feature, i would say "truely" blackboxing rom is kinda tricky, as you need to feed somehow some uniquness to the blackbox itself. 
So i would say, let's keep this part out of the scope for now, and just focus on getting the SpinalConfig(switchRom=true)  parameter added and functional ^^ (switch refer to Verilog case). This would be an alternative to the pre-existing inlineRom: 
https://github.com/SpinalHDL/SpinalHDL/blob/2a34c667e8b72568a8b6c25f4a926863b8e068ed/core/src/main/scala/spinal/core/Spinal.scala#L149

> do you know if there is a way I can detect if a ROM is blackboxed at the codegen stage?

So, the intended design of the memory blackboxing in SpinalHDL, is that the Mem declaration removed (via the removeMem function, which isn't in call in your PR, but should be, see : 
https://github.com/SpinalHDL/SpinalHDL/blob/ba79ca0136aef8915abfeab477ef6024fb5e1f79/core/src/main/scala/spinal/core/internals/Phase.scala#L993) and its behaviour replaced by a new component instance.

To keep trace of which Mem the blackbox replace, what is done, is add a special tag to the new blackbox, see for example : 
https://github.com/SpinalHDL/SpinalHDL/blob/ba79ca0136aef8915abfeab477ef6024fb5e1f79/core/src/main/scala/spinal/core/internals/Phase.scala#L975

Which can then be retreived at generation via 

```scala
  myMem.getTag(classOf[MemBlackboxOf]).get.mem
```

Let's me know if you need any help of any sort :)

Regards
Charles

bunnie:
Got it. I'll try to take a swag at splitting this up and re-doing the PR. It's good for me to get some practice at SpinalHDL, this has been a nice exercise so far for me to get more familiar with the system.

Might take me a couple days before I can get to re-doing this, when I put the new ones up I'll close this one.

Dolu1990:
Ok ^^

The verilog backend code is able for a few memory layout to handle mixed width / byte enable memories.
This complexify quite a bit the code which emit the verilog, but in your case, for ROM, you don't need to care about those, you can just assert those case out.

Let's me know if you have any issue

core/src/main/scala/spinal/core/internals/Phase.scala
```diff
@@ -898,7 +898,26 @@ class PhaseMemBlackBoxingDefault(policy: MemBlackboxingPolicy) extends PhaseMemB
...
+            wordCount = mem.wordCount*mem.width/port.width,
+            technology = mem.technology
+          )
+
+          rom.io.addr.assignFrom(port.address)
+          rom.io.en.assignFrom(port.clockDomain.isClockEnableActive)
+
+          wrapConsumers(port, rom.io.data)
+
+          rom.setName(mem.getName())
```
bunnie, line 916: I've got a question about how to approach a problem I'm having refactoring this.

I understand that there should be a `removeMem()` call right around here.

However, when I do that, `emitMem()` is no longer called. 

In order to generate `emitMem()`, I need the phase context and some other state in order to figure out the name of the file and directory to put the black box in. 

I think the conflict lies in this difference:
- black box RAM models are meant to be static files provided by the user, and thus no generation is required
- for the black box ROM model, I'd like it to be generated by the system, because the ROM contents are actually inside SpinalHDL.
- Furthermore, I'd like the black box itself to contain a `case` statement, so that if there is a failure to replace the blackbox model with an actual hardware macro, in the worst case the ROM is tossed directly into the synthesizer and it's generated "as gates" - in a modern process this might be preferable for ROMs that are less than a few hundred bytes in size. Either way, you at least get a "correct" circuit even if the back-end team makes a mistake and fails to replace the black box with a hard macro file.

Can you suggest how I could generate the ROM file at this point in the output compilation phase?

bunnie:
BTW, I added https://github.com/SpinalHDL/SpinalHDL/pull/1485 which addresses the situation where you don't want any black box but just a case statement in-line.

I'm leaving this PR open because I think actually the code flow for generating a black box with ROM contents is almost correct in this PR, but the main issue is the `emitRom()` statement versus how the black box instance is generated.

If in the end it turns out that this is just not a thing that can fit well into SpinaHDL, I'm also fine with the PR being rejected and I maintain this generator in a branch that I patch in. I recognize this is a bit of a niche use case...

core/src/main/scala/spinal/core/internals/Phase.scala
Dolu1990: Hi,

> However, when I do that, emitMem() is no longer called.

Yes that is on purpose.
The idea of the SpinalHDL blackboxing  phase is to only emit it as a blackbox durring the generation phase, so SpinalHDL will only see a blackbox in the future, no more Mem => no emitMem

> black box RAM models are meant to be static files provided by the user, and thus no generation is required

Right. What i did a few time, is to add another SpinalHDL transformation phase, which will look at all the emited memory blackbox and transform them into something more specific (for ASIC), for instance : 
- automaticaly adding some test IO to the blackbox
- changing their name to depend on their parameter, for instance transforming the name Ram_1w_1rs into Ram_1w_1rs_1024x32, aswell as generating scripts for openram to generate them.

> for the black box ROM model, I'd like it to be generated by the system, because the ROM contents are actually inside SpinalHDL. 

I think here a custom blackboxing phase could be added to either generate a verilog of that blackboxed rom, or either export the rom content into some binary file for other usages.

Here is an example : 
https://github.com/SpinalHDL/SpinalHDL/commit/1d86c31bfa90d6c328e6498d264f8c97221ed8ee
It is based on your branch.

That will generate for each rom a verilog file with things as : 

```verilog
module xy_rom (
  input  wire clk,
  input  wire en,
  output wire [7:0] addr,
  input  wire [31:0] data
);
  always @(posedge clk) begin
    if(en) begin
      case (address)
        8'h0: data <= 32'hff
        8'h1: data <= 32'hfe
        8'h2: data <= 32'hfd
...
        8'hfc: data <= 32'h3
        8'hfd: data <= 32'h2
        8'hfe: data <= 32'h1
        8'hff: data <= 32'h0
      endcase
    end
  end
endmodule
```

Also, carefull your branch miss this : 
https://github.com/SpinalHDL/SpinalHDL/commit/e57f89dfb90b40cc13306418197a3000d88eaef7

Next week i will be in travel / holiday, so away from the keyboard :)

Dolu1990:
> I'm leaving this PR open because I think actually the code flow for generating a black box with ROM contents is almost correct in this PR, but the main issue is the emitRom() statement versus how the black box instance is generated.

to me i think this PR without the ComponentEmitterVerilog.scala changes, and with the https://github.com/SpinalHDL/SpinalHDL/commit/e57f89dfb90b40cc13306418197a3000d88eaef7 fix would be good for merge.

(so in short, keeping the rom blackbox + blackboxing code + the fix)

Thanks for https://github.com/SpinalHDL/SpinalHDL/pull/1485, i will take a look now :D

Dolu1990:
Ahhh the one thing would be to have rom blackboxing option, to ensure we do not start to blackbox rom on code which was running fine before (keeping the default as it was)

bunnie:
looks like there is a regression. how do i run the unit test that failed on its own?

looks like the failure is in spinal.simromtester, so it's definitely this patch that's causing the problem.

bunnie:
> Ahhh the one thing would be to have rom blackboxing option, to ensure we do not start to blackbox rom on code which was running fine before (keeping the default as it was)

Right, I'll add an option for that, which is `false` by default.

core/src/main/scala/spinal/core/internals/Phase.scala
bunnie: I *think* I vaguely follow the approach you're suggesting. I will need to poke at it a bit more.

> Next week i will be in travel / holiday, so away from the keyboard :)

No worries, same here as well. I don't think I'll be able to make the changes so soon, but also, this isn't urgent (I've got a tape-out candidate generated from my own branch that passes the tests I need).

I'm mostly just trying to merge the changes as an exercise to get me more comfortable with SpinalHDL, while hopefully adding a minor feature to the code base. Hopefully in the future I can do more meaningful contributions.

I appreciate the review and guidance, we can pick this up again after you're back from holiday!

Dolu1990:
> looks like there is a regression. how do i run the unit test that failed on its own?

[info] *** 2 TESTS FAILED ***
[error] Failed tests:
[error] 	spinal.core.SpinalSimRomTester
[error] (tester / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 306 s (05:06), completed Jul 12, 2024, 7:28:38 AM

=>

``` 
cd SpinalHDL
sbt
testOnly spinal.core.SpinalSimRomTester
```

> so it's definitely this patch that's causing the problem.

Need to remove all the ComponentEmitterVerilog.scala changes, so the PR would only be about blackboxing the rom without the verilog backend changes :)

core/src/main/scala/spinal/core/internals/Phase.scala
Dolu1990: > I'm mostly just trying to merge the changes as an exercise to get me more comfortable with SpinalHDL, while hopefully adding a minor feature to the code base. Hopefully in the future I can do more meaningful contributions. 

No worries ^^

> I appreciate the review and guidance, we can pick this up again after you're back from holiday!

Yes :)

bunnie:
OK, assuming these CI tests pass, I think I have at least added the flag for rom black boxing (`romBlackBox`) correctly.

I think I'm still a little confused about how you want to approach the actual specific generation of the blackbox ROM file data (as initialized with the unique ROM contents), but I think what you mean is this:

- remove any unique ROM instance generation from the ComponentEmitterVerilog.scala
- make it a function inside the user's code to create the specific ROM model that has the switch-case data in it that encodes the ROM as RTL

So I think maybe then in that case, some example code on how to use the ROM black boxing feature would be useful because I definitely would not know how to do it if I turned on the option without having written the patch. Is there a place in the documentation where I should include such examples?

No hurries on response, I recall you're on holiday...

bunnie:
Alright, I added the final patches I think to get this in line with the outlined vision. There is no blackbox generator in Spinal itself, and instead, it should be part of the code that defines the CPU (so it becomes part of the user code's hyothetical SoC generator).

Here's the code I ultimately used to generate the AES ROM blackboxes:

```scala
class CaseRomPhase extends PhaseNetlist{
  override def impl(pc: PhaseContext): Unit = {
    pc.walkComponents{
      case c : Rom_1rs => {
        c.setDefinitionName(c.getRtlPath("_")) // provide a unique name
        val rom = c.getTag(classOf[MemBlackboxOf]).get.mem
        val content = rom.initialContent

        c.genericElements.clear()
        val addressWidth = log2Up(c.wordCount)
        val lines = for ((data, addr) <- content.zipWithIndex) yield {
          f"            $addressWidth'h${addr}%x: data <= ${c.wordWidth}%d'h${data}%x;"
        }
        // Emit a verilog blackbox template for the ROM
        val template =
          """
            |`resetall
            |`timescale 1ns / 1ps
            |`default_nettype none
            |
            |module {modname} #(
            |    parameter wordCount = {wordCount},
            |    parameter wordWidth = {wordWidth},
            |    parameter addrWidth = $clog2(wordCount)
            |)
            |(
            |    input  wire                             clk,
            |    input  wire                             en,
            |    input  wire [addrWidth - 1:0]           addr,
            |    output reg  [wordWidth - 1:0]           data
            |);
            |
            |always @(posedge clk) begin
            |    if (en) begin
            |        case (addr)
            |{romvals}
            |
            |            default: data <= {wordWidth}'h0;
            |        endcase
            |    end else begin
            |        data <= data;
            |    end
            |end
            |
            |endmodule
            |`resetall
            |""".stripMargin

        val verilog = template
          .replace("{romvals}", s"${lines.mkString("\n")}")
          .replace("{modname}", s"${c.definitionName}")
          .replace("{wordCount}", s"${c.wordCount}")
          .replace("{wordWidth}", s"${c.wordWidth}")

        val fileWriter = new FileWriter(new File(c.definitionName + ".v"))
        fileWriter.write(verilog)
        fileWriter.flush()
        fileWriter.close()
      }
      case _ =>
    }
  }
}
```

The function above merely defines the black box generator. Calling the generator requires inserting it into the memory black boxers list with a line something like this:

```scala
    config.memBlackBoxers += new CaseRomPhase
```

where the `config` is the `spinal.core.SpinalConfig`-derived object that defines the CPU configuration.

@Dolu1990 lmk if this is what you intended, and also, lmk if there's some place documentation examples on a blackboxer like this normally goes. And of course, no hurries on getting back on this. I'll be a bit slow closing the loop myself.

Dolu1990:
Hi Bunnie,

> Alright, I added the final patches I think to get this in line with the outlined vision. There is no blackbox generator in Spinal itself, and instead, it should be part of the code that defines the CPU (so it becomes part of the user code's hyothetical SoC generator).

Yes :D, as it is can be quite user specific.

>  lmk if this is what you intended

Yes it is.

>  if there's some place documentation examples on a blackboxer like this normally goes

I would say it could be added as an subchapter of https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Sequential%20logic/memory.html#automatic-blackboxing ?

>  I'll be a bit slow closing the loop myself

No worries ^^
From the 9 august until the 25 i will be in travel (work). So, i will have a looooow bandwidth XD

Cheers
Charles

bunnie:
OK, looks like https://github.com/SpinalHDL/SpinalDoc-RTD/blob/master/source/SpinalHDL/Sequential%20logic/memory.rst is a different repo, so I'll make changes there an open a PR for that eventually.

Enjoy your travels! Get some rest and unplug from all the work. I see the volume of github issues and PRs you deal with, you put a heroic amount of effort into maintaining this, you deserve a break.

Dolu1990:
> I'll make changes there an open a PR for that eventually.

Thanks ^^

> Get some rest and unplug from all the work

Ahh, I just got 2 weeks of holiday, my GF did not missed one occasion to keep me away from email / work :D

Cheers

PR #1484, opened 2024-07-17
set rst name for default clock domain by ClockDomainConfig

yportne13:
I have to name rst signal to `rst_n` instead of `resetn`

### DIFF
--- a/core/src/main/scala/spinal/core/ClockDomain.scala
+++ b/core/src/main/scala/spinal/core/ClockDomain.scala
@@ -70,7 +70,7 @@ case class ClockEnableTag(clockDomain: ClockDomain) extends ClockDomainBoolTag
 
 // Default configuration of clock domain is :
 // Rising edge clock with optional asynchronous reset active high and optional active high clockEnable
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

I would say, the ClockDomainConfig and , resetName: String = null arent things which realy should be together ?

More generaly, maybe what you would prefer, is to have a configuration in the SpinalConfig which specify which flavor of naming convention you would prefer ?
That way, naming convention like  clk vs clock , reset vs rst would be applied globaly.

yportne13:
> Hi,
> 
> I would say, the ClockDomainConfig and , resetName: String = null arent things which realy should be together ?
> 
> More generaly, maybe what you would prefer, is to have a configuration in the SpinalConfig which specify which flavor of naming convention you would prefer ? That way, naming convention like clk vs clock , reset vs rst would be applied globaly.

yes, if want to apply globaly, just:
```scala
SpinalConfig(defaultConfigForClockDomains = ClockDomainConfig(
    resetName = "rst_n"
))
```

Dolu1990:
The issue comes when you have multiple clock domain, they will al try to use the same reset name

Maybe what you would want instead, is to have some custom SpinalHDL phase, which look for component with single clockdomain, and override the reset / clock signal names there ?

PR #1485, opened 2024-07-18
Add `caseRom` option

bunnie:
Related to #1475 

# Impact on code generation

Per @Dolu1990 's request, this PR creates an alternative to in-line ROM where the code is generated in a switch/case form, instead of using an `initial begin` with in-line code.

The purpose of this is for better support in back-end tool flows that can't honor `initial begin` statements in verliog: in particular, the ASIC flow I'm working with simply ignores them. The ASIC tool flow won't synthesize a ROM, because there's a decision that should be made between doing a hard macro ROM or a ROM-in-gates. If the intention is ROM-in-gates, then `caseRom` should be selected and the ROM will be described as a case statement directly in the verilog code. If the intention is a ROM-as-macro, then blackboxing should be selected, see #1475 for the PR to generate this.

The blackbox flow ideally also generates a blackbox with the ROM-in-gates inside it as stand-in for simulation and a back-stop in case the back-end team can't generate the hard ROM macro. For small ROMs (< couple hundred bytes), it's a bit of a coin toss whether a hard macro or ROM in gates is better, and this form gives the back end team the latitude to experiment with both formats without a risk to correctness.

# Checklist

- [ ] Unit tests were added


Dolu1990:
Thanks ^^
I just pushed one fix for the emitRead rom detection.
If it passes regression we can merge.

PR #1491, opened 2024-07-26
fix incorrect right-shift of signed AFix

### LINKED ISSUE #1490: signed `AFix` right shift does not sign-extend
Right-shifting a signed `AFix` by another `AFix`, or when using `>>|`, does not sign-extend. It doesn't act like a divide-by-power-of-2 like one would expect.

thajohns:
Closes #1490

As mentioned in the issue, using `>>` between two `AFix`es, or using `>>|` between an `AFix` and an `AFix` or an `AFix` and an `Int`, does not behave as expected when the left `AFix` is signed. It can change a negative `AFix` into a positive one. This PR adds a check to each of the three relevant functions and sign-extends if the left `AFix` is signed.

# Impact on code generation

Right-shifting signed `AFix`es will now sign-extend.

# Checklist

- [x] Unit tests were added
  - none needed

### DIFF
--- a/core/src/main/scala/spinal/core/AFix.scala
+++ b/core/src/main/scala/spinal/core/AFix.scala
@@ -586,13 +586,13 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
     ret
   }
 
-  // Shift bits and decimal point right, adding padding bits left
+  // Shift bits and decimal point right, truncating bits on the right
   def >>|(shift: Int): AFix = {
     val shiftBig = BigInt(2).pow(shift)
     val ret = new AFix(this.maxRaw / shiftBig, this.minRaw / shiftBig, this.exp)
 
     if (this.signed)
-      ret.raw := (this.raw.asSInt >> shift).asBits
+      ret.raw := this.raw.dropLow(shift)
     else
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -592,7 +592,7 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
     val ret = new AFix(this.maxRaw / shiftBig, this.minRaw / shiftBig, this.exp)
 
     if (this.signed)
-      ret.raw := this.raw.dropLow(shift)
+      ret.raw := (this.raw.asSInt >> shift).asBits
```
Dolu1990, line 595: I would say this change isn't required.
Are you sure the old implementation was bugy ?
I mean, ret.raw := this.raw.dropLow(shift)  only remove the LSB bits. No padding is required. (the operator comment is wrong XD about "adding padding bits left")

Dolu1990: Else it is good for merge :)

thajohns: I'll take another look at this when I'm done with my current task.

Dolu1990:
Thanks :D

PR #1493, opened 2024-07-28
Fix timescale of the trace files generated by the Verilator backend

### LINKED ISSUE #1492: Timescale of the trace file generated by Verilator does not match the value defined in the scala simulation setup
*Problem Description*
I have a testbench which configures the timescale to 1ns/1ps.
grafik
Based on this setting I would expect the timescale in the .vcd file to be set to:
> "$timescale 1ps $end"

Unfortunately, what I see in the .vcd file is a timescale of "1s":
grafik
The same issue seems to also happen with .fst files.

I'm using a recent Verilator version (v5.026) and the most recent SpinalHDL version (v1.10.2a).

*Problem Analysis*
I debugged into the Verilator Backend and it seems that the issue is caused by the fact that the VerilatedContext instance is created via the default constructor of that class (I couldn't find any definition of that constructor in the Verilator repo, so I assume it sets all members to zero). Later on in the Verilator Backend the timeprecision is then set via a dedicated setter method. Neverthless, at that point the "VerilatedVcd" is already created and the .vcd file opened with a default value of "1s" as timeprecision.

RolinBert:
Explicitly set the time precision for trace files (.vcd and .fst) before opening the trace file.

Closes #1492

The Verilator Backend  seems to create the VerilatedContext instance via the default constructor of that class (I couldn't find any definition of that constructor in the Verilator repo, so I assume it sets all members to zero). Later on in the Verilator Backend the timeprecision is then set via a dedicated setter method. Nevertheless, at that point the "VerilatedVcd" is already created and the .vcd file opened with a default value of "1s" as timeprecision. The same issue is also seen for .fst files. For more details refer to #1492

# Impact on code generation

Timescale for .vcd and .fst files is set correctly. Correctly, means that the set values match the Scala simulation config. The change has been tested for .vcd and .fst files with different time specs set via the "withTimeSpec(...)" statement.

# Checklist

- [x] Unit tests were added
-> I did not find any unit tests which would check the resulting .vcd files (I guess most of the vcd generation is checked on the Verilator side)

### DIFF
--- a/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
+++ b/sim/src/main/scala/spinal/sim/VerilatorBackend.scala
@@ -293,6 +293,7 @@ ${    val signalInits = for((signal, id) <- config.signals.zipWithIndex) yield {
       #ifdef TRACE
       Verilated::traceEverOn(true);
       top->trace(&tfp, 99);
+      tfp.set_time_resolution(${if (useTimePrecision) "Verilated::threadContextp()->timeprecisionString()" else "VL_TIME_PRECISION_STR" });
       tfp.open((std::string(wavePath) + "wave" + ".${format.ext}").c_str());
       #endif
       this->name = name;


Dolu1990:
Nice thanks :D

When the regression test finish, we can merge it.

PR #1495, opened 2024-07-31
Allow randomizing TKEEP/TSTRB for Axi4StreamMaster

KireinaHoro:
A nice-to-have for the `Axi4StreamMaster` simulation agent.  Randomizing TKEEP is useful to test if an AXIS component is standard-compliant and correctly discards the null bytes in the stream (TSTRB = 0 && TKEEP = 0).

Note that this does not support randomizing TSTRB as well (it's tied to TKEEP) since I don't have a use for it (yet); it can be implemented at a later date.


Dolu1990:
Thanks ^^

PR #1505, opened 2024-08-13
Add mill template link to readme

mrcmry:
Also use double level list for base projects to make it more readable.

# Impact on code generation

None


Dolu1990:
We will also have to update those base projects (https://github.com/SpinalHDL/SpinalTemplateMill / https://github.com/SpinalHDL/SpinalTemplateGradle) They are realy realy old.
Currently, the https://github.com/SpinalHDL/SpinalTemplateSbt was kinda be used as a Mill template aswell, but overall i think having separeted repo isn't a bad thing.
I'm in a work trip, i will have to take some time.

PR #1506, opened 2024-08-17
Mill support on windows

Readon:
Add mill support for Windows.
It also helps to find the compiling fail by compiler 2.11.

Closes #

# Impact on code generation

# Checklist

 - [ ] Unit tests were added 


Dolu1990:
Thanks ^^

KireinaHoro:
This in turn breaks having spinalhdl as a submodule, since you're looking for `project/version.conf` with the assumption that the current working directory; this is not true if it's in a submodule.

Dolu1990:
Right, i just reverted #1506 to fall back in the previous state.
@Readon Do you think the PR would be good without that version.conf file ?

PR #1516, opened 2024-08-26
Fix regression from #1506: allow mill to locate version config file

KireinaHoro:
PR #1506 introduced a `$PWD`-relative lookup to find the compiler version config file, which would break submodule setups.  Use ammonite's capability to locate the source file to find the right path.

Note that the same issue is also there for sbt, but I'm not using it and do not have an idea how to fix it there.


Dolu1990:
For now, i reverted #1506

PR #1529, opened 2024-09-07
Fix FSM bug

### LINKED ISSUE #1528: Outer FSM `whenIsActive` statements always win vs. nested FSM `onEntry` statements for combinational logic
If an FSM drives a combinational signal from the `whenIsActive` block of one of its states, and then transitions to a nested FSM state whose entry state has an `onEntry` block which also drives the combinational signal, then the outer FSM's statement always wins even if it appears earlier in the file.

For example:
```scala
case class BrokenFsm() extends Component {
  val io = new Bundle {
      val go = in Bool()
      val a = out Bool()
      val b = out Bool()
  }

  output := False
  io.a := False
  io.b := False

  val fsm = new StateMachine() {
      val idle: State = new State with EntryPoint() {
          whenIsActive {
              io.b := False
              when (io.go) { goto(par) }
          }
      }
      val par = new StateFsm(
          new StateMachine {
              val nested = new State() with EntryPoint {
                  onEntry {
                      io.a := True
                      io.b := True
                  }
                  whenIsActive { exit() }

              }
          }
      ) {
          whenCompleted { goto(idle) }
      }
  }
}
```

In the above FSM, we expect to see `io.a` and `io.b` act the same way, since the second driver on `io.b` (the only one which `io.a` does not have) is redundant. However, `io.b` never goes high, but `io.a` does.

The generated Verilog also demonstrates the issue. The (correct) block produced for `io.a` is:
```scala
  always @(*) begin
    io_a = 1'b0;
    if(when_StateMachine_l253) begin
      io_a = 1'b1;
    end
  end
```
whereas the (incorrect) block for `io.b` is:
```scala
  always @(*) begin
    io_b = 1'b0;
    if(when_StateMachine_l253) begin
      io_b = 1'b1;
    end
    case(fsm_stateReg)
      fsm_enumDef_idle : begin
        io_b = 1'b0;
      end
      fsm_enumDef_par : begin
      end
      default : begin
      end
    endcase
  end
```
In the latter, the branch for the `idle` state comes after the branch for the `onEntry` state (the `if` above), which if I understand Verilog's semantics correctly means that it wins.

thajohns:
Closes #1528

See issue for details.

# Impact on code generation

`StateFsm` and `StateParallelFsm` now generate `onEntry` statements duplicating the `onEntry` blocks of the entry points of their nested state machines.

# Checklist

- [x] Unit tests were added
- [-] API changes are or will be documented (bugfix - none needed)

### DIFF
--- a/lib/src/main/scala/spinal/lib/fsm/State.scala
+++ b/lib/src/main/scala/spinal/lib/fsm/State.scala
@@ -126,6 +126,8 @@ class StateFsm[T <: StateMachineAccessor](val fsm:  T)(implicit stateMachineAcce
 
   onEntry{
     fsm.startFsm()
+    fsm.getEntry().onEntryTasks.foreach(_())
+    fsm.getEntry().whenIsNextTasks.foreach(_())
   }
 
   whenIsActiveWithPriority(1){
@@ -171,7 +173,11 @@ class StateParallelFsm(val _fsms: StateMachineAccessor*)(implicit stateMachineAc
   val fsms = _fsms
   
   onEntry{
-    fsms.foreach(_.startFsm())
+    fsms.foreach { fsm =>
+      fsm.startFsm()
+      fsm.getEntry().onEntryTasks.foreach(_())
+      fsm.getEntry().whenIsNextTasks.foreach(_())
+    }
   }
 
   whenIsActiveWithPriority(1){
--- a/tester/src/test/scala/spinal/lib/SpinalFsmTester.scala
+++ b/tester/src/test/scala/spinal/lib/SpinalFsmTester.scala
@@ -1,0 +1,68 @@
+package spinal.lib
+
+import spinal.core._
+import spinal.core.sim._
+import spinal.lib.fsm._
+import spinal.tester.SpinalAnyFunSuite
+
+class SpinalFsmTester extends SpinalAnyFunSuite {
+  test("nested fsm onEntry") {
+    SimConfig
+      .compile(new Component {
+        val io = new Bundle {
+          val go = in Bool()
+          val a = out Bool()
+          val b = out Bool()
+          val aReg = out(Reg(Bool()) init(False))
+          val bReg = out(Reg(Bool()) init(False))
+        }
+
+        io.a := False
+        io.b := False
+        io.aReg := False
+        io.bReg := False
+
+        val fsm = new StateMachine() {
+          val idle: State = new State with EntryPoint() {
+            whenIsActive {
+              io.b := False
+              io.bReg := False
+              when (io.go) { goto(par) }
+            }
+          }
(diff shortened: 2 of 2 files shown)


thajohns:
I actually have a couple of questions related to this PR. First, there must already be code to test FSMs, right? I couldn't find any so I made a new file for it. Second, is it ever a problem when a block appears in two places, like would happen to the `onEntry` blocks here? If so, a more sophisticated solution is probably required.

Dolu1990:
Thanks ^^

> First, there must already be code to test FSMs, right?

Kinda but not realy. The FSM stuff is through other modules which happen to use it in the SpinalLib. Mainly the UsbOhci.

> Second, is it ever a problem when a block appears in two places, like would happen to the onEntry blocks here? If so, a more sophisticated solution is probably required

It is a big issue i would say.

More generaly, i think the issue that you are facing is more that between the statments of the parent FSM and the children FSM, the ordering isn't realy preserved between isActive onEntry.
Currently, the parrent statemachine first do `childStateMachines.foreach(_.build())` and then build itself.
Instead what should may be done, is to stage the child state machine build into multiple sub step called at different stages of the parrent FSM.

thajohns:
It seems straightforward enough to move the child state machine's states into the parent state machine when using `StateFsm`. However, it's less obvious to me how it should be done for `StateParallelFsm`. (I have some thoughts, but I need to think about them more carefully before I can write them out here. I think it's achievable with nested `case`s.)

Looking at the FSM code again (although I haven't written something to test this), it looks like all `onEntry` statements always take priority over all `whenIsActive` statements, which is either a bug or should be documented. (tbh i kind of like it that way.) Changing that would be a substantial effort, and I think the best way to do it might involve substantial changes to the internals, so that a block that only gets evaluated once can appear in multiple scopes. That would probably also involve making the FSM code much less ... nice. Is that a bad enough idea to be discarded out-of-hand?
 
I'll keep thinking about this and might contribute some code when I get the time.

Dolu1990:
>  move the child state machine's states into the parent state

Should not be done ^^

> it looks like all onEntry statements always take priority over all whenIsActive statements, which is either a bug or should be documented.

hmm to me is is normal behaviour. Here is the expected ordering inside a given FSM
stateA isActive -> stateA on Exit -> stateB on entry -> stateB isActive

The issue you are hitting, isn't the way a given FSM build itself, but more that between the main FSM and its inner FSM, the inner FSM get fully builded before the parent FSM, meaning that all the callbacks of the children FSM lose against the callback of the main FSM.

So, to fix that, the build function of the FSM should be splited into multiple phases. and when the main FSM want to build the children FSM, it has to be done chunk by chunk in sync with the main FSM (instead of all at once)

thajohns:
I don't think I have time to pursue this right now.

PR #1530, opened 2024-09-09
Add Support for IPXACT 2022 and Vivado-Specific 2009 IP Core Packaging in Spinal

ZhaokunHu:
# Introduction

IPXACT, known as the IEEE1685 standard, is primarily used for packaging reusable IP cores. The latest version is the 2022 edition, but tools supporting this version are currently limited. Vivado, for instance, still uses the 2009 version of the standard. Although this version is technically outdated, it remains widely used due to Vivado's compatibility constraints.
Here are the official IEEE documents for IPXACT:
IEEE1685-2022: https://ieeexplore.ieee.org/document/10054520
IEEE1685-2009: https://ieeexplore.ieee.org/document/5417309
In this project, I've developed code that supports both the 2022 and Vivado (2009) versions of IPXACT. The 2022 version implementation strictly adheres to the IPXACT standard, while the 2009 version includes Vivado-specific extensions, making it usable exclusively in Vivado.

# Purpose of the Feature
This feature adds support for IPXACT in Spinal, with the 2022 version mainly targeting future potential use cases, while the Vivado version facilitates certain simulation functions within Vivado. Compared to Vivado's native IP packaging tool, this functionality can automatically connect buses and clocks, greatly simplifying the wiring process.

# Usage Instructions
First, you need to add a library dependency in sbt:

```scala
libraryDependencies += "io.github.zhaokunhu" %% "ipxactscalacases" % "0.0.1"
```
The source URL is: https://central.sonatype.com/artifact/io.github.zhaokunhu/ipxactscalacases
Instantiate an IPXACTGenerator object as follows:
```scala
val ipxactGenerator = new IPXACTGenerator(SpinalVerilog(_))
```
To generate an IP core in the IEEE1685-2022 format, use the function:
```scala
ipxactGenerator.generate2022Component()
```
Optional parameters include the vendor  and version, with the default being "SpinalHDL" and "1.0". This function will create a folder named after the vendor, containing the IP core defined using the 2022 version of IPXACT. You can use Kactus2 to open and view the IPXACT content of the 2022 version: https://github.com/kactus2/kactus2dev
If you want to use kactuc, follow the steps below, and the selected folder should be the one generated and named after the vendor.
image

Note that since the generated files are pure IPXACT without specific support for Kactus, the display may not be optimal in Kactus. In theory, any software that supports IPXACT 2022 should be able to open the files. If you find any software that does support IPXACT 2022, please let me know so I can test it.

To generate an IP core that can be directly used in Vivado, use the function:
```scala
ipxactGenerator.generateVivadoComponent()
```

Optional parameters include the vendor and version, defaulting to "SpinalHDL" and "1.0". This function will create a folder named VivadoIpFolder containing IP cores that can be opened and used directly in Vivado. The generated IPXACT file will follow the 2009 version and include some Vivado-specific content, which means it can only be opened in Vivado.
If you want to use Vivado, follow the steps below, and the selected folder should be VivadoIpFolder.
image
# Design Rationale
IPXACT files are essentially XML files. The standard format is defined by an XML Schema Definition (XSD) file, which can be found here: https://github.com/edaa-org/IPXACT-Schema

By using scalaxb (https://github.com/eed3si9n/scalaxb) to process these XSD files, we can generate corresponding Scala classes. By instantiating these classes and filling in the appropriate parameters, we can generate XML files that conform to the IPXACT standard. The library that needs to be added to sbt, as mentioned earlier, contains Scala classes that I auto-generated using scalaxb.

Currently, the code only supports packaging a single IP core, i.e., the Component entity defined in the IPXACT standard. Other features are still under development.

# Differences Between the Two Versions
In the 2022 version, when a bus needs to be packaged, two additional files defining the bus are first generated. These correspond to abstractionDefinition and busDefinition in the IPXACT standard, and the bus is then referenced within the Component to package the signals.

For the Vivado version, Xilinx provides predefined abstractionDefinition and busDefinition files. When defining an IP core, users can select the appropriate bus and map the signals to the bus lines. This design follows the same process by referencing the existing Vivado buses and connecting the signals generated by Spinal to the corresponding Vivado bus signals. If there is a bus in Spinal that Vivado does not support, the signals will remain unconnected to any bus but will still be exposed.

# Future Development Plans
For the 2022 version, only single IP core packaging (i.e., the Component entity in IPXACT) is currently supported. I plan to add support for other features such as design, designcfg, and catalog as defined in the IPXACT standard.

For the Vivado version, there are a few issues that need to be addressed:
1. The buses defined in Vivado and Spinal are not always directly compatible. For example, the AHb bus may result in an extra signal line after packaging. Additionally, some buses like Tilelink exist in Spinal but do not have corresponding definitions in Vivado. There may also be buses that can be packaged but are not currently supported. If you find such a bus or discover mapping errors, please inform me.
2. We need to discuss how to handle buses that are not supported by Vivado. Should we follow the approach used in the 2022 version and define custom buses, or is it sufficient to leave the signal lines exposed without bus packaging?

# Impact on code generation

This feature only supports Verilog code generation. 
# Checklist

- [x]  Unit tests were added


ZhaokunHu:
This is the test I ran in tester/src/test/scala/spinal/lib/tools/IPXACTGeneratorTest.scala.
In Kactus, this feature generates some buses, and you can click on the corresponding bus's absDef file to view the related definitions. By clicking on the Component name file, you can view the interface definitions of the corresponding IP, as well as some related information.
image
image
image
image

In Vivado, after opening the created IP core, you can directly apply it to the block design, or right-click to further edit and modify it.
image
image
In the IP editing interface, you can view and modify the relevant information of the IP.
image
image
Since this project is not yet fully completed, there may be some bugs. If you encounter any, please let me know, and I will make sure to correct them carefully.

lib/src/main/scala/spinal/lib/tools/IPXACTGenerator/IPXACT2022ComponentGenerator.scala
```diff
@@ -0,0 +1,233 @@
...
+        if (port.isInput) {
+          InValue
+        } else if (port.isOutput) {
+          OutValue
+        } else {
+          InoutValue
+        }
+      val leftUnsignedIntExpression = UnsignedIntExpression((port.getBitsWidth - 1).toString)
+      val rightUnsignedIntExpression = UnsignedIntExpression("0")
+      val vector1 = Vector4(left = leftUnsignedIntExpression, right = rightUnsignedIntExpression)
```
Readon, line 113: what's the purpose of this variable?

ZhaokunHu:
Now, by using the generateDesignView function, you can generate the design and designcfg files. With these two files, you can use Kactus to view the internal connections of the module. One important point to note is that if you want to use this function, you need to set the oneFilePerComponent parameter in spinalConfig to true.
image

jijingg:
@ZhaokunHu   @Readon 

 `io.github.zhaokunhu:ipxactscalacases_2.12:0.0.1`    dependency for `org.scala-lang.modules:scala-xml_2.12` and `org.scala-lang.modules:scala-parser-combinators_2.12`  is too old.  Please update to a newer version. 

Additionally, please handle such complex dependencies with caution to avoid causing a wider impact.

```sh
[error] stack trace is suppressed; run 'last update' for the full output
[error] stack trace is suppressed; run 'last ssExtractDependencies' for the full output
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error] 	* org.scala-lang.modules:scala-xml_2.12:2.1.0 (early-semver) is selected over {1.3.0, 1.0.6}
[error] 	    +- org.scala-lang:scala-compiler:2.12.18              (depends on 2.1.0)
[error] 	    +- net.databinder.dispatch:dispatch-core_2.12:0.13.4  (depends on 1.0.6)
[error] 	    +- io.github.zhaokunhu:ipxactscalacases_2.12:0.0.1    (depends on 1.3.0)
[error]
[error] 	* org.scala-lang.modules:scala-parser-combinators_2.12:2.1.1 (early-semver) is selected over 1.1.2
[error] 	    +- migen.com:migen_2.12:1.4.1                         (depends on 2.1.1)
[error] 	    +- io.github.zhaokunhu:ipxactscalacases_2.12:0.0.1    (depends on 1.1.2)
```

ZhaokunHu:
I sincerely apologize for this issue. I have resubmitted a new version, io.github.zhaokunhu
.12:0.0.2, updating the versions of scala-xml and scala-parser-combinators. Please try it again to see if it works; if there are any issues, I will continue to make adjustments.

Directly placing IPXACT into spinal.lib is not very feasible, as the code generated by scalaxb is quite substantial. I have uploaded the generated code to this repository: https://github.com/ZhaokunHu/IPXACTScalaCases.

Additionally, I am not very familiar with the process of submitting to Maven. I don’t know how to use sbt for submissions, so I submitted the code to Maven Central using Maven. If there’s a better submission method, please let me know, and I will make further adjustments.

> @ZhaokunHu @Readon
> 
> `io.github.zhaokunhu:ipxactscalacases_2.12:0.0.1` dependency for `org.scala-lang.modules:scala-xml_2.12` and `org.scala-lang.modules:scala-parser-combinators_2.12` is too old. Please update to a newer version.
> 
> Is it possible to include IPxact directly into spinal.lib? If there are any issues, we can fix them in Spinal. The dependencies of Spinal should ideally be those that are more popular and well-vetted by the community. For newer or smaller libraries, we should be cautious to avoid the risk of losing maintenance support.
> 
> ```shell
> [error] stack trace is suppressed; run 'last update' for the full output
> [error] stack trace is suppressed; run 'last ssExtractDependencies' for the full output
> [error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
> [error]
> [error] 	* org.scala-lang.modules:scala-xml_2.12:2.1.0 (early-semver) is selected over {1.3.0, 1.0.6}
> [error] 	    +- org.scala-lang:scala-compiler:2.12.18              (depends on 2.1.0)
> [error] 	    +- net.databinder.dispatch:dispatch-core_2.12:0.13.4  (depends on 1.0.6)
> [error] 	    +- io.github.zhaokunhu:ipxactscalacases_2.12:0.0.1    (depends on 1.3.0)
> [error]
> [error] 	* org.scala-lang.modules:scala-parser-combinators_2.12:2.1.1 (early-semver) is selected over 1.1.2
> [error] 	    +- migen.com:migen_2.12:1.4.1                         (depends on 2.1.1)
> [error] 	    +- io.github.zhaokunhu:ipxactscalacases_2.12:0.0.1    (depends on 1.1.2)
> ```

Dolu1990:
Hmm maybe the best then would be to keep the whole utility in a separated repo ? A bit like SpinalCrypto ?
 That way we avoid coupling things too much ?

Readon:
> Hmm maybe the best then would be to keep the whole utility in a separated repo ? A bit like SpinalCrypto ? That way we avoid coupling things too much ?

That could be a good idea.

PR #1538, opened 2024-09-16
Fix the wrong directory when generating .lst of Verilog/SV

Tikifire:
A small fix. There's a missing "/" in the file path when generating .lst in Verilog/SystemVerilog mode. Everything works fine in VHDL mode.

For example. If "targetDirectory" in the config is "output/rtl", the full path of .lst should be "output/rtl/mytoplevel.lst", instead of "output/rtlmytoplevel.lst".

### DIFF
--- a/core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/PhaseVerilog.scala
@@ -145,7 +145,7 @@ class PhaseVerilog(pc: PhaseContext, report: SpinalReport[_]) extends PhaseMisc 
       }
 
       if(pc.config.printFilelist){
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^ 

I completly missed this github issue from my emails, sorry for that !
The issue was closed because if the delay ? Or there is something about it ?
To me, it look like it is good for merge. Agree ?

Tikifire:
> Hi ^^
> 
> I completly missed this github issue from my emails, sorry for that ! The issue was closed because if the delay ? Or there is something about it ? To me, it look like it is good for merge. Agree ?

I closed it only for the delay.
Any way, it's happy to receive your reply :). You can merge it.

Dolu1990:
Merged ^^
Thanks !

PR #1543, opened 2024-09-19
StreamCCByToggle Altera attributes

g0t00:
StreamCCByToggle: Add `altera_attribute` Attributes to the registers that get max_skew requirement to disable optimization. Otherwise Quartus sometimes moves comb logic between the registers making them illegit targets for most timing constraints. 

Also add crossClockFalsePath to the relevant BufferCC registers. (analog to PulseCCByToggle)

# Impact on code generation

# Checklist

- [x] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/internals/Phase.scala
+++ b/core/src/main/scala/spinal/core/internals/Phase.scala
@@ -259,6 +259,28 @@ class PhaseDeviceSpecifics(pc : PhaseContext) extends PhaseNetlist{
 //        if(hit) mem.addAttribute("ram_style", "distributed") //Vivado stupid ganbling workaround Synth 8-6430
 //      case _ =>
 //    }
+    walkComponentsExceptBlackbox{component =>
+      import spinal.lib.AnalysisUtils
+      var statements = ArrayBuffer[DataAssignmentStatement]()
+      component.dslBody.walkStatements {
+        case da: DataAssignmentStatement if da.target.isInstanceOf[SpinalTagReady] && da.target
+          .asInstanceOf[SpinalTagReady]
+          .hasTag(classOf[crossClockMaxDelay]) =>
+          statements += da
+        case _ =>
+      }
+      for (statement <- statements) {
+        var source = statement.source.asInstanceOf[BaseType]
+        AnalysisUtils.seekNonCombDrivers(source) { case b: BaseType =>
+          source = b
+        }
+        val target = statement.target.asInstanceOf[BaseType]
+        val regAttribute = new AttributeString("altera_attribute", "-name ADV_NETLIST_OPT_ALLOWED NEVER_ALLOW")
+
+        source.addAttribute(regAttribute)
+        target.addAttribute(regAttribute)
+      }
+    }
   }
 }
 
@@ -3057,4 +3079,3 @@ object SpinalVerilogBoot{
     report
   }
 }
-
--- a/core/src/main/scala/spinal/core/internals/Phase.scala
+++ b/core/src/main/scala/spinal/core/internals/Phase.scala
@@ -259,45 +259,7 @@ class PhaseDeviceSpecifics(pc : PhaseContext) extends PhaseNetlist{
 //        if(hit) mem.addAttribute("ram_style", "distributed") //Vivado stupid ganbling workaround Synth 8-6430
 //      case _ =>
 //    }
-    def seekRegDriver(that : BaseType): Iterable[BaseType] ={
-      var buffer = ArrayBuffer[BaseType]()
-        that.foreachStatements{ s =>
-          def forExp(e : Expression) : Unit = e match {
-            case s : Statement => s match {
-              case s : BaseType if !s.isReg => {buffer ++= seekRegDriver(s) }
-              case s : BaseType if s.isReg => buffer += s
-              case s =>
-            }
-            case e: MemReadSync =>
-            case e: MemReadWrite =>
-            case e : Expression => e.foreachDrivingExpression(forExp)
-          }
-          s.walkParentTreeStatementsUntilRootScope{sParent =>
-            sParent.foreachDrivingExpression(forExp)
-          }
-          s.foreachDrivingExpression(forExp)
-        }
-        buffer
-      }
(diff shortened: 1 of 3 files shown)


Dolu1990:
Hi,

`val regAttribute = new AttributeString("altera_attribute", "-name ADV_NETLIST_OPT_ALLOWED NEVER_ALLOW")`
is very specific to altera.

Wouldn't a KEEP attribute work here aswelll (on most platforms aswell) ?

Dolu1990:
Also, how did you found that issue XD ?

g0t00:
> Hi,
> 
> `val regAttribute = new AttributeString("altera_attribute", "-name ADV_NETLIST_OPT_ALLOWED NEVER_ALLOW")` is very specific to altera.
> 
> Wouldn't a KEEP attribute work here aswelll (on most platforms aswell) ?

No, Keep is what we tried before sadly this does not work.
I think the reason is that keep just makes sure the Register/FF it self stays there. But it still allows the logic to be moved.
In our case I think for some reason it removes clock enable on the register and replaces it by some LUT.
(Which itself is not allowed in the path of the timing constraint)
I think normally Quartus is not supposed to make these "optimizations" when the constraint exists. But apperantly this does not always work...
I have less experience with other vendor platforms, so I do not if these problem exists there as well, and what the appropriate work around is.

Thinking about it the problem is actually not specifc to `StreamCCByToggle`, but more to the `crossClockMaxDelay`.
I could probably create a new Phase that sets the attribute on the source and target registers of the `crossClockMaxDelay` tag.
I do not know if the phases are already vendor specifics, but this could be then made altera specific if need be.

Talking about how we found.... We were chasing some random issue which we suspect to be related to wrong timing constraints. Quartus provides some timing DRC, and we were going on a wild chase and look at post PnR result viewer.

Dolu1990:
> Talking about how we found.... We were chasing some random issue which we suspect to be related to wrong timing constraints. Quartus provides some timing DRC, and we were going on a wild chase and look at post PnR result viewer.

Ahhhh pain XD
gg :D

> No, Keep is what we tried before sadly this does not work.

Then ok for the PR, the idea is to get minimal pain.

> Thinking about it the problem is actually not specifc to StreamCCByToggle, but more to the crossClockMaxDelay.

Yes right, those need to be exported one way or another onto the synthesis tool

> I do not know if the phases are already vendor specifics, but this could be then made altera specific if need be.

Hmm, i think that we could have a Phase base class which would provide callbacks to allllll the timing related things on the netlist.
Then people would just need to implement the device specific stuff in those callbacks to make things work.

g0t00:
Yeah the exercise was fun :)
Okay, I moved the tagging for Attributes to a Phase. I found a device specific Phase (which actually does not checking for the device). Unsure if this is a clever place for this though.
I am not sure what  you want to put in this base class though.

Dolu1990:
Hmmm, i will try to come up with something tomorrow. This kind of device specific stuff is realy something we need to sort out ^^

Dolu1990:
Hi,
I added in the SpinalConfig a devicePhaseHandler, with a default implementation here : 
https://github.com/SpinalHDL/SpinalHDL/blob/913575779192354e2c09bbfaa7a0b35772d1075b/core/src/main/scala/spinal/core/internals/Phase.scala#L1401

So i guess we could add that by default, it add the following 
new AttributeString("altera_attribute", "-name ADV_NETLIST_OPT_ALLOWED NEVER_ALLOW")

on onCrossClockBuffer.

Question is, does this attribute need to be on the register clocked on the destination ?
Or it also need to be on the register clocked on the source ?

g0t00:
Okay thank you.
`onCrossClockBuffer` is a bit different though, as it specific to `crossClockDomain` Tag. This probably applies the attributes to more register than is required.
To be honest, I am not sure how altera handles this attributes, there documentations sadly is slim on this.
I will test if this attribute needs to be set on the source as well (I was kind of assuming this, but might be worth trying out.)

core/src/main/scala/spinal/core/internals/Phase.scala
```diff
@@ -1420,6 +1422,14 @@ class PhaseDeviceDefault extends PhaseDeviceHandler{
       mem.addAttribute("ramstyle", "no_rw_check")
     }
   }
+  override def onCrossClockMaxDelay(config: SpinalConfig, sources: Iterable[BaseType], target: BaseType) = {
+    if(config.device.vendor == Device.ALTERA.vendor){
```
Dolu1990, line 1426: I would say, that it could also be enabled for config.device.isVendorDefault. that would make things safer for everyone ^^ unless it can break some flow ?

Dolu1990:
> I will test if this attribute needs to be set on the source as well

Ahhh make sense, some logic leaking futher.

core/src/main/scala/spinal/core/internals/Phase.scala
Dolu1990: Else i think this can be merged, i may touch things a bit.

g0t00: Yes good point. I do not see this clearly altera marked attribute confusing anything else.
I am good with merging this as well.

Dolu1990: Ok ^^ 

I should be able to merge monday :)

likewise:
"Also add crossClockFalsePath to the relevant BufferCC registers. (analog to PulseCCByToggle)"

Adding a false path to a cross clock domain path is not the best way to ensure timing. There was a large discussion in another topic (specifically regarding automatically generatin constraint files) that I do not have the issue number for (should re-edit when I find it).

I refered to the excellent work of Alex Forencich building the correct constraints for a few CDC primitives.

(I need to add more references to my comment).

Dolu1990:
I agree that adding crossClockFalsePath isn't realy the best way to handle it.
Better add a 1 cycle max delax constraint, it would at least constraint it (more deterministic)

g0t00:
> I agree that adding crossClockFalsePath isn't realy the best way to handle it. Better add a 1 cycle max delax constraint, it would at least constraint it (more deterministic)

Seems good to me

PR #1552, opened 2024-10-02
fix xsim on Linux.

### LINKED ISSUE #1501: XSIM backend not working on Linux
I tried to run the following simple example:

```scala
import spinal.core._

// Identity takes n bits in a and gives them back in z
class Identity(n: Int) extends Component {
  val io = new Bundle {
    val a = in Bits(n bits)
    val z = out Bits(n bits)
  }

  io.z := io.a
}

import spinal.core.sim._

object TestIdentity extends App {
  // Use the component with n = 3 bits as "dut" (device under test)
  SimConfig.withWave.withXSim.compile(new Identity(3)).doSim{ dut =>
    // For each number from 3'b000 to 3'b111 included
    for (a <- 0 to 7) {
      // Apply input
      dut.io.a #= a
      // Wait for a simulation time unit
      sleep(1)
      // Read output
      val z = dut.io.z.toInt
      // Check result
      assert(z == a, s"Got $z, expected $a")
    }
  }
}
```

It is from spinalhdl docs. I only added `withXSim`. I am getting the following error (replaced project path with ${PATH_TO_PROJECT}):
```
source {"${PATH_TO_PROJECT}/./simWorkspace/Identity/xsim/spinal_xsim.tcl"}
couldn't read file ""${PATH_TO_PROJECT}/./simWorkspace/Identity/xsim/spinal_xsim.tcl"": no such file or directory
```

The file does exist, but the problem is that this string is getting double quoted when passed to vivado. I narrowed the problem to #1246 which was fixing XSim for windows. The problem is in [line 243](https://github.com/SpinalHDL/SpinalHDL/pull/1246/files#diff-279ddd318041d78f9dd90c955e9bdc0ca9474da6f2e55d77aa075a22bd962c90R243):

```scala
val command = s""" vivado -mode batch -source "$vivadoScriptPath" """.trim
```

The original code did not have a quote around `$vivadoScriptPath`

Readon:
Closes #1501

Still needs to be tested in different platforms and different Vivado versions.

- [ ] MacOS
- [x] Linux
- [x] Windows
- [x] MSys2

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/sim/src/main/scala/spinal/sim/XSimBackend.scala
+++ b/sim/src/main/scala/spinal/sim/XSimBackend.scala
@@ -240,7 +240,7 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
     outFile.close()
 
     val vivadoScriptPath = scriptPath.replace("\\", "/")
-    val command =  s""" vivado -mode batch -source "$vivadoScriptPath" """.trim
+    val command =  s""" vivado -mode batch -source $vivadoScriptPath """.trim
     doCmdVivado(command,
       new File(workPath),
       "Generation of vivado script failed")
@@ -254,7 +254,7 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
         s"${cmd}.bat"
       }
     } else {
-      s"${cmd}.sh"
+      s"${simulatePath}/${cmd}.sh"
     }
   }
 
--- a/sim/src/main/scala/spinal/sim/XSimBackend.scala
+++ b/sim/src/main/scala/spinal/sim/XSimBackend.scala
@@ -249,11 +249,7 @@ class XSimBackend(config: XSimBackendConfig) extends Backend {
 
   def getScriptCommand(cmd: String) = {
     if (isWindows) {
-      if (isMsys){
-        s"${simulatePath}\\${cmd}.bat"  
-      } else {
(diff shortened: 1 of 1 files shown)


Readon:
Tested on MSYS2 which might meet problem on Vivado's settings64.sh which use windows's style path as default.
So that in PATH there would be "C:\Xilinx\Vivado" like string, and lead to an invalid searching path for vivado.
Write one your own or convert paths in .setting64-Vivado.sh file into MSYS2 or Unix like style could help.
I have written one [python script](https://gist.github.com/Readon/d0f662e193ec429efd08bd2d55f7ecde) to help with that.

oletf:
Tested on Windows, works without problem!

I've also tested modifying [the line](https://github.com/SpinalHDL/SpinalHDL/blob/0b1d1e9222755cda45793bfa2403b48cbf5c0936/sim/src/main/scala/spinal/sim/XSimBackend.scala#L254) to full path removing the "not msys2" branch :
 ```scala
   def getScriptCommand(cmd: String) = {
    if (isWindows) {
      s"${simulatePath}\\${cmd}.bat"
    } else {
      s"${simulatePath}/${cmd}.sh"
    }
  }
 ```
works well too !

PR #1555, opened 2024-10-06
Watch unassigned inputs

### LINKED ISSUE #1006: The missing initialization of signals in simulation is better to be checked and warned.
The existing simulation engine allows the presence of uninitialized signals without prompting.
The uninitialized signals will be assigned a random value determined by the seed, thus the errors will be overwritten or cause the simulation to fail.

It would be a good solution to check the uninitialized signals through the engine at the beginning of the simulation.

Readon:
Closes #1006 

# Impact on code generation

# Checklist

- [x] Unit tests were added


Dolu1990:
Seems ok to me ^^

PR #1560, opened 2024-10-08
Fix isUnnamed for NameableByComponent

### LINKED ISSUE #1551: Issue with component naming
The fix for #1020 commit id 1c25b6c482e9e46f47b12601a75a64bc47ccb93b has introduced some issue with naming of components.
We get a component without a name/label emitted to the rtl. 

I now spend a while trying to extract a MWE but fail to do so...
But the problem always appears when I instantiate a StreamCCByToggle component while calling rework on a existing component. This makes a BufferCC be instantiated that does not get a name. (I think from calling `outputClock.withOptionalBufferedResetFrom`).
As I use the getRtlPath function to generate SDC constraints the label need to be deterministic (and not empty).

I do not fully unerstand what happen in `PhaseAllocateNames` and the change seems minimal.
I tried changing ` for (c <- componentsReversed)` to ` for (c <- componentsReversed.reverse)`  (the second loop) this seems to fix the issue. But not sure if this is actually correct...

If required I can spend more time trying to extract a code example, but the bug seems to be rather finicky.

g0t00:
Closes #1551

I seem to have found the problem for issue #1551:
`NameableByComponent` extends `Nameable`.
`Nameeable` defines `isUnnamed` and defines `isNamed` as the opposite, but `NameableByComponent` defines isNamed, without overwriting `isUnnamed` therefore `isUnnamed` has some wrong content.
This does funky stuff in `Component.allocateNames` (`child.isUnnamed` is checked there)
I reversed the definition in NameableByComponent.

Sadly, it takes some complicated combination to trigger this edge case which I was not able to destille a unit test from.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/Trait.scala
+++ b/core/src/main/scala/spinal/core/Trait.scala
@@ -386,7 +386,7 @@ trait Nameable extends OwnableRef with ContextUser{
     case OWNER_PREFIXED        => refOwner == null || refOwner.asInstanceOf[Nameable].isUnnamed
   }
 
-  def isNamed: Boolean = !isUnnamed
+  final def isNamed: Boolean = !isUnnamed
 
   def getName(): String = getName("")
   def getPartialName() : String = name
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks ^^

PR #1561, opened 2024-10-09
add fp-interpolator to avoid errors in 2.13

### LINKED ISSUE #1559: Physical Number representation wrong on Scala 2.13
Scala 2.13's f-string parser would not call formatTo function in Formattable. 
This lead following lines failed.
``` scala
assert(f"${HertzNumber(5123000)}" == "5.123 MHz")
assert(f"${HertzNumber(5123000)}%s" == "5.123 MHz")
assert(f"${HertzNumber(5123000)}%S" == "5.123MHz")
.....
``` 

Also fails the test spinal.core.PhysicalNumbers

Readon:
Closes #1559 

avoid use official f-interpolator, use fp"xxxxx". 
This will change API.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/core.scala
+++ b/core/src/main/scala/spinal/core/core.scala
@@ -599,4 +599,19 @@ package object core extends BaseTypeFactory with BaseTypeCast {
   object LITTLE extends Endianness
   /** Big-Endian */
   object BIG    extends Endianness
+
+  implicit class PhysicalNumberPimper(val sc: StringContext) extends AnyVal {
+    def fp(args: Any*): String = {
+      val strings = sc.parts.iterator
+      val expressions = args.iterator
+
+      var formatted = Seq(strings.next())
+      formatted ++= strings.map { part =>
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks ^^ 
One thing is why fp ?
As that interpolator is type agnostic, could be sf instead ? (Spinal Formater)

Readon:
> Thanks ^^ One thing is why fp ? As that interpolator is type agnostic, could be sf instead ? (Spinal Formater)

it's formatter for physical number^^
spinal formatter seems can cover this.

Dolu1990:
i would say, let's go for sp instead of fp, as in the future other issues like that could be workaround the same way.

Readon:
> i would say, let's go for sp instead of fp, as in the future other issues like that could be workaround the same way.

sf is better IMO, this code could be worked not only for physical number but also other class extended java.util.Formattable

PR #1562, opened 2024-10-10
Allow user to swap out module dependencies when included as a git submodule

KireinaHoro:
Due to the current way the `build.sc` file is written, when a user includes SpinalHDL as a submodule, the `lib`, `sim` and compiler plugin modules will be compiled multiple times since they depend on the instantiated module inside the build.sc and does not allow overriding.  This PR would allow the user to override these modules and supply their own instance:

https://github.com/KireinaHoro/enzian-pio-nic/blob/634ac4be382f57613339797f55ef3aed16dd0583/build.sc#L18-L39

Draft for now due to [a limitation](https://github.com/com-lihaoyi/mill/issues/3715) in Mill, leading to the ugly `Some(...)` hack.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -39,14 +39,14 @@ trait SpinalPublishModule extends PublishModule {
 }
 
 object idslpayload extends Cross[IdslPayload](Version.SpinalVersion.compilers)
-trait IdslPayload extends SpinalModule with SpinalPublishModule with CrossSbtModule {
+trait IdslPayload extends SpinalModule with SpinalPublishModule {
   def mainClass = Some("spinal.idslpayload")
   override def artifactName = "spinalhdl-idsl-payload"
   def ivyDeps = super.ivyDeps() ++ Agg(ivy"org.scala-lang:scala-reflect:${scalaVersion}")
 }
 
 object idslplugin extends Cross[IdslPlugin](Version.SpinalVersion.compilers)
-trait IdslPlugin extends SpinalModule with SpinalPublishModule with CrossSbtModule {
+trait IdslPlugin extends SpinalModule with SpinalPublishModule {
   def mainClass = Some("spinal.idslplugin")
   override def artifactName = "spinalhdl-idsl-plugin"
   def moduleDeps = Seq(idslpayload(crossScalaVersion))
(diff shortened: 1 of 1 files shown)


KireinaHoro:
The solution from Mill is to use `ModuleRef`; changed and ready for review.

Readon:
I am wondering that why it's necessary to use SpinalHDL as subproject but a dependency?
This modification would cause the building of SpinalHDL standalone failed at 2.11 version with command.
``` bash
mill __.compile
```

On other hand, there is no CI for enzian-pio-nic, i could not tell if it works for all cases.

KireinaHoro:
I believe it’s a common use case since one would want to work on a private fork—to develop new features, fix bugs, etc.  Dolu does the same with Vex/VexiiRiscv (albeit not as a submodule) using [sbt](https://github.com/SpinalHDL/VexiiRiscv/blob/dev/build.sbt#L3), so I don’t think this would be that niche of a use case.

The 2.11 breaking is unfortunate and should be caught if we had any CI checks for mill.  Only if someone would port the sbt ones…. I can put some time aside for that but it would have to be a separate PR.

I totally didn’t understand what you meant with the comment about my project that depend on SpinalHDL.  I don’t understand why you would want to enforce a downstream project to have CI—I linked the project as an example, not suggesting that it should be the golden standard.  I do have CI, just not on GitHub, and it’s for a totally different purpose—obviously not for testing multiple Scala version being compatible.

Readon:
> I believe it’s a common use case since one would want to work on a private fork—to develop new features, fix bugs, etc. Dolu does the same with Vex/VexiiRiscv (albeit not as a submodule) using [sbt](https://github.com/SpinalHDL/VexiiRiscv/blob/dev/build.sbt#L3), so I don’t think this would be that niche of a use case.
> 
> The 2.11 breaking is unfortunate and should be caught if we had any CI checks for mill. Only if someone would port the sbt ones…. I can put some time aside for that but it would have to be a separate PR.
> 
> I totally didn’t understand what you meant with the comment about my project that depend on SpinalHDL. I don’t understand why you would want to enforce a downstream project to have CI—I linked the project as an example, not suggesting that it should be the golden standard. I do have CI, just not on GitHub, and it’s for a totally different purpose—obviously not for testing multiple Scala version being compatible.

Ok, I will solve the problem of 2.11 and 2.13 then.

Readon:
I just want to make sure the PR works for various cases.

PR #1563, opened 2024-10-11
Generate more meaningful signal name in lib.fsm

Konecho:
generate more meaningful & understandable signal name in fsm

# Impact on code generation

Only signal name changed, behaviour unchanged.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/fsm/State.scala
+++ b/lib/src/main/scala/spinal/lib/fsm/State.scala
@@ -28,6 +28,7 @@ package spinal.lib.fsm
 import spinal.core._
 
 import scala.collection.mutable.ArrayBuffer
+import spinal.lib.IntRicher
 
 /**
   * This trait indicate the entry point of the state machine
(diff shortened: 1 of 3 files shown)


lib/src/main/scala/spinal/lib/fsm/State.scala
```diff
@@ -219,18 +219,18 @@ class StateMachineSharableRegUInt {
   * }}}
   *
   */
-class StateDelay(cyclesCount: UInt)(implicit stateMachineAccessor: StateMachineAccessor) extends State with StateCompletionTrait {
```
Readon, line 222: This modification seems to make the StateDelay's delay count unchangable.

Konecho: The motivation for this change is that spinal doesn't seem to have something like constant folding, causing the counter reset of the generated hdl to be split into two parts - the reset value comes at the very beginning of the file while the assigement at the end.

Konecho: I've changed the delay count api to be compatible with the previous `Int`, `UInt` or `TimeNumber` inputs.

Dolu1990: Ahhh interresting ^^ 
Seems ok to me now

Dolu1990:
thanks ^^

PR #1567, opened 2024-10-14
VHDL basic assert message formatting

knapheide:
Closes #1532

Allow format strings as assert messages for VHDL, just like the Verilog emitter does.

# Impact on code generation

Formatted assert messages are emitted as VHDL. This shouldn't change anything for code that built previously.

The new `pkg_toString` function might lead to collisions.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVhdl.scala
@@ -639,8 +639,8 @@ class ComponentEmitterVhdl(
               .map {
                 case m: String =>
                   "\"" +
-                    m.replace("\n", "\\n")
-                      .replace("\"", "\\\"") +
+                    m .replace("\"", "\"\"")
+                  .replace("\n", "\" & LF & \"")+
                   "\""
                 case m: SpinalEnumCraft[_] =>
                   require(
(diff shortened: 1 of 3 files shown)


knapheide:
A couple of open questions:
- Are there any signal types missing?
- Is the name `pkg_toString` unique enough to avoid collisions? Should that function reside somewhere else?
- I would like to write a test for this. Is there a way to access the output of the simulator (so i can check what gets printed)?

Dolu1990:
Hi ^^

> Are there any signal types missing?
> Is the name pkg_toString unique enough to avoid collisions? Should that function reside somewhere else?

Seems all good to me.

> I would like to write a test for this. Is there a way to access the output of the simulator (so i can check what gets printed)?

Hmmm, that is the kind of things which can be tricky.
Maybe via Console.withOut. here is an example : 

```scala
    val compiled = SimConfig.withIVerilog.compile(new Component {
      val x = out(Reg(UInt(8 bits))) init(0)
      x := x + 1
      report(L"miaou $x")
    })


    var buffer = new StringBuilder()
    Console.withOut(new OutputStream {
      override def write(i: Int) = {
        buffer.+=(i.toChar)
      }
    }) {
      compiled.doSim { dut =>
        val cd = dut.clockDomain
        cd.forkStimulus(10)
        cd.waitSampling(42)
        println(dut.x.toInt)
      }
    }
    println("#########")
    println(s"Got : $buffer")
  }
  ```

knapheide:
> Maybe via Console.withOut. here is an example :

Nice, that works!

I would call this ready then

Dolu1990:
Thanks ^^

PR #1584, opened 2024-10-30
Add missing initialization of `portLowSpeed` in UsbLsFsPhy

petrmikheev:
Add missing initialization of `portLowSpeed` in UsbLsFsPhy.

Currently `portLowSpeed` is assigned only in [this](https://github.com/SpinalHDL/SpinalHDL/blob/70750fc0e74703cd821cf3a31bcfea7eb38ddf79/lib/src/main/scala/spinal/lib/com/usb/phy/UsbHubPhy.scala#L624) line and happens only on reset of a already connected device, that seems to be bug.

### DIFF
--- a/lib/src/main/scala/spinal/lib/com/usb/phy/UsbHubPhy.scala
+++ b/lib/src/main/scala/spinal/lib/com/usb/phy/UsbHubPhy.scala
@@ -409,7 +409,7 @@ case class UsbLsFsPhy(portCount : Int, sim : Boolean = false) extends Component 
 
   val ports = for((usb, ctrl, management) <- (io.usb, io.ctrl.ports, io.management).zipped) yield new Area{
 //    val connected = Reg(Bool()) init(False) //TODO
-    val portLowSpeed = Reg(Bool())
+    val portLowSpeed = Reg(Bool()) init(False)
 //    val enable = Reg(Bool())
 
 //    ctrl.connected := connected
@@ -646,6 +646,9 @@ case class UsbLsFsPhy(portCount : Int, sim : Boolean = false) extends Component 
         when(timer.DISCONNECTED_EOI){
           ctrl.connect := True
           goto(DISABLED)
+        }
+        when(!ctrl.reset.valid && filter.io.filtred.dm =/= filter.io.filtred.dp) {
+          portLowSpeed := !filter.io.filtred.d
         }
       }
 


Dolu1990:
Hi,

Here is a quote from the USB 2.0 spec : 

---
11.8.2 Speed Detection
At the end of reset, the bus is in the Idle state for the speed recorded in the port status register. Speed
detection is described in Section 7.1.7.5.
If the device connected at the downstream facing port is high-speed, the repeater (rather than the
Transaction Translator) is used to signal between this port and the upstream facing port.
Due to connect and start-up transients, the hub may not be able to reliably determine the speed of the device
until the transients have ended. The USB System Software is required to "debounce" the connection and
provide a delay between the time a connection is detected and the device is used (see Section 7.1.7.3). At
the end of the debounce interval, the device is expected to have placed its upstream facing port in the Idle
state and be able to react to reset signaling. The USB System Software must send a
SetPortFeature(PORT_RESET) request to the port to enable the port and make the attached device ready for
use.
The downstream facing port monitors the state of the D+ and D- lines to determine if the connected device
is low-speed. If so, the PORT_LOW_SPEED status bit is set to one to indicate a low-speed device. If not,
the PORT_LOW_SPEED status bit is set to zero to indicate a full-/high-speed device. Upon exit from the
reset process, the hub must set the PORT_HIGH_SPEED status bit according to the detected speed. The
downstream facing port performs the required reset processing as defined in Section 7.1.7.5. At the end of
the Resetting state, the hub will return the bus to the Idle state that is appropriate for the speed of the
attached device and transition to the Enabled state

---

Overall, the speed detection seems to be described to be handled around the resetting state. 
The disconnected state is never mentionned.

So i would say the current implementation seems fine ?

About the missing initialization, looking at the Open Host Controller Interface Specification for USB,

HcRhPortStatus.LowSpeedDeviceAttached bit under the root hub reset is specified to be "Xb". Seems like the spec doesn't specify any specific reset value for that one.

Dolu1990:
By the way, how did you ended up reading the code of this SpinalHDL / OHCI controller :D ?

petrmikheev:
Oh... You are right. I read the spec not carefully enough and missed the "at the end of reset" part. Sorry.

> By the way, how did you ended up reading the code of this SpinalHDL / OHCI controller :D ?

In my usual way of doing weird stuff. Tried to use it with dma interface connected to a small on-chip ram rather than to the main memory (didn't want to add additional master to interconnect). Got errors from OHCI driver (probably because of incorrect dma region specification in DTS, haven't looked deep into it yet). Then tried to write something manually to OHCI registers and see if it works. Faced X-prop in simulation (yeah, because I haven't realized that I must reset the port before everything else), and then started looking from where this X comes from.

Dolu1990:
Ahhhhh keep in mind, i never did simulations of the controller with x-prop, instead i did plenty simulation with verilator which only simulate 1 and 0.
So it is possible you are facing x-prop issues (which are simulation artefacts, not hardware bugs)

Let's me know how things goes :)

With which memory bus are you using it ?

petrmikheev:
apb3 :)

https://github.com/petrmikheev/endeavour/blob/master/rtl/spinal/src/main/scala/endeavour/usb.scala

Though I haven't managed to make it work in linux. Likely I will switch to tilelink and real dma to main memory.

petrmikheev:
> instead i did plenty simulation with verilator which only simulate 1 and 0.

I probably also should tried verilator, but it doesn't seem easy: my testbench includes sdram and sdcard verilog models which are connected via inout wires. I afraid verilator would require me to change how this all is structured.

Dolu1990:
> my testbench includes sdram and sdcard verilog models which are connected via inout wires. I afraid verilator would require me to change how this all is structured.

Right it will not play well at all there ^^

petrmikheev:
> Let's me know how things goes :)

Unfortunately I haven't managed to get it to work :(

[Here](https://github.com/petrmikheev/endeavour/blob/try_usb/rtl/spinal/src/main/scala/endeavour/usb.scala) is scala code where I instantiate SpinalHDL OHCI controller

This is my test program in C. OHCI controller makes DMA request to an invalid address and the simulation is terminated with "FAILURE Tilelink decoder miss". Full code is [here](https://github.com/petrmikheev/endeavour/blob/try_usb/software/usbtest3.c).
Do you see here anything that I do obviously wrong?

```c
#define SIMULATION

  volatile struct OHCICtrl* ctrl = (void*)USB_OHCI_BASE;
  bios_printf("HcRevision: 0x%x\n", ctrl->HcRevision);
  ctrl->HcControl = 0x0;  // hub reset

#ifndef SIMULATION
  wait(10000); // keep hub reset for 10ms
#else
  wait(20); // during simulation 10ms is too long
#endif

  ctrl->HcFmInterval = (1<<31 /*FIT=1*/) | (0x2000 << 16 /*FSLargestDataPackage=1KB*/) | (0x2edf /*FrameInterval for 1ms frame*/);
  ctrl->HcPeriodicStart = 0x2a2f; // 90% of FrameInterval
  ctrl->HcInterruptEnable = 0x80000000; // disable all

  volatile struct HCCA* hcca = malloc(256);  // 0x80500000 - 0x805000ff
  for (int i = 0; i < 32; ++i) hcca->interrupt_table[i] = 0;
  ctrl->HcHCCA = (long)hcca;

  volatile struct ED *ed = malloc(16);       // 0x80500100 - 0x8050010f
  volatile struct TD *td = malloc(16 * 3);   // 0x80500110 - 0x8050013f

  // Initialize default endpoint for control transfers
  ed->flags = 64 << 16; // max 64 byte, full-speed, get direction from TD, USB address = 0, endpoint number = 0
  ed->head = ed->tail = &td[0]; // head==tail, so it has no TDs
  ed->nextED = 0;

  // Set to operational state, enable control lists.
  // But ed.head == ed.tail, so it shouldn't do anything yet.
  ctrl->HcControlHeadED = (unsigned)ed;
  ctrl->HcControl = 0x80 + (1<<4);

#ifndef SIMULATION
  wait(500);

  // reset both ports
  ctrl->HcRhPortStatus[0] = 1<<4;
  ctrl->HcRhPortStatus[1] = 1<<4;

  wait(100000);
  for (int i = 0; i < 2; ++i) {
    int state = ctrl->HcRhPortStatus[i];
    bios_printf("port%u: connected=%u enabled=%u\n", i, state & 1, (state>>1) & 1);
  }
#else
  // skip port reset because it is too slow for simulation
#endif
```

Questions:
* Should HcFmInterval initialization be done during hub reset, or after switching to operational state?
* What is the appropriate value for FSLargestDataPackage?
* In my case there are two downstream usb ports. How to control which port will be used for a specific endpoint? Or controller sends same message to all enabled ports at the same time?

```c
struct UsbRequest *request = malloc(sizeof(struct UsbRequest));                 // 0x80500140 - 0x80500147
  struct UsbDeviceDescriptor *descr = malloc(sizeof(struct UsbDeviceDescriptor)); // 0x80500148 - 0x80500159
  for (char* p = descr; p < (char*)(descr + 1); ++p) *p = 0; // fill with zeros

  // Request UsbDeviceDescriptor
  request->bmRequestType = 0x80;
  request->bRequst = 6; // GET_DESCRIPTOR
  request->wValue = 1;
  request->wIndex = 0;
  request->wLength = sizeof(struct UsbDeviceDescriptor);

  td[0].flags = (7<<21 /*no interrupt*/) | (0<<19 /*SETUP*/);
  td[0].nextTD = &td[1];
  td[0].buf_start = request;
  td[0].buf_end = (char*)request + sizeof(struct UsbRequest)-1;

  // Receive UsbDeviceDescriptor
  td[1].flags = (7<<21 /*no interrupt*/) | (2<<19 /*IN*/);
  td[1].nextTD = &td[2];
  td[1].buf_start = descr;
  td[1].buf_end = (char*)descr + sizeof(struct UsbDeviceDescriptor)-1;

  for (int i = 0; i < 2; ++i) {
    bios_printf("td[%u] = { %08x %08x %08x %08x }\n", i, td[i].flags, td[i].buf_start, td[i].nextTD, td[i].buf_end);
  }
  // td[0] = { 00E00000 80500140 80500120 80500147 }
  // td[1] = { 00F00000 80500148 80500130 80500159 }

  // Start processing UsbDeviceDescriptor request
  ed->tail = &td[2];
  ctrl->HcCommandStatus = 2;  // ControlListFilled=1

  wait(1000);  // Simulation fails with "FAILURE Tilelink decoder miss ???"
```

When I look what goes through ohci dma bmb bus I see that controller for some reason clears the topmost byte in `td[0].buf_start` (i.e. changing buf_start from `80500140` to `00500140`) and then tries to read from this invalid address (main memory starts from `0x80000000`).

petrmikheev:
> When I look what goes through ohci dma bmb bus I see that controller for some reason clears the topmost byte in td[0].buf_start (i.e. changing buf_start from 80500140 to 00500140) and then tries to read from this invalid address (main memory starts from 0x80000000).

Nevermind. Actually it was a bug in my ddr sdram controller! Write mask comes with an offset from data and wrong bytes are being written. Didn't notice before because vexiiriscv lsu L1 makes only unmasked writes. This is what happens with people who are too lazy to write unit tests for every module...

petrmikheev:
OK, this bug is fixed, but can't make anything working anyway.

In the test program that I linked above the first transfer (SETUP with GET_DESCRIPTOR request) is handled successfully (CC=0), but the second one (IN that should receive the answer) gets STALL (CC=4) and receives no data. Maybe I somehow create UsbLsFsPhy incorrectly. Can't find a reasonable way to debug: with iverilog I can simulate only a few milliseconds, end even usb reset takes much longer.

petrmikheev:
Found an option `sim=true` in UsbLsFsPhy that reduces reset time. Added in testbench [this core](https://github.com/ultraembedded/core_usb_cdc) as a dummy usb device. And...

image

In simulation it works and successfully reads device descriptor. On real device it doesn't. I hate such situations! Damn io timing constraints! I hoped that I won't face this kind of problem again after I spend days in attempts to make ddr1 sdram controller work stable.

petrmikheev:
It works! Well, not fully works, but at least I got something.

```
[    0.979469] mmcblk  reg 0x3000  rca 0x59b4  size 29007 MB
[    1.006237]  mmcblk0: p1 p2
[    1.031086] hub 1-0:1.0: USB hub found
[    1.117851] Freeing initrd memory: 4096K
[    1.120771] hub 1-0:1.0: 2 ports detected
[    1.127058] Initializing display driver
[    1.430675] debug_vm_pgtable: [debug_vm_pgtable         ]: Validating architecture page table helpers
[    1.442136] Legacy PMU implementation is available
[    1.444639] clk: Disabling unused clocks
[    1.445112] ALSA device list:
[    1.445477]   No soundcards found.
[    1.448471] set_termios ttyS0 CFG=0x1019f
[    1.467144] RAMDISK: ext2 filesystem found at block 0
[    1.467668] RAMDISK: Loading 4096KiB [1 disk] into ram disk...  
[    1.757315] usb 1-1: new full-speed USB device number 2 using ohci-platform
[    1.763287]  
[    2.142553] input: Logitech USB Receiver as /devices/platform/a000.ohci/usb1/1-1/1-1:1.0/0003:046D:C52B.0001/input/input0
[    2.148324]  
[    2.213479] hid-generic 0003:046D:C52B.0001: input: USB HID v1.11 Keyboard [Logitech USB Receiver] on usb-a000.ohci-1/input0
[    2.223604]  
[    2.626721] ohci-platform a000.ohci: frame counter not updating; disabled
[    2.627965] ohci-platform a000.ohci: HC died; cleaning up
```

Dolu1990:
Nice ^^

Hmm one question is how is the memory coherency implemented between the OHCI and the CPU ?

in the VexiiRiscv Tilelink and VexRiscv BMB SoC, it is implemented in hardware.
But in your case how did you connected the OHCI to your system ?

Dolu1990:
>  I moved ohci to cpu clock domain and connected dma to tilelink hub via BmbToTilelink bridge

 BmbToTilelink => 
 
 You used it manualy ?
 
 Else, i also ported OHCI to integrate well with tilelink : 
 https://github.com/SpinalHDL/SpinalHDL/blob/85e9f5246515c83c7e7c0f9e48cae159475d557e/lib/src/main/scala/spinal/lib/com/usb/ohci/UsbOhciTilelinkFiber.scala#L13
 
 >  Unfortunately it added about 2000 LE to used area.
 
That is weird, quite a lot XD

petrmikheev:
> You used it manualy ?

Yes (there is a link to the code in one of my comments above). I looked at UsbOhciTilelinkFiber, but decided to keep ctrl interface to be apb3 for now, since currently everything that I have connected to dBus is apb3.

> That is weird, quite a lot XD

Don't know why. It is `UsbHubLsFs.CtrlCc`, `BmbToTilelink`, and a new master added to tilelink hub. Maybe the verilog compiler in quartus decided to optimize something by duplicating parts of the design.

> To make linux use this RAM block I tried to add to dts: ...

Update: it is actually easy to do, requires a small driver that just calls `usb_hcd_setup_local_mem` and forwards everything to generic ohci.
Example: https://github.com/torvalds/linux/blob/master/drivers/usb/host/ohci-sm501.c#L160

Later I'll try again this approach with a small dedicated dma ram.

petrmikheev:
I am stuck again.

Currently OHCI DMA is connected to main memory via tilelink hub.

When I try to boot linux kernel I always get a deadlock right after USB device detection. Sometimes the deadlock affects tilelink hub - I see it because video controller starts filling the whole screen with the line it has in buffer, means it can't read next line from memory.

Unfortunately I haven't managed to reproduce the deadlock in simulation and have no idea how to debug it.

Also I tried to use OHCI interface manually, without linux. I can send control messages (get descriptors, set address, set configuration, etc.) and it works perfectly. But for some reason interrupt transfers (trying keyboard poll) don't work.

What I do:
- allocate 256 bytes (aligned by 256 byte border) for HCCA, fill with zeros, set HcHCCA register
- set HcControl to `0x80 | (1<<4) | (1<<2)` - operational mode, allow control transfers and periodic transfers
- do other initialization, reset port, read descriptors, detect endpoint ids, set address, set configuration
- create a new ED for an IN interrupt endpoint
- create a new TD with a 8-byte IN transfer
- set ed->head to the new TD; set ed->tail to a dummy TD; set td->nextTD to the same dummy TD.
- fill first 128 bytes of HCCA with 32 links to the new ED.

If I understand correctly, after that OHCI should try to do the transfer in every frame. But nothing happens - it doesn't even set any error code in the TD. Do I miss something obvious?

Also during experimenting I found something that might be a bug:
- If I set FSLargestDataPackage to a big value and enable periodic transfers, there is a deadlock on next access to HcFmInterval:
```c
ctrl->HcFmInterval = (1<<31 /*FIT=1*/)
                     | (0x2000 << 16 /*FSLargestDataPackage=1KB*/)  // with 0x200 << 16 works fine
                     | (0x2edf /*FrameInterval for 1ms frame*/);
ctrl->HcPeriodicStart = 0x2a2f; // 90% of FrameInterval
...
ctrl->HcControl = 0x80 | (1<<4) | (1<<2); // without 1<<2 also works fine
...
ctrl->HcControl = 0x0;  // reset hub
printf("%08x\n", ctrl->HcFmInterval);  // deadlock in this line
ctrl->HcFmInterval = (1<<31 /*FIT=1*/)  // or in this line if I comment the printf
                     | (0x2000 << 16 /*FSLargestDataPackage=1KB*/)
                     | (0x2edf /*FrameInterval for 1ms frame*/);
```

Another note:
- Sometimes controller increments scheduling overrun counter in HcCommandStatus even if HCCA interrupts table is completely empty. Is it normal?

Dolu1990:
Hi,

> Unfortunately I haven't managed to reproduce the deadlock in simulation and have no idea how to debug it.

I just double check on hardware using the tilelink ohci, i got all to work in linux debian rv64gc (mouse, keyboard, audio out, usb drive, hub)

How the is the OHCI DMA connected to the system right now ?
It is going to the hub fiber ?
Overall, the OHCI tilelink DMA will only do the following kind of accesses : 
read/wrte aligned 4/16/32/64 bytes

So, to debug such case, i would say you would need to put a logic analyser in the FPGA. What FPGA / toolchain are you using ?
There is typicaly one signal which is interresting, it is the : 
dmaCtx.pendingCounter.
This is a 4 bits counter to track the progress of DMA transaction. depending the broken behaviour, it could be used to derive a trigger when it stay stuck too long.

> If I set FSLargestDataPackage to a big value and enable periodic transfers, there is a deadlock on next access to HcFmInterval:

Can you send me a wave of the sim were it dead lock ? (with all the signal from the OHCI stored)

> But for some reason interrupt transfers (trying keyboard poll) don't work.

Can you send a simulation wave aswell ?

> Sometimes controller increments scheduling overrun counter in HcCommandStatus even if HCCA interrupts table is completely empty. Is it normal?

Seems not normal. Can you sent a simulation wave aswell ?


In general, i worked on the USB OHCI controller quite some time ago, i'm a bit rusty XD
So, a simulation wave + you expectation of what should happen can help me looking at the status of the simulated controller, and then cross checking with the spec :D

petrmikheev:
> How the is the OHCI DMA connected to the system right now ?
> It is going to the hub fiber ?

Yes, connected to the hub. The code is [here](https://github.com/petrmikheev/endeavour/blob/master/rtl/spinal/src/main/scala/endeavour/EndeavourSoc.scala#L128).

> What FPGA / toolchain are you using ?

Intel MAX10 10M50SAE144. Quartus Prime Lite.

> So, to debug such case, i would say you would need to put a logic analyser in the FPGA.

Well... It is kind of complicated in my case. My FPGA board has integrated non-standard programmer based on FTDI chip. The only implemented functionality is to write SVF file (exported from quartus) via jtag. No IDE integration, no on-chip debugging. So I am trying to simulate everything in icarus verilog.

---

I did a few more simulation experiments and seems I found the issue!

1.

I reduced reset timings in UsbLsFsPhy and tried, reduced frame interval, and tried to reproduce it in simulation. Found that it doesn't send SOF regularly.

Then I looked to UsbOhci source and found that:
- Operation StateMachine can go into `SOF` state only from `WAIT_SOF`.
- `WAIT_SOF` can happen only after `limitCounter` reaches zero.
- [limitCounter](https://github.com/SpinalHDL/SpinalHDL/blob/dev/lib/src/main/scala/spinal/lib/com/usb/ohci/UsbOhci.scala#L529) is set to `FSLargestDataPacket` every time when `hcFmRemaining` reaches zero.
- If `FSLargestDataPacket` > `FrameInterval` (as I had in this simulation), then SOF never happens (except the first one).

**It is not the direct source of the deadlock, but it is probably should be considered a bug - description of `FSLargestDataPacket` doesn't mention that incorrect value can stop SOFs.**

2.

Tried simulation with `FrameInterval=0x2edf`, `FSLargestDataPackage=0x200` and I caught it!

image

Explanation:
- 32bit-wide OHCI DMA makes 32 bit request at address 0x8050003c (aligned by 4 bytes, but not 8 bytes)
- OHCI DMA is connected via 64bit tilelink bus to a hub. The hub is connected to memory controller (via tilelink.fabric.Axi4Bridge). Memory controller has 64bit AXI4 bus.
- Memory controller receives request to read 8 byte at unaligned address 0x8050003c and returns
```
   0x44  0x40   0x3c
64'hXXXXXXXX8050060
```
- Tilelink hub relies that the address was aligned (0x---38) and extracts 0x44-0x40 part instead of 0x40-0x3c part.
- OHCI received 32'hXXXXXXXX and uses it as address for the next request.
- Invalid address leads to deadlock.

**If I am not mistaken AXI4 standard doesn't forbid unaligned access. Then either tilelink hub or tilelink.fabric.Axi4Bridge should zero the last bits in the address if they rely that the operation is aligned.**

Dolu1990:
Hi,

> If FSLargestDataPacket > FrameInterval (as I had in this simulation), then SOF never happens (except the first one).

Ahhh so, it isn't something that should be fixed, as it would eat hardware ressources for something that should never be done.

> If I am not mistaken AXI4 standard doesn't forbid unaligned access. Then either tilelink hub or tilelink.fabric.Axi4Bridge should zero the last bits in the address if they rely that the operation is aligned.

Hmm i think you are wrong on what the AXI4 spec says : 
image
Unaligned access on AXI are kinda dump, see the alignement of the first access above
You should return this instead  :  (?)

```scala
   0x40  0x3c   0x38
64'h8050060XXXXXXXX
``` 

Pretty sure this is the bug you have on hardware.

petrmikheev:
> Ahhh so, it isn't something that should be fixed, as it would eat hardware ressources for something that should never be done.

By the way, what is the optimal value of FSLargestDataPacket? Should it be exactly equal to FrameInterval?

The only explanation I found in OpenHCI spec is:
```
This field specifies a value which is loaded into the Largest
Data Packet Counter at the beginning of each frame. The
counter value represents the largest amount of data in bits
which can be sent or received by the HC in a single
transaction at any given time without causing scheduling
overrun. The field value is calculated by the HCD.
```
But the spec doesn't say how exactly is should be calculated by the HCD.
For me it was completely not obvious that if `FSLargestDataPacket > FrameInterval` the controller will not send SOF at all, and `FSLargestDataPacket < FrameInterval - PeriodicStart` means that periodic transfers will not work as controller enters WAIT_SOF state before reaching PeriodicStart.

> Hmm i think you are wrong on what the AXI4 spec says :

Thanks! It is funny that the fact that I read AXI spec not carefully enough, made me spent days on reading USB specs and digging into internals of the controller :-D
Sorry for bothering you with bugs on my side.

Dolu1990:
> By the way, what is the optimal value of FSLargestDataPacket? Should it be exactly equal to FrameInterval?

Hmm from the ohci spec : Interval |= (((Interval - MAXIMUM_OVERHEAD) * 6) / 7) << 16;
MAXIMUM_OVERHEAD is 210

> made me spent days on reading USB specs and digging into internals of the controller :-D

XDXDXD

> Sorry for bothering you with bugs on my side.

no worries

PR #1586, opened 2024-11-01
fix xsim backend improperly converting signed values to bigint

Pyromuffin:
Closes #

I noticed that I was never getting negative numbers back from simulation when using the xsim backend, regardless of type.
One reason was that non-32 or 64 bit types were having vectors returned with one byte per bit, and then the last byte wasn't being sign extended. 

# Impact on code generation

no codegen changes, this just fixes how xsim retrieves signed numbers from simulation


Dolu1990:
Thanks ^^

PR #1592, opened 2024-11-06
Fixes minimum time for Timeout and StateDelay

### LINKED ISSUE #1588: FSM StateDelay and Timeout Rounding Issues
In the [FSM StateDelay](https://github.com/SpinalHDL/SpinalHDL/blob/8ee0e9c942676fd740b22c888110f9fa696735da/lib/src/main/scala/spinal/lib/fsm/State.scala#L234) and [Timeout](https://github.com/SpinalHDL/SpinalHDL/blob/8ee0e9c942676fd740b22c888110f9fa696735da/lib/src/main/scala/spinal/lib/Utils.scala#L652), the arithmetic used does a truncation type rounding as they are kept as integer values.  When doing timing that requires a Minimum amount of time, this can lead to issues.  This is somewhat resolved in a few place in simulation code like [here](https://github.com/SpinalHDL/SpinalHDL/blob/8ee0e9c942676fd740b22c888110f9fa696735da/core/src/main/scala/spinal/core/sim/package.scala#L215), but forcing these to round up could break existing code.

I can add changes to fix this, but I would like some direction on how to best do so.  Here are my thoughts on how to resolve this:
1. Change the rounding in FSM State and Timeout to always just round up (potentially breaking existing code).
2. Add an overloaded function that takes a BigDecimal Rounding Type that can be used to round how the user sees fit.
3. Add new methods that somehow state minimum timing in their name, like "StateDelayMin" and "TimeoutMin".

What do you suggest?

mrberman87:
Closes #1588 

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/Utils.scala
+++ b/lib/src/main/scala/spinal/lib/Utils.scala
@@ -649,7 +649,10 @@ class Counter(val start: BigInt,val end: BigInt) extends ImplicitArea[UInt] {
 object Timeout {
   def apply(cycles: BigInt): Timeout = new Timeout(cycles)
 
-  def apply(time: TimeNumber): Timeout = new Timeout((time * ClockDomain.current.frequency.getValue).toBigInt)
+  def apply(time: TimeNumber): Timeout = new Timeout(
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks ^^

PR #1607, opened 2024-12-03
Fix incorrect interconnect addressWidth generation

cherrypiejam:
Closes #

When the master of a connection uses `SizeMapping` to describe its address mapping, the current Bmb interconnect generator assigns the minimal number of bits that describe the range of the mapping to be the address width of that connection (i.e., `log2Up(sizeMapping.size)`). However, this doesn't make sense as the hit function used by `SizeMapping` implies that the address used by it is a true address instead of an offset. This is because the given address (parameter used by the hit function) is compared with `base + size`. I didn't fully walk myself through how the hit function interacts with the address parameter in the interconnect, so I am not sure if I missed anything. If so, please let me know, thanks!

# Impact on code generation
I simply removed the spacial case for `SizeMapping` and use the addressWidth parameter described in the master's `accessRequirements` instead.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/bmb/Bmb.scala
+++ b/lib/src/main/scala/spinal/lib/bus/bmb/Bmb.scala
@@ -403,7 +403,7 @@ case class BmbCmd(p : BmbParameter) extends Bundle{
     def s = this
     WeakConnector(m, s, m.source,  s.source,  defaultValue = null, allowUpSize = true , allowDownSize = false, allowDrop = false)
     WeakConnector(m, s, m.opcode,  s.opcode,  defaultValue = null, allowUpSize = false, allowDownSize = false, allowDrop = false)
-    WeakConnector(m, s, m.address, s.address, defaultValue = null, allowUpSize = false, allowDownSize = true , allowDrop = false)
+    WeakConnector(m, s, m.address, s.address, defaultValue = null, allowUpSize = true, allowDownSize = true , allowDrop = false)
     WeakConnector(m, s, m.length,  s.length,  defaultValue = null, allowUpSize = true, allowDownSize = false , allowDrop = false)
     WeakConnector(m, s, m.data,    s.data,    defaultValue = () => Bits(m.p.access.dataWidth bits).assignDontCare() , allowUpSize = false, allowDownSize = false, allowDrop = true )
     WeakConnector(m, s, m.mask,    s.mask,    defaultValue = () => Bits(m.p.access.maskWidth bits).assignDontCare() , allowUpSize = false, allowDownSize = false, allowDrop = true)
@@ -456,7 +456,7 @@ case class BmbInv(p: BmbParameter) extends Bundle{
   def weakAssignFrom(m : BmbInv): Unit ={
     def s = this
     WeakConnector(m, s, m.source , s.source , defaultValue = null, allowUpSize = false, allowDownSize = false, allowDrop = false)
-    WeakConnector(m, s, m.address, s.address, defaultValue = null, allowUpSize = false, allowDownSize = false, allowDrop = false)
+    WeakConnector(m, s, m.address, s.address, defaultValue = null, allowUpSize = true, allowDownSize = false, allowDrop = false)
     WeakConnector(m, s, m.length , s.length , defaultValue = null, allowUpSize = false, allowDownSize = false, allowDrop = false)
     WeakConnector(m, s, m.all    , s.all    , defaultValue = null, allowUpSize = false, allowDownSize = false, allowDrop = false)
   }


Dolu1990:
Hi, can you provide an self contained example where the bug appear ?
Thing is, c.mapping is there to enforce the address space.

cherrypiejam:
> Hi, can you provide an self contained example where the bug appear ?

Yes, please check out this branch https://github.com/cherrypiejam/VexRiscv/tree/interconnect-width-mismatch-bug, where I made a small change to the interconnect topology of the smp cluster (by adding two bus spliters and an aggregator, which shouldn't introduce any variation in their logical behavior). To run the example and trigger the error, use the following command:

```
sbt "runMain vexriscv.demo.smp.VexRiscvLitexSmpClusterCmdGen --cpu-count=1 --netlist-name=VexRiscvBridgeLitexSmpCluster_Cc1_test --netlist-directory=build"
```

As a result, you should be get the following error:
```
[error] Exception in thread "main" spinal.core.SpinalExit: 
[error]  Error detected in phase PhaseCreateComponent
[error] ********************************************************************************
[error] ********************************************************************************
[error] (toplevel/iArbiterAggregator_bmb_arbiter/io_inputs_1_cmd_payload_fragment_address : in UInt[28 bits]) can't drive (toplevel/iArbiterAggregator_bmb_arbiter/memory_arbiter/io_inputs_1_payload_fragment_address : in UInt[32 bits]) because the width do not match
[error]     spinal.lib.bus.bmb.WeakConnector$.apply(Bmb.scala:23)
[error]     spinal.lib.bus.bmb.BmbCmd.weakAssignFrom(Bmb.scala:406)
[error]     spinal.lib.bus.bmb.BmbArbiter$$anon$3.$anonfun$new$3(BmbArbiter.scala:40)
[error]     spinal.lib.bus.bmb.BmbArbiter$$anon$3.$anonfun$new$3$adapted(BmbArbiter.scala:37)
[error]     spinal.lib.bus.bmb.BmbArbiter$$anon$3.<init>(BmbArbiter.scala:37)
[error]     spinal.lib.bus.bmb.BmbArbiter.$anonfun$memory$1(BmbArbiter.scala:28)
[error]     spinal.lib.bus.bmb.BmbArbiter.<init>(BmbArbiter.scala:28)
...
```

> .. Thing is, c.mapping is there to enforce the address space.

Isn't it the purpose of the `hit` function defined for all AddressMapping subclasses? Presumably all illegal addressing will be blocked through the `hit` function. Please correct if I am wrong.

Dolu1990:
Hmm, i would say, it may be more about fixing the BmbArbiter, to allow a 28 bits bus to be connected to a 32 bits one ?
(instead of fixing the BMB node address width)

So there, after https://github.com/SpinalHDL/SpinalHDL/blob/fb2a8df445631b3da65e84f7ee254d0c4b4c4c67/lib/src/main/scala/spinal/lib/bus/bmb/BmbArbiter.scala#L40 
adding something like : 
s.address.removeAssignments() := m.address.resized

cherrypiejam:
It would fix the compile-time error. However, I _think_, keeping the original implementation doesn't make sense, and it has an implication of a potential unintended misbehavior.

Suppose we want to instantiate an interconnect topology like the following, using `BmbInterconnectGenerator`. Let's also assume both A and B's address width are 32 bits. Finally, we use `SizeMapping` to route traffic from the decoder to the arbiter, where the base address is `0x4000_0000` and the size is `0x2` (supposing an extreme case)

```
                   +---------------------...
---A--> [decoder] -+
                   +---B--> [arbiter] -C-...
------------------------------^

```

As a result, both A and C's address fields have the size of 32 bits (this is by assumption). However, B's address will be downsized to 1 (`log2Up(0x2)`) bit due to this line https://github.com/SpinalHDL/SpinalHDL/blob/fb2a8df445631b3da65e84f7ee254d0c4b4c4c67/lib/src/main/scala/spinal/lib/bus/bmb/BmbDecoder.scala#L49

Unfortunately, I couldn't track which `resize` function this line ends up calling, so I am not sure what would happen. But I assume the extra bits will be discarded (because it only resize the `cmd.payload`, not the Bmb bus). If this is the case, the arbiter only has the knowledge of the 1st bit of the actual address (from B) from A to pass to C, leading to incorrect addressing. This matters because such misbehavior could either crash a running program or, more serious, result in unauthorized memory overwrites. If this is not the case that the extra bits are not discarded but streaming into the downstream arbiter instead, then it is an unintended performance bottleneck.

cherrypiejam:
I understand the intention and the reducing area part, and I believe you that BMB indeed uses relative address for indexing. However, there is one more thing I am confused very much. I assume when converting BMB to other buses (e.g., Wishbone), we need to recover the absolute address from the relative one for correctness. If this is true, would you show me a pointer of an example where this is implemented? 

Specifically, when instantiating a sequence of decoders such as `[decoder1] ---> [decoder2] ---->` and assuming all decoders use `SizeMapping` to route the traffic to their subsequent downstream decoder, is the base used in `SizeMapping` absolute or relative? For example, if I want a logical mapping from 0x1000_0800 to 0x1000_1000 and I want to use two decoders to achieve that. Should I use `SizeMapping(0x1000_0000, 0x1000)` and `SizeMapping(0x1000_0800, 0x800)`, or `SizeMapping(0x1000_0000, 0x1000)` and `SizeMapping(0x800, 0x800)` to achieve that? 

This is a hypothetical but important question. The reason I am asking is because, if the base is a relative address, to recover to the actual (absolute) address, we need to add all relative bases along the same bus route (which I couldn't find such an implementation). Alternatively, if the base is an absolute address, then the implementation of `hit(address)` implies that the `address` parameter must be absolute, which further implies `bmb.io.output/input.cmd.address` is absolute due to https://github.com/SpinalHDL/SpinalHDL/blob/fb2a8df445631b3da65e84f7ee254d0c4b4c4c67/lib/src/main/scala/spinal/lib/bus/bmb/BmbDecoder.scala#L44

This contradicts "address aren't absolute" statement, which makes me very confused and I'd like to know what I am missing here

cherrypiejam:
That makes sense if all addresses are relative. The new fix allows `WeakConnector` to resize`BmbInv` and `BmbCmd`'s addresses.

Dolu1990:
Thanks ^^

PR #1608, opened 2024-12-03
lib/bus/bmb/BmbDecoder: Fix missing wirings when invalidation is enabled

cherrypiejam:
Closes #

The current implementation of BmbDecoder ignores `canInvalidate` and `canSync` parameters found in the access parameters. This becomes an issue when instantiating multiple (or a single) VexRsicv cores with a stripped SRAM.

Changes made in this PR simply wires `inv`, `ack`, and `sync` ports between input and outputs.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/bmb/BmbDecoder.scala
+++ b/lib/src/main/scala/spinal/lib/bus/bmb/BmbDecoder.scala
@@ -63,9 +63,38 @@ case class BmbDecoder(p : BmbParameter,
       val context = RegNextWhen(input.context, input.fire)
       val counter = p.access.canRead generate RegNextWhen(input.transferBeatCountMinusOne, input.fire)
     }
+    val hitIndex = OHToUInt(rspHits)
 
     io.input.rsp.valid := io.outputs.map(_.rsp.valid).orR || (rspPending && rspNoHitValid)
-    io.input.rsp.payload := io.outputs.map(_.rsp.payload).read(OHToUInt(rspHits))
+    io.input.rsp.payload := io.outputs.map(_.rsp.payload).read(hitIndex)
+
+    if(p.access.canInvalidate) {
+      val outputInv = io.outputs.map(_.inv).read(hitIndex)
+      val outputAck = io.outputs.map(_.ack).read(hitIndex)
+
+      io.input.inv.weakAssignFrom(outputInv)
+      io.input.inv.valid := outputInv.valid
+
+      outputAck.weakAssignFrom(io.input.ack)
+      io.input.ack.ready := outputAck.ready
+
+      for ((output, index) <- io.outputs.zipWithIndex) {
+        output.inv.ready := io.input.inv.ready && hitIndex === index
+        output.ack.valid := io.input.ack.valid && hitIndex === index
+      }
+    }
+
+    if(p.access.canSync) {
+      val outputSync = io.outputs.map(_.sync).read(hitIndex)
+
+      io.input.sync.weakAssignFrom(outputSync)
+      io.input.sync.valid := outputSync.valid
+
+      for ((output, index) <- io.outputs.zipWithIndex) {
+        output.sync.ready := io.input.sync.ready && hitIndex === index
+      }
+    }
+
     if(!hasDefault) when(rspNoHit.doIt) {
       io.input.rsp.valid := True
       io.input.rsp.setError()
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

Right.
For the code change, i think it is bugy ?
rspHits can change as soon as there is no pending cmd -> rsp transactions.
Thing is, the PR inv/ack/sync logic uses the rspHits to route traffic aswell.

So i would say, to support inv/ack/sync, the rspPendingCounter would need to be changed, to take in acount the whole round trip
 - read : cmd -> rsp
 - write : cmd -> rsp + sync
 
Or some other changes :)

cherrypiejam:
Sorry for the late response. I push a fix to take sync into account the round trip. Would you take a look at it. Thanks!

Dolu1990:
Hi ^^

There is a few issues : 

1) U(input.lastFire) * 2   will not work well with  
cmdWait := (rspPending && (hitsS1 =/= rspHits || rspNoHitValid)) || rspPendingCounter === pendingMax
because  it can skip the rspPendingCounter === pendingMax case

2) Memory read request will only generate a BMB rsp, but no BMB sync. So the rspPending will be buggy on read

I would say, do only the  * 2 for memory writes, and fix the rspPendingCounter === pendingMax condition should be good.

cherrypiejam:
Alright I just updated the patch accordingly. However, there are one assumption I made and I'd like to make it explicit here. 

I am assuming`input.lastFire`, `io.input.rsp.lastFirst`, and `io.input.sync.valid` will only be true for one cycle. This assumption is essential for the correctness of updating `rspPendingCounter`, as we want both rsp + sync, instead of mistakenly taking sync/rsp into account twice. This assumption isn't there before the patch, as `rspPendingCounter := rspPendingCounter + U(input.lastFire) - U(io.input.rsp.lastFire)` implies `io.input.rsp..lastFire` will be true for more cycles then `input.lastFire`, which is a more relaxed assumption.

lib/src/main/scala/spinal/lib/bus/bmb/BmbDecoder.scala
```diff
@@ -51,7 +51,11 @@ case class BmbDecoder(p : BmbParameter,
     input.ready := (hitsS1, io.outputs).zipped.map(_ && _.cmd.ready).orR || noHitS1
 
     val rspPendingCounter = Reg(UInt(log2Up(pendingMax + 1) bits)) init(0)
-    rspPendingCounter := rspPendingCounter + U(input.lastFire) - U(io.input.rsp.lastFire)
+    if (p.access.canSync)
+      rspPendingCounter := (rspPendingCounter + U(input.lastFire) - U(io.input.rsp.lastFire)
+                           + U(input.lastFire && input.isWrite) - U(io.input.sync.valid && input.isWrite))
```
Dolu1990, line 56: I think - U(io.input.sync.valid && input.isWrite) is wrong. input.isWrite isn't related.
- U(io.input.sync.valid)    should be enough, as sync only come for writes already.

lib/src/main/scala/spinal/lib/bus/bmb/BmbDecoder.scala
```diff
@@ -90,7 +123,8 @@ case class BmbDecoder(p : BmbParameter,
     }
     for(output <- io.outputs) output.rsp.ready := io.input.rsp.ready
 
-    cmdWait := (rspPending && (hitsS1 =/= rspHits || rspNoHitValid)) || rspPendingCounter === pendingMax
+    cmdWait := ((rspPending && (hitsS1 =/= rspHits || rspNoHitValid))
+               || ((rspPendingCounter ^ pendingMax) & ~U(input.isWrite, log2Up(pendingMax + 1) bits)) === 0)
```
Dolu1990, line 127: I think ((rspPendingCounter ^ pendingMax) & ~U(input.isWrite, log2Up(pendingMax + 1) bits)) is wrong

For instance, let's say :
- isWrite = 1
- rspPendingCounter = 0xFF
- pendingMax = 0x100

the result will say " all good dont wait.
Maybe instead, just do : 
rspPendingCounter === pendingMax || rspPendingCounter === pendingMax-1

That should be good enough.

lib/src/main/scala/spinal/lib/bus/bmb/BmbDecoder.scala
Dolu1990: Also, io.input.sync.valid isn't right, instead, do io.input.sync.fire to ensure it is high only once cycle

Dolu1990:
Thanks ^^

PR #1611, opened 2024-12-07
add customized prefix for phaseObfuscate

facebreeze:
Closes #1595

#1595 Change `SpinalConfig.obfuscateNames` to `SpinalConfig.obfuscateNamesPrefix` to enable customizable prefix for PhaseObfuscate.

Remove always false logic of `Mem.streamReadSync`.

# Impact on code generation
No direct impact on RTL code generation.

# Checklist


lib/src/main/scala/spinal/lib/Mem.scala
```diff
@@ -48,9 +48,6 @@ class MemPimped[T <: Data](mem: Mem[T]) {
     val retValid = RegInit(False)
     val retData = mem.readSync(cmd.payload, cmd.fire, clockCrossing = crossClock)
 
-    when(ret.ready) {
-      retValid := Bool(false)
```
Readon, line 52: Is there any reason to remove this block?

core/src/main/scala/spinal/core/Spinal.scala
```diff
@@ -167,7 +167,7 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         removePruned                   : Boolean = false,
                         allowOutOfRangeLiterals        : Boolean = false,
                         dontCareGenAsZero              : Boolean = false,
-                        obfuscateNames                 : Boolean = false,
+                        obfuscateNamesPrefix           : String  = null,
```
Readon, line 170: how about set this to "" empty string to avoid errors?

lib/src/main/scala/spinal/lib/Mem.scala
facebreeze: Because of the **last win** rule,  the first condition `ret.ready` of `retValid` only wins when the second condition`cmd.ready` becomes false, however,  `cmd.ready` = `ret.isFree` = **False** means `ret.valid && !ret.ready`. It means when the `cmd.ready` is false,  `ret.ready` will be **False** as well. Therefore, the first condition will always be false for retValid, it should be removed.

core/src/main/scala/spinal/core/Spinal.scala
facebreeze: Since I'm a newbie of scala, can you explain which kind of errors will it leads to ?  D:)

lib/src/main/scala/spinal/lib/Mem.scala
Dolu1990: The function was written in a very generic way. But when we take everything toggether, yes right, that when(ret.ready) can be removed.

Reasons is : 

If ret.ready == True, then cmd.ready is True aswell (as ret.isFree = !ret.valid || ret.ready)
So, the retValid := Bool(false) will always be erased by retValid := cmd.valid

core/src/main/scala/spinal/core/Spinal.scala
Dolu1990: I think we should keep as it was before.
obfuscateNames is the feature
obfuscateNamesPrefix is a configuration for the feature, there can be more to comes.

So i think we should keep obfuscateNames, and provide a default value for obfuscateNamesPrefix ("oo_")

facebreeze: I totally agree.

lib/src/main/scala/spinal/lib/Mem.scala
Readon: Then I suggest split this modification into another PR, to keep this PR concentrate on specific topic.

Issue #1605, opened 2024-11-29
currentTestPath and Simulation Workdir Structure


distributed:
In a simulation, which by default uses Verilator, I have the following line: 

```scala
val writer = new BufferedWriter(new FileWriter(Paths.get(currentTestPath, filename).toString))
```

This works fine. However if I change to simulate `.withIverilog` the simulation crashes as follows:

```
[info] - triggerbank *** FAILED ***
[info]   java.io.FileNotFoundException: ./simWorkspace/TriggerBank/triggerbank/ctx.tcl (No such file or directory)
[info]   at java.base/java.io.FileOutputStream.open0(Native Method)
[info]   at java.base/java.io.FileOutputStream.open(FileOutputStream.java:295)
[info]   at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:236)
[info]   at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:124)
[info]   at java.base/java.io.FileWriter.<init>(FileWriter.java:67)
[info]   at simutils.WriteSimContextFile$.apply(WriteSimContextFile.scala:11)
[info]   at oscar.imxctrl.TriggerBankSuite.$anonfun$new$2(TriggerBankSuite.scala:55)
[info]   at spinal.core.sim.SimCompiled.$anonfun$doSimApi$1(SimBootstraps.scala:614)
[info]   at spinal.sim.SimThread.$anonfun$jvmThread$1(SimThread.scala:90)
[info]   at spinal.sim.JvmThread.run(SimManager.scala:51)
```

It seems to me that directory structure is different between the simulators. Here it is for Verilator:

```
$ tree simWorkspace/TriggerBank/
simWorkspace/TriggerBank/
├── rtl
│   └── TriggerBank.v
├── triggerbank
│   ├── ctx.tcl
│   └── wave.vcd
```

And here for Iverilator:

```
$ tree simWorkspace/TriggerBank/
simWorkspace/TriggerBank/
├── rtl
│   ├── __simulation_def.v
│   └── TriggerBank.v
├── shmem_name
├── test.vcd
└── TriggerBank.vvp
```

Dolu1990:
Hi,

Right, the other simulation backend would need to be updated to follow the same layout

distributed:
I've tried my hand at this, albeit not successfully. 

[...]

PR #1614, opened 2024-12-08
IVerilog and GHDL backends: put wave file in per-test paths

distributed:
Partial resolution to #1605, at least for IVerilog and GHDL backends.

Verilator backed simulation runs put their wavefiles into a per-test directory. This PR adapts the IVerilog and GHDL backends to follow this pattern as well.

This is my first SpinalHDL PR, so please bear with me :upside_down_face: I'm happy to incorporate feedback.

For instance, I'm unsure whether the route chosen is OK. Also I've modified some tests along the way (e.g. `TestIVerlilog.scala`) - I discovered these through compile errors (maybe on some test command?) but I can't figure out how to run those.

Also I lack the means to to extend this implementation to e.g. the VCS backend as I can't test it.

# Impact on code generation

None on the RTL code, a small impact (reading the wave file name from a $plus arg in the IVerilog simulation file) on simulation.

# Checklist

- [x] Unit tests were added (check that The Big Three Simulators put wavefiles in expected location)

### DIFF
--- a/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
+++ b/core/src/main/scala/spinal/core/sim/SimBootstraps.scala
@@ -295,6 +295,7 @@ object SpinalGhdlBackend {
     }
     vconfig.runFlags = config.runFlags.mkString(" ")
     vconfig.logSimProcess = config.enableLogging
+    vconfig.testPath = config.testPath
 
     val signalsCollector = SpinalVpiBackend(config, vconfig)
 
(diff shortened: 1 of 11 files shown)


Dolu1990:
Hi,

All the sim/src/test/scala/spinal/sim things are for dev only, they aren't realy made for regression, so no worries there.

> Also I lack the means to to extend this implementation to e.g. the VCS backend as I can't test it.

I think it is fine to merge this PR with VCS not careing about the feature. If somebody needs it with VCS, a new PR can be open ^^ 

>  Unit tests were added (no, should there be some test that checks that a given simulator run generates file in the correct location?)

Hmm idealy yes, for all the open simulators.
If you have time for it => nice. Else, it isn't a big deal.

Cheers :D 
Charles

distributed:
I added a unit test for the wave file locations for IVerilog, Verilator and GHDL. From my side I'm done unless you have more comments.

Dolu1990:
Thanks :D

PR #1616, opened 2024-12-09
Apb3.m2sPipe() test and fix

distributed:

I noticed that `Apb3.m2sPipe()` would have an oscillating `PENABLE` signal at its output if the input `PREADY` and `PENABLE` signals are both asserted outside of a transaction. I believe the enable and ready signals may both always be asserted outside of PSEL. If an input transaction arrives at the time the output `PENABLE` is asserted, the bus transfer will be malformed as it has no setup phase on the output side.

I added a test that reproduces the issue. Then I added a fix to `Apb3.m2sPipe()`. I also took the liberty of removing the "not tested" comment there.

# Impact on code generation

None, change is in library code.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba3/apb/APB3.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba3/apb/APB3.scala
@@ -86,15 +86,13 @@ case class Apb3(config: Apb3Config) extends Bundle with IMasterSlave {
 
   def m2sPipe(): Apb3 ={
     val that = Apb3(config)
-    // The first PENABLE of a transaction needs to be delayed by one clock cycle in any case.
-    val penableDelay = RegInit(True) clearWhen (this.PSEL.orR) setWhen (this.PENABLE && this.PREADY)
-    that.PADDR    := RegNext(this.PADDR  )
-    that.PWRITE   := RegNext(this.PWRITE )
-    that.PWDATA   := RegNext(this.PWDATA )
-    that.PSEL     := RegNext(this.PREADY ? B(0) | this.PSEL)
-    that.PENABLE  := RegNext(this.PENABLE && !this.PREADY && !penableDelay)
+    that.PADDR    := RegNext(this.PADDR)
+    that.PWRITE   := RegNext(this.PWRITE)
+    that.PWDATA   := RegNext(this.PWDATA)
+    that.PSEL     := RegNext(this.PSEL.andMask(!this.PREADY)).init(0)
+    that.PENABLE  := RegNext(this.PENABLE)
     this.PRDATA   := that.PRDATA
-    this.PREADY   := that.PREADY && that.PENABLE
+    this.PREADY   := that.PREADY && that.PENABLE && that.PSEL.orR
     if(PSLVERROR != null) { this.PSLVERROR := that.PSLVERROR }
     that
   }
--- a/lib/src/main/scala/spinal/lib/bus/amba3/apb/APB3.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba3/apb/APB3.scala
@@ -84,14 +84,15 @@ case class Apb3(config: Apb3Config) extends Bundle with IMasterSlave {
     }
   }
 
-  // !! no tested !!
   def m2sPipe(): Apb3 ={
     val that = Apb3(config)
+    // The first PENABLE of a transaction needs to be delayed by one clock cycle in any case.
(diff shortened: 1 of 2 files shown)


Dolu1990:
Hmm the one thing which seems sus to me in the overall implementation is : 
that.PSEL := RegNext(this.PREADY ? B(0) | this.PSEL)

It seems to assume that PREADY is 1 ~in idle phases~. I think that PSEL shouldn't depend on PREADY (but may have additional logic with penableDelay)

distributed:
>  I think that PSEL shouldn't depend on PREADY

Because you expect PSEL to be more "feed forward"? Are you also worried about potential combinatorial loops depending on slave implementation?

While we're at it :grin:  Where do the names `m2spipe/s2mpipe` come from? Master to Slave Pipe? Why? I have always accepted them as just names and never understood the intention.

What about the following? It uses the stall (ex penableDelay) on both sides of the register stage.

```scala
  def m2sPipe(): Apb3 ={
    val that = Apb3(config)
    val thatStall = RegInit(True) clearWhen (this.PSEL.orR) setWhen (this.PENABLE && this.PREADY)
    that.PADDR    := RegNext(this.PADDR  )
    that.PWRITE   := RegNext(this.PWRITE )
    that.PWDATA   := RegNext(this.PWDATA )
    that.PSEL     := thatStall ? B(0) | RegNext(this.PSEL)
    that.PENABLE  := RegNext(this.PENABLE)
    this.PRDATA   := that.PRDATA
    this.PREADY   := that.PREADY && that.PENABLE && !thatStall
    if(PSLVERROR != null) { this.PSLVERROR := that.PSLVERROR }
    that
  }
```

Dolu1990:
> Master to Slave Pipe

Right

> that.PSEL := RegNext(this.PREADY ? B(0) | this.PSEL)

Ahhh now i understand it XD
all good with it.

By trying to understand, i came up with this implementation, which seems to avoid having an additional penableDelay register ? : (untested)
```scala
  def m2sPipe(): Apb3 ={
    val that = Apb3(config)
    val fire = that.PENABLE && that.PREADY
    that.PADDR    := RegNext(this.PADDR)
    that.PWRITE   := RegNext(this.PWRITE)
    that.PWDATA   := RegNext(this.PWDATA)
    that.PSEL     := RegNext(this.PSEL.andMask(!fire)) init(False)
    that.PENABLE  := RegNext(this.PENABLE && !fire) init(False)
    this.PRDATA   := that.PRDATA
    this.PREADY   := that.PREADY && that.PENABLE
    if(PSLVERROR != null) { this.PSLVERROR := that.PSLVERROR }
    that
  }
  ```
  
  Note, this.PREADY ? B(0) | this.PSEL is equivalent to this.PSEL.andMask(!this.PREADY)

**Comment by distributed** [comment]
(Note I fixed an issue with my TB and I rebase-i-d it into the original commit)

Using your code (with the small change of using `.init(0)` instead of `.init(False)` for `that.PSEL`), the test fails, see the following trace for reference:

image

This has both oscillating `this.PREADY` and `that.PENABLE`, apart from the showstopper of wrongly suppressing `that.PSEL` by a clock cycle.

How would you like to proceed? From visual inspection of the above trace I'm pretty convinced that the new proposed design isn't working as intended.

**Comment by Dolu1990** [comment]
Ahhh i understand now. I assumed that penable should always be low after a given transaction. but looking at the APB3 spec, it isn't the case.

So, i did checkout your testbench, and got the following version to pass your tests : 

```scala
    val that = Apb3(config)
    that.PADDR    := RegNext(this.PADDR)
    that.PWRITE   := RegNext(this.PWRITE)
    that.PWDATA   := RegNext(this.PWDATA)
    that.PSEL     := RegNext(this.PSEL.andMask(!this.PREADY)).init(0)
    that.PENABLE  := RegNext(this.PENABLE)
    this.PRDATA   := that.PRDATA
    this.PREADY   := that.PREADY && that.PENABLE && that.PSEL.orR
    if(PSLVERROR != null) { this.PSLVERROR := that.PSLVERROR }
    that
```

Would it be good for you ?
If yes, feel free to push it ^^

distributed:
I also think that `PENABLE` doesn't have to go low outside of transactions. I find the diagrams in the spec to be supoptimal as they seem to always show the enable signal going low even when it's X. The State Machine diagrams say nothing of it, so I do expect X.

I think it looks good to me. I pushed your version!

From my side we are good to go.

Dolu1990:
Thanks :D

PR #1618, opened 2024-12-10
Fix simulation failure with Verilator v5.x on Windows

### LINKED ISSUE #1572: Failed to run simulation using verilator on Windows
[The doc](https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Simulation/install/Verilator.html#windows) says:
> Add `bin` and `usr\bin` of MSYS2 into your windows PATH (ie : `C:\msys64\usr\bin;C:\msys64\mingw64\bin`)

I followed this instruction and got the following error when executing `verilator_bin.exe`:
```powershell
Program 'verilator_bin.exe' failed to run: The specified executable is not a valid application for this OS platform.
```

However, Verilator works fine in the MSYS2 shell.
It turns out that `C:\msys64\mingw64\bin\verilator_bin.exe` is some kinds of symbolic link with a size of around 1 KB, and cannot be properly recognized by the native Windows system.
The actual executable is located at `C:\msys64\mingw64\share\verilator\bin\verilator_bin.exe`, and it runs correctly in PowerShell.

We need to modify the PATH as follows to allow Spinal to locate Verilator successfully:
```
C:\msys64\usr\bin;C:\msys64\mingw64\share\verilator\bin;C:\msys64\mingw64\bin
```
Note that `C:\msys64\mingw64\share\verilator\bin` is placed first to ensure the correct executable is found.

However, during compilation, a new error occurred:
```powershell
%Error: Cannot find verilated_std.sv containing built-in std:: definitions: /mingw64/share/verilator\include\verilated_std.sv
%Error: This may be because there's no search path specified with -I<dir>.
        ... Looked in:
             /mingw64/share/verilator\include\verilated_std.sv
             /mingw64/share/verilator\include\verilated_std.sv.v
             /mingw64/share/verilator\include\verilated_std.sv.sv
             verilator\/mingw64/share/verilator\include\verilated_std.sv
             verilator\/mingw64/share/verilator\include\verilated_std.sv.v
             verilator\/mingw64/share/verilator\include\verilated_std.sv.sv
%Error: Exiting due to 2 error(s)
Exception in thread "main" java.lang.AssertionError: assertion failed: Verilator invocation failed
```
The file is present at `C:\msys64\mingw64\share\verilator\include\verilated_std.sv`.
The error seems to be related to the path handling of the Verilator compiled for MSYS2 when running in a native environment.

---
Has anyone successfully run a Verilator simulation on Windows?

du33169:
Closes #1572

Simulation with verilator v5.x on Windows will fail with the following error:

```powershell
Program 'verilator_bin.exe' failed to run: The specified executable is not a valid application for this OS platform.
```

It is caused by invoking `verilator_bin.exe` directly to get its version:
https://github.com/SpinalHDL/SpinalHDL/blob/fb2a8df445631b3da65e84f7ee254d0c4b4c4c67/sim/src/main/scala/spinal/sim/VerilatorBackend.scala#L576

However, since verilator v5.x, the `verilator_bin.exe` file in `/mingw64/bin` installed with the MSYS2 package is actually a perl script. It runs as a normal perl script in the Unix-like MSYS2 shell, but will fail if run directly on Windows.

So it might be better to specify the shell, just like how spinalHDL runs `verilatorScript.sh` later:

https://github.com/SpinalHDL/SpinalHDL/blob/fb2a8df445631b3da65e84f7ee254d0c4b4c4c67/sim/src/main/scala/spinal/sim/VerilatorBackend.scala#L688-L691

This way we can also use a single `val verilatorBinFilename = "verilator"` for both Windows and Linux. 

# Impact on code generation

No impact, ideally.

# Checklist

- [ ] Unit tests were added

No changes to unit tests or API.

---

# Additional Contents

It works for simple cases on my machine. But there are still two problems:

- Even without this patch, some tests will fail on Windows for the dev branch, and I failed to build tag v1.10.2. Is there a stable branch to start with?
- The task hangs strangely after the simulation run is done. But debugging and testing can finish successfully.


Readon:
So, the assumption is that sh.exe is installed in the environment.
It would be best to test this on various cases. Windows with MSYS2 setting, MSYS2, Linux, or even MacOS.

du33169:
This PR is not the first to introduce sh.exe, it has been used by spinalHDL to run verilatorScript.sh for a long time. If it worked before, it should not be a problem here.

Dolu1990:
> This PR is not the first to introduce sh.exe, it has been used by spinalHDL to run verilatorScript.sh for a long time. If it worked before, it should not be a problem here.

Ahhh right sh.exe is alread used when generating the cache, should be all good then.

All good for merge @Readon ?

Readon:
Unfortunately, I could not make it work on MSYS2 environment.
It reports 
```
Can't exec "$MSYS2_ROOT$/share/verilator//verilator_bin": No such file or directory at /mingw64/bin/verilator line 188.er / Test / testOnly 90s
```
However, in dev branch it works.

Readon:
> Unfortunately, I could not make it work on MSYS2 environment. It reports
> 
> ```
> Can't exec "$MSYS2_ROOT$/share/verilator//verilator_bin": No such file or directory at /mingw64/bin/verilator line 188.er / Test / testOnly 90s
> ```
> 
> However, in dev branch it works.

So I find that when this PR is applied, then VERILATOR_ROOT should not be set.
After VERILATOR_ROOT is unset, the verilator in 4.228 and 5.024 works.

du33169:
Yes, in this way the original documentation still applies. Forget about any additional operations from the issue.

The only problem is, the simulation task won't exit on my machine. Is that reproducible on yours? This only happens for Verilator v5.x.

Readon:
In some cases, it would not stop too. That's why I always recommand do not use the version 5.

PR #1627, opened 2024-12-16
Add another auxiliary constructor for StreamDemux

facebreeze:

Add one more method `two` for StreamDemux, which could demux input stream by another input stream. 

# Impact on code generation
None

# Checklist


Dolu1990:
Thanks ^^

PR #1628, opened 2024-12-18
remove unnecessary 'return' of some methods

facebreeze:

Refine Stream.scala
# Impact on code generation
None

# Checklist

None


Dolu1990:
Thanks ^^

PR #1631, opened 2024-12-23
`Interface`-related SystemVerilog Code Emission Enhancements

fl4shk:

SystemVerilog `interface`s do not strictly have to have `modport`s. They can be much like a SpinalHDL `Bundle`s (without the implementation detail where `Bundle`s are emitted as flattened-out signals in the SystemVerilog code).

# Impact on code generation

Only SystemVerilog code generation was changed in this PR. SystemVerilog `interface`s can be emitted, using the changes this of pull request, to not have `modport`s. Emitting `modport`s  can be prevented from being emitted with a config option, or per-`interface`. I still need to test some corner cases where

I made it so that SystemVerilog `struct`s can be chosen to be emitted if the `Interface` has no generics (though admittedly there *is* a way to have a `struct` with generics in SystemVerilog, but it's not as well supported as I'd prefer (in various SV implementations, though it's supported in Verilator and `sv2v`)).

For SV `struct`s as ports, I also need to make it so that it's checked whether all fields of the `struct` are the same direction. That is partially implemented, but only in the implementation of `modport`s

I also made it so that it's possible for a `Vec[Interface]`  that is inside of another `Interface` to be implemented as a SystemVerilog array of `interface`s or `struct`s if the `emitInterface(...)` output is entirely the same text for each member of the `Vec`. Since Verilator (and probably other implementations of SystemVerilog) don't support 2D (or higher dimensions) of arrays of SV `interface`s, this is only done for the lowest `Vec`s in a `Vec[Vec[Interface]]`.

```scala
case class MyOuterInterface() extends Interface {
  val myVec = Vec.fill(3)(Vec.fill(2)(MyInterface()))
}
```

becomes
```verilog
interface MyOuterInterface ();
  MyInterface myVec_0[2]();
  MyInterface myVec_1[2]();
  MyInterface myVec_2[2]();
endinterface
```

For what I've implemented so far, this translation is **not** done inside of a `Component`/SystemVerilog `module`, only inside of other `Interface`. 

# Checklist
NOTE: I still need to develop the parts of this checklist. The API changes I've implemented are not much.

- [ ] Unit tests were added


fl4shk:
more debugging needed, as I just found a bug for `notSVIF()`.

Dolu1990:
Hi ^^ 

Overall, i'm not the dev/maintainer of the Interface feature. @yportne13 is :)

A few things about the Interface feature (in general), it should try to disrupt as little as possible the non-Interface code. To keep things as cleanly separated as possible.

fl4shk:
> Hi ^^
> 
> Overall, i'm not the dev/maintainer of the Interface feature. @yportne13 is :)
> 
> A few things about the Interface feature (in general), it should try to disrupt as little as possible the non-Interface code. To keep things as cleanly separated as possible.

So I've actually primarily *not* modified the non-`Interface` code already! I did have to change `Data` `trait`, but only a little, and nothing intended to be API-facing.

fl4shk:
> > Hi ^^
> > Overall, i'm not the dev/maintainer of the Interface feature. @yportne13 is :)
> > A few things about the Interface feature (in general), it should try to disrupt as little as possible the non-Interface code. To keep things as cleanly separated as possible.
> 
> So I've actually primarily _not_ modified the non-`Interface` code already! I did have to change `Data` `trait`, but only a little, and nothing intended to be API-facing.

I also changed `NamingScope` a little, but again nothing API-facing as far as I know.

PR #1632, opened 2024-12-24
[RegIf] Fix the "BusInterface" and add a new function to change the default error state

HaroldZ32:
#1636 

**BugFix**:

The implementation of "BusInterface" associated with "AxiLite4BusInterface" is incomplete.

https://github.com/SpinalHDL/SpinalHDL/blob/04bb74bc81a540798621e9fbf4e56fe512f0a015/lib/src/main/scala/spinal/lib/bus/regif/BusInterface.scala#L41

**Add Function**:

The default error bits for accessing reserved addresses can be modified now.

You can use this function to change it.

https://github.com/HaroldZ32/SpinalHDL/blob/c483e9d508902ec92c9852677a68e37d9aceabf2/lib/src/main/scala/spinal/lib/bus/regif/BusIf.scala#L80

# Impact on code generation

- BusInterface can now generate a AxiLite4Interface with SecFireWall.
- If the accessDefaultError has been set, the error signal when the master hit the reserved address would be set.

# Checklist

- [x] Unit tests were added
- [ ] API changes are or will be documented

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/regif/BusIf.scala
+++ b/lib/src/main/scala/spinal/lib/bus/regif/BusIf.scala
@@ -2,19 +2,18 @@ package spinal.lib.bus.regif
 
 import spinal.core._
 import spinal.lib._
-import spinal.lib.bus.localbus.{MemBus, MinBus}
 import spinal.lib.bus.misc.SizeMapping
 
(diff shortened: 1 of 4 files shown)


jijingg:
thanks  <3  @HaroldZ32

PR #1641, opened 2025-01-02
Add convenience properties and overloads to CounterUpDown

jdavidberger:
Split from #1637 

- Adds underflow check properties to CounterUpDown in the same style of the overflow ones.
- Adds `apply` overload to allow for slightly less verbose instantiation when handleOverflow is false. 

# Impact on code generation

No impact on code generation

# Checklist

It seemed small enough to not need unit tests but I can add if wanted.


Dolu1990:
Thanks ^^

PR #1646, opened 2025-01-07
Improved scala doc

mrcmry:

A few weeks ago, I was newcomer to HDL, Spinal and Scala with no prior verilog of VHDL knowledge (but with a electronic and functional programming background).

This PR is an improvement of the scaladoc that I missed when I started to learn SpinalHDL. I added scaladoc for the operator and the basic constructs like `mux`, `when`, etc. with a mention of SpinalHDL. The reason is when you start both Scala and SpinalHDL it's difficult to tell them apart, for example is `===` as Scala construct, are UInt a Scala type. Of course with an IDE you can jump to the class file and check its location, but you are then distracted and it's difficult to concentrate on the code meaning. With the IDE mouse hover text it's much easier (at is on vscode).

1b9c154ba119cc7e is also a correction of typo in a text error. Perhaps it can break some workflows that parse the error. I don't mind if it's not merged, it's not that important.

I also noted there is a `@deprecated` that is not active due to the use of only `//` here:
https://github.com/SpinalHDL/SpinalHDL/blob/2c93394cedce78e885df5cf2b7c6ebebc702ce03/core/src/main/scala/spinal/core/when.scala#L117

I don't know if I should be reactivated or removed.

# Impact on code generation

No impact on code generation.

### DIFF
--- a/core/src/main/scala/spinal/core/Bool.scala
+++ b/core/src/main/scala/spinal/core/Bool.scala
@@ -24,20 +24,20 @@ import spinal.core.internals.Operator.Formal
 import spinal.core.internals._
 import spinal.idslplugin.Location
 
-/** Bool factory used for instance by the IODirection to create a in/out Bool() */
+/** `Bool` factory used for instance by the IODirection to create a in/out Bool() */
 trait BoolFactory {
   @deprecated("Use `Bool()` (with braces) instead")
   def Bool: Bool = Bool()
 
-  /** Create a new Bool */
+  /** Create a new `Bool` */
   def Bool(u: Unit = ()): Bool = new Bool
 
-  /** Create a new Bool with a value */
+  /** Create a new `Bool` with a value */
   def Bool(value: Boolean)(implicit loc: Location): Bool = BoolLiteral(value, Bool().setAsTypeNode())
 }
 
 /**
-  * The Bool type corresponds to a boolean value (True or False)
+  * The `Bool` type corresponds to a SpinalHDL boolean value (`True` or `False`)
   *
   * @example {{{
   *     val myBool = Bool()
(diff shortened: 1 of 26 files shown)


core/src/main/scala/spinal/core/Bool.scala
```diff
@@ -58,15 +58,15 @@ class Bool extends BaseType with DataPrimitives[Bool]  with BaseTypePrimitives[B
   private[spinal] override def _data: Bool = this
 
   /**
-    * Logical AND
+    * SpinalHDL logical AND
```
Dolu1990, line 61: Hmm, to be more explicit, i think it should be : 
Hardware logical AND
(same for the ones bellow)

core/src/main/scala/spinal/core/core.scala
```diff
@@ -262,13 +262,13 @@ package object core extends BaseTypeFactory with BaseTypeCast {
   }
 
 
-  /**
-    * True / False definition
-    */
+  /** SpinalHDL true value
```
Dolu1990, line 265: "Hardware Bool literal" instead ?

mrcmry:
I replaced "SpinalHDL" by "hardware" in the scaladoc.

In some place, I used "data" instead of "signal", because it's more intuitive. I understand that `Data` is the parent type of all SpinalHDL "signals", so in a context of very deep internal doc of SpinalHDL it make sense to use `Data` (probably better that Data or data visually) but for basic stuf "signal" make more sense. This is also widely used in SpinalDoc-RTD (420 occurences vs 97 of "wire" and very few usage of that should be `Data`).

Also, what Do you think of remplacing "wire" by "signal" in SpinalDoc-RTD ? Of course not when we speak directly of the `wire` of verilog. I find it a bit confusing to use both. And signal is a better name, in hardware a "wire" is a single voltage, the proper name would be a "cable" :).

mrcmry:
> Good for me ^^ Maybe in another PR ? (so i merge this one as it is now)

Yes, please merge this one. the wire -> signal is mostly in SpinalDoc-RTD anyway.

Dolu1990:
Thanks ^^

PR #1658, opened 2025-01-31
fix `AFix` bitwise op alignment bug

thajohns:
No associated issue

It seems like the bitwise ops for `AFix`es have never been tested. The intent was clear from reading the code, but it didn't work.

# Impact on code generation

# Checklist

- [x] Unit tests were added
- [x] API changes are or will be documented (none needed)


Dolu1990:
Thanks :D

PR #1659, opened 2025-02-06
Add core features and classes to allow development of formal testing code

jdavidberger:
Split from #1637 

This adds the basic traits and classes required around that formal testing PR.

The expected usage is for components and areas to implement the various overridable functions to fully define their valid internal state so that induction based solvers can prove the assertions.

Additionally, this PR introduces traits that allow for adding logic to bundles about whether their internal state is valid. This enables `ComponentWithFormalAsserts` to have reasonable default behaviors which minimizes the work required to prove out a custom component. 

The long term goals for this functionality is that Spinal lib components and bundles that are commonly used are covered by these traits in such a way that library users can reuse the formal logic and easily validate their own designs. Additionally, since the internal components of the library can be `assume`d correct, this should enable fast run-times for even complicated user components.

# Impact on code generation

Code generation is not directly impacted

# Checklist

- [x] Unit tests were added


lib/src/main/scala/spinal/lib/formal/FormalMasterSlave.scala
```diff
@@ -0,0 +1,50 @@
...
+   * @param that The object to assert equivalence with.
+   */
+  def formalAssertEquivalence(that : Self): Unit = {}
+}
+
+/**
+ * For bundles which implement IMasterSlave, this can be used to add logic which informs on if the master or slave of
+ * that bundle are using it correctly.
+ */
+trait FormalMasterSlave extends IMasterSlave with FormalBundle {
```
jdavidberger, line 39: https://github.com/SpinalHDL/SpinalHDL/pull/1637#discussion_r1944779654

> If you want the FormalMasterSlave to be a traits which is required to be used with IMaterSlave, then is that possible to use Self-Type Annotation? So that it is not necessary to change the design related code.
> 
> ```scala
> trait FormalMasterSlave extends FormalBundle { self: IMasterSlave =>
>   ......
> }
> ```
> 
> Moreover, what is the purpose of `formallsValid` and `formalAssertEquivalence` methods? It's a bit hard to me.

I added comments for what those methods are for. Let me know if that makes sense.

I have it `extends IMasterSlave` there so that [this](https://github.com/SpinalHDL/SpinalHDL/pull/1659/files#diff-fd42c1518d1f6881227480ea7870b3795dfb8e3c9507c20aa8847a6fa56acc42R42) can use `isMasterInterface` and related functionality. Otherwise it's unclear what to do if something has FormalMasterSlave logic but doesn't have the appropriate semantics around it.

jdavidberger:
Apologies if this is not what you meant to do. I've pulled from that branch only the things required for the core components and the related test functions around it. I also cleaned up the imports from the other branch to make it a much smaller PR. 

My rough plan here is to get this PR to an acceptable state, then, in order:
- Stream related assertions / logic
- PMB related assertions / logic
- WB related assertions / logic

Mostly because the stream ones are used for basically everything else. I'll reference feedback from the other PR when I get to those PRs.

lib/src/main/scala/spinal/lib/formal/FormalMasterSlave.scala
Readon: You can use Self-Type Annotation I describe above, so that isMasterInterface is still valid inside FormalMasterSlave.

lib/src/main/scala/spinal/lib/formal/HasFormalAsserts.scala
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  /**
+   * Helper function; uses the implicit useAssumes variable to either emit an assert or assume
+   *
+   * @param cond       Condition to assert or assume
+   * @param msg        Some backends with asserts will print out a message when an assert fails. Ignored for assumes
+   * @param useAssumes True to emit an assume
+   */
+  def assertOrAssume(cond: Bool, msg: Any*)(implicit loc: Location, useAssumes: Boolean): Unit =
```
Readon, line 137: What does this method designed for? I think the assert, assume switching is not that useful now. Would you please refer the comments, and reply to that?

lib/src/main/scala/spinal/lib/formal/FormalMasterSlave.scala
jdavidberger: I tried that and it didn't resolve those functions directly. I played around with it and found that I could add a function like this: 

```
def asIMasterSlave = self
```

to resolve it through. That is the best way I could find to do it. I'm not super well versed in scalas type system, so if there is a better way to do it please let me know

lib/src/main/scala/spinal/lib/formal/HasFormalAsserts.scala
jdavidberger: I missed this comment from before since I was just pulling comments from the core files.
 https://github.com/SpinalHDL/SpinalHDL/pull/1637#discussion_r1944363638

> according to previous disscusion the assume is not usable in hierachical verification, what does this assertOrAssume used for?

I added relevant comments to the function doc, but I wanted to go into the details about hierarchical verification. 

`assume` _is_ usable and useful in hierarchical verification and part of the benefits to this PR I think is that it handles a lot of these details transparently. [The points made here are true](https://zipcpu.com/formal/2018/12/18/skynet.html), but this is resolved with this:

```
 when(formalValidInputs) {
  formalAssertsChildren()(useAssumes = true)
  formalChecks()(useAssumes = true)
}
```

The `assume`s are predicated on the inputs being true. We aren't just swapping asserts and assumes. 

For something like the toy example from the zipcpu article (which is the basis of the `FormalHasFormalTester` test), there isn't a major benefit to this -- the internal assume is identical to the `formalValidInput` so no effort is saved. 

But for something like a FIFO, proving the inputs to the FIFO are valid is a much smaller task than proving the internals of the FIFO. This massively cuts down on the run-time required to solve a component that uses the FIFO. For a component with many such components, the time savings are exponential. 

> BTW, it seems that formalIsValid do what formalAsserts do?

It is, in general, not enough to just assert that a bundle is valid -- we need to have a signal which tells us if it is valid or not. This is what feeds into `formalValidInputs`, and ultimately predicates the assumption blocks. If we just had:

```
formalAssertValidInputs()
formalAssertsChildren()(useAssumes = true)
formalChecks()(useAssumes = true)
```

this would _not_ work because the assumptions would take precedence over the assertion.

```
 assert(formalValidInputs)
 when(formalValidInputs) {
  assume(someProperty)
}
```

is logically equivalent to

```
assert(formalValidInputs)
assume(!formalValidInputs || someProperty)
```

and so the solver must show the first assertion to use the assumption that `someProperty` is true. Alternatively, The problematic logic the ZipCPU blog points at is:

```
// Bad logic; doesn't assert anything if someProperty implies formalValidInputs!
assert(formalValidInputs)
assume(someProperty)
```

This understanding motivates the existence of `formalValidInputs`, `formalIsValid`, `formalIsConsumerValid` and `formalIsProducerValid` as things that return a Bool rather than simply assert the condition.

Readon:
Sorry for the late reply, I am in a tight bandwidth.
On other hand, is that I could not understand your description and code as:
1. What is the scenario assume is useful, would you give me an example? You code and function name some time seems misleading.
2. Why it must be a wrapped assumeOrAssert function but not a assume or assert directly. To me if I want to assume inputs I can just call assume, if it is used as a children, it could be automatically called in formalAssertChildren and call formalAssertInputs directly. An implicit function with status make the difficulties of understanding on the situation to developers, especially auto run things.
3. I strongly enourage you to retide the name of the function, such as formalAssertTags, does it designed for assert tags? 
4. What does the formalIsValid actutally mean? Valid is a noun which could leading to misunderstanding of valid. Do you mean if the IMasterSlave bundle is verified by the design? 
5. The name HasFormalAsserts seems could not cover the functions. 
6. formalConfigureForTest this config seems weird too.

I know I have asked too much. However, this PR is more like to introduce a framework but a library. There are requirements on using this framework, I am also have concerns on this type of requirements. Some times it's best to let the user understand those as quick as possible.
We should think it though clearly, so that we can build more on it.

jdavidberger:
> What is the scenario assume is useful, would you give me an example? You code and function name some time seems misleading.

Assume is useful as a mechanism to run formal testing faster. If formal testing ran instantly no matter how large of a design you had, there would be no point. But since it can take a while for large designs, it lets you dramatically limit the space the formal tester has to explore. 

If you have a design which has a StreamFIFO in it, and you want to use inductive proofs to demonstrate that it works, you do not want the solvers to spend time trying to break the StreamFIFO. But those formal properties are still necessary in general as assumptions so the inductive solver is constrained to something sane.

> However, this PR is more like to introduce a framework but a library

I've considered having this as just a library; but it isn't ideal -- the `spinal.lib` constructs need to be annotated one way or another with formal properties if you want to formally prove something that uses them. 

> What does the formalIsValid actutally mean? Valid is a noun which could leading to misunderstanding of valid. Do you mean if the IMasterSlave bundle is verified by the design?

`formalIsValid` for those traits is meant to return true if and only if the bundles state follows whatever it's contract is. So for stream, it codifies these semantics: https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Libraries/stream.html#semantics. The `formalIsConsumerValid` / `formalIsProducerValid` functions are there to determine whether the consumer or producer are violating that semantic contract. 

For instance, on the producer side of a `PipelinedMemoryBus`, it must follow the stream rules for `cmd`. On the consumer side, it must never trigger `rsp.fire` more than it saw `cmd.fire && !cmd.write`. 

These return Bool's rather than `assert` directly for a few reasons. First, it lets as `assume` the inputs into a component we are testing are valid, and then `anyseq` tests all valid cases. Secondly, it lets us predicate using assumptions for a given component based on it's inputs being driven in a valid state. In general, we should have no expectations for a component if it's inputs are invalid. 

In terms of the naming -- `valid` is overloaded for sure. For me, the `formal` prefix to that member makes it distinct from the `valid` as used in Stream and Flow but maybe this isn't clear. `formalIsSemanticallyValid` seems too long. `formalIsCorrect`? `formalIsProper`? 

So from this feedback, I think it makes sense to make the following changes:
- Instead of the `assertOrAssume`'s, rewrite it somewhat to make it `addFormalProperty(cond : Boolean, msg: Any*)`.  `formalChecks` becomes `formalProperties`. The implicit parameter is dropped.  
- The calls for `formalAsserts` or `formalAssumes` largely remain the same; they invoke the collected formal properties as either `assume` calls or `assert` calls.
- Rename HasFormalAsserts -> HasFormalProperties
- Rename formalConfigureForTest -> formalConfigureForFormalTesting
- Swap `formalIsValid` and associated functions to use Correct instead of Valid. Open to other ideas; I don't have strong feelings about any one given name here.

Readon:
Ok, If I understand right, the thing could be describe as below.
These core is to enable Component's IO assert or assume swappable by the situation that if the target signal is IO of DUT under verification or belongs to submodule. 
   - In a DUT component, the assert and assume can be remained as it is originally.
   - In submodule, to assert should be changed to assume, and assume should be changed to assert, due to the concentrated input and output direction changed. This could provide enough constraints to speed up.
   This might be done during the FormalPhase, by checking if the Component's isFormalTester is true.
   If the target IO signal belongs to the DUT Component keep it's asserts and assumes, if not swapped them.

In this case, user just define the asserts and assumes once while verify submodule standalone, and if it is embedded into other component than it will swap automatically.

Readon:
In the second step, we could solve how to build a hierarchy that can call all check IO functions of submodules.
I think it should be another PR to concentrate on the target.
The overall function, maybe named as formalCheckChildren, could be added into the Component, I suppose.

Readon:
> Ok, If I understand right, the thing could be describe as below. These core is to enable Component's IO assert or assume swappable by the situation that if the target signal is IO of DUT under verification or belongs to submodule.
> 
> * In a DUT component, the assert and assume can be remained as it is originally.
> * In submodule, to assert should be changed to assume, and assume should be changed to assert, due to the concentrated input and output direction changed. This could provide enough constraints to speed up.
>   This might be done during the FormalPhase, by checking if the Component's isFormalTester is true.
>   If the target IO signal belongs to the DUT Component keep it's asserts and assumes, if not swapped them.
> 
> In this case, user just define the asserts and assumes once while verify submodule standalone, and if it is embedded into other component than it will swap automatically.
> 
> This might be much simpler than adding a framework to the lib.

This method is on a different angle of what you do, you have added assertOrAssume function in application layer, and utilizing a useAssume variable and mechanism to propagate when and where to swap.
This method is just to check the assumed and asserted target signals, and act depending on what is their property.
The limitation might be wrongly swap, especially change those `assert` to `assume`, more specifically rules should be thought carefully.

Readon:
I have rethink on this topic and go through the ZipCPU's blog again, I believe that just swapping assert and assume would take a bad effect on correctness of formal verification, which is non-trivial for this function. 

I think we should stand on method 3 that they give out. "3.  Replace the assumptions with assertions, but leave the original assertions untouched.  Hence we would have only  `assert`  properties within the submodule design.". 

So the thing would be just change `assume` inside the child component to be `assert` only.

jdavidberger:
> I have rethink on this topic and go through the ZipCPU's blog again, I believe that just swapping assert and assume would take a bad effect on correctness of formal verification, which is non-trivial for this function.

> I think we should stand on method 3 that they give out. "3. Replace the assumptions with assertions, but leave the original assertions untouched. Hence we would have only assert properties within the submodule design.".

I think a key point which must be understood for this effort to make sense is that we completely avoid the issue in the zipcpu blog by only `assume`ing formal properties when their inputs are valid. ie:

```
when(formalValidInputs) {
    assumeFormalProperties() // Psuedo code
}
```

The formal properties defined by this class are NOT saying these properties must always hold -- they are saying IF the inputs are valid as we have defined, then these properties hold.

Readon:
I just looked over the code generated by `test("Check Parent")` inside `FormalHasFormalTester`。
``` verilog
  initial begin
    assume(reset); // FormalHasFormalAssertsTester.scala:L71
  end

  assign dut_formalValidInputs_validInputs = (&(dut__zz_dut_formalValidInputs_validInputs != 32'hdeadbeef));
  assign dut_io_contract_input = $anyseq(32);
  always @(posedge clk) begin
    if(reset) begin
    end else begin
      assert(dut_formalValidInputs_validInputs); // HasFormalAsserts.scala:L65      ===========> this one is right.
      if(dut_formalValidInputs_validInputs) begin
        assume((dut__zz_1 == 16'h0)); // FormalHasFormalAssertsTester.scala:L37
        assume((dut__zz_2 == 1'b0)); // ComponentWithFormalAsserts.scala:L56
      end
      assume(dut_formalValidInputs_validInputs); // HasFormalAsserts.scala:L65    ========>this one keeps upper one never failed.
    end
  end
```

Those two lines are make sure it pass. if you comment out the second line, it failed.

Readon:
> I think a key point which must be understood for this effort to make sense is that we completely avoid the issue in the zipcpu blog by only `assume`ing formal properties when their inputs are valid. ie:

The key point is that the `assume` of output + logic inside the submodule would take the cases that make the `assert` fail from the solving spaces.

That's why they proposed method 1 "Remove all of the logic within Skynet, leaving only the formal properties behind.Because there’s no logic between the input assumptions, now turned into assertions, and the output assertions, now turned into assumptions, the output could be anything depending upon the choice of the solver."

jdavidberger:
I've renamed / refactored for clarity given the earlier points. I commented the code heavily so it can be discussed inline.

> That's why they proposed method 1 "Remove all of the logic within Skynet, leaving only the formal properties behind.Because there’s no logic between the input assumptions, now turned into assertions, and the output assertions, now turned into assumptions, the output could be anything depending upon the choice of the solver."

In general I think this would be fantastic capability to add -- a mechanism to remove all internal logic to a component and just have it driven by `assume` and `anyseq`. But it is a much more stringent test than keeping the child components logic and just assuming it's properties. 

For instance, lets say you have an internal FIFO, which has enough properties defined to be provably valid via an inductive method. And the component under test also has enough properties defined to pass the inductive solver too. But if the component you are testing somehow has a reliance on the latency of the FIFO being less than 2, and the fifo doesn't have that as a defined property, swapping out the FIFO for a blackbox will fail.

The benefits to a blackbox test like this though is it would, in some cases, offer another speed up opportunity -- it's entirely possible that it takes longer to simulate complicated design internals than it is to just enumerate the assumptions. 

I'm interested in getting something like this working eventually -- I think you just have to remove all assignments inside a component and swap them with anyseq? Yosys would see whatever is left as not being connected and remove them? -- but want to focus on the core functionality for now.

Readon:
> I've renamed / refactored for clarity given the earlier points. I commented the code heavily so it can be discussed inline.
> 
> > That's why they proposed method 1 "Remove all of the logic within Skynet, leaving only the formal properties behind.Because there’s no logic between the input assumptions, now turned into assertions, and the output assertions, now turned into assumptions, the output could be anything depending upon the choice of the solver."
> 
> In general I think this would be fantastic capability to add -- a mechanism to remove all internal logic to a component and just have it driven by `assume` and `anyseq`. But it is a much more stringent test than keeping the child components logic and just assuming it's properties.

Unfortunately, it is hard to write the properties with good equivalency. There should be an equivalency check first.

jdavidberger:
> Unfortunately, it is hard to write the properties with good equivalency. There should be an equivalency check first.

I'm not sure what you are saying there should be equivalency between for this -- but I want to try focusing on just what is in the PR for now since anything in that vein is probably much later.

Readon:
> > Unfortunately, it is hard to write the properties with good equivalency. There should be an equivalency check first.
> 
> I'm not sure what you are saying there should be equivalency between for this -- but I want to try focusing on just what is in the PR for now since anything in that vein is probably much later.

For this PR, the base function is to swap `assume` and `assert`, to me
1. swap `assume` to `assert` only is preferable.
2. The way to implement the swap is best on phase process, the framework seems too overloaded.
3. It would be better to split these into add swap capability and propagate the swap operation. Just a suggestion.

jdavidberger:
> For this PR, the base function is to swap assume and assert, to me

That isn't what this PR is at all though. 

I've been trying to contribute some version of this for two months and it sort of seems like it's going nowhere. I keep making changes as suggested and trying to understand concerns but it seems like we aren't on the same page or something. It's frustrating because I put a lot of work into making the interface as coherent, clearly named, and documented as I could and the feedback doesn't engage with that at all. 

I think the only way forward here is to put it into a separate repo and add a facilities to attach the formal properties to components external to them. In places were key state isn't accessible externally, I may open PRs to make those accessible in a way that doesn't impact anything else if that is acceptable.

Readon:
> were key state isn't accessible exter

I am so sorry for this, sometimes I could not understand you well, and not express the thing clearly.

PR #1662, opened 2025-02-14
add StreamMux.merge to merge multiple streams into one.

facebreeze:

When multiple streams become valid in turn,  it can be directly merged into a single one without an arbiter.

# Impact on code generation

None
# Checklist

- [ ] Unit tests were added
No

### DIFF
--- a/core/src/main/scala/spinal/core/UInt.scala
+++ b/core/src/main/scala/spinal/core/UInt.scala
@@ -87,6 +87,7 @@ class UInt extends BitVector with Num[UInt] with MinMaxProvider with DataPrimiti
   override def <=(right: UInt): Bool = wrapLogicalOperator(right, new Operator.UInt.SmallerOrEqual)
   override def >=(right: UInt): Bool = right <= this
   override def >>(that: Int): UInt   = wrapConstantOperator(new Operator.UInt.ShiftRightByInt(that))
+  def *<(that: Int): UInt            = (this * U(that)).resize(log2Up(this.maxValue * that) bits)
   override def <<(that: Int): UInt   = wrapConstantOperator(new Operator.UInt.ShiftLeftByInt(that))
   override def +^(right: UInt): UInt = this.expand + right.expand
   override def -^(right: UInt): UInt = this.expand - right.expand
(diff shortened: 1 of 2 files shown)


Dolu1990:
So overall, that function implement an arbitrator between the inputs streams right ?
Have you seen StreamArbiterFactory ?
I'm not saying its API is good. Could probably improved (StreamArbiterFactory).

facebreeze:
Hi, 
     It's not a arbitrator between all input streams. The use case is that only one stream get valid at the same time, therefore, it's no need to do arbitrate. Although an arbitrator can do this well, my concern is that area of an arbitrator get larger then this piece of logic. And if it is, then put this logic into StreamArbiterFactory  is fine.

Dolu1990:
> The use case is that only one stream get valid at the same time, therefore, it's no need to do arbitrate

The thing is that the StreamMux, isn't taking advantage of that assmuption. It will only ready the selected stream instead of passing through the ready.
```scala
  for ((input, index) <- io.inputs.zipWithIndex) {
    input.ready := io.select === index && io.output.ready
  }
```

> def merge

Hmm i'm trying to find a more explicit name for what this function does. merge is not specific enough i think ?

facebreeze:
A more explicit name is fine. And how about put this in StreamArbiter ? Since the **select** of StreamMux does nothing.

Dolu1990:
>  And how about put this in StreamArbiter ? Since the select of StreamMux does nothing.

Ahh sorry i don't understand exactly ?

facebreeze:
I mean we could move this piece of logic in StreamArbiter, and add one more argument *merge* to it ?

Dolu1990:
> I mean we could move this piece of logic into StreamArbiter, and add one more argument merge to it ?

Ahhh, yes, could be done.

likewise:
I have followed this issue but am not quite understanding why it's useful.

A stream merger that assumes that at most one input is valid at a time, is a dangerous block.

The safe approach is an arbiter.

Does it really reduce resources to go from arbiter to a dumber but smaller merger?

Basically for a two-input, you demux data1 if valid1==1 and data2 otherwise in a priority arbiter, which is equally simple to a merger, right?

Dolu1990:
If you can assume that only one input stream will be valid at a given time, then your arbitration do not need to make the inputs[..].ready := functionOf(inputs[..].valid)

It reduce the arbitration dependencies. Also, the selection of the input can be done via a simple OhToUInt instead of a lowerFirst/roundrobin. Which is also quite lighter aswell.

So overal, it has much faster arbitration timings.

PR #1665, opened 2025-02-20
Fix buffer name conflict in spinal.lib.misc.pipeline.S2MLink

JunyiLiu1994:

`S2MLink` assigns the same name to buffers of different payloads, which causes naming conflict, as shown in the following example:

```scala
object S2MLinkBufferNameConflict extends App {
  case class Top() extends Component {
    val n1, n2 = Node()
    val p1, p2 = Payload(Bool())
    val o = out port Bool()

    val a1 = new n1.Area {
      p1 := True
      p2 := True
    }
    val a2 = new n2.Area {
      o := p1 && p2
    }

    n2.ready := True
    val l = S2MLink(n1, n2)
    Builder(List(l))
  }

  SpinalVerilog(Top())
}
```

# Impact on code generation

We can use a name associated with the payload instead.
```scala
    val rData = matches.map(e => RegNextWhen(up(e), up.ready).setCompositeName(up(e), "s2mBuffer"))
```

# Checklist

- [ ] Unit tests were added


Dolu1990:
Nice thanks ^^

PR #1671, opened 2025-03-05
extract onMismatch method in ScoreboardInOrder

etchone:

If we want to change Scoreboard mismatch message currently, we need to override the whole `check` method, and copy the compare logic again.

I wish there could be a better way, so in this commit, the relavant code are copied to a new method.

ps. maybe it's better to leave `simFailure()` outside the new method? Someone may forget to call it when overriding.

# Example usage

```scala
// By this modification, we can use the code below:
val scoreboard = new ScoreboardInOrder[Int] {
  override def onMismatch(ref: Int, dut: Int) = {
    println(s"DUT $dut != REF $ref")
    simFailure()
  }
}

// Currently I have to use this:
val scoreboard = new ScoreboardInOrder[Int] {
  override def check(): Unit ={
    if(ref.nonEmpty && dut.nonEmpty){
      val dutHead = dut.dequeue()
      val refHead = ref.dequeue()
      if(!compare(refHead, dutHead)){
        println(s"DUT $dut != REF $ref")
        simFailure()
      }
      matches += 1
    }
  }
```

# Impact on code generation

The modification will not change the existing code generated.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/sim/Scoreboard.scala
+++ b/lib/src/main/scala/spinal/lib/sim/Scoreboard.scala
@@ -43,17 +43,21 @@ class ScoreboardInOrder[T]() {
 
   def compare(ref : T, dut : T) = !(ref != dut)
 
+  def onMismatch(ref: T, dut: T) = {
+    println("Transaction mismatch :")
+    println("REF :")
+    println(ref)
+    println("DUT :")
+    println(dut)
+    simFailure()
+  }
+
   def check(): Unit ={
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

> ps. maybe it's better to leave simFailure() outside the new method? Someone may forget to call it when overriding.

Hmmm, i think that is fine to keep it in the method, maybe somebody want to keep going.
The PR look good to me.

Why did you closed it ?

PR #1681, opened 2025-03-17
add exclusive arbitration method for StreamArbiterFacotory

facebreeze:
Closes #1662 

When several streams assert only one at the same time , a simple arbitration logic results in smaller area and better timing.

# Impact on code generation

# Checklist

- [ ] Unit tests were added


Dolu1990:
Hi,

"def exclusive" isn't an explicit enough area. why not "assumeOhInput" ?

facebreeze:
Hi，
     renamed to "assumeOhInput" 😄

Dolu1990:
Thanks ^^

PR #1682, opened 2025-03-17
Merge various changes

Dolu1990:


core/src/main/scala/spinal/core/MemBlackBox.scala
azaparov, line 1: rdLatency & rd_dataEn seem to be missing from documentation. It requires to update Verilog implementations for the architecture.

Dolu1990: I pushed a bit more doc about it.

So, overall, you can ignore the values of those two things. because unless you explicilty enable the MemReadBufferPhase, it will always behave like before (1 cycle latency ram, rd_dataEn unused)

PR #1683, opened 2025-03-18
Introduce Flow.discardWhen() and Stream.discardWhen()

Tectu:
As discussed separately, this PR introduces `Flow.discardWhen()` and `Stream.discardWhen()`.

The behavior is exactly the same as `Flow.throwWhen()` and `Stream.throwWhen()` (in fact, these are just wrappers) but the function names are semantically a lot clearer / self-explaining.

This does not affect any existing code whatsoever.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Flow.scala
+++ b/lib/src/main/scala/spinal/lib/Flow.scala
@@ -106,6 +106,20 @@ class Flow[T <: Data](val payloadType: HardType[T]) extends Bundle with IMasterS
     return next
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


Tectu:
I'm new to scala - please check whether the 'this' return is correct.

Dolu1990:
The code is correct.
But the wording "the 'this' return is correct." isn't right ^^
```scala
this throwWhen(cond)
is equivalent to 
this.throwWhen(cond)
```

Dolu1990:
thanks <3

PR #1689, opened 2025-03-24
register Stream.s2mPipe.rData when necessary to save power

facebreeze:

The action that `rData` always registering payload leads to uncessary toggle, which could waste power. To avoid this behavior, `rData` should be registered when necessary (only when s2mPipe == 0) as show below. 
`rData0, rData1, rData2, rData3 and rData5`  won't be used by next pipe.

image

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/.github/workflows/run-tests-2_11.yml
+++ b/.github/workflows/run-tests-2_11.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - dev
+      - streamArb
   pull_request:
     branches:
       - dev
--- a/.github/workflows/run-tests-2_11.yml
+++ b/.github/workflows/run-tests-2_11.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - dev
-      - streamArb
   pull_request:
     branches:
       - dev
--- a/.github/workflows/run-tests-2_13.yml
+++ b/.github/workflows/run-tests-2_13.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - dev
+      - streamArb
   pull_request:
     branches:
       - dev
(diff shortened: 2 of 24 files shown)


facebreeze:
Hi, 
     
     
When run scala 2.13.12 StreamTester,  it fails:

```
raise TestFailure("FAIL %s    %d != %d" % (name,int(a),int(b)))
                                                       cocotb.result.TestFailure: FAIL io_master0_payload_a    98 != 133
```

In order to debug this, I dumped a vcd file and seems that the `dut` works correctly as show below:

Screenshot from 2025-04-03 21-17-07

[StreamTester_verilog.vcd.zip](https://github.com/user-attachments/files/19587938/StreamTester_verilog.vcd.zip)


And finally, from the below debug log I found the issue in  `tester/src/test/python/spinal/StreamTester/StreamTester.py`.

For the `Fifo.push` method, it pushed two payloads into the **queue** while **dut** got only one `@74.5ns`

Screenshot from 2025-04-03 22-13-03

And I tried to fix this, but failed, need help.

Dolu1990:
Thanks ^^

Would it be possible to have this as a parameters ?
Thing is, here we save power at the cost of combinatorial path which can impact FMax. And this S2m pipe is specificaly designed to help FMax.
Default being FMax, not area.

facebreeze:
Of course, add one more parameter will give more fine control over it.  

And how about name this parameter to `holdPayload` just like that in `m2sPipe` ?  

The reason is when `holdPayload = true` , it will also try to hold output payload when downstream didn't fire, the default value can be `false` in order to get max frequency.

Dolu1990:
Hmm, hold payload isn't good, the reason is that the s2mPipe will not always behave in that way from a down stream perspective (when the buffer isn't being used)

Maybe here we can just be very explicit about the intend. something like "savePower : Bool = false"

PR #1690, opened 2025-03-25
Fixes for when axi stream isnt a fragment

jdavidberger:
There were a few places related to creating an AXI crossbar that assumes the AXI stream is a fragment stream. This isn't necessarily true all the time. These changes prevent the null pointer exception from happening.


Dolu1990:
<3

PR #1691, opened 2025-03-25
Add scaladoc from the RTD doc and fix spelling

mrcmry:
No change of the code.

# Impact on code generation

None


Dolu1990:
Thanks ^^

PR #1692, opened 2025-03-26
Fix strb of Axi4Master (spinal.lib.bus.amba4.axi.sim)

Nik-Sch:

The `strb` for the `Axi4Master` (sim) was wrong for aligned transfers with `len`>0. I fixed this and added a simple testbench that verifies the `Axi4Master` against the `AxiMemorySim`.

# Impact on code generation
None

# Checklist

- [x] Unit tests were added
- [ ] ~~API changes are or will be documented:~~
  - ~~using Scaladoc comments: `/** */`?~~
  - ~~on [RTD](https://github.com/SpinalHDL/SpinalDoc-RTD)?~~
  - ~~thanks to a [new tracking issue on RTD](https://github.com/SpinalHDL/SpinalDoc-RTD/issues/new?title=Document%20XXX&body=Do%20not%20merge%20until%20SpinalHDL/SpinalHDL%23XXX%20has%20not%20been%20merged)?~~


Dolu1990:
Hi,

(~((BigInt(1) << padBack) - 1)) & ((BigInt(1) << bytePerBeat) - 1)

Does it work for burst where size is smaller than the whole bus width ?
Seems like & ((BigInt(1) << bytePerBeat) - 1) will not work in all cases (last beat not finishing at the start of the data word.

Nik-Sch:
Yes, you are correct, there were some weird bugs...
I added tests for unaligned transfers and fixed the problems. Should I test something else?

lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/Axi4Master.scala
```diff
@@ -274,35 +273,40 @@ case class Axi4Master(axi: Axi4, clockDomain: ClockDomain, name: String = "unnam
     }
     assert(len <= 255, s"max burst in one transaction is 256")
     val bytePerBeat = 1 << size
-    val bytes = (len + 1) * bytePerBeat
+    val actualLen = ((data.length / bytePerBeat.toFloat).ceil.toInt - 1) min len
```
Dolu1990, line 276: I don't understand why there is actualLen ? 
What is the purpose of the len argument then ?

Nik-Sch: When the `len` a larger transaction than the `data.length` (i.e. `(len + 1) * bytePerBeat > data.length`), the axi4master would issue more beats (transfer in axi notion) after the last data transfer. This would lead to zeros (at least in verilator) being written after the actual last byte.

This often happens when using the `write` or `writeCB` function where multiple transfers are initiated until the data is exhausted. I could also move the `actualLen` calculation to `writeCB` but I found it quite unexpected behavior that the master would write (a lot of) zeros (with active strb) after the data I want it to write.

Nik-Sch: Yes, len specifies the number of transfers (beats in this naming here) in one transaction (off by one per axi standard). But the old version of the AXI Master would always issue a full transaction regardless of how many data bytes are provided. 
An example:
- `data.length = 20` (20 bytes)
- `size = 3` -> `bytePerBeat = 8`
- `len = 7` -> 1 transaction is 8 transfers of 8 bytes each, resulting in 64 written bytes.

In the old implementation this would have written the 20 input bytes at the specified address and then additionally 44 bytes `0x00`. I fixed this by using the `len` parameter rather as a `maxLen` and then calculating the `actualLen` to be 2 (3 transfers) where the `strb` in the last transfer is set such that only 4 of the 8 bytes are written.

Dolu1990: The implementation assumes that the data.lenght has some data for each of the beats specified by len.
There should be no "actualLen". But instead maybe some assert to ensure the above property is always true.

Nik-Sch: Ok. I added the asserts and fixed the len calculation in `writeCB` for the last transaction.

lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/Axi4Master.scala
```diff
@@ -293,16 +298,15 @@ case class Axi4Master(axi: Axi4, clockDomain: ClockDomain, name: String = "unnam
...
-            case 0 => (BigInt(1) << (bytePerBeat - padFront)) - 1
-            case `len` => ~((BigInt(1) << padBack) - 1)
-            case _ => BigInt(1) << busConfig.bytePerWord
-          }
+          val fullStrb = (BigInt(1) << bytePerBeat) - 1
+          val strb = (beat match {
+            case 0     => fullStrb << padFront
+            case `len` => fullStrb >> padBack
+            case _     => fullStrb
+          }) & fullStrb
```
Dolu1990, line 306: Overall i think there is still two cases which aren't handled (and were not before the pr either ?)

- Single beat burst with address not being word aligned.
- burst with size smaller than word size

@KireinaHoro
Can you double check ?

Nik-Sch: I added a test for that and it was already working in my branch. This was handled by `padData` in line 284.
Did I understand your points correctly?

lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/Axi4Master.scala
```diff
@@ -249,7 +249,12 @@ case class Axi4Master(axi: Axi4, clockDomain: ClockDomain, name: String = "unnam
         case _ => data.take(bytes)
       }
       val remaining = data.drop(slice.length)
-      writeSingle(addr, slice, id, burst, len, size)(handleTransaction(addr, transactionId, remaining))
+      val thisLen = if (remaining.length == 0) {
+        ((slice.length / bytePerBeat.toFloat).ceil.toInt - 1) min len
+      } else {
+        len
+      }
+      writeSingle(addr, slice, id, burst, thisLen, size)(handleTransaction(addr, transactionId, remaining))
```
KireinaHoro, line 257: Not sure if silently down-capping `AxLEN` is a good idea, with the current semantics of `len`.  I see one of two ways: 
- change `len` to something like `maxLen`, so that it's used to limit the max burst size, or
- assert false here to complain that the user specified too big of `AxLEN`

lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/Axi4Master.scala
KireinaHoro: Looks like narrow bursts will be handled in the proposed version (since you use `fullStrb`).

Not so sure about the unaligned address case -- address is rounded at the beginning through `padData`... @Dolu1990 can you elaborate?

KireinaHoro (changes requested):
Thanks a lot and sorry for the delay!  I don't have time to test this locally yet -- will get back to you.

lib/src/main/scala/spinal/lib/bus/amba4/axi/sim/Axi4Master.scala
Nik-Sch: changed the parameter name to maxLen with a comment in the scaladoc.

I really don't like to assert here because this would require the user to specify a len that is compatible with the length of the data array and size.

Nik-Sch:
@Dolu1990 @KireinaHoro do you have any updates on this?

KireinaHoro:
Sorry for the delay, I've been traveling and only got a chance to test this locally.  LGTM

PR #1696, opened 2025-04-05
Add method to return a seq of tags of a given type

jdavidberger:
There wasn't a convenience function to grab a list of spinal tags and return them by type; this adds that in. 

I needed it for some code I am writing and thought others might also need something like it eventually

### DIFF
--- a/core/src/main/scala/spinal/core/Trait.scala
+++ b/core/src/main/scala/spinal/core/Trait.scala
@@ -29,6 +29,8 @@ import scala.collection.mutable.{ArrayBuffer, Stack}
 import spinal.core.internals._
 import spinal.idslplugin.Location
 
+import scala.reflect.ClassTag
+
 trait DummyTrait
 object DummyObject extends DummyTrait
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^

Yes, why not.
Before merging, can you check the .toSet ? it probably doesn't preserve deterministic order

jdavidberger:
toSet doesn't; I took the the set behavior of entirely since Seq is more general here

Dolu1990:
There is the toMapLinked which does that for Map, but it is missing for Set.
https://github.com/SpinalHDL/SpinalHDL/blob/3f49633cfdfb50e99f2084ceb0915a6c594015fb/lib/src/main/scala/spinal/lib/Utils.scala#L1205

Could be added aswell in this PR

jdavidberger:
I think having the function return an Iterable matches the filter function above it and so fits in better. Since it's based on a set it should already be unique AFAICT anyway. I can make it do the LinkedSet if you want but I'm not sure it adds anything to the function

Dolu1990:
hmmm your are right, let's keep it as a seq

Dolu1990:
Thanks ^^

PR #1700, opened 2025-04-07
Improve assignment overlap error message with previous location

pluveto:

**Context:** The `PhaseCheck_noLatchNoOverride` phase in Spinal HDL performs checks for potential issues in combinatorial logic, including detecting when a signal is fully assigned multiple times within the same scope (assignment overlap).

**Motivation:** The previous error message for `ASSIGNMENT OVERLAP` only reported the source code location of the *second* (or subsequent) assignment that caused the overlap. Users had to manually search backwards in their code to find the *first* assignment that was being completely overwritten. This process was time-consuming and frustrating, especially in complex logic blocks.

**Description:** This PR enhances the usability of the `ASSIGNMENT OVERLAP` error reporting.

1.  It modifies the `PhaseCheck_noLatchNoOverride` phase to track the assigned bits of a signal within a scope and the source code location of the **most recent assignment operation** (`lastLocation`) that affected that signal in the scope. This is managed using a new internal `AssignmentInfo` case class.
2.  When a complete assignment overlap is detected (i.e., a signal is fully assigned again within the same scope, and it's not explicitly allowed via the `allowAssignmentOverride` tag), the generated `PendingError` message is updated.
3.  The new error message now presents **both**:
    *   The source code location of the **offending assignment** (the one causing the overlap).
    *   The source code location of the **previous assignment operation** that occurred just before the offending one within that scope.

This provides developers with immediate context, pointing directly to the two relevant lines of code involved in the overlap, and accelerates the debugging process.

# Impact on code generation

This change affects a compile-time check phase (`PhaseCheck_noLatchNoOverride`) and its error reporting mechanism. It does not alter the logic analysis or the final generated VHDL, Verilog, or SystemVerilog code.

# Checklist

- [ ] Unit tests were added.
  - using Scaladoc comments: `/** */`.
  - on [RTD](https://github.com/SpinalHDL/SpinalDoc-RTD).
  - thanks to a [new tracking issue on RTD](https://github.com/SpinalHDL/SpinalDoc-RTD/issues/new?title=Document%20XXX&body=Do%20not%20merge%20until%20SpinalHDL/SpinalHDL%23XXX%20has%20not%20been%20merged).

### DIFF
--- a/core/src/main/scala/spinal/core/internals/Phase.scala
+++ b/core/src/main/scala/spinal/core/internals/Phase.scala
@@ -2467,7 +2467,7 @@ class PhaseCheck_noRegisterAsLatch() extends PhaseCheck{
   }
 }
 
-class PhaseCheck_noLatchNoOverride(pc: PhaseContext) extends PhaseCheck {
+class PhaseCheck_noLatchNoOverride(pc: PhaseContext) extends PhaseCheck { // Renamed class for clarity
   case class AssignmentInfo(bits: AssignedBits, lastLocation: Option[String])
 
   override def impl(pc : PhaseContext): Unit = {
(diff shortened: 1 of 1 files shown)


pluveto:
Thanks, I'll try to fix these test fails

core/src/main/scala/spinal/core/internals/Phase.scala
```diff
@@ -2476,152 +2477,214 @@ class PhaseCheck_noLatchNoOverride(pc: PhaseContext) extends PhaseCheck{
...
               val unassignedBits = new AssignedBits(bt.getBitsWidth)
-
               unassignedBits.add(bt.getBitsWidth - 1, 0)
               unassignedBits.remove(assignedBits)
 
               if (!unassignedBits.isEmpty) {
                 if (bt.dlcIsEmpty) {
                   if(!bt.hasTag(allowFloating)) {
-                    PendingError(s"NO DRIVER ON $bt, defined at\n${bt.getScalaLocationLong}")
+                    PendingError(s"NO DRIVER ON $bt (combinatorial signal with no logic connection), defined at\n${bt.getScalaLocationLong}")
```
Dolu1990, line 2673: Hmm maybe this could be "(Combinatorial signal with no driver)" ? Or something else ? 
no logic connection isn't precise enough

Dolu1990:
Look good, thanks :D

PR #1702, opened 2025-04-07
sim: Extend logic to retrieve make binary path

Tectu:
Some time ago, the environment variable `SPINAL_MAKE_CMD` was introduce on my request to allow overriding the filesystem path to the `make` binary which was previously hardcoded to `"make"`.
This is due to the fact that SpinalHDL requires GNU make. On non-GNU platforms, this is typically `gmake`.

This PR adds logic to default to `"gmake"` on platforms that are known to be non-GNU (FreeBSD in this case).

This change should not affect any existing users as:
1. We always use `SPINAL_MAKE_CMD` if it's defined
2. We always default to `"make"`

Please review this carefully - still new to Scala.

### DIFF
--- a/sim/src/main/scala/spinal/Env.scala
+++ b/sim/src/main/scala/spinal/Env.scala
@@ -1,5 +1,34 @@
 package spinal
 
 object SpinalEnv {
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks ^^

PR #1707, opened 2025-04-14
Add an easy way to enable sync resets in FormalConfig

pftbest:

Recent versions of yosys do not like to have async resets in the design when running in single clock mode. The error message looks like this:

ERROR: $check cell ... with TRG_WIDTH > 1 is not support by async2sync, use clk2fflogic.

See https://github.com/YosysHQ/yosys/issues/4424 and https://github.com/YosysHQ/yosys/issues/4231

In most cases the issue can be avoided by setting resetKind=SYNC on the default clock domain to make the whole design synchronous.

This PR adds an easy way to do that when creating new FormalConfig

Also related:
#1314
#1435
#1430

# Impact on code generation

# Checklist

- [ ] Unit tests were added


Readon:
Thanks. 
Is it possible to add a simple test case.

PR #1711, opened 2025-04-20
Support various settings in formalRamCheck

jdavidberger:
formalCheckRam in StreamFifo can currently crash in various conditions. This fixes that by adding logic for 1 and 0 element fifos. 

It also adds a 'formalCheckLastPush' which can be useful to assert something about the last valid push.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1495,45 +1495,61 @@ class StreamFifo[T <: Data](val dataType: HardType[T],
     }
   }
 
-
+  def formalCheckLastPush(cond: T => Bool) : Bool = new Composite(this) {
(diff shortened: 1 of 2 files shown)


Dolu1990:
@Readon

Readon:
Good catch, I think it's necessary to discuss on more cases.
The design purpose of formalCheckRam is "check a condition against all valid payloads in the FIFO RAM".

I wonder that if formalCheckRam should be applicapable withByPass on, if so the check logic for depth == 1 withByPass on might need to be carefully designed.
In this case, the push and pop payload would be connected directly in cases, which payload would not be saved in the ram.
This also applied to depth == 0 case.

The cases are also the same in `formalCheckLastPush`.

jdavidberger:
I've added unit tests for the depth == 0 and bypass cases. I believe they work as intended.

PR #1712, opened 2025-04-22
add Stream.delay and Flow.delay

facebreeze:
Closes #

Compared to the  `StreamDelay` component in the lib, this provides a simple implementation to delay the `Stream` with a given number of cycles, which should save some area for small delay cycles.

# Impact on code generation

# Checklist


Dolu1990:
Nice thanks ^^

PR #1719, opened 2025-04-30
More scaladoc

mrcmry:
A serie of improvement of the scaladoc, mainly taken from the RTD Spinal doc. Also some spelling and typo in code comments.

# Impact on code generation

No change in lib code, only doc


Dolu1990:
Thanks ^^

PR #1723, opened 2025-05-03
Fix typo in 'isFireing' to 'isFiring' across multiple sources

pluveto:

Fix typo in 'isFireing' to 'isFiring' across multiple source files.

NOTE: Old API isFireing is still available but marked `@deprecated`.

# Impact on code generation

As @Readon said, if a signal has "firing" in its name it will be replaced with "firing"

# Checklist

- [N/A] Unit tests were added
- [N/A] API changes are or will be documented:

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Aligner.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axi/Axi4Aligner.scala
@@ -108,7 +108,7 @@ class Axi4WriteOnlyAligner(upConfig: Axi4Config, bytesMax : Int, slotsCount : In
       context.write.address := ptr.cmd.resized
       context.write.data.last := LAST
       context.write.data.id := AW.id
-      when(isFireing) {
+      when(isFiring) {
         context.write.valid := True
         slots.onSel(ptr.cmd.resized) { slot =>
           slot.done := False
@@ -194,7 +194,7 @@ class Axi4WriteOnlyAligner(upConfig: Axi4Config, bytesMax : Int, slotsCount : In
       valid := ptr.cmd =/= ptr.fetch
       context.read.cmd.valid := False
       context.read.cmd.payload := ptr.fetch.resized
-      when(isFireing){
+      when(isFiring){
         context.read.cmd.valid := True
         ptr.fetch := ptr.fetch + 1
       }
@@ -204,7 +204,7 @@ class Axi4WriteOnlyAligner(upConfig: Axi4Config, bytesMax : Int, slotsCount : In
       val reader = slots.reader(ptr.rsp.resized)
       val done = reader(_.done)
       val error = reader(_.error)
-      val errorAcc = RegInit(False) setWhen(isFireing && error)
+      val errorAcc = RegInit(False) setWhen(isFiring && error)
 
       io.up.b.valid := False
       io.up.b.id := CTX.id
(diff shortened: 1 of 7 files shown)


mrcmry:
Hi !

I was doing some typo and scaladoc improvement lately and was thinking also to fix this spelling in the name error. But I don't know how much of a problem it is to deprecate method like this. I don't see any problem, even with tool parsing the generated HDL. The only problem would be to change the parameter name of a method, because of the named parameter functionality is Scala.

On top of my head there is also:

* Symplify -> simplify
* ElementLanda -> ElementLambda in misc.database
* a landa package in test (at this point I'm not really sure it's a typo, I have found no doc anywhere on the web about "landa")

I don't see any problem doing this incrementally but perhaps it's better to group these change in a branch and merge before a minor instead of a bugfix release.

What do you think ?

mrcmry:
I would add that it look like a trivial and even pedantic change, but as a beginner these kind of misspelled names are puzzling. You are never really sure if it's done on purpose to name something specific (like Closjure, Mosquitto, functor, Reqwest, Rediculous, etc.) and spend of few minutes on internet every time to try to educate yourself without success.

pluveto:
Hi @mrcmry!

Thanks so much for the review and for finding those other potential typos!

Good point on the Scala method/param naming – glad to hear you don't see an issue with this specific fix regarding that.

Regarding the strategy for handling these: I really like your idea of having a dedicated branch for cleanup efforts and merging before a minor release – that makes a lot of sense for systemizing these fixes.

Given this PR is a very small, clear typo fix, could we perhaps merge this one as a bugfix now, and then start that dedicated cleanup branch (maybe named chore/text-cleanup or similar) for the other typos you found and future ones, targeting a minor release as you suggested? You can paste the link to that branch later.

What do you think is the best approach? Happy to proceed as you advise.

mrcmry:
@Dolu1990 what do you think ? 

Should I open a branch on my fork and start to refactor this (I can easily merge pluveto change in this branch) ?

Dolu1990:
> Should I open a branch on my fork and start to refactor this (I can easily merge pluveto change in this branch) ?

Hmm the issue with having branches which stay unmerged long is that they will get out of date and may increase workload. Also, manpower is limited and there is the risk that branches just get lost, so I would say let's merge into dev. Not a big deal ^^

All good ?

Readon:
I think there would be impact on generated code, the prefix would be changed to is_firing_xxxx, isn't it?

pluveto:
> I think there would be impact on generated code, the prefix would be changed to is_firing_xxxx, isn't it?

Good catch

mrcmry:
Perhaps one way to integrate the change but without changing generated code is to add progressively:
```
  /** This spelling will be deprecated in a future minor release, use `isFiring`
  // SPELLING_MIGRATION_TODO
  def isFireing: Bool =  signalCache(this -> "isFiring")(isValid && isReady).setCompositeName(this, "isFiring")

  /** Return if it is in the process of being on fire :) */
  def isFiring: Bool = isFireing
```

Without renaming to `isFireing` in the lib.

Then, when we think we have fixed most of the spelling problems, change to:

```
 @deprecated("Use isFiring instead", "SpinalHDL 1.14.0")
  def isFireing: Bool =  isFiring

  /** Return if it is in the process of being on fire :) */
  def isFiring: Bool = signalCache(this -> "isFiring")(isValid && isReady).setCompositeName(this, "isFiring")
```

and rename all usages in the lib.

This way we break names in generated code only one time and people fight against deprecated warnings in code only one time.

Dolu1990:
Hmm, as it is the pipelinine API, and we still keep backward compatibility via depeciated API, i think it is realy fine to just merge as it is.
Just that the next SpinalHDL release will be tagged as 1.13.0 at least.

> This way we break names in generated code 

I would say, this isFiring signal should never be used as a interface, but only interaly to a the module it is created, so, not realy a breaking change.

PR #1726, opened 2025-05-07
Introduce a null-check within the PhaseRemoveUselessStuff phase

pluveto:

This change introduces a null-check within the `PhaseRemoveUselessStuff` compilation phase to improve error reporting.

Occasionally, a `Statement` object (`s`) processed within this phase might have a `null` reference for its `s.component` field. If `s.component.pulledDataCache(...)` is subsequently called, a `NullPointerException` occurs. The default NPE lacks specific details about the problematic `Statement`, making it very hard to debug.

This modification adds a check immediately before `s.component.pulledDataCache(...)` is accessed. If `s.component` is `null`, a `NullPointerException` is thrown with an enhanced message, including details about the `Statement` (type, name, and source location, if available) and the specific `pulledDataCache` operation being attempted.

The goal is to provide more actionable diagnostic information when this specific internal inconsistency arises, aiding in debugging. The change is localized to the point of access and does not alter other behaviors of the phase.

# Impact on code generation

This change does not affect the generated hardware.

# Checklist

- [N/A] Unit tests were added (This change improves diagnostics for an existing internal error condition).
- [N/A] API changes are or will be documented (Internal change to a compilation phase, no public API alteration).

### DIFF
--- a/core/src/main/scala/spinal/core/internals/Phase.scala
+++ b/core/src/main/scala/spinal/core/internals/Phase.scala
@@ -2084,80 +2084,128 @@ class PhaseRemoveUselessStuff(postClockPulling: Boolean, tagVitals: Boolean) ext
 
     val okId = globalData.allocateAlgoIncrementale()
 
-    def propagate(root: Statement, vital: Boolean): Unit = {
-      if(root.algoIncrementale == okId) return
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks ^^

PR #1729, opened 2025-05-12
fix width issue in AFix negate()

thajohns:

The `AFix` `negate` method on signed `AFix`es always tries to assign the result's `raw` to a new bit vector that is one bit wider, but negating the `AFix` doesn't always make it wider.

# Impact on code generation

Now, when an `AFix`'s `raw` is assigned during negation, it is resized to fit. This should be fine since if it doesn't use the entire available range in that many bits, then it's just a sign extension bit being removed.

# Checklist

- [x] Unit tests were added
- [x] API changes are or will be documented (none needed)


Dolu1990:
Thanks ^^

PR #1730, opened 2025-05-12
Fix AFix <<| bug

thajohns:

The `AFix` `<<|` operator (where the right operand is a constant) did not behave in a way that made sense, and was not symmetric with the `AFix` `>>|` operator or other types' `<<|` and `>>|` operators. It shifted both the bit representation and the exponent of the `AFix`, when shifting either one of those would have produced a left-shift-like change in the value of the `AFix`.

# Impact on code generation

`AFix`'s `<<|` operator (where the right operand is a constant) no longer changes the resolution of the left operand.

# Checklist

- [x] Unit tests were added


Dolu1990:
Nice catch ^^
thanks

PR #1744, opened 2025-06-15
add assignUnassignedFromBits for MultiData

facebreeze:
Closes #

Assign some uninterested bits to zeros or ones is quite common in engineering. 
# Impact on code generation
None

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/MultiData.scala
+++ b/core/src/main/scala/spinal/core/MultiData.scala
@@ -246,6 +246,21 @@ abstract class MultiData extends Data {
     }
   }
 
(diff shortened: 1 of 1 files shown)


Readon:
I think this function might lead to misunderstanding, user might need to trace which signals are unassigned manually, which is hard.
Give out info to explain which signals connected to which bits is a good helper, but might flush the terminal.

facebreeze:
Yes, this function should be used with caution, and potentially leads to bugs. Maybe functions named **assignUnassignedFromOnes** or **assignUnassignedFromZeros** will do ?

PR #1751, opened 2025-06-26
Port Vivado constraint writer to the new TimingExtractor framework

KireinaHoro:
Ported the existing `VivadoConstraintWriter` to the new `TimingExtractor` framework.  This way we can share the DSL walking logic with other constraint writers, for example the `TimingExtractorSdc`.  Expanded the error messages when encountering a unknown driver for a signal.

Adapted the existing test case that bit-rotted some time ago.  Is there a way to include this in the CI?

Regarding the input / output delay generation:  I believe the existing code doesn't generate useful constraints.  Here's a [reference](https://www.programmersought.com/article/18569207623/) on how to write correct input/output delay constraints.  I included a comment for what we still need to do.  In summary, we still need to:
- introduce a separate tag (instead of just `ClockDomainReportTag`) to allow specifying input/output delays on signals
- walk DSL to find these tags
- in Tcl, locate the actual top-level port (in case the module is included in a bigger design)
- `set_input_delay`

I'll mark this as ready to review when I've tested the generated constraints in Vivado, for my private project.  But for now @andreasWallner @Dolu1990 what are your thoughts?  Thanks!

# Impact on code generation

None

# Checklist

- [x] generated constraint with `spinal.lib.eda.xilinx.Test`
  - [ ] add checks on generated result
  - [ ] include checks in CI
- [x] generated constraint with local project
  - [ ] test constraints with Vivado


Dolu1990:
TimingExtractorXdc sound good. I would say to preserve the old implemention in, for backward compatibilty. (it is fine to have some legacy duplicata)

KireinaHoro:
I've tested the new thing and added one change to find the source and destination clocks in max delay constraints from the clock pins of the data drivers directly.  This previously relied on having access to the component from the `DataAssignStatement` -- this is now not possible.

Simply locating the clock domain is not sufficient, since the top level does not have a `KEEP_HIERARCHY` attribute and the clock pins will be renamed.  I've tested the current approach in Vivado and it seems to be working.

@Dolu1990 I'm rather against keeping the old `VivadoConstraintWriter`: after the changes of `TimingExtractor`, the old netlist walker somehow became broken and would output a lot of `??` on stdout.

Dolu1990:
> I'm rather against keeping the old VivadoConstraintWriter: after the changes of TimingExtractor, the old netlist walker somehow became broken and would output a lot of ?? on stdout.

Ok then.

Let me know when i shoud merge :)

KireinaHoro:
@Dolu1990 ready to merge :)

PR #1756, opened 2025-07-04
CI:Add disk space optimization for Verilator 5.x CI tests

jaynerlin:
Closes #

## Problem
Verilator 5.x generates significantly more intermediate files during simulation compilation compared to earlier versions, causing "No space left on device" errors in GitHub Actions CI runners. This particularly affects the `tester-test` and `lib-test` jobs which involve extensive simulation testing.

## Solution
This PR introduces a targeted disk space optimization system that:

### New Components
1. **Disk Cleanup Action** (`.github/actions/disk-cleanup/action.yml`):
   - Background monitoring process that tracks disk usage every 3 minutes
   - Automatic cleanup of Verilator intermediate files when disk usage exceeds 90%
   - System optimizations specific to Verilator 5.x (Java heap tuning, coredump disabling)
   - Comprehensive logging and reporting

2. **Enhanced CI Workflow** (`.github/workflows/sbt-tests.yml`):
   - Conditional activation: only enables disk optimization for Verilator 5.x versions
   - Targeted integration in `tester-test` and `lib-test` jobs (the most disk-intensive tasks)
   - Post-test cleanup and detailed disk usage reporting

### Key Features
- **Smart Detection**: Automatically detects Verilator version and enables optimization only when needed
- **Non-intrusive**: Runs as background process without affecting test execution
- **Comprehensive Cleanup**: Removes Verilator-generated files (*.cpp, *.gch, *.o, *.vcd, etc.)
- **Detailed Reporting**: Provides disk usage statistics before, during, and after tests
- **Fail-safe Design**: Includes proper error handling and cleanup even if tests fail

### Files Cleaned(Impact only when testing CI)
- Verilator C++ compilation artifacts (`V*__ALL.cpp`, `V*__Trace*.cpp`)
- Precompiled headers (`*.gch`)
- Object files (`*.o`, `*.d`)
- Simulation workspace directories (`simWorkspace`)
- VCD dump files (`*.vcd`)
- Temporary Verilator headers in `/tmp`

This solution specifically addresses the disk space challenges introduced by Verilator 5.x while maintaining backward compatibility and not affecting the actual test logic.

# Impact on code generation

**No impact on HDL code generation.** This PR only modifies CI/CD infrastructure to handle disk space management during testing. The actual SpinalHDL compilation and code generation processes remain unchanged.

The optimization affects only the CI environment cleanup and does not modify:
- VHDL/Verilog/SystemVerilog output
- SpinalHDL compiler behavior
- Test execution logic
- Generated hardware descriptions

# Checklist

- [x] Unit tests were added - *N/A: This is CI infrastructure change*
- [N ] API changes are or will be documented - *N/A: No API changes*
  - using Scaladoc comments: `/** */`? - *N/A*
  - on [RTD](https://github.com/SpinalHDL/SpinalDoc-RTD)? - *N/A*
  - thanks to a [new tracking issue on RTD](https://github.com/SpinalHDL/SpinalDoc-RTD/issues/new?title=Document%20XXX&body=Do%20not%20merge%20until%20SpinalHDL/SpinalHDL%23XXX%20has%20not%20been%20merged)? - *N/A*

### DIFF
--- a/.github/actions/disk-cleanup/action.yml
+++ b/.github/actions/disk-cleanup/action.yml
@@ -1,0 +1,116 @@
+name: 'Disk Space Optimization for Verilator CI'
+description: 'Monitor disk usage and clean intermediate files during SpinalHDL CI tests'
+
+inputs:
+  cleanup_interval:
+    description: 'Cleanup interval in seconds'
+    required: false
+    default: '180'  # 3 minutes
+
+runs:
+  using: "composite"
+  steps:
+    # Check initial disk space and set environment variables
+    - name: Setup disk optimization environment
+      shell: bash
+      run: |
+        echo "=== Initial Disk Space Check ==="
+        df -h /
+        
+        # Set monitor parameters
+        echo "CLEANUP_INTERVAL=${{ inputs.cleanup_interval }}" >> $GITHUB_ENV
+        echo "DISK_CRITICAL_THRESHOLD=90" >> $GITHUB_ENV
+        
+    # Start background disk monitor and cleanup process
+    - name: Start background disk monitor
+      shell: bash
+      run: |
+        cat > /tmp/spinal_disk_monitor.sh << 'EOF'
+        #!/bin/bash
+        
+        LOG_FILE="/tmp/disk_monitor.log"
+        PID_FILE="/tmp/disk_monitor.pid"
+        echo $$ > $PID_FILE
+        
+        log_with_time() {
+            echo "[$(date '+%H:%M:%S')] $1" | tee -a $LOG_FILE
+        }
+          
+        cleanup_spinal_files() {
+            # Clean up Verilator generated temporary files
+            log_with_time "Cleaning Verilator generated files"
+            find . -name "V*__ALL.cpp" -mmin +2 -delete 2>/dev/null || true
+            find . -name "V*__Trace*.cpp" -mmin +2 -delete 2>/dev/null || true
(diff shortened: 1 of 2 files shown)


jaynerlin:
I have completed the modification

jaynerlin:
I forgot to fix the way to get the verilator_version version. In the test branch I established myself, in order to facilitate the testing of multiple versions of verilator, I used the method that can be specified.
I forgot to modify this. The disk cleanup link I repaired is not enabled by default, so the ci test fails. The failed information is consistent as I analyzed (insufficient disk space), and the memory of the intermediate compilation file of the higher version of verilator is even larger.

Readon:
It's weird that some tests would pass even when there is no #1757 applied with verilator 5.

jaynerlin:
Yes, after my testing, only in verilator v5.022 (excluding v5.022) and later versions will only not pass the lib-test and tester-test workflows, and only some of the tests in these two workflows are for a single test instance to call multiple times dosim{} for emulation, and only in this way it will result in memory access violation.causing the process to crash.
It's worth mentioning that this scenario was already being sent before verilator v5.022: concurrent ci tests with insufficient disk space caused the workflow to fail.
It's worth mentioning that this situation was already being sent before verilator v5.022: concurrent ci tests with insufficient disk space caused the workflow to fail.
I haven't been able to pinpoint exactly which subversion of verilator has the specific update that is causing problem [#1757](https://github.com/SpinalHDL/SpinalHDL/pull/1757) and this issue

PR #1768, opened 2025-07-18
Add optional 32-bit sysbus to RiscV DebugModule

craigjb:
# Description
This change adds an optional system bus `sysBus` to the RISC-V DebugModule and implements 32-bit read+write accesses per RISC-V Debug Specification v1.0.

When enabled using `withSysBus = true` in `DebugModuleParameter`, a simple `DebugSysBus` is added to the DebugModule's IO. It is up to the user to handle commands and responses. I've implemented support in VexRiscV's EmbeddedRiscvJtag plugin, and I'll raise a separate PR for that.

Only 32-bit addresses and accesses are supported for now, but this change could be extended for 128, 64, 16, or 8 bit accesses.

# Motivation and Context
Without system bus support, debug memory accesses must go through the CPU. I noticed this when building Rust firmware for a VexRiscV based system with RTT logging. When running the firmware with probe-rs, the host continuously looks for new logs in a ring buffer in memory. Without system bus support, this causes delays in the firmware's execution during those memory polls--not helpful when you're working on something with real-time requirements!

# Impact on code generation
Library change only

# Checklist
- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugModule.scala
+++ b/lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugModule.scala
@@ -12,7 +12,8 @@ case class DebugModuleParameter(version : Int,
                                 harts : Int,
                                 progBufSize : Int,
                                 datacount : Int,
-                                hartsConfig : Seq[DebugModuleCpuConfig])
+                                hartsConfig : Seq[DebugModuleCpuConfig],
+                                withSysBus: Boolean = false)
 
 
 object DebugDmToHartOp extends SpinalEnum{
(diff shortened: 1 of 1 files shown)


Dolu1990:
Note, i'm in holydays / trip until the 26 :)

craigjb:
> Note, i'm in holydays / trip until the 26 :)

Enjoy!

Dolu1990:
Hi,
I pushed two changes : 

[DebugModule unified p.withSysBus generate](https://github.com/SpinalHDL/SpinalHDL/pull/1768/commits/a1e5a77dfc92cbe07bd25b786e621050041dc1c3)
So here the idea is to preserve the old behaviour of just "ignoring it all", as it was working  well with all de debugger tested to far, as well as syimplifying the code base / risk

[DebugModule bus now use flow as a respons](https://github.com/SpinalHDL/SpinalHDL/pull/1768/commits/fa2ca2b5bfbb98cac42d567b61688793114abe82)
which swap the "ready" for a "valid" flag.
That valid flag should only be high a single cycle per response.

Does the changes look good for you ?

craigjb:
@Dolu1990 I think your commits look good.

The previously used "ready" because it matched `DBusSimpleRsp` in VexRiscv's DBusSimpleBusPlugin. Connecting sysBus to an AXI or other bus is really simple with that infrastructure already there.

With this change to a Flow, I think I'd just have to map `DBusSimpleRsp.ready` to the `valid` signal of the Flow, and then the other pieces from the payload. So, not a big deal at all.

Dolu1990:
Thanks ^^

PR #1773, opened 2025-07-28
 Fixed the problem of process hanging indefinitely after the simulation is over in Windows environment

### LINKED ISSUE #1775: Simulation hangs indefinitely in Windows/MSYS2 after completion with Verilator v5.004+
In Windows/MSYS2 environments, SpinalHDL simulations using Verilator (v5.004 and later) complete logically but fail to exit the process, hanging indefinitely after displaying "Simulation done" and requiring manual termination. This issue does not occur in Verilator v5.002 or earlier, nor in Linux environments.

To illustrate, in normal scenarios (e.g., on Linux or older Verilator versions), the simulation follows a complete workflow, ending with a success message like this (simplified):

```
[info] [Progress] Simulation workspace in ...\simWorkspace\MySim
[info] [Progress] Verilator compilation started
[info] [Progress] Verilator compilation done in 601.934 ms
[info] [Progress] Start MyAdder 1 simulation with seed ...
[info] doSim1 Simulation successful!
[info] [Done] Simulation done in 5.191 ms
[info] [Progress] Start MyAdder 2 simulation with seed ...
[info] doSim2 Simulation successful!
[info] [Done] Simulation done in 0.943 ms
[success] Total time: 14 s, completed ...
```

However, in the affected Windows/MSYS2 setup, the simulation runs normally up to "Simulation done," but then hangs without the final "[success] Total time" output, as shown below (simplified):

```
[info] [Progress] Simulation workspace in ...\simWorkspace\MyAdder
[info] [Progress] Verilator compilation started
[info] [Progress] Verilator compilation done in 601.934 ms
[info] [Progress] Start MyAdder 1 simulation with seed ...
[info] doSim1 Simulation successful!
[info] [Done] Simulation done in 4.650 ms
[info] [Progress] Start MyAdder 2 simulation with seed ...
[info] doSim2 Simulation successful!
[info] [Done] Simulation done in 0.922 ms
```

The key issue is that although the simulation logic executes successfully, the process does not exit automatically and lacks the final end-of-simulation when the sbt tool prints the `[success] Total time` end-of-simulation message, requiring manual intervention.

jaynerlin:
Closes #1775

In Windows/MSYS2 environments, SpinalHDL simulations using Verilator (v5.004+) complete logically but fail to exit, hanging indefinitely after "Simulation done" and requiring manual termination. This does not occur in v5.002 or earlier, nor in Linux.
Root cause: Verilator v5.004 forces thread-safe architecture, removing --no-threads and unifying code paths. SpinalHDL's JNI integration uses global Verilated methods (e.g., Verilated::randReset) without setting thread_local context, leading to nullptr returns in Verilated::threadContextp() under MSYS2's strict thread management. This breaks $finish signal propagation, causing the Java-side loop to hang.
Solution: Introduce explicit VerilatedContext instance in VerilatorBackend.scala for state management, using instance methods (e.g., contextp->randReset) and ensuring proper cleanup. This bypasses thread_local dependency, enabling reliable status synchronization and process exit. Verified on Windows/MSYS2 and Linux; compatible with all Verilator versions.

# Impact on code generation

This is a simulation backend change only, with zero impact on VHDL/Verilog/SystemVerilog code generation, RTL output, or simulation logic.

# Checklist

- [x] Unit tests were added


jaynerlin:
@Dolu1990 While looking at the fix code, I noticed that the original code had comments //VerilatedContext* contextp; //Buggy in multi threaded spinalsim and //contextp = new VerilatedContext;, which seem to have been there originally but commented out, and thechose to use the global Verilated method.May I ask why Verilated was chosen in the first place instead of using a VerilatedContext instance?Is this because of a specific bug in a multi-threaded environment?Sharing more context would help me understand this design decision better.

Readon:
I have tested this PR on MSYS2, it passed. However, I am not sure why it is commented out in previous version, too.

jaynerlin:
> Hi,
> 
> > I have tested this PR on MSYS2, it passed. However, I am not sure why it is commented out in previous version, too.
> 
> So, the tricky thing with verilator (or that i didn't properly understood) is how to run multiple instances of it in the same process, while keeping control of the verilator randomizer. I did hit quite some issue getting that right, (and it is probably still wrong), to be able to properly control seeds, and avoid multiple instances to use the same seed. Mostly the verilator randomizer is used to give initial values to all the signal / registers. Before 5.0, it was working well with the current SpinalHDL implementation, but since 5.0, things when out of hands XD

Through testing, I have verified that the following two management approaches are equivalent in different scenarios for giving initial values to all signals/registers in (Verilator v5.004, v5.022, v5.024, v5.036 versions):

**Approach One**: Generate a VerilatedContext management context for each simulation instance

```c++
VerilatedContext* contextp;  
contextp = new VerilatedContext;                                
contextp->randReset(2);                                              
contextp->randSeed(seed);
```

**Approach Two**: Directly use Verilated's static methods (managed through static VerilatedContext methods)

```c++
// VerilatedContext* contextp; 
// contextp = new VerilatedContext;
Verilated::randReset(2);
Verilated::randSeed(seed); 
```

VerilatedContext and global Verilated static methods perform completely equivalently in terms of signal/register initialization. Both approaches rely on the same Verilator VL_RAND_RESET mechanism, producing identical initial values for the same seed, with consistent behavior in seed isolation, reproducibility, and concurrency safety.

However, Approach Two causes many problems, such as access to already freed memory errors when calling the doSim method multiple times on the same SimCompiled object, as well as processes hanging indefinitely after the simulation is over in Windows environments... I suspect this is due to version updates in Verilator's global static information class Verilated.

Dolu1990:
Tested, seems to work well with verilator 5.
Verilator 4.2xx seems it was already broken before in that regards. 

Thanks for the PR :D

Issue #1259, opened 2023-12-14
Issues with verilator 5


andreasWallner:
After discussing some verilator behavior with Gitter user solater we came up with this example that has some (I'd say)  weird behavior with verilator 5:
```scala

case class Repro() extends Component {
  val io = new Bundle {
    val address = in port UInt(1 bit)
    val value = in port UInt(8 bit)
    val cond = in port Bool()
  }
  val init = Seq.fill(2){U(0)}
  val mem = Mem(UInt(8 bits), init)
  when(io.cond) {
    mem(io.address) := io.value
    report(Seq("writing ", io.address, " = ", io.value))
  }
}

object Repro extends App {
  val dut = SimConfig.withFstWave.compile{
    val dut = Repro()
    dut.mem.simPublic()
    dut
  }
  for(_ <- 0 until 100)
    dut.doSim { dut =>
      dut.io.cond #= false
      dut.clockDomain.fallingEdge()

      dumpData()

      for(i <- 0 until 2) {
        assert(dut.mem.getBigInt(i) == 0)
      }

      def dumpData(): Unit = {
        println(simTime())
        for (ii <- (0 until dut.mem.wordCount).grouped(8)) {
          print(f"${ii(0)}%04d  ")
          for (i <- ii) {
            print(f"${dut.mem.getBigInt(i)}%08x ")
          }
          println()
        }
      }
    }
}
```
Which normally simulates fine, but will fail randomly like so:
```
[info] [Progress] Start Repro test_1 simulation with seed 1906628651
[info] 0
[info] 0000  00000000 00000000 
[info] [Done] Simulation done in 6.013 ms
[info] [Info] Test 'test' was reallocated as 'test_2' to avoid collision
[info] [Progress] Start Repro test_2 simulation with seed 1361515301
[info] NOTE writing 0 = 89
[info] 0
[info] 0000  00000089 00000000 
[info] [Error] Simulation failed at time=0
[error] Exception in thread "main" java.lang.AssertionError: assertion failed
[error]         at scala.Predef$.assert(Predef.scala:156)
[error]         at spinal.core.package$.assert(core.scala:497)
[error]         at vexiiriscv.scratchpad.Repro$$anonfun$6$$anonfun$apply$mcVI$sp$2$$anonfun$apply$4.apply$mcVI$sp(Play.scala:217)
[error]         at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:160)
[...]

PR #1777, opened 2025-08-04
Add Workaround for RAM with init values in sim

g0t00:
This is a workaround for #1259
The issue being with a RAM that has an init value, when using verilator it gets randomly overwritten at the initialization of the simulation.
The Init of the sim gets interpreted by the sim as a clock event if understand correctly, and then it is random if write enable is high or not. This is workaround by ANDing the enable with a wide register that gets reset to all ones. Thus being always one except for in init where it is very unlikely to be all ones.
I think this is quite hacky, but I do not have a better idea.

# Checklist

- [X] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/Mem.scala
+++ b/core/src/main/scala/spinal/core/Mem.scala
@@ -389,7 +389,13 @@ class Mem[T <: Data](val wordType: HardType[T], val wordCount: Int) extends Decl
   def writeImpl(address: UInt, data: Data, enable: Bool = null, mask: Bits = null, allowMixedWidth: Boolean = false): Unit = {
 
     val whenCond =  if(enable == null) ConditionalContext.isTrue() else enable
-    val writePort = MemWrite(this, address, data.asBits, mask, whenCond, if(allowMixedWidth) data.getBitsWidth else getWidth ,ClockDomain.current)
+    val whenGuarded =
+      if (GenerationFlags.simulation) {
+        val dummy = RegInit(B(True, 256))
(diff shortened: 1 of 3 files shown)


Dolu1990:
Hi,

What version of verilator do you used and observed this behaviour ?

g0t00:
@Dolu1990 Tested this with `Verilator 5.021 devel rev v5.020-79-gc702fc944 (mod)`
My assumption was this is same for other Verilator 5 versions..

Dolu1990:
> My assumption was this is same for other Verilator 5 versions..

So far i avoided verilator 5 because it has some bad behaviour which was triggering rising edge at time 0 before anything realy happened.

It may more be a verilator bug than a SpinalHDL one.

Can you try with verilator 4.2xx ?

jaynerlin:
> @Dolu1990 I just checked with verilator 4.226 there it works as expected. To be honest I do not fully understand the `--x-initial-edge` situation so not sure if this bug is in verilator or in SpinalHDL... But verilator 5 has for use case some clear performance advantages, so I would prefer to use verilator 5.

Can I ask you questions about verilator 5's performance advantages?I'm working on this aspect recently. Maybe it's a problem with my test case. I added verilator's simulation thread but the performance is still reduced.
In what way does verilator 5 show better performance?It would be great if you could provide example files!

g0t00:
Okay, so this ugly workaround is not required anymore. I think we can keep the test case anyways.

Dolu1990:
Thanks :D

PR #1780, opened 2025-08-07
Mask tilelink bus valid bit when the bus is marked as assignDontCare

inochisa:

When the tilelink bus is marked as assignDontCare, it will be assigned as random bits. This will break the system if the valid bit is assigned as true and the channel will send an invalid request.

Mask the valid bit of the channel A and B if the bus is a master port.

# Impact on code generation

This make the code generation always produce the `1b'0'` instead of `1b'x'` for the valid line.

# Checklist

- [x] Unit tests were added


inochisa:
@Dolu1990

Dolu1990:
Ahh i would say, assignDontCare is more an "raw" API, with no intends to be extended for specific busses.
Maybe you would prefer something like "def driveIdle(){ ... }" ? or "def setIdle(){ ... }"

inochisa:
> Ahh i would say, assignDontCare is more an "raw" API, with no intends to be extended for specific busses. Maybe you would prefer something like "def driveIdle(){ ... }" ? or "def setIdle(){ ... }"

Yeah, I think this is OK for the bus. Or apply this for the all base type? 
I found this is caused by the valid bits of `Stream`, and I think it may cause trouble for any unassigned master `Stream`.

Issue #1476, opened 2024-07-11
Observations about AFix


thajohns:
I'm currently working on a project involving a lot of fixed-point math, and I noticed a few things that annoyed me about the `AFix` API that I wanted to share.

It would be nice if there were more functions that were bit and `exp`-oriented, similar to `.trim()` but with the power to:
- round, saturate, or truncate
- keep the same max or min value or exponent, or specify either
- specify the number of bits in the result, rather than the number being removed
- specify a `HardType[AFix]` to become

Such functions would help with
- creating `HardType[AFix]`es for component port types etc.
- using e.g. DSP blocks, which need a specific number of bits of input

It seems like there are some inconvenient ways in which `AFix`es do not interact nicely with Scala types -- for example, I can't say `a := 1.0 - b`.

Sometimes it is necessary to deal with the bits directly instead of the `AFix` type. Using `.raw` works well enough, but it would be nice to be able to index the bits by their exponent.

Sometimes I want to truncate/saturate/round, but I know that if I saturate away the MSB, I've done something wrong. (Analogously, it could be that rounding the LSB means I've done something wrong.) Functions like `.truncated(...)`, but which only allow removing bits on one side (otherwise producing an error), would help me detect such mistakes. Another `RoundType` that just causes an error if the LSB needs to be removed would allow the existing `.truncate(...)` to do that.

I'd be interested in hearing other peoples' thoughts on the issues presented above.

Dolu1990:
> I'm currently working on a project involving a lot of fixed-point math, and I noticed a few things that annoyed me about the AFix API that I wanted to share.

Thanks for the feedback :D

> It would be nice if there were more functions that were bit and exp-oriented, similar to .trim() but with the power to:

API extentions are welcome (PR), as long as things as things are well named there should be no problems. We also have to be carefull to not create duplication (it is easy to forget about one part of the API is already existing)

> HardType[AFix]es for component port types 

Hmm maybe HardType[AFix] should be used with care, because it kinda force you to "craft" the type in order to access to the AFix parameters.
I think we would need a AFixParam class to allow capturing the AFix val maxRaw: BigInt, val minRaw: BigInt, val exp: Int parameters.

> a := 1.0 - b

One issue here is that the resolution of the substraction is kinda unknown.
Something more like : 
a := 1.0.as(b) - b  could be a possible API ? function "as" meaning "The double value 1.0 with the resolution of b"


So, overall, all the points you bring up are very likely valids.

[...]

PR #1798, opened 2025-09-09
Add `RoundType.NONE`

thajohns:
Related to ideas in #1476 

Sometimes a function will have a `RoundType` argument that decides how rounding is done, but you want to make sure that you don't actually lose any bits. This adds a new `RoundType` which enforces that by throwing an error if bits are lost.

 This is pretty much done except for tests, but I wanted to get some thoughts on it before I bother to figure out how to write a test that checks for an error :P

# Checklist

- [ ] Unit tests were added - not yet
- [ ] API changes are or will be documented - needs a new row on rounding modes table [here](https://github.com/SpinalHDL/SpinalDoc-RTD/blob/master/source/SpinalHDL/Data%20types/Int.rst) ; I will make a PR if people like this idea.

### DIFF
--- a/core/src/main/scala/spinal/core/AFix.scala
+++ b/core/src/main/scala/spinal/core/AFix.scala
@@ -329,6 +329,11 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
             _roundFixedBigInt(i, exp, ROUNDUP)
           }
         }
+      case NONE => {
+        assert(exp <= 0, "RoundType is NONE but rounding requires losing precision")
+        i
+      }
+
     }
   }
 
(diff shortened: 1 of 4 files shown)


thajohns:
@Dolu1990 is this a dumb idea?

Dolu1990:
This is a good idea :)
I missed this PR.

Let me know when i can merge ^^

Tectu:
@thajohns Any progress on this regarding adding tests?

thajohns:
No. It shouldn't be that difficult, I just didn't know off the top of my head how to make a test pass when an elaboration fails. I probably won't have the time or inclination to work on this for a long time -- if you need it, you should probably finish it up.

Tectu:
I don't think that it's wise to merge this without adding any corresponding tests.

PR #1800, opened 2025-09-09
add AFix contructors from width and resolution

thajohns:
Related to ideas in #1476 

Adds a variant of each of `AFix.S` and `AFix.U` based on the width and resolution of the `AFix`.

It seems like there was a commented-out variant with the same parameter types, but which used the first `BitCount` to specify the number of bits in the integer part of the fixed-point. That seems inconsistent with the other variants of the same name, so I thought it would be best to get rid of it.

# Checklist

- [x] Unit tests were added


thajohns:
@Dolu1990 do you have any thoughts on this PR?

Dolu1990:
Thanks, look good to me ^^

PR #1801, opened 2025-09-09
AFix `resize()` bug

thajohns:

It was previously impossible for `AFix.resize()` to be a no-op -- there was a check that it always strictly resized. That seemed non-obvious and needlessly restrictive, so I have changed it.

# Checklist

- [x] Unit tests were added
- [x] API changes are or will be documented

### DIFF
--- a/core/src/main/scala/spinal/core/AFix.scala
+++ b/core/src/main/scala/spinal/core/AFix.scala
@@ -666,8 +666,12 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
     ret
   }
 
+  /** Changes the resolution of the AFix without changing its represented value, by adding low bits. */
   def resize(newExp : ExpNumber): AFix ={
-    assert(newExp.value < exp, s"AFix resize loses precision -- use a rounding function instead") //for now
+    assert(newExp.value <= exp, s"AFix resize loses precision -- use a rounding function instead") //for now
+    if (newExp.value == exp) {  // no-op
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spinal/core/AFix.scala
```diff
@@ -666,8 +666,12 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
     ret
   }
 
+  /** Changes the resolution of the AFix without changing its represented value, by adding low bits. */
   def resize(newExp : ExpNumber): AFix ={
-    assert(newExp.value < exp, s"AFix resize loses precision -- use a rounding function instead") //for now
+    assert(newExp.value <= exp, s"AFix resize loses precision -- use a rounding function instead") //for now
+    if (newExp.value == exp) {  // no-op
```
Readon, line 672: according to line above this line takes no effect.

is this PR necessary?

thajohns: The important part for my use-case is the inequality in the assert. I guess the conditional is superfluous, since a shift-left by zero is a no-op -- not sure what I was thinking. I can remove it when I get the chance.

Dolu1990: Ahh i missed this PR.
the PR is necessary
if (newExp.value == exp) .. combinit isn't necessary, but it works, and as it is already there, why not keeping it, as it also generate a bit less verilog.

PR #1806, opened 2025-09-11
Implement automatic initial reset and signal analysis in Verilator ba…

jaynerlin:
…ckend

Closes 
#1777 
#1259 

This PR implements a comprehensive RTL-based auto-reset sequence for the Verilator backend, replacing Verilator's "--x-initial-edge" option with a SpinalHDL native solution, which changed its behavior in version 5.x. The PR also implements a comprehensive RTL-based auto-reset sequence for the Verilator backend.

The current Verilator backend relies on the `--x-initial-edge` option for initial reset, which:
- In verilator version v5.x there is an unintended signal jumping behavior, which leads to some unintended errors.
- Doesn't provide precise control over reset sequences
- Trigger assertion judgment at time zero
[ref](https://github.com/verilator/verilator/issues/6377):

RTL analysis-based automatic reset system that:
1. **Analyzes the entire SpinalHDL design** using `GraphUtils.walkAllComponents` to discover all registers and their clock domains
2. **Extracts precise reset and clock signal information** including:
   - Reset signal polarity (HIGH/LOW active)
   - Reset type (ASYNC/SYNC)
   - Clock edge sensitivity (RISING/FALLING)
3. **Generates C++ reset sequences** that:
   - Properly handle multiple reset signals with different polarities
   - Support both asynchronous and synchronous resets
   - Generate appropriate clock edges for synchronous reset propagation
   - Ensure all registers start from a deterministic state
Can be enabled/disabled via `SpinalSimConfig`

# Impact on code generation

No effect-this PR does not modify any Verilog/VHDL/SystemVerilog code generation logic, all improvements are at the simulation tool level.
This PR also not affect the user's subsequent simulation operations, and the auto-reset operation will not be recorded by verilator.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
+++ b/core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
@@ -287,6 +287,7 @@ class ComponentEmitterVerilog(
   def emitInitials() : Unit = {
     var withRandBoot = ArrayBuffer[(BaseType, String)]();
     var withInitBoot = ArrayBuffer[(BaseType, String)]();
+    var withSimInit = ArrayBuffer[(BaseType, String)]();
     component.dslBody.walkDeclarations {
       case bt: BaseType => {
         if (!bt.isSuffix) {
@@ -298,12 +299,16 @@ class ComponentEmitterVerilog(
             case null =>
             case str  => withInitBoot += bt -> str
           }
+          getBaseTypeSignalSimInit(bt) match {
+            case null =>
+            case str  => withSimInit += bt -> str
+          }
         }
       }
       case _ =>
     }
 
-    if(initials.isEmpty && withRandBoot.isEmpty && withInitBoot.isEmpty) return
+    if(initials.isEmpty && withRandBoot.isEmpty && withInitBoot.isEmpty && withSimInit.isEmpty) return
     logics ++= "  initial begin\n"
     emitLeafStatements(initials, 0, c.dslBody, "=", logics , "    ")
 
@@ -314,6 +319,11 @@ class ComponentEmitterVerilog(
         logics ++= s"${theme.maintab + theme.maintab}${name}${str};\n"
       }
       logics ++= "  `endif\n"
+    }
+
+    for((bt, str) <- withSimInit){
+      val name = emitReference(bt, false)
+      logics ++= s"${theme.maintab + theme.maintab}${name}${str};\n"
     }
 
     for((bt, str) <- withInitBoot){
@@ -1153,6 +1163,50 @@ class ComponentEmitterVerilog(
       }
     }
     null
+  }
+
+  def getBaseTypeSignalSimInit(signal: BaseType): String = {
+    if(signal.isReg){
+      signal.getTag(classOf[SimInitTag]) match {
+        case Some(tag) =>
+          try {
+            val result = tag.value match {
+              case bvl: BitVectorLiteral =>
+                val targetWidth = signal.getBitsWidth
+                val value = bvl.getValue()
+
+                // Handle negative values
+                val unsignedValue = if (value >= 0) {
+                  value
+                } else {
+                  (BigInt(1) << targetWidth) + value
+                }
+
+                // Use hex for width > 4, binary for width <= 4
+                if (targetWidth > 4) {
+                  val hexDigits = (targetWidth + 3) / 4
+                  val hexValue = unsignedValue.toString(16)
+                  val paddedHex = ("0" * (hexDigits - hexValue.length)) + hexValue
+                  s"${targetWidth}'h${paddedHex}"
+                } else {
+                  val binValue = unsignedValue.toString(2)
+                  val paddedBin = ("0" * (targetWidth - binValue.length)) + binValue
+                  s"${targetWidth}'b${paddedBin}"
+                }
+              case _ =>
+                emitExpressionNoWrappeForFirstOne(tag.value)
+            }
+            " = " + result
+          } catch {
+            case e: Exception =>
+              SpinalError(s"Failed to process SimInit for signal $signal: ${e.getMessage}. " +
+                s"SimInit value must be a compile-time constant. Tag value: ${tag.value}, type: ${tag.value.getClass}")
+          }
+        case None => null
+      }
+    } else {
+      null
+    }
   }
 
   var memBitsMaskKind: MemBitsMaskKind = MULTIPLE_RAM
--- a/.github/workflows/mill-tests.yml
+++ b/.github/workflows/mill-tests.yml
@@ -133,7 +133,7 @@ jobs:
     lib-test:
         needs: compile
         runs-on: ubuntu-latest
-        timeout-minutes: 90
+        timeout-minutes: 120
         container:
           image: ghcr.io/spinalhdl/docker:latest
         steps:
--- a/.github/workflows/sbt-tests.yml
+++ b/.github/workflows/sbt-tests.yml
@@ -139,7 +139,7 @@ jobs:
   lib-test:
     needs: compile
     runs-on: ubuntu-latest
-    timeout-minutes: 90
+    timeout-minutes: 120
     container:
       image: ghcr.io/spinalhdl/docker:${{ inputs.docker_image_version }}
     env:
--- a/core/src/main/scala/spinal/core/Reg.scala
+++ b/core/src/main/scala/spinal/core/Reg.scala
@@ -20,6 +20,7 @@
 \*                                                                           */
 package spinal.core
 
+import spinal.core.internals.Expression
 import spinal.idslplugin.Location
 
 
@@ -88,4 +89,13 @@ object RegInit {
   def apply[T <: Data](init: T): T = Reg(init, init)
 
   def apply[T <: SpinalEnum](init : SpinalEnumElement[T]) : SpinalEnumCraft[T] = apply(init())
+}
+
+
+/**
+ * SimInit tag for register simulation initialization
+ * Sets simulation-only initial values via Verilog initial blocks
+ */
+case class SimInitTag(value: Expression) extends SpinalTag {
+  override def allowMultipleInstance = false
 }
--- a/core/src/main/scala/spinal/core/core.scala
+++ b/core/src/main/scala/spinal/core/core.scala
@@ -661,4 +661,50 @@ package object core extends BaseTypeFactory with BaseTypeCast {
       formatted.mkString
     }
   }
+
+  /**
+   * Extension methods for register simulation initialization
+   */
+  implicit class DataSimInitPimper[T <: Data](data: T) {
+    /**
+     * Set simulation initial value for registers
+     *
+     * @param that Initial value for simulation (supports same types as init())
+     * @return The register itself for chaining
+     *
+     * @example {{{
+     * val counter = Reg(UInt(8 bits)).simInit(0x32)
+     * val reg = Reg(UInt(8 bits)).init(0x00).simInit(0x32)
+     * }}}
+     */
+    def simInit(that: T): T = {
+      require(data.flatten.forall(_.isReg), "simInit can only be applied to registers")
+      
+      // Extract literal values from expressions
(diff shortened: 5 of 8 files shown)


jaynerlin:
After testing, this solution can be well compatible with verilator v4.x and v5.x. I have tested and passed it on v4.228,v5.002, and v5.038

I failed the ci tests many times before because I was too careless and forgot to consider the situation  without reset signals, which led to C++ syntax errors in the newly added functions

Dolu1990:
Hi ^^

Nice :D
So, the only thing i think need to be discussed is : 
var _enableRtlAutoReset: Boolean = true

I don't know if it should be true by default, because this could hide hardware bugs where part of the design aren't getting reseted, (but will appear as reseted because of the simulation workaround for verialtor), am i right ?

jaynerlin:
I think this scheme is an upper-level alternative to the "-x-initial-edge" option. "-x-initial-edge" will trigger all level conversions. This PR scheme is achieved by automatically triggering the reset signal of the registers that need to be initialized before simulation
> this could hide hardware bugs where part of the design aren't getting reseted

I didn't understand this passage. Could you give an example

jaynerlin:
Perhaps I submit a solution that only performs reset initialization for the reset controller

Dolu1990:
> I didn't understand this passage. Could you give an example

Let's say i have some peripherals, with some hardware which can be reseted via a register (a flip flop drive the reset pin of some hardware).

```scala
val doReset = Reg(Bool()) // User forget to init that register
val logicCd = ClockDomain(clk, doReset)
val logic = logicCd on new Area{ ... }
```

The question is what is the behaviour of such hardware when simulating with this PR ? Without it, doReset will have a random value (correct).
Will this PR automaticaly assert the doReset register ? And so, hide this hardware bug ?

jaynerlin:
> Let's say i have some peripherals, with some hardware which can be reseted via a register (a flip flop drive the reset pin of some hardware).
> 
> ```scala
> val doReset = Reg(Bool()) // User forget to init that register
> val logicCd = ClockDomain(clk, doReset)
> val logic = logicCd on new Area{ ... }
> ```
> 
> The question is what is the behaviour of such hardware when simulating with this PR ? Without it, doReset will have a random value (correct). Will this PR automaticaly assert the doReset register ? And so, hide this hardware bug ?

Yes, after my test, it does hide this hardware error,

> Nice :D So, the only thing i think need to be discussed is : var _enableRtlAutoReset: Boolean = true

And if it is set to var enableRtlAutoReset by default: Boolean = false cases of spinal. Lib. SpinalSimDmaSg2ReadOnlyTester this test case will fail, because the test case has reset controller, need a similar "-- x - initial - edge" option to reset the activation,[ref: test in verilator v4,228 without "-- x - initial - edge"](https://github.com/jaynerlin/SpinalHDL/actions/runs/17576367610/job/49922734824)

jaynerlin:
> he only thing i think need to be discussed is : var _enableRtlAutoReset: Boolean = true

Hi,
Currently, I have changed it to disable the auto-reset method by default.  Additionally, if the reset signal is not initialized when the automatic reset function is performed, a warning will be reported

jaynerlin:
## Alternative Solution: RegSimInit

There is another viable alternative solution:
Create a new register initialization method `RegSimInit` that directly adds `initial` blocks in the generated Verilog, for example:

```scala
// SpinalHDL code
val counter = RegSimInit(UInt(8 bits), 0x32)
```

```verilog
// Generated Verilog code
reg [7:0] counter;
initial begin
    counter = 8'h32;  // Direct initial value assignment
end

always @(posedge clk, negedge reset) begin  // Reset logic can be configured as needed
    if(!reset) begin
        counter <= 8'h00; 
    end else if(enable) begin
        counter <= counter + 1;
    end
end
```

This approach avoids power-on reset signal issues by allowing users to set the power-on initial values for individual registers, directly providing a known initial circuit state.

Although `resetKind = BOOT` reset type also supports register power-on initialization, the key differences between this method and BOOT reset are:

- **BOOT Reset**: Operates at the entire clock domain level. When `clockDomain.config.resetKind == BOOT`, all registers within that clock domain use the BOOT reset strategy. This is mutually exclusive with traditional reset signals - you cannot simultaneously use async/sync reset and BOOT initialization within the same clock domain.

- **RegSimInit**: Operates at the individual register level, allowing selective initial value setting for specific registers within the same clock domain.

### Enhanced Flexibility

RegSimInit provides greater flexibility by enabling mixed usage within the same clock domain:

```scala
val clockDomain = ClockDomain(clk, reset, config = ClockDomainConfig(resetKind = ASYNC))
val clockArea = new ClockingArea(clockDomain) {
  val counterReg1 = Reg(UInt(8 bits)) init(0)        // Needs reset function
  val counterReg2 = RegSimInit(UInt(8 bits), 0xFF)   // Only needs deterministic initial value
}
```

This solution is particularly suitable for designs that need to simultaneously satisfy the following requirements:

- **Critical registers** need reliable reset protection
- **Debug registers** need deterministic initial states  
- **Simulation** needs predictable starting conditions

This approach bridges the gap between hardware reset requirements and simulation initialization needs, providing designers with fine-grained control over register initialization behavior.

Readon:
How about adding a function called 'simInit' to regs and wires (or regs only), who's init value can be set and shown at initial block?

Dolu1990:
@Readon  
> How about adding a function called 'simInit' to regs and wires (or regs only), who's init value can be set and shown at initial block?

A bit like randBoot(), but with an explicit value ?

Readon:
> @Readon
> 
> > How about adding a function called 'simInit' to regs and wires (or regs only), who's init value can be set and shown at initial block?
> 
> A bit like randBoot(), but with an explicit value ?

Yes, I think this can give the reset signal an appropriate initial value, so it doesn't have to rely on the functionality of the validator.

Dolu1990:
Ahhhh some updates on the verilator side : 
https://github.com/verilator/verilator/issues/4424#issuecomment-3315454618

jaynerlin:
> Ahhhh some updates on the verilator side :啊，验证器方面的一些更新： [verilator/verilator#4424 (comment)](https://github.com/verilator/verilator/issues/4424#issuecomment-3315454618)

I checked the updates within the last month and found no updates related to the --x-initial option.

Dolu1990:
I reopened a verilator issue : 
https://github.com/verilator/verilator/issues/6476

Also @jaynerlin What's about we just remove --x-initial-edge and change nothing else ?
Did you tried with the master verilator version ?

jaynerlin:
> Also @jaynerlin What's about we just remove --x-initial-edge and change nothing else ? 

Some CI test cases include an asynchronous reset controller and rely on the init values of the registers for testing. When only the --x-initial-edge option is removed, these tests will fail.
And subsequently, the simulation program with an asynchronous reset controller relies on the user to reset it by themselves. Is this a bit inconvenient?

Dolu1990:
> Yes, I think this can give the reset signal an appropriate initial value, so it doesn't have to rely on the functionality of the validator.

So i think the best would be to go that route, as it would allow to handle embedded reset sources, by forcing their value on simulation start. (and so, no more  --x-initial-edge)

Does that sound right ?

jaynerlin:
> So i think the best would be to go that route, as it would allow to handle embedded reset sources, by forcing their value on simulation start. (and so, no more --x-initial-edge)

I have added a new method, simInit(), to the register object.


The following is a brief introduction:

`simInit()` is a method in SpinalHDL used to set simulation initial values for registers. It only affects the simulation environment and does not impact hardware synthesis.

### Basic Syntax

```scala
val register = Reg(DataType).simInit(initialValue)
```

### Examples

```scala
// Basic usage
val counter = Reg(UInt(8 bits)).simInit(0x42)

// Combined with hardware reset
val reg = Reg(UInt(8 bits)).init(0x00).simInit(0x32)

// Supports all BitVector types
val sintReg = Reg(SInt(8 bits)).simInit(-5)
```

### Key Features

- Non-breaking changes
- Supports all clock domain reset types
- Fully compatible with existing code

Dolu1990:
Nice :D

So, as far my understanding goes, with this simInit and no --x-initial-edge, we should be all good ?

No need anymore for the autoInitialReset ?

jaynerlin:
> So, as far my understanding goes, with this simInit and no --x-initial-edge, we should be all good ?

Yes
> No need anymore for the autoInitialReset ?

However, for this submissions, some ci test objects rely on autoInitialReset. If the autoInitialReset function is not needed, I can modify these test cases to the simInit() method

Dolu1990:
>  If the autoInitialReset function is not needed, I can modify these test cases to the simInit() method

Yes please :)

core/src/main/scala/spinal/core/internals/ComponentEmitterVerilog.scala
```diff
@@ -320,6 +325,11 @@ class ComponentEmitterVerilog(
       val name = emitReference(bt, false)
       logics ++= s"${theme.maintab + theme.maintab}${name}${str};\n"
     }
+
+    for((bt, str) <- withSimInit){
```
Dolu1990: I think the withInitBoot should have priority over withSimInit, as withInitBoot is the design intend, while withSimInit a simulation workaround.
So those two for loop would need to be swap.

jaynerlin:
> If the autoInitialReset function is not needed, I can modify these test cases to the simInit() method

  @Dolu1990 I tried  and found this method extremely challenging, because many SpinalHDL library modules and CI test cases rely on registers having deterministic initial values at the start of simulation. After removing --x-initial-edge, if we also remove autoInitialReset and rely entirely on simInit for power‑on initialization, we would need to annotate registers with simInit point by point inside a large number of library modules.
  
  Take the SpinalSimDmaSg2ReadOnlyTester test case as an example: it directly reuses DmaSgReadOnlyComp and involves numerous registers (control/status/pointer/interrupt/counter registers, with both synchronous and asynchronous reset semantics). I tried adding simInit(...) broadly within DmaSgReadOnly, which does eliminate unknown states, but the scope of changes is large, the maintenance cost and risk of errors are high, and I ultimately had to roll back.
  
  To keep such modules stably passing CI after removing autoInitialReset, we would need systematic, per‑module, fine‑grained annotations across all relevant library modules, which is very difficult in the short term. In my view, simInit is suited for fine‑grained reset control, while autoInitialReset is more appropriate for users who are not deeply familiar with SpinalHDL’s library (like me).

Dolu1990:
> So, as far my understanding goes, with this simInit and no --x-initial-edge, we should be all good ?
> Take the SpinalSimDmaSg2ReadOnlyTester test case as an example:

I checked.
So the reason why this one fail (and others aswell) is because they have little reset controllers in each crossclock fifo. When those are init randomized by verilator to their asserted level, this won't create reset edge.

The fix is in lib/src/main/scala/spinal/lib/CrossClock.scala: 

```scala
object ResetCtrl{
  def asyncAssertSyncDeassert(input : Bool,
                              clockDomain : ClockDomain,
                              inputPolarity : Polarity = HIGH,
                              outputPolarity : Polarity = null, //null => inferred from the clockDomain
                              bufferDepth : Option[Int] = None,
                              inputSync : Bool = False) : Bool = {
   ...
    val ret = samplerCD(BufferCC(
      input       = ((if(solvedOutputPolarity == HIGH) False else True) ^ inputSync).setCompositeName(input, "asyncAssertSyncDeassert", true),
      init        = if(solvedOutputPolarity == HIGH) True  else False,
      bufferDepth = bufferDepth,
      allBufAttributes = falsePathAttrs)
    )
    ret.getDrivingReg().simInit(if(solvedOutputPolarity == HIGH) False else True)
    ret
  }
``` 

Can you give a try ?

jaynerlin:
> I checked. So the reason why this one fail (and others aswell) is because they have little reset controllers in each crossclock fifo. When those are init randomized by verilator to their asserted level, this won't create reset edge.

This fix is effective. Thank you very much for your answer. It can work normally now

Dolu1990:
> This fix is effective. Thank you very much for your answer. It can work normally now

Nice :D
Thanks for the fixes ^^

PR #1812, opened 2025-09-20
lib: misc: aia: APlicSource: fix wrong mode access when update sourcecfg

inochisa:

When split the mode support from APlicSource, the parameter "update mode" of driveConfigUpdateCtx is missing, and the function access the global mode, which is different from the update mode and make the ip detection broken.

Add the mode parameter to fix the detection behavior.

# Impact on code generation

`APlicSource` now will detect interrupt line and update the ip state when updating `config` register.

# Checklist

- [x] Unit tests were added


Dolu1990:
Thanks ^^

PR #1819, opened 2025-10-01
Fix simulator waveform paths for GHDL and IVerilog

louiecaulfield:
The waveform path for IVerilog and GHDL is a concatenation of the user's home directory and the absolute version path of the testPath.

This patch ensures the waveforms get stored together with the rest of the simulation data for these backends, bringing them in line with the Verilator backend.


louiecaulfield:
@Dolu1990 Can you have a look at this?

Dolu1990:
Thanks ^^

PR #1822, opened 2025-10-08
AxiLite4Master: do not allow more than 1 outstanding transaction

KireinaHoro:
Just observed that under heavy load, the simulation master will interleave AW+W and B, as well as AR and R.  This is not allowed by standard (even though seemingly not complained about by `AxiLite4SlaveFactory`), but disallow nonetheless.


KireinaHoro:
No idea why the test failed -- it doesn't even use the simulation master so I don't think I introduced the failure...

Dolu1990:
Ahhh loool
I didn't know that AXI lite wasn't allowing multiple pending transactions XD
damned XD

Thanks !

KireinaHoro:
Thanks!  Basically you can't match the interleaved transaction results back to the requests, since AXI-Lite does not have AWID/ARID.

PR #1824, opened 2025-10-10
Mark unused data line as dont care; prevents unassigned errors in certain test modes

jdavidberger:

WIthout this annotation, when used in a formal test, spinal issues an error that the R data line isn't ever assigned. This fixes this.

The formal test I have inserts instrumentation signals on all streams to verify that the payload data does not change when it is stalled, which causes the signal to be read so it is a niche error and does not trigger in normal usage. But the fix should have no impact on generated code and signals the intent to not set the data so I think the change is a net positive overall.


Dolu1990:
Thanks ^^

PR #1832, opened 2025-11-04
fix typo

thajohns:
# Context, Motivation & Description

It seems like there's a typo in the method name.

# Impact on code generation

 None

# Checklist

- [ ] Unit tests were added
- [ ] API changes are or will be documented:

PR #1835, opened 2025-11-10
Fix `asReversedBits`

thajohns:

At the time of my last PR ( #1832 ) , I did not realize that `asReversedBits` was broken and had the same behavior as `asBits`. This PR fixes that, and also moves the method from `TupleBundle` to `MultiData`, so that it can be used with `Vec`s etc.

# Checklist

- [ ] Unit tests were added


Dolu1990:
Thanks ^^

PR #1837, opened 2025-11-11
add simSetAll method for simulation.

Readon:
Closes #

Add set all method to simulation.

There are two concerns:
1. If it is okay to mix the simulation function with design.
2. It will report fail while assign 255 to SInt, I think it should be act like that due to the method name is set all.

# Impact on code generation

# Checklist

- [x] Unit tests were not added, but have been tested internally.


Dolu1990:
Thanks ^^

Issue #1839, opened 2025-11-13
about removePruned = true


fengwu0805:
In the SpinalConfig settings, setting removePruned = true will prune connections between modules during RTL generation, especially if those signals are not considered essential for the functionality. However, since these signals are analog signal, you need to ensure they are not pruned. To prevent these signals from being optimized away, you can explicitly mark them as important or critical in your configuration. This can be done by using specific attributes or directives that instruct the toolchain to retain these signals. Check the documentation for SpinalConfig or the specific synthesis tool you are using to find out how to mark signals as non-prunable.

Below is the test code, and I want to retain the signals B2 and A2.

// SpinalHDL Code
```
case class A() extends BlackBox {
    val io = new Bundle {
        val A1 = inout(Analog(Bool()))
        val A2 = inout(Analog(Bool()))
    }
    noIoPrefix()
}

case class B() extends BlackBox {
    val io = new Bundle {
        val B1 = inout(Analog(Bool()))
        val B2 = inout(Analog(Bool()))
    }
    noIoPrefix()
}

case class C() extends Component{
    val io = new Bundle {
        val A1 = inout(Analog(Bool()))
        val B1 = inout(Analog(Bool()))
    }

    val a = A()
    val b = B()

    a.io.A1 <> io.A1
    b.io.B1 <> io.A1

    // inner connneted
    a.io.A2 <> b.io.B2
}

```

// RTL
```
module C (
  inout  wire          io_A1,
  inout  wire          io_B1
);


  A a_1 (
    .A1 (io_A1)  //~
  );
  B b_1 (
    .B1 (io_A1)  //~
  );

endmodule
```

Dolu1990:
Hi,

> removePruned = true

Ahhh this is a settings which is mostly a legacy from the early days of SpinalHDL. I'm never using it, because it just make debugging harder and can have surprising effects (without even considering the bug you described).

Likely, to fix that behaviour

I pushed a workaround : 
https://github.com/SpinalHDL/SpinalHDL/pull/1841

let's see how it goes.

PR #1841, opened 2025-11-14
#1839 workaround

Dolu1990:
Prevent analog signal from being pruned


fengwu0805:
> Prevent analog signal from being pruned
Thank you — your modification actually preserves the analog signal.

PR #1849, opened 2025-12-02
add Clip utility

facebreeze:
Closes #

Add widely used **clip** function based on existing Max&Min.

# Impact on code generation

# Checklist


Dolu1990:
Hi ^^
What's about naming it Clamp instead of Clip ?
(glm::clamp / https://registry.khronos.org/OpenGL-Refpages/gl4/html/clamp.xhtml)

Dolu1990:
thanks ^^

PR #1855, opened 2025-12-11
Replace hardcoded 32 with width parameter in extendedMantissa

jooyeon111:
Replaced hardcoded 32-bit width with the `width` parameter in `extendedMantissa` 
to make the function more flexible and consistent with the input width.

# Impact on code generation
No impact on code generation behavior. This is a code cleanup that makes 
the mantissa size properly match the width parameter instead of being 
hardcoded to 32 bits.

# Checklist
- [ ] Unit tests were added
- [ ] API changes are or will be documented


Dolu1990:
Thanks ^^

PR #1859, opened 2025-12-19
Add support for automated Vivado IP TCL generation and simulation integration

skylayer:
## Description

This PR introduces a new mechanism to handle Vivado IP generation scripts directly within SpinalHDL. It allows users to define TCL scripts for `BlackBox` components, which are then automatically collected, generated, and integrated into simulation flows (specifically XSim).

### Key Changes:

1.  **`VivadoIpTcl` Trait**: Added a new trait that can be mixed into any `BlackBox`. It requires implementing `vivadoIpTcl(): String`, which returns the TCL commands needed to create and configure the IP in Vivado.
2.  **`PhaseVivadoIpTcl`**: A new compiler phase that:
    *   Walks the component hierarchy to find all `BlackBox` instances using `VivadoIpTcl`.
    *   Generates TCL scripts in the target directory.
    *   Supports two modes via `SpinalConfig`:
        *   **Consolidated (Default)**: Generates a single `${TopLevel}_ips.tcl` containing all IP definitions (deduplicated by `definitionName`).
        *   **Separate**: Generates one `.tcl` file per unique IP definition.
3.  **`SpinalConfig` Extensions**:
    *   `ipTclTargetDirectory`: Optional custom path for generated TCL scripts.
    *   `separateIpTcl`: Boolean flag to toggle between consolidated and separate file generation.
4.  **XSim Simulation Integration**:
    *   Updated `SpinalSimConfig` to automatically copy generated IP TCL scripts into the simulation workspace.
    *   Updated `XSimBackend` to `source` these TCL scripts during the Vivado project creation phase, ensuring IPs are correctly initialized before simulation.
5.  **Testing**: Added `VivadoIpTclTester` covering:
    *   Default consolidated TCL generation.
    *   Custom path and separate file generation.
    *   End-to-end integration with XSim.

## Motivation

Currently, integrating Xilinx IPs (like Floating Point IP, DSP48, etc.) into SpinalHDL designs often requires manual management of TCL scripts or XCI files. This PR automates the "TCL-based IP generation" flow, making it easier to maintain IP configurations in Scala code and ensuring they are always in sync with the RTL and simulation environment.

## Checklist
- [x] Code follows the style guidelines of this project
- [x] New tests added to verify changes
- [x] All tests passed (`sbt "tester/testOnly spinal.core.VivadoIpTclTester"`)


core/src/main/scala/spinal/core/internals/Phase.scala
```diff
@@ -2883,7 +2883,7 @@ class PhaseGetInfoRTL(prunedSignals: mutable.Set[BaseType], unusedSignals: mutab
  * Phase that collects all BlackBoxes mixing in VivadoIpTcl and generates Vivado IP TCL scripts,
  * either as a single consolidated script or as separate per-IP scripts depending on configuration.
  */
-case class PhaseVivadoIpTcl(report: SpinalReport[_]) extends Phase {
```
skylayer: To fix Scala 2.11 compile error

PR #1861, opened 2025-12-22
Fix spelling mistakes

Tectu:
This replaces some spelling mistakes of "unamed" with the correct spelling "unnamed".

I tried to ignore this for a while but it's grinding my gears. Apparently I am not the first one: 8abe10d267578ab157ba5a2186e26e6021cf4138


Dolu1990:
<3

PR #1862, opened 2025-12-25
Bump mill to version 1.1.0

### LINKED ISSUE #1803: Upgrade mill support to newer version
SpinalHDL is currently still using mill `0.11.x` and thus the `build.sc` format.  Mill upstream has now moved to a new build file format, `build.mill` for top-level projects and `package.mill` for submodules that other projects can depend on, since `0.12`.  I am thinking of having a go at upgrading the mill definitions, so that newer features and new plugins can be used (e.g. [mill-ivy-fetcher](https://github.com/Avimitin/mill-ivy-fetcher) that would allow building a mill-based project inside Nix).

I'm opening this to collect some opinions on what features / use cases our mill build should still support:

- allow including the source tree as a git submodule, to facilitate a private branch (as in [my project here](https://github.com/KireinaHoro/enzian-pio-nic/blob/49f7ec02d697a61adc1b1415a0455dea6b22ba43/build.sc#L18-L38))
- allow generating publish artifacts for Maven: @Dolu1990 is this needed, or do you primarily use sbt for this task?
- allow running CI tests with Mill: same question as above
- maintain commandline-level compatibility with instructions in the doc (or update the doc?)

TODOs I see right now:

- [ ] port `build.sc` to `build.mill`:
  - [ ] build
  - [ ] publish
  - [ ] doc
  - [ ] test
- [ ] refactor out common logic into `package.mill` for direct dependencies
- [ ] update [project template](https://github.com/SpinalHDL/SpinalTemplateSbt) (or split from Sbt template?)
- [ ] update docs

@Dolu1990 @Readon what are your thoughts?  Thanks!

inochisa:
Closes #1803

Since the mill is bump to 1.1.0 release version, I think it is a good time point to bump the mill version. As it is supported nested build.

# Impact on code generation

I am not sure, but I think this PR should have no impact on code generation

# Checklist

- [ ] Unit tests were added


Dolu1990:
@Readon @KireinaHoro 
Any objection ?
If i remember well, this could break the SpinalHDL integration from source in peoples mill project ? (big mill API change)

Readon:
if i remember right, it would.
we need to take more time to have it tested.



---Original---
From: ***@***.***&gt;
Date: Sun, Jan 4, 2026 17:58 PM
To: ***@***.***&gt;;
Subject: Re: [SpinalHDL/SpinalHDL] Bump mill to version 1.0.6 (PR #1862)


Dolu1990 left a comment (SpinalHDL/SpinalHDL#1862)
 
@Readon @KireinaHoro
 Any objection ?
 If i remember well, this could break the SpinalHDL integration from source in peoples mill project ? (big mill API change)

KireinaHoro:
Yes -- @inochisa can you please check if we would still be able to include SpinalHDL as a submodule?  Ideally the bumped version should still allow using it like [in this project](https://github.com/KireinaHoro/enzian-pio-nic/blob/4b4dad65c492f77462782fa50c27e4ed391195ce/build.sc#L18-L26), sorry that I don't have time to give you a more stripped-down version...

KireinaHoro:
Thanks -- @inochisa can you check to make sure that the workflow of using SpinalHDL as a git submodule would still work?  Sorry I don't have a cut-down example at the moment, a big example would be in [this project](https://github.com/KireinaHoro/enzian-pio-nic/blob/4b4dad65c492f77462782fa50c27e4ed391195ce/build.sc#L18-L26).

inochisa:
@KireinaHoro I guess it is broken at least for now. I got the following error

```
[build.mill-48] generatedScriptSources Package declaration "package build.project" in ext/SpinalHDL/project/package.mill does not match folder structure. Expected: "package build.ext.SpinalHDL.project"
Package declaration "package build" in ext/SpinalHDL/package.mill does not match folder structure. Expected: "package build.ext.SpinalHDL"
Package declaration "package build.project" in ext/SpinalHDL/project/Version.mill does not match folder structure. Expected: "package build.ext.SpinalHDL.project"
Package declaration "package build.project" in ext/SpinalHDL/project/package.mill does not match folder structure. Expected: "package build.ext.SpinalHDL.project"
Package declaration "package build" in ext/SpinalHDL/build.mill does not match folder structure. Expected: "package build.ext.SpinalHDL"
Package declaration "package build" in ext/SpinalHDL/package.mill does not match folder structure. Expected: "package build.ext.SpinalHDL"
```

It seems like this is mentioned in com-lihaoyi/mill#5091, as you have known before.
But I have failed to overcome this by get rid of the code mentioned in this issue.

KireinaHoro:
@inochisa thanks for the check!  So my idea when I opened the issue was that, I’ll have some time to downgrade the error as mill upstream suggested, and then do the version bump on our side.  Unfortunately I just didn’t get the time.  

Would you have the interest to tinker with mill a bit?  IIRC the change should be small, just disabling a check.  Thanks and happy new year!

inochisa:
@KireinaHoro At least for now, with PR com-lihaoyi/mill#6504, I have managed to compile my code with SpinalHDL repository.

You can check https://github.com/project-inochi/misc-ips/tree/mill-subproject for reference

KireinaHoro:
> @KireinaHoro At least for now, with PR [com-lihaoyi/mill#6504](https://github.com/com-lihaoyi/mill/pull/6504), I have managed to compile my code with SpinalHDL repository.
> 
> You can check https://github.com/project-inochi/misc-ips/tree/mill-subproject for reference

Thanks!  I really like the simplicity of the syntax and structure.  I haven't attempted to build locally just yet, since the mill PR is not merged yet.  I'll test locally as soon as that one is merged upstream.

inochisa:
@KireinaHoro @Dolu1990  See PR com-lihaoyi/mill#6655, it is merged and it will be include the incoming 1.1.0 release.I think you can give a test and to see whether there is something I have missed. I have updated the version pointer to the incoming release, which will be out in a few day according to the author.

KireinaHoro:
@inochisa LGTM, I took a glance at the upstream PR and noticed:

https://github.com/com-lihaoyi/mill/pull/6655/files#diff-a60dca7fc0cab99696ad5149b9c861c793795a694e8329ba3f14436dcd9485b6R15-R16

They say that we can't import traits from the nested `build.mill` directly, but you apparently did just that.  Is this a case that upstream allowed explicitly, or did they forget a check (which they might add back as a bugfix in the future...)?

**Update**: ah no apparently you didn't import any traits, you only imported a function that produced the module references.  This is pretty clever and it looks like it can work on my side as well.  Thanks!

Emin017:
> https://github.com/com-lihaoyi/mill/pull/6655/files#diff-a60dca7fc0cab99696ad5149b9c861c793795a694e8329ba3f14436dcd9485b6R15-R16
> 
> They say that we can't import traits from the nested `build.mill` directly, but you apparently did just that. Is this a case that upstream allowed explicitly, or did they forget a check (which they might add back as a bugfix in the future...)?
> 
> Update: ah no apparently you didn't import any traits, you only imported a function that produced the module references. Let me check my use case if this would suffice.

@KireinaHoro I think some upstream examples have not been updated yet. I’ll run through the tests first and then follow up with the Mill maintainers.

Dolu1990:
Let me know when this PR should be merged ^^
(I'm hands off when it is about mill)

inochisa:
Great, I will reminder you once the mill is released.

inochisa:
@KireinaHoro @Emin017 I have seen the mill is released, can you try the latest release and verify it is fine.
With this check I think it will be good to be merged.

Emin017 (approved):
```shell
372] hw.compile compiling 3 Scala sources to out/hw/compile.dest/classes ...
372] done compiling
378] hw.runMain
[Runtime] SpinalHDL dev    git head : 54e133b16a06cb6d38e036e627d3c45b445676a9
[Runtime] JVM max memory : 30208.0MiB
[Runtime] Current date : 2026.01.29 07:25:58
[Progress] at 0.000 : Elaborate components
[Progress] at 0.020 : PhaseCreateComponent
[Progress] at 0.577 : PhaseDummy
[Progress] at 0.577 : Checks and transforms
[Progress] at 0.577 : PhaseMemBlackBoxingDefault
[Progress] at 0.581 : PhaseDeviceSpecifics
[Progress] at 0.581 : PhaseApplyIoDefault
[Progress] at 0.583 : PhaseNameNodesByReflection
[Progress] at 0.585 : PhaseCollectAndNameEnum
[Progress] at 0.590 : PhaseCheckIoBundle
[Progress] at 0.592 : PhaseCheckHierarchy
[Progress] at 0.595 : PhaseAnalog
[Progress] at 0.599 : PhaseNextifyReg
[Progress] at 0.600 : PhaseRemoveUselessStuff
[Progress] at 0.608 : PhaseRemoveIntermediateUnnameds
[Progress] at 0.612 : PhasePullClockDomains
[Progress] at 0.616 : PhaseInferEnumEncodings
[Progress] at 0.623 : PhaseInferWidth
[Progress] at 0.627 : PhaseNormalizeNodeInputs
[Progress] at 0.632 : PhaseRemoveIntermediateUnnameds
[Progress] at 0.633 : PhaseSimplifyNodes
[Progress] at 0.636 : PhaseCompletSwitchCases
[Progress] at 0.636 : PhaseRemoveUselessStuff
[Progress] at 0.638 : PhaseRemoveIntermediateUnnameds
[Progress] at 0.638 : PhaseCheck_noLatchNoOverride
[Progress] at 0.648 : PhaseCheck_noRegisterAsLatch
[Progress] at 0.649 : PhaseCheckCombinationalLoops
[Progress] at 0.652 : PhaseCheckCrossClock
[Progress] at 0.656 : PhasePropagateNames
[Progress] at 0.658 : PhaseObfuscate
[Progress] at 0.658 : PhaseAllocateNames
[Progress] at 0.665 : PhaseDevice
[Progress] at 0.666 : PhaseGetInfoRTL
[Progress] at 0.667 : PhaseDummy
[Progress] at 0.667 : Generate Verilog to gen
[Progress] at 0.667 : PhaseVerilog
[Done] at 0.713
378/378, SUCCESS] ./mill hw.runMain io.inochisa.gen.clock.ClockDetectedVerilog 61s
```
The build can be completed with **Mill 1.1.0**.

```shell
❯ ./mill --version
Mill Build Tool version 1.1.0
java.version: 21.0.9
java.vendor: Azul Systems, Inc.
java.home: /home/emin/.cache/coursier/arc/https/cdn.azul.com/zulu/bin/zulu21.46.19-ca-jdk21.0.9-linux_x64.tar.gz/zulu21.46.19-ca-jdk21.0.9-linux_x64
file.encoding: UTF-8
os.name: Linux
os.version: 6.8.0-86-generic
os.arch: amd64
```

inochisa:
Great, @Dolu1990 let's merge this.

KireinaHoro:
Hold on a bit -- I've been struggling to get my stuff working.  So far I have two questions:

- In your `build.mill`, I don't see any `crossValue` any more
  - Scala version in the module deps graph is passed through this Cross[...].apply and then just discarded inside
  - please verify that if you build the downstream with another Scala version, that indeed no mixture of versions happen
- I can't seem to run any tests in my downstream module (that will also get included as a dep).  Hold tight I'll push an example.

build.mill
```diff
@@ -0,0 +1,149 @@
...
+    mvn"com.github.oshi:oshi-core:6.4.0"
+  )
+  def publishVersion = SpinalVersion.sim
+}
+
+object lib extends Cross[Lib](SpinalVersion.compilers){
+  def defaultCrossSegments = Seq(SpinalVersion.compilers.head)
+}
+trait Lib extends SpinalModule with SpinalPublishModule {
+  import moduleRefs._
```
KireinaHoro: I think here you'd need a

```scala
override def scalaVersion = crossValue
```

If I understand [how Cross works in mill](https://mill-build.org/mill/fundamentals/cross-builds.html) correctly.  Same argument applies to all other modules.

inochisa:
> Hold on a bit -- I've been struggling to get my stuff working. So far I have two questions:
> 
>     * In your `build.mill`, I don't see any `crossValue` any more
>       
>       * Scala version in the module deps graph is passed through this Cross[...].apply and then just discarded inside
>       * please verify that if you build the downstream with another Scala version, that indeed no mixture of versions happen
> 

I do not think it is  discarded at all, it is used by the moduleRefs, which includes the reference of the module with the same scala version.

I have prepared a mill build for VexiiRiscv to resolve my intellisence on vscode, and you can check out in https://github.com/project-inochi/VexiiRiscv/tree/mill, which can build in both 2.12 and 2.13 (you need to change the SpinalHDL branch by yourself, as I haven't update the reference since this is not merged.)

>     * I can't seem to run any tests in my downstream module (that will also get included as a dep).  Hold tight I'll push an example.

KireinaHoro:
> > Hold on a bit -- I've been struggling to get my stuff working. So far I have two questions:
> > ```
> > * In your `build.mill`, I don't see any `crossValue` any more
> >   
> >   * Scala version in the module deps graph is passed through this Cross[...].apply and then just discarded inside
> >   * please verify that if you build the downstream with another Scala version, that indeed no mixture of versions happen
> > ```
> 
> I do not think it is not discarded at all, it is used by the moduleRefs, which includes the reference of the module with the same scala version.

Ok good point, I missed the `crossScalaVersion` thing; I assume this takes the value from `crossValue` then.

> I have prepared a mill build for VexiiRiscv to resolve my intellisence on vscode, and you can check out in https://github.com/project-inochi/VexiiRiscv/tree/mill, which can build in both 2.12 and 2.13 (you need to change the SpinalHDL branch by yourself, as I haven't update the reference since this is not merged.)

Yeah that's good to know.  I managed to build my [repo of common stuff](https://github.com/KireinaHoro/spinal-blocks/blob/master/build.mill) -- so <del>LGTM now</del>.  Thanks!

KireinaHoro:
I seem to have encountered a version of the diamond problem -- my `spinal-blocks` repo (similar to your `misc-ips`) includes `spinalhdl` as a dependency; the big project that includes `spinal-blocks` also includes its own version of `spinalhdl`.  In our earlier setup, `spinal-blocks` will be built against the `spinalhdl` of the *parent*.  Right now, though, I had a crude attempt at this:

- blocks repo: https://github.com/KireinaHoro/spinal-blocks/blob/master/build.mill
- final project repo: https://github.com/KireinaHoro/enzian-pio-nic/blob/e31f214444204ff746ab742289c58fca0b87064a/build.mill#L19 (haven't finished upgrading yet)

But I see this error message:

image

This is a bit concerning, since it looks like what I'm trying to do (pass the SpinalModuleRef to the blocks build) won't work.  @inochisa do you have any ideas on this?

inochisa:
> I seem to have encountered a version of the diamond problem -- my `spinal-blocks` repo (similar to your `misc-ips`) includes `spinalhdl` as a dependency; the big project that includes `spinal-blocks` also includes its own version of `spinalhdl`. In our earlier setup, `spinal-blocks` will be built against the `spinalhdl` of the _parent_. Right now, though, I had a crude attempt at this:
> 
>     * blocks repo: https://github.com/KireinaHoro/spinal-blocks/blob/master/build.mill
> 
>     * final project repo: https://github.com/KireinaHoro/enzian-pio-nic/blob/e31f214444204ff746ab742289c58fca0b87064a/build.mill#L19 (haven't finished upgrading yet)
> 
> 
> But I see this error message:
> image
> 
> This is a bit concerning, since it looks like what I'm trying to do (pass the SpinalModuleRef to the blocks build) won't work. @inochisa do you have any ideas on this?

@KireinaHoro I think it is like alias problem, IIRC you dependency is like that

```
A -> (B, SpinalHDL)
B -> SpinalHDL
```

There is a problem, even if the project A and B reference the same project, the mill treats it as different one since they have different path (A.SpinalHDL and A.B.SpinalHDL). I guess this is thing that the mill should fix, or at least they should give some advice about this scenario. I suggest opening an issue in mill to see if they have some ideas about how to fix this.

KireinaHoro:
> There is a problem, even if the project A and B reference the same project, the mill treats it as different one since they have different path (A.SpinalHDL and A.B.SpinalHDL). I guess this is thing that the mill should fix, or at least they should give some advice about this scenario. I suggest opening an issue in mill to see if they have some ideas about how to fix this.

The thing is, the fundamental dependency model in mill is ok -- everything is a `Module`; this is how the old dependency replacement worked.  What's going on here is that we are actually passing around a `case class SpinalModuleRef` -- this is aliasing.

I think mill 1.x is now Scala 3 no?  Let me try a bit if [structural types](https://docs.scala-lang.org/scala3/book/types-structural.html) can work here...

inochisa:
> > There is a problem, even if the project A and B reference the same project, the mill treats it as different one since they have different path (A.SpinalHDL and A.B.SpinalHDL). I guess this is thing that the mill should fix, or at least they should give some advice about this scenario. I suggest opening an issue in mill to see if they have some ideas about how to fix this.
> 
> The thing is, the fundamental dependency model in mill is ok -- everything is a `Module`; this is how the old dependency replacement worked. What's going on here is that we are actually passing around a `case class SpinalModuleRef` -- this is aliasing.

Yes, have you tried to pass all the module to the next level? I guess it should work, in fact, I guess there are some problems since we have two copy of the same tree.... maybe something like out of sync.

> 
> I think mill 1.x is now Scala 3 no? Let me try a bit if [structural types](https://docs.scala-lang.org/scala3/book/types-structural.html) can work here...

IIRC it is.

KireinaHoro:
> Yes, have you tried to pass all the module to the next level? I guess it should work, in fact, I guess there are some problems since we have two copy of the same tree.... maybe something like out of sync.

Well if you look at the history of `build.sc` in the main tree, this is what I worked earlier to fix.  You'll have to choose one copy if we have multiple copies of `spinalhdl`, and previously I decided that it should be the one in the very top level.

In fact now I look at your `misc-ips` project, you're not actually overriding `moduleRefs` anywhere no?  This is where magic happens in the old implementation; as it is right now in this PR, if the downstream project depends on both core and lib, we will have two versions of core (albeit from the same tree -- not sure if mill dedups in this case).

> IIRC it is.

Just took a quick look, it requires defining a type with `Selectable` (that will be shared across everyone; the example in the doc is `Record`), and then all ducktyping classes inherit that.  For our case, I think we could do something like that in in an ivy module, which hopefully is deduplicated by mill?  Then all copies of `spinalhdl` should have `SpinalModuleRef` all extend this record, and then it should work.  Sounds quite hacky though...

We could also just go all the way back and use a `Map[String, ModuleRef]` to pass these stuff.  But you need to actually override this as well, see the previous comment.

inochisa:
> > Yes, have you tried to pass all the module to the next level? I guess it should work, in fact, I guess there are some problems since we have two copy of the same tree.... maybe something like out of sync.
> 
> Well if you look at the history of `build.sc` in the main tree, this is what I worked earlier to fix. You'll have to choose one copy if we have multiple copies of `spinalhdl`, and previously I decided that it should be the one in the very top level.
> 
> In fact now I look at your `misc-ips` project, you're not actually overriding `moduleRefs` anywhere no? This is where magic happens in the old implementation; as it is right now in this PR, if the downstream project depends on both core and lib, we will have two versions of core (albeit from the same tree -- not sure if mill dedups in this case).
> 

It is true, I switch this magic as I found the dependency now always require a module instead of the moduleRef. It seems we need to switch to a more generic things.

> > IIRC it is.
> 
> Just took a quick look, it requires defining a type with `Selectable` (that will be shared across everyone; the example in the doc is `Record`), and then all ducktyping classes inherit that. For our case, I think we could do something like that in in an ivy module, which hopefully is deduplicated by mill? Then all copies of `spinalhdl` should have `SpinalModuleRef` all extend this record, and then it should work. Sounds quite hacky though...
> 

It is very hacky for using something like reflection, I do not thing it is a good idea for this as it is hard to understand.

> We could also just go all the way back and use a `Map[String, ModuleRef]` to pass these stuff. But you need to actually override this as well, see the previous comment.

Great idea, I will switch to a generic things.

KireinaHoro:
@inochisa yup you got the idea, thanks a lot.  I think passing a `ModuleRef` should still work, I don't think that has changed too much with the new mill release (support was implemented upstream [here](https://github.com/com-lihaoyi/mill/issues/3715)).  IMO you can maybe pass/override a `def moduleRefs: Map[String, ModuleRef]` (for example with keys `core`, `lib`, ...) and this should still be less verbose than the old one where I have to manually override every `coreMod` `libMod` etc.

For upstream mill I think we don't need anything more from them, the only issue we had was the nested project stuff anyways and that's taken care of nicely now.

inochisa:
> @inochisa yup you got the idea, thanks a lot. I think passing a `ModuleRef` should still work, I don't think that has changed too much with the new mill release (support was implemented upstream [here](https://github.com/com-lihaoyi/mill/issues/3715)). IMO you can maybe pass/override a `def moduleRefs: Map[String, ModuleRef]` (for example with keys `core`, `lib`, ...) and this should still be less verbose than the old one where I have to manually override every `coreMod` `libMod` etc.
> 
> For upstream mill I think we don't need anything more from them, the only issue we had was the nested project stuff anyways and that's taken care of nicely now.

@KireinaHoro Thanks, I have resolved almost all the thing, but there is only one thing I have no good idea to to do: the object "idslplugin" has a `pluginOptions` function, which provides plugin options and is required to use a precise type to invoke this function, 
I now think it is workable to just add a new `pluginOptions` function to "SpinalModuleRef" hold this options, is this OK for you? (By the way, it is necessary to hold an object or a case class to group all the target if you do not want to import the separate ones, since the mill does not allow us to import trait directly).

KireinaHoro:
> > @inochisa yup you got the idea, thanks a lot. I think passing a `ModuleRef` should still work, I don't think that has changed too much with the new mill release (support was implemented upstream [here](https://github.com/com-lihaoyi/mill/issues/3715)). IMO you can maybe pass/override a `def moduleRefs: Map[String, ModuleRef]` (for example with keys `core`, `lib`, ...) and this should still be less verbose than the old one where I have to manually override every `coreMod` `libMod` etc.
> > For upstream mill I think we don't need anything more from them, the only issue we had was the nested project stuff anyways and that's taken care of nicely now.
> 
> @KireinaHoro Thanks, I have resolved almost all the thing, but there is only one thing I have no good idea to to do: the object "idslplugin" has a `pluginOptions` function, which provides plugin options and is required to use a precise type to invoke this function, I now think it is workable to just add a new `pluginOptions` function to "SpinalModuleRef" hold this options, is this OK for you? (By the way, it is necessary to hold an object or a case class to group all the target if you do not want to import the separate ones, since the mill does not allow us to import trait directly).

Are you keeping the `SpinalModuleRef` still?  So long as you have the named type, there's gonna be aliasing and you won't be able to override it.  You could potentially just replace the `case class` constructor with a package-level `def` that returns a map...

I think we could really just go back to the base types and add two values to override: one `Map[String, ModuleRef]` for `moduleDeps`, and one `Seq[String]` for `scalacOptions`.  This way they can be replaced rather easily, no?

inochisa:
> > > @inochisa yup you got the idea, thanks a lot. I think passing a `ModuleRef` should still work, I don't think that has changed too much with the new mill release (support was implemented upstream [here](https://github.com/com-lihaoyi/mill/issues/3715)). IMO you can maybe pass/override a `def moduleRefs: Map[String, ModuleRef]` (for example with keys `core`, `lib`, ...) and this should still be less verbose than the old one where I have to manually override every `coreMod` `libMod` etc.
> > > For upstream mill I think we don't need anything more from them, the only issue we had was the nested project stuff anyways and that's taken care of nicely now.
> > 
> > 
> > @KireinaHoro Thanks, I have resolved almost all the thing, but there is only one thing I have no good idea to to do: the object "idslplugin" has a `pluginOptions` function, which provides plugin options and is required to use a precise type to invoke this function, I now think it is workable to just add a new `pluginOptions` function to "SpinalModuleRef" hold this options, is this OK for you? (By the way, it is necessary to hold an object or a case class to group all the target if you do not want to import the separate ones, since the mill does not allow us to import trait directly).
> 
> Are you keeping the `SpinalModuleRef` still? So long as you have the named type, there's gonna be aliasing and you won't be able to override it.
> 

Yes, but I switch to apply function of object, so it can return a map function.

> I think we could really just go back to the base types and add two values to override: one `Map[String, ModuleRef]` for `moduleDeps`, and one `Seq[String]` for `scalacOptions`. This way they can be replaced rather easily, no?

This is what I have done, I have provide something like this:
```
object SpinalModuleRef {
  def apply(crossScalaVersion: String): Map[String, ModuleRef[PublishModule]] = Map(
    "core"        -> ModuleRef(core(crossScalaVersion)),
    "lib"         -> ModuleRef(lib(crossScalaVersion)),
    "sim"         -> ModuleRef(sim(crossScalaVersion)),
    "idslpayload" -> ModuleRef(idslpayload(crossScalaVersion)),
    "idslplugin"  -> ModuleRef(idslplugin(crossScalaVersion)),
    "tester"      -> ModuleRef(tester(crossScalaVersion))
  )

  def pluginOptions(crossScalaVersion: String) = idslplugin(crossScalaVersion).pluginOptions
}
```
It is simple to rename the `apply` to `moduleDeps`.

I have now encounter halt when resolving target, I will push a new version once I have resolve this.

KireinaHoro:
uh that does look neat, and then I assume in every trait you'd have a def that calls this function, and then the consumer should override that def in every trait?  Well two defs, one for your `SpinalModuleRef.apply` and one for `.pluginOptions`; they then get overridden

KireinaHoro:
I just had a slightly better idea, we actually can't import traits from the sub project anymore -- so how about in the apply function, you pass the function itself to the cross module as a separate cross axis?  And you'd take that function from the `crossValue` in the trait in `spinalhdl`.  Also to avoid having to override two functions, just make the function return a tuple `(Map[..], Seq[String])`

inochisa:
I just reference this in the base type `SpinalModule`.

```
  def moduleRefs = SpinalModuleRef.moduleDeps(crossScalaVersion)
  def modulePluginOptions = SpinalModuleRef.scalacOptions(crossScalaVersion)
```

and all the target can directly call this.

Finally I drop the `ModuleRef` and record the `PublishModule` directly, as it is to weird to write something like this: `moduleRefs("core")()`

I also update the build script in my `misc-ip` repo.

inochisa:
> I just had a slightly better idea, we actually can't import traits from the sub project anymore -- so how about in the apply function, you pass the function itself to the cross module as a separate cross axis? And you'd take that function from the `crossValue` in the trait in `spinalhdl`. Also to avoid having to override two functions, just make the function return a tuple `(Map[..], Seq[String])`

I think you mean just group this two functions and return a tuple? So we just use a apply function?

inochisa:
@KireinaHoro Update, I have encounter a Circular task dependency problem, I think this maybe a hard thing if these two are grouped.
Because the `idslplugin(crossScalaVersion).pluginOptions` is a Task, which is de-referenced if it is merge to the `moduleRefs`, and this will cause a Circular dependency. Any suggestions?

KireinaHoro:
> I just reference this in the base type `SpinalModule`.
> 
> ```
>   def moduleRefs = SpinalModuleRef.moduleDeps(crossScalaVersion)
>   def modulePluginOptions = SpinalModuleRef.scalacOptions(crossScalaVersion)
> ```
> 
> and all the target can directly call this.

Yes they can call this, but you need to offer the consumer a way to _override_ this to their own version that returns modules in their tree, otherwise you end up with two copies of `core`.  The way you have right now, there's no way for the consumer to override this `def moduleRefs` and `def modulePluginOptions`.

That's why I suggested you should pass them in as another Cross axis:

```scala
trait SpinalModule extends SbtModule with CrossSbtModule with Cross.Module3[String, Map[String, PublishModule], T[Seq[String]]] { outer =>
  // access here crossValue2 and crossValue3

object SpinalModuleRef {
def apply(crossVersion: String): (Map[String, PublishModule], T[Seq[String]]) = {
  val rm = mutable.Map[String, PublishModule]()
  val opTask = idslplugin(crossVersion).scalacOptions
  
  rm += "core" -> core(crossVersion, rm, opTask)
  rm += "lib" -> lib(crossVersion, rm, opTask)
  ...

  (rm, opTask)
}
```

I think using a `T[Seq[String]]` might have a chance solving the issue.  Sorry I'm asking for a lot, but would you give it a crack?

inochisa:
> > I just reference this in the base type `SpinalModule`.
> > ```
> >   def moduleRefs = SpinalModuleRef.moduleDeps(crossScalaVersion)
> >   def modulePluginOptions = SpinalModuleRef.scalacOptions(crossScalaVersion)
> > ```
> >   
> > and all the target can directly call this.
> 
> Yes they can call this, but you need to offer the consumer a way to _override_ this to their own version that returns modules in their tree, otherwise you end up with two copies of `core`. The way you have right now, there's no way for the consumer to override this `def moduleRefs` and `def modulePluginOptions`.
> 

I can not understand what you mean about the "offer the consumer a way to _override_ this to their own version", the consumer should select their own version of `SpinalModuleRef` to get everything it needed. And these two function are just used internal and can not be accessed by the top build script. The only interface for the top build script is `SpinalModuleRef` and all the cross object.

Anyway, I find a way to merge them with a little magic:
```scala
object SpinalModuleInfo {
  def apply(crossScalaVersion: String) = (moduleDeps(crossScalaVersion), () => scalacOptions(crossScalaVersion))

  def moduleDeps(crossScalaVersion: String): Map[String, PublishModule] = Map(
    "core"        -> core(crossScalaVersion),
    "lib"         -> lib(crossScalaVersion),
    "sim"         -> sim(crossScalaVersion),
    "idslpayload" -> idslpayload(crossScalaVersion),
    "idslplugin"  -> idslplugin(crossScalaVersion),
    "tester"      -> tester(crossScalaVersion)
  )

  def scalacOptions(crossScalaVersion: String) = idslplugin(crossScalaVersion).pluginOptions
}
```
And so you can access this in the following way
```scala
  def moduleInfo = SpinalModuleInfo(crossScalaVersion)
  def moduleRefs = moduleInfo._1
  def modulePluginOptions = moduleInfo._2()

```


> That's why I suggested you should pass them in as another Cross axis:
> 
> ```scala
> trait SpinalModule extends SbtModule with CrossSbtModule with Cross.Module3[String, Map[String, PublishModule], T[Seq[String]]] { outer =>
>   // access here crossValue2 and crossValue3
> 
> object SpinalModuleRef {
> def apply(crossVersion: String): (Map[String, PublishModule], T[Seq[String]]) = {
>   val rm = mutable.Map[String, PublishModule]()
>   val opTask = idslplugin(crossVersion).scalacOptions
>   
>   rm += "core" -> core(crossVersion, rm, opTask)
>   rm += "lib" -> lib(crossVersion, rm, opTask)
>   ...
> 
>   (rm, opTask)
> }
> ```
> 
> I think using a `T[Seq[String]]` might have a chance solving the issue. Sorry I'm asking for a lot, but would you give it a crack?

Using this `Cross.Module3` does give us a hard thing to implement (and I suspect there may something more), This definition requires to define
```scala
ToSegments[scala.Tuple.Elem[Tuple3[String, mutable.Map[String, PublishModule], Task.Simple[Seq[String]]], 1]]
```
which is definitely a magic and I think no one wants to check it....

KireinaHoro:
Say a consumer depends on core and lib.  It calls SpinalModuleRef, gets core and lib and adds them to its moduleDeps.

But the lib it just pulled in also depends on core, so it pulls in core‘ — at least in the old mill this results in two separate (but identical) core being compiled.  

What we need to do is to override lib‘s dependency and pass the outer core to replace the inner core‘.  This also solves the diamond dependency problem where you have aliasing and can end up with three copies of core.

This is why the old codebase has these stuff possible to override in the base trait (and consumers actually do).  Can you show that you have only one copy of core?  The new mill might have solved the first case by some module dedup, but the second case is definitely still there since two copies of core are from different trees and just can’t be merged.

inochisa:
> Say a consumer depends on core and lib. It calls SpinalModuleRef, gets core and lib and adds them to its moduleDeps.
> 
> But the lib it just pulled in also depends on core, so it pulls in core‘ — at least in the old mill this results in two separate (but identical) core being compiled.
> 
> What we need to do is to override lib‘s dependency and pass the outer core to replace the inner core‘. This also solves the diamond dependency problem where you have aliasing and can end up with three copies of core.
> 
> This is why the old codebase has these stuff possible to override in the base trait (and consumers actually do). Can you show that you have only one copy of core? The new mill might have solved the first case by some module dedup, but the second case is definitely still there since two copies of core are from different trees and just can’t be merged.

I have checked you code base and found the override is meaningless, the override in the `spinal-block` repo has point to the dependency in the `spinal-block` repo, and the override in the you final project has point to the final project repo. As a result, there are two copies and I do not think the usage for the override is meaningful.

IIRC there is almost no way for diamond dependency for source code dependency (not only limited to mill, but also sbt, gradle, maven). IIRC The only ways is to re-design your project and make the dependency flat. Take your case as an example, merge the `spinal-block` repo into the final project as a subproject, and adapt the project as a subproject instead of a standalone one. So the `spinal-block` can get the SpinalHDL for the whole project. Or just let you final project depends on the SpinalHDL `spinal-block` from In my option, a normal project should only focus on itself and its dependency, not the project it is used.

If you can use the the same Map from `SpinalModuleInfo`, I am sure there is only one copy, otherwise there is no guarantee.  

In fact, I think it is almost impossible to use complex type like `Map` and `Task` for the `Cross.Module`, not only it is hard to transform these type into a meaningful segment, but also may suffer recursive dependency as the Map is de-reference in the parsing stage.

If you do have a better way, just raise an alternative PR and I will drop mine. Thanks.

KireinaHoro:
> I have checked you code base and found the override is meaningless, the override in the spinal-block repo has point to the dependency in the spinal-block repo, and the override in the you final project has point to the final project repo.

That’s the whole point!  You want to override the original dependency on core to use the core provided by the consumer.  If you don’t do this, you will end up with two copies of core.  I don’t see what’s so difficult here…

> As a result, there are two copies and I do not think the usage for the override is meaningful.

I don’t see how you drew this conclusion.  The PR I gave earlier (on mill 0.11) removes the extra copy(-ies) — you can check in out/.  

This has taken so many rounds.   Sure the Cross stuff are just wild guesses, and it’s ok if they just won’t work; but the module duplication is a dealbreaker — I can’t give my approval as is.

inochisa:
> > I have checked you code base and found the override is meaningless, the override in the spinal-block repo has point to the dependency in the spinal-block repo, and the override in the you final project has point to the final project repo.
> 
> That’s the whole point! You want to override the original dependency on core to use the core provided by the consumer. If you don’t do this, you will end up with two copies of core. I don’t see what’s so difficult here…
> 
> > As a result, there are two copies and I do not think the usage for the override is meaningful.
> 
> I don’t see how you drew this conclusion. The PR I gave earlier (on mill 0.11) removes the extra copy(-ies) — you can check in out/.

By checking https://github.com/KireinaHoro/enzian-pio-nic/blob/8a4b7d9ee74c544dc40c2b3d3287616a350ae507/build.sc#L35-L41

I think I know how the override works, you defined a new object that setups all parameter with new dependency instead of using the predefined object in your subproject. Right? 

If this is right, using a new object with right dependency does reduce the copies and can avoid some dependency problem when using predefined object target directly, so the problem can be convert as the follows: *How can we simplify the object creation without importing trait?* I think it is the thing we should explore, instead of arguing about how to provide some override things, right?

> 
> This has taken so many rounds. Sure the Cross stuff are just wild guesses, and it’s ok if they just won’t work; but the module duplication is a dealbreaker — I can’t give my approval as is.

KireinaHoro:
We are taking over each other; let me take a step back.

> how to provide some override things

I feel like the term „override“ is overloaded here — at a higher level it means to replace the dependencies one way or another; it could also just mean the Scala keyword.

Let’s resync on what we agree on (an what we don’t):

- the goal is to keep the old behavior: for an intermediate IP project and a final consumer (the final design) project, there should be only one copy of everything (core, lib, etc.) of SpinalHDL in the entire build tree
- to do this, we need to somehow pass not only Scala version, but also the list of SpinalHDL components to the construction of each component (allow passing core to lib)
- The old codebase imported the trait directly and `override def`‘ed these dependencies
  - this is not possible any more in Mill 1.1
- Cross.Module3 is a potential way to do this but it looks like it’s too complicated

If you agree with the above, then we can think of more ways to pass the SpinalHDL components to the construction process.  If you don’t, I can defend them separately :)

inochisa:
> We are taking over each other; let me take a step back.
> 
> > how to provide some override things
> 
> I feel like the term „override“ is overloaded here — at a higher level it means to replace the dependencies one way or another; it could also just mean the Scala keyword.
> 

Yes, in my view it is more like scala keyword, and I agree with a mechanism for replacing the dependencies, or at least, providing a simple way to construct a object for nested sources in the way allowed by the mill.

> Let’s resync on what we agree on (an what we don’t):
> 
>     * the goal is to keep the old behavior: for an intermediate IP project and a final consumer (the final design) project, there should be only one copy of everything (core, lib, etc.) of SpinalHDL in the entire build tree
> 
>     * to do this, we need to somehow pass not only Scala version, but also the list of SpinalHDL components to the construction of each component (allow passing core to lib)
> 
>     * The old codebase imported the trait directly and `override def`‘ed these dependencies
>       
>       * this is not possible any more in Mill 1.1
> 
>     * Cross.Module3 is a potential way to do this but it looks like it’s too complicated
> 
> 
> If you agree with the above, then we can think of more ways to pass the SpinalHDL components to the construction process. If you don’t, I can defend them separately :)

The only thing I want you to clarify is the point 2, the construction is at the top build script, or in the subproject (in this scenario it is more like something predefined), I think this may be the thing needed to check.

For the others, I couldn't agree more. And you do not need to defend them :)

KireinaHoro:
> The only thing I want you to clarify is the point 2, the construction is at the top build script, or in the subproject (in this scenario it is more like something predefined), I think this may be the thing needed to check.

Yeah here I'm referring to both -- basically anywhere that would result in the construction of a `PublishModule` for every SpinalHDL component (core, lib, tester, ...).  There are a couple cases:

- construction in `build.mill` in SpinalHDL itself
  - here the dependencies should be allowed to just be the other modules defined in the same file.  This case is useful when for example you're running a included test case
- construction in `build.mill` in a consumer project; this can either be a _final_ consumer (like enzian-pio-nic) or a _intermediate_ consumer (like spinal-blocks)
  - for both cases, the consumer will construct the SpinalHDL components in its own `build.mill`
  - for the _intermediate_ consumer, it needs to expose some other mechanism (e.g. the `BlocksModule` trait in the old version) to allow the _final_ consumer to also replace the SpinalHDL components, both in constructing the intermediate consumer itself, and its SpinalHDL dependencies

The mechanism we come up with should accommodate both cases -- in case 1 the default defined in SpinalHDL's `build.mill` (which refers to the deps locally) should work already.  Importing the trait and mixing in `MySpinal` (the old way) satisfies this.

Right now I see two alternatives with the new Mill, both keep using the trait overriding method:
- Instead of asking the consumer to import the trait and override it themselves, define functions that do the trait overriding in `build.mill` in SpinalHDL
  - this is closest to how your PR is right now, just need to bring back the trait overriding in the `SpinalModuleInfo.apply`
- Split `build.mill` up into a sub package.  Importing traits from subprojects is possible, just not directly from `build.mill`

Feel free to propose other methods as well :)

build.mill
```diff
@@ -0,0 +1,150 @@
...
+  }
+  def testOnly(args: String*) = test.testOnly(args*)
+
+  def moduleInfo = SpinalModuleInfo(crossScalaVersion)
+  def moduleRefs = moduleInfo._1
+  def modulePluginOptions = moduleInfo._2()
+}
+
+object SpinalModuleInfo {
+  def apply(crossScalaVersion: String) = (moduleDeps(crossScalaVersion), () => scalacOptions(crossScalaVersion))
```
KireinaHoro, line 34: Might be slightly more idiomatic Mill to return a Task.Simple or Task.Anon instead of a lambda function -- not entirely sure if that works though...

inochisa: Using `Task.Simple` or `Task.Anon` type will  cause a circular dependency. Using a function pointer is the only way I found to avoid this circular at least for now.

inochisa:
> > The only thing I want you to clarify is the point 2, the construction is at the top build script, or in the subproject (in this scenario it is more like something predefined), I think this may be the thing needed to check.
> 
> Yeah here I'm referring to both -- basically anywhere that would result in the construction of a `PublishModule` for every SpinalHDL component (core, lib, tester, ...). There are a couple cases:
> 
>     * construction in `build.mill` in SpinalHDL itself
>       
>       * here the dependencies should be allowed to just be the other modules defined in the same file.  This case is useful when for example you're running a included test case
> 
>     * construction in `build.mill` in a consumer project; this can either be a _final_ consumer (like enzian-pio-nic) or a _intermediate_ consumer (like spinal-blocks)
>       
>       * for both cases, the consumer will construct the SpinalHDL components in its own `build.mill`
>       * for the _intermediate_ consumer, it needs to expose some other mechanism (e.g. the `BlocksModule` trait in the old version) to allow the _final_ consumer to also replace the SpinalHDL components, both in constructing the intermediate consumer itself, and its SpinalHDL dependencies
> 
> 
> The mechanism we come up with should accommodate both cases -- in case 1 the default defined in SpinalHDL's `build.mill` (which refers to the deps locally) should work already. Importing the trait and mixing in `MySpinal` (the old way) satisfies this.
> 

Great, it is very detailed and I agree with this.

> Right now I see two alternatives with the new Mill, both keep using the trait overriding method:
> 
>     * Instead of asking the consumer to import the trait and override it themselves, define functions that do the trait overriding in `build.mill` in SpinalHDL
>       
>       * this is closest to how your PR is right now, just need to bring back the trait overriding in the `SpinalModuleInfo.apply`
> 
>     * Split `build.mill` up into a sub package.  Importing traits from subprojects is possible, just not directly from `build.mill`
> 
> 
> Feel free to propose other methods as well :)

I think the second one worth a try, the first one is too hard to change the dependency according to the current situation. Anyway, I will take a look again to see if there is a way. Thanks.

KireinaHoro:
Yeah great, sorry for being a bit grumpy earlier, I'm on a plane and only just took my laptop out...  Thanks!

inochisa:
@KireinaHoro 
According to https://mill-build.org/mill/large/multi-file-builds.html#_limitation
I found the trait can be put in a separate `module.mill`, can you check it is possible to use this? I guess the only thing may be the source path problem. Other thing are fine.

KireinaHoro:
@inochisa can you also push to your `misc-ips` repo?  I am a bit lost on how you intend the consumer to use these modules -- should they:

- copy the definition of `SpinalModuleInfo` into their own (so that all `core`, `lib` point to what they define below), and
- define a bunch of `object` that extend these traits with a mixin that overrides `def moduleInfo` to use the defined object above?

I have a feeling this could be streamlined a bit...

inochisa:
> @inochisa can you also push to your `misc-ips` repo? I am a bit lost on how you intend the consumer to use these modules -- should they:
> 
>     * copy the definition of `SpinalModuleInfo` into their own (so that all `core`, `lib` point to what they define below), and
> 
>     * define a bunch of `object` that extend these traits with a mixin that overrides `def moduleInfo` to use the defined object above?
> 
> 
> I have a feeling this could be streamlined a bit...

In fact I found it is simple to use this, and no need to change the SpinalHDL repo.

Take the following as an example, We have two projects: A and B, A depends on B, and both of them depends on SpinalHDL, So the module structure is like this:
```
A/
A/deps/SpinalHDL
A/deps/B
A/deps/B/deps/SpinalHDL
```
All the directories has its own `build.mill` and other necessary files.

So let do something like
```scala
// A/deps/B/module.mill

trait XModule extends ... {
  def moduleInfo: Tuple2[Map[String, PublishModule], () => T[Seq[String]]]
  def moduleRefs = moduleInfo._1
  def modulePluginOptions = moduleInfo._2()

  // other definitions
}

// real one
trait XX extends XModule 
```

```scala
// A/deps/B/build.mill

trait XXXModule { this: XModule 
  def moduleInfo = build.deps.SpinalHDL.SpinalModuleInfo (xxx)
}

object xx extends XX with XXXModule 
```

So you can make a target `xx` in the B repo.

For the top one

```scala
// A/build.mill

trait XXXModule { this: build.deps.B.XModule 
  // this means you override the module for the build.deps.B.XX and redirect it to the A/deps/SpinalHDL
  override def moduleInfo = build.deps.SpinalHDL.SpinalModuleInfo (xxx) 
}

object xx extends build.deps.B.XX with XXXModule 

// other objects require xx
```

This should work since the mill allow us to import trait from helper files. The only thing is you may need to check is the `moduleDir` for source path (But I have no time to test today, I have spilled water on the my table, I had to shutdown down my computer so I can clean them....)

With this example, I think now we only need to expose `SpinalModuleInfo `, because the SpinalHDL requires no git dependency.

KireinaHoro:
Ok looks like this would properly replace the spinalHDL dependencies in the intermediate (B) — but you sure about this wont result in two copies of core?  You’re not replacing the core deps in SpinalHDL that B pulled in.  

I’ll experiment a bit as well and tell you later.

inochisa:
> Ok looks like this would properly replace the spinalHDL dependencies in the intermediate (B) — but you sure about this wont result in two copies of core? You’re not replacing the core deps in SpinalHDL that B pulled in.
> 
> I’ll experiment a bit as well and tell you later.

I am sure this will be only one copy if you only use `moduleInfo`. In repo A there is a dedicated `xx` object to reference the source code in repo B, I got this idea from you repo at here: https://github.com/KireinaHoro/enzian-pio-nic/blob/8a4b7d9ee74c544dc40c2b3d3287616a350ae507/build.sc#L35-L41. Correct me if I understand incorrectly.

KireinaHoro:
Looks like I got it to work:

- I merged your `module.mill` back into `build.mill` and used a type alias https://github.com/KireinaHoro/SpinalHDL/commit/cc1226ccae0734407ecb9f8f7ad7e06d52fcf854
- the blocks repo allow overriding the `BlocksModuleInfo` tuple, which contains both all Spinal modules and the blocks module itself (so that the blocks tester can have its dependency swapped out as well) https://github.com/KireinaHoro/spinal-blocks/blob/c49b7255185cb9ebde1b6127f1013b179be97850/module.mill#L13
- finally the top level overrides the BlocksModuleInfo in the blocks traits
  - here I had to override moduleDir as well, since it looks like moduleDir is resolved based on where the final `object` is defined

This might look a bit ugly still.  Feel free to cherry pick my commit to your PR branch -- I think then it should be ready from my perspective.  I should then be able to dogfood it and fix any other minor stuff as they come up.

Thanks a lot for your effort!

inochisa:
> Looks like I got it to work:
> 
>     * I merged your `module.mill` back into `build.mill` and used a type alias [KireinaHoro@cc1226c](https://github.com/KireinaHoro/SpinalHDL/commit/cc1226ccae0734407ecb9f8f7ad7e06d52fcf854)
> 

I can just to rollback the last commit and then add the types, it is OK for me.

>     * the blocks repo allow overriding the `BlocksModuleInfo` tuple, which contains both all Spinal modules and the blocks module itself (so that the blocks tester can have its dependency swapped out as well) https://github.com/KireinaHoro/spinal-blocks/blob/c49b7255185cb9ebde1b6127f1013b179be97850/module.mill#L13
> 
>     * finally the top level overrides the BlocksModuleInfo in the blocks traits
>       
>       * here I had to override moduleDir as well, since it looks like moduleDir is resolved based on where the final `object` is defined
> 
> 

That's right. It is the thing I want to tell.

> This might look a bit ugly still. Feel free to cherry pick my commit to your PR branch -- I think then it should be ready from my perspective. I should then be able to dogfood it and fix any other minor stuff as they come up.
> 
> Thanks a lot for your effort!

I think it is kind of weird too, IIRC, using git dependency is not a recommended way. I will also update the README so anyone can match the lasted one. Thanks.

KireinaHoro:
Yeah I agree it’s mainly for people that has changes downstream.  I’ll spend some time in the future to upstream / drop the stuff I have…

inochisa:
@KireinaHoro 
Update: 
1. I have rename the alias type to `SpinalModuleInfoType` to make it straightforward. and the previous module.mill is merged by revoking the last commit.
2. I also add an example to the README for ones who want to use new mill, but I leave the gap of using with source code since I think it is the thing for experts.

Do you think it is good to go?

KireinaHoro:
LGTM.  Thanks a lot!

inochisa:
@KireinaHoro Great, could you approval this PR so the @Dolu1990 can merge this?

inochisa:
Oh, a small force push to fix a typo in the comment.

KireinaHoro:
I don’t have write perms and on GitHub I don’t need to click the button (can’t find it on the phone as well…); @Dolu1990 you can go ahead.

Dolu1990:
gg ^^

PR #1865, opened 2026-01-02
tilelink: Add M2sTransfers for GET only

Tectu:
This adds a `getOnly()` convenience function similar to the existing `allGetPut()` to `tilelink.M2sTransfers`.

Question: Should we name it `onlyGet()` instead for consistency (and in case we'd ever have onlyPut()` for whatever reason)?


Dolu1990:
The allXxxx was refering to the SizeRange, not the kind of transfers.
So, onlyXxxx or xxxOnly isn't right i think.

Best would be to just have 
- def get(range : SizeRange)
- def putFull(range : SizeRange)
- def putPartial(range : SizeRange)
- ...

?

PR #1866, opened 2026-01-03
Fix Fragment.reduce accumulator not initialized

fayalalebrun:

I was using Fragment.reduce in one of my designs but found a bug with it. The accumulator register is not initialized, so the first time reduction is done the output will be garbage.

I also added a unit test, and checked it failed before the fix while working after the fix..

# Impact on code generation

# Checklist

- [x] Unit tests were added


Dolu1990:
Nice :D

PR #1867, opened 2026-01-03
Add clear input signal to StreamShiftChain

fayalalebrun:

I was using StreamShiftChain in order to implement a sliding window on a Fragment-based stream. In that case, you want to clear out the shift chain after the last fragment comes into it. Since otherwise you will be mixing elements from two different packets.

Without this, you would need to wait for the shift chain to be flushed out, which could be a serious cost if packets are relatively small, and also requires more logic.

# Impact on code generation

# Checklist

- [x] Unit tests were added


Dolu1990:
Thanks ^^

PR #1870, opened 2026-01-18
Add Bcd data type

fayalalebrun:

I implemented this data type a while ago and think others might find it useful. This is a binary-coded decimal implementation, with methods to transform from and into UInt. A few other operators like addition and shifting are also implemented.

# Impact on code generation

# Checklist

- [x] Unit tests were added


Dolu1990:
Hi,

I would feel more confortable with it being in the spinal.lib, as in some ways it is kinda a specific feature.

Maybe there : https://github.com/SpinalHDL/SpinalHDL/tree/dev/lib/src/main/scala/spinal/lib/math ?

fayalalebrun:
@Dolu1990 Makes sense, I moved it over.

Dolu1990:
Thanks :D

PR #1872, opened 2026-01-24
Enhance the plugin system

Aqwyh123:

All build phases predefined in spinal.core.fiber are allowed.
This should be able to reduce the use of explicit locks.
Example:
1. In the patch phase, pipeline.build() can be called to simplify the negotiation of the calling timing.
2. In the check phase, reports can be printed to check whether the negotiation was successful.

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/misc/plugin/Fiber.scala
+++ b/lib/src/main/scala/spinal/lib/misc/plugin/Fiber.scala
@@ -94,5 +94,27 @@ class FiberPlugin extends Area with Hostable {
         }
       }
     }
+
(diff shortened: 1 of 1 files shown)


Aqwyh123:
I don't understand why simply adding methods that have never been called caused the test to fail. This seems to be related to the comparability of the object.
The failure occurred in the ram2 test of the spinal.lib.bus.tilelink.InterconnectTester module.
Exception: java.lang.IllegalArgumentException: The comparison method violates its general contract

Dolu1990:
Ahhh that is a weird one XD

I'm restarting the CI to see.

PR #1873, opened 2026-01-26
feat: optionally include the call-site location (file and line) in the report output

### LINKED ISSUE #1737: Feature Request: Add Source Location to `report` for Better IDE Navigation
Hi SpinalHDL Developers,

I propose enhancing the `report` function to include source location (file path and line number) in its output.

**Motivation** Modern IDEs allow clicking terminal logs to jump to source code. Adding source location to `report` output enables this, significantly improving debugging by linking reported messages directly to their origin.

**Feasibility** I've confirmed this is feasible and implemented it on a local branch.

Image

**Non-Breaking Integration Discussion**

I'd like to discuss how to integrate this without breaking users' existing code. Here are options:

1.  **Optional Parameter:** Add `includeSourceLocation: Boolean = false` to `report` (e.g., `report(msg, includeSourceLocation = true)`).
2.  **Global Configuration:** Introduce a `SpinalConfig.reportIncludeSourceLocation: Boolean` setting.
3.  **Context Object:** Utilize a context object (similar to Scala's `println` interception) to implicitly provide source location when `report` is called within a specific context. This could be a powerful and clean way to inject the information.
4. **Context Logger:** We can even wrap it as a logger to support different logging level and color options.

We should also agree on a standard output format (e.g., `file_path:line_number: message`) for IDE compatibility.

Looking forward to your thoughts on the best approach.

Thanks,

Zijing Zhang

pluveto:
Closes #1737

This PR adds an opt-in way to include the Scala call-site source location (file + line) in `report(...)` messages, so IDEs/terminals can hyperlink simulation logs and jump directly to the code.

Two non-breaking opt-in mechanisms are provided:
- Per-call: `report("msg", includeSourceLocation = true)` (also available on the `severity` overloads)
- Global: `SpinalConfig(reportIncludeSourceLocation = true)` to apply to all `report(...)` calls during elaboration

When enabled, the emitted simulation print string is prefixed by a configurable format string.
The default format is: `SEVERITY($FILE:$LINE) `, e.g. `NOTE(path/to/File.scala:123) `.

Users can override the format globally via `SpinalConfig.reportSourceLocationFormat` or the `SPINAL_REPORT_SOURCE_LOCATION_FORMAT` environment variable.

A regression test is included to ensure the prefix is present only when requested and that the location points to the call site (not to `core.scala`).

image

# Impact on code generation

- Verilog/SystemVerilog: when enabled, the simulation print format string is prefixed with `reportSourceLocationFormat`; when disabled, generated code is unchanged.
- VHDL: when enabled, the `report(...)` message expression is prefixed with `reportSourceLocationFormat`; when disabled, generated code is unchanged.
- This feature only affects simulation-time messages; synthesis output is unchanged.

# Checklist

- [x] Unit tests were added
- [x] API changes are documented [here](https://github.com/SpinalHDL/SpinalDoc-RTD/pull/285)

### DIFF
--- a/core/src/main/scala/spinal/core/Spinal.scala
+++ b/core/src/main/scala/spinal/core/Spinal.scala
@@ -220,6 +220,7 @@ case class SpinalConfig(mode                           : SpinalMode = null,
                         var printFilelist              : Boolean = true,
                         var emitFullComponentBindings  : Boolean = true,
                         var reportIncludeSourceLocation: Boolean = false,
+                        var reportSourceLocationFormat : String = SpinalConfig.defaultReportSourceLocationFormat,
                         var svInterface                : Boolean = false
 ){
   def generate       [T <: Component](gen: => T): SpinalReport[T] = Spinal(this)(gen)
@@ -326,6 +327,8 @@ object SpinalConfig{
   }
 
   var defaultTargetDirectory: String = System.getenv().getOrDefault("SPINAL_TARGET_DIR", ".")
+  var defaultReportSourceLocationFormat: String =
+    System.getenv().getOrDefault("SPINAL_REPORT_SOURCE_LOCATION_FORMAT", "$SEVERITY($FILE:$LINE) ")
 }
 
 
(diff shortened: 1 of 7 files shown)


Dolu1990:
Thanks :D

PR #1881, opened 2026-02-15
Add CounterOH

korbin:
This PR adds `CounterOH`, a one-hot encoded counter complementing the existing `Counter` and `CounterUpDown`

- This counter allows overflowing increment/decrement via bit rotation, avoiding adder/comparator logic
- Simultaneous increment + decrement cancels out (no-op), matching CounterUpDown semantics
- Supports load, clear, freeRun, init/reset, overflow/underflow detection, and comparison operators (===/=/=/!==) against Bits, Int, BigInt, and UInt
- All `Bits` mode operations are assumed to be one-hot encoded, `UInt`/`BigInt`/`Int` are converted implicitly.


Dolu1990:
Thanks ^^

PR #1885, opened 2026-03-14
Add method alternatives with spelling and typo fix

mrcmry:
This PR adds several method alternatives with corrected names. It also fix spelling errors in comments, local variables and error messages.

There is no change in the interface, and currently it doesn't emit any deprecation warning. The deprecation annotations are commented and ready for the next minor release:

```
  // TODO enable deprecation
  //@deprecated("Use correctly spelled 'hasAssignment' instead", since = "1.15.0")
  def hasAssignement : Boolean = hasAssignment

  def hasAssignment : Boolean = !this.dlcIsEmpty
```

I can set it to `1.14.2` and turn it on if it's a better. On a side note, I have seen that there is a 1.14.1 tag in git but no release on github, I don't know if it's normal.

# Impact on code generation

No change in code generation.

### DIFF
--- a/core/src/main/scala/spinal/core/AFix.scala
+++ b/core/src/main/scala/spinal/core/AFix.scala
@@ -425,7 +425,7 @@ class AFix(val maxRaw: BigInt, val minRaw: BigInt, val exp: Int) extends MultiDa
   }
 
   /**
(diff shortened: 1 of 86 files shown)


Dolu1990:
Thanks ^^

PR #1890, opened 2026-03-20
Add scaladoc for the FSM lib based on RTD doc

mrcmry:
I used the SpinalDoc-RTD to add scaladoc the the state machine library.

This way a developer can quickly access the doc when coding.

Still open but not blocking for PR:

* `StateMachineSlave` is not documented
* the method `innerFsm` is not used anywhere (at least in this repos code). It's a bit confusing, because some example use classes or variables using this name. Perhaps we should remote it or deprecate it.
* I added a scaladoc description as "used internally" for `getStateMachineAccessor` `whenIsActiveWithPriority`. Perhaps we should make them protected or private, but it could break things, I don't know

# Impact on code generation

Do not change the scala code an thus the generation as well.

### DIFF
--- a/lib/src/main/scala/spinal/lib/fsm/State.scala
+++ b/lib/src/main/scala/spinal/lib/fsm/State.scala
@@ -30,13 +30,16 @@ import spinal.core._
 import scala.collection.mutable.ArrayBuffer
 
 /**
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks ^^

PR #1891, opened 2026-03-22
Remove trivial warnings

mrcmry:
I'm a big fan of warning free production code. My opinion is when there is too much warnings you stop looking at them and they become distractions. But they are very useful to catch potential bugs at a very low dev cost.

So this PR is a first step to have a warning free `sbt compile`. I started with the trivial ones, where there is no effect on the rtl generation, the simulation behavior or user code compilation.

This can be merged as is.

For the next PR with more complicated warning fix, I have the following question: one warning that comes frequently is this one:

```
[warn] SpinalHDL/lib/src/main/scala/spinal/lib/memory/sdram/xdr/Xdr.scala:310:38: Adaptation of argument list by inserting () is deprecated: this is unlikely to be what you want.
[warn]         signature: Data.randBoot(u: Unit): Data.this.type
[warn]   given arguments: <none>
[warn]  after adaptation: Data.randBoot((): Unit)
[warn]       mapper.drive(ODTend, 0x34,  8) randBoot()
```

`randBoot` has `u : Unit` as argument. This is different compared to others chainable modifiers without arguments like `allowPruning()`, `setAsReg()`, etc. Is there a reason for this ?

If not, I propose to define a `def randBoot()` and make `def randBoot(u : Unit)` point to it with a deprecation. This way people what fixed the scala warning with randBoot(()) will still have a compiling code.

# Impact on code generation

No impact.

### DIFF
--- a/core/src/main/scala/spinal/core/BitVector.scala
+++ b/core/src/main/scala/spinal/core/BitVector.scala
@@ -124,7 +124,7 @@ abstract class BitVector extends BaseType with Widthable {
 
     var result = cloneOf(this).asInstanceOf[T]
     result := this.asInstanceOf[T]
-    for(i <- that.range){
+    for(i <- that.bitsRange) {
       result \= (that(i) ? result.rotateLeft(1 << i).asInstanceOf[T] | result)
     }
     result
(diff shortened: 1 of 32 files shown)


Dolu1990:
Thanks :D

Dolu1990:
> randBoot has u : Unit

This was likely added as a workaround to make it work across all scala versions.
Maybe allowing stuff like calling the function without the bracket    "randBoot" aswell as "randBoot()"
(my guess)

> I propose to define a def randBoot() and make def randBoot(u : Unit) point to it with a deprecation.

let's try. sound good to me.

PR #1893, opened 2026-03-26
Remove easy warnings

mrcmry:
This PR follows #1891 and fix the remaining warnings what should change nothing to the engine of spinalHDL or the generated HDL. 

As discussed in #1891 I added the deprecation of `def randBoot(u : Unit)` replaced by `def randBoot()`. The deprecation is commented and should be turned on before the next minor.

with this build config:
```diff
-  scalacOptions ++= Seq("-unchecked","-target:jvm-1.8"/*, "-feature" ,"-deprecation"*/),
+  scalacOptions ++= Seq("-unchecked","-target:jvm-1.8", "-deprecation", "-feature"),
```

After this PR I get 56 warnings after a `sbt "clean; compile"`. `v1.14.1` had 123 warnings, so it's already a good start.

I will pause my warning hunting campaign for now, I need to gain a better understanding of scala, hdl and the internals of spinal before fixing safely the others.

Perhaps turning warning on again in the `build.sbt` could help by having the developers of the lib fixing the warning of the code they know when they add new feature/fix bugs. But perhaps it create noise for users of the lib, I don't know, I don't really master the build system.

# Impact on code generation

None

### DIFF
--- a/core/src/main/scala/spinal/core/Data.scala
+++ b/core/src/main/scala/spinal/core/Data.scala
@@ -661,11 +661,16 @@ trait Data extends ContextUser with NameableByComponent with Assignable with Spi
     this
   }
 
+
+  // TODO enable deprecation
+  //@deprecated("use randBoot() instead", since = "1.15.0")
+  def randBoot(u : Unit): this.type = randBoot()
+  
   /**
     * Useful for register that doesn't need a reset value in RTL,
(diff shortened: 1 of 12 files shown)


Dolu1990:
Thanks ^^

Dolu1990:
About the -deprecation, sure it could be added

PR #1896, opened 2026-04-06
lib: blackboxes: Add PhaseNetlist for IHP SRAMs

dnltz:
IHP currently provides some SRAM macros. Add a PhaseNetlist class to automatically use those blackboxes, if the SRAM option exists.

It currently doesn't support the BIST interface and disables those ports by default.


Dolu1990:
Thanks ^^

PR #1897, opened 2026-04-06
lib: bus: WishboneSlaveFactory: Improve Error Handling

dnltz:
The Wisbone error signal (ERR) is currently not driven when the Wishbone parameter "useERR" is set. Extend the WishboneSlaveFactory to generate error signals under two scenarios:

Generate an error on access to unmapped registers. This is an optional feature and can be disabled via the "errorOnUnmapped" flag.

Additionally, the slave factory has a new function
  setError(condition: Bool)
to register error conditions. If one of those conditions is true, the slave factory will generate an error event. For example, a bus subordinate (slave) can generate an error when the FIFO is full.


Dolu1990:
thanks ^^

PR #1898, opened 2026-04-07
Add Companion trait for Bundles and MultiData types

cyndquill:
Just so they can be less lonely :)

Rationale: Imagine a circuit for a clock with 3 primary signals: an output clock, a feedback clock, and an enable line. In my mind, something like the following would be a nice way of organizing these signals:

```scala
val io = new Bundle {
  val clock = out Bool()
  val clock = new Bundle {
    val enable = out Bool()
    val feedback = in Bool()
  }
}
```

Obviously this doesn't work, but I found it appealing being able to treat `io.clock` as a separate "entity" from `io.clock._`. I *sort of* accomplished this via adding a new trait, `Companion`, that a `MultiData` type can extend. The only caveat of this feature being it overwrites the no-parameter `apply()` method. To use this in a bundle, simply do:

```scala
val io = new Bundle {
  val clock = new Bundle with Companion[Bool] {
    @companion val output = out Bool()
    val enable = out Bool()
    val feedback = in Bool()
  }
}
```

Then, calling the empty apply `io.clock()` will net you the value of `io.clock.output`, thereby (almost) replicating the original problem I presented!

# Another Example

With the Companion trait, something like the following is possible:

```scala
class CompanionStream[T <: Data](plType: HardType[T]) extends Stream(plType) with Companion[T] {
  @companion val plWhenFire = RegNextWhen(this.payload, this.fire)
}
```

Then, you may simply use `val streamData = compStream()`, instead of `... = RegNextWhen(compStream.payload...`. While this isn't the ideal usage of a `Stream`, this is just an example of how a `Companion` can be used (especially in larger projects, where the verbosity of complex Bundles and functions can get overwhelming).

# Impact on code generation

None to my knowledge, though I have considered adding the functionality of making the generated companion signal have no partial name (i.e., as in the first example, `io_clock_output` will be renamed to `io_clock`). This isn't exactly ideal, though, as considering my second example, it would make more sense to name it `compStream_plWhenFire`, as opposed to just `compStream`. Any suggestions on this would be appreciated (though, using `setPartialName("")` directly may just be the best solution).

# Other Miscellany

I mostly implemented this for my own use cases, but thought it would be worth at least opening a pull request for. I don't expect it to be upstreamed, but at least the exposure here could be beneficial if others deem this functionality suitable to their needs and use it in their own projects. There are some quirks with it, notably that the execution code doesn't happen at runtime, but rather whenever the `apply()` method is called. Also, nesting `Companion` bundles is weird and doesn't *exactly* work as I'd like (as it only provides `this.companion` and doesn't resolve to the lowest-level). Any feedback/suggestions/comments are welcome! :)

# Checklist

- [X] Unit tests were added
- [X] API changes are or will be documented:

*Scala docs are added to commited files, I can add RTD documentation should this feature be desired!

### DIFF
--- a/lib/src/main/java/spinal/lib/companion.java
+++ b/lib/src/main/java/spinal/lib/companion.java
@@ -1,0 +1,8 @@
+package spinal.lib;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface companion {
+}
--- a/lib/src/main/scala/spinal/lib/Companion.scala
+++ b/lib/src/main/scala/spinal/lib/Companion.scala
@@ -1,0 +1,56 @@
+package spinal.lib
+
+import spinal.core.{Data, MultiData, dontName}
+
+/** A trait that a `MultiData` can implement, to make it less lonely :) */
+trait Companion[T <: Data] {
+
+  /** Make sure we are implemented on a MultiData object only */
+  assert(this.isInstanceOf[MultiData], s"Error: Companion ${this} must be of Type spinal.core.MultiData")
+
+  /** The value to act as a companion
+    * Initialized as lazy so as to not fail during object creation
+    */
+  lazy private val companion = getCompanion()
+
+  /** Parse object to find the field with the companion attribute
+    * Ensure that there is one (and only one) field with said attribute
+    * @return The field in this with the companion attribute as an instance of [T <: Data]
+    */
+  private def getCompanion(): T = {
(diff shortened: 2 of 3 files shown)


Dolu1990:
Hi ^^
One question, in what sence this differs from :

```scala
val io = new Bundle {
  val clock = new Bundle {
    val output = out Bool()
    val enable = out Bool()
    val feedback = in Bool()
    
    def apply() = output
  }
}
```

?
Is the main usage of the variant you proposed to have a Companion trait to be able to pass it in function and hiding the Bundle type itself ?

PR #1899, opened 2026-04-07
lib: bus: WishboneDecoder: use byte address for slave selection

### LINKED ISSUE #1846: BmbToWishbone: Wishbone is missing LSBs
I'm migrating my MCU from AXI4 to BMB right now. So far great experience with the simplified command and response channels.
I'm replacing APB3 with Wishbone and using the BmbToWishbone bridge. I'm asking myself why, for example in a 32-bit context, the two lower bits are missing on the Wishbone bus?

https://github.com/SpinalHDL/SpinalHDL/blob/c1d2a76e9d1c62906f9c51785a9a294d6052e831/lib/src/main/scala/spinal/lib/bus/bmb/BmbToWishbone.scala#L37

I'm using the WishboneSlaveFactory with `factory.read(reg, 0x4)`, but this is wrong now sincethe bridge will convert `0x0004 -> 0x0001`.

So, curious about the reason behind this remapping.

dnltz:
Convert ADR to byte address before hitting AddressMapping, consistent with WishboneSlaveFactory. Removes the manual shift parameter which was error-prone and defaulted silently to 0.

Closes #1846

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/wishbone/WishboneDecoder.scala
+++ b/lib/src/main/scala/spinal/lib/bus/wishbone/WishboneDecoder.scala
@@ -12,15 +12,26 @@ object WishboneDecoder {
   * @param decodings it will use for configuring the partial address decoder
   * @return a [[spinal.lib.bus.wishbone.WishboneDecoder]] instance
   */
-  def apply(config: WishboneConfig, decodings: Seq[AddressMapping]) = new WishboneDecoder(config,decodings)
+  def apply(config: WishboneConfig, decodings: Seq[AddressMapping]) = new WishboneDecoder(config, decodings)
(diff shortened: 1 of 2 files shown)


dnltz:
@Dolu1990 can you please restart the Verilator test?

Dolu1990:
Restarted ^^

Dolu1990:
Hi,

hmmm this change the default behaviour right ? maybe the best then would be to add a parameter with no default value, where the user can specify if he want address in byte or word and so on ?

(to avoid silently breaking stuff in other people code)

dnltz:
> Hi,
> 
> hmmm this change the default behaviour right ? maybe the best then would be to add a parameter with no default value, where the user can specify if he want address in byte or word and so on ?
> 
> (to avoid silently breaking stuff in other people code)

Added a new argument and the core change of this PR is disabled by default to not break any existing designs.

dnltz:
@Dolu1990 gentle ping ;)

Dolu1990:
Thanks ^^

PR #1902, opened 2026-04-16
Add default for uninitialized register in jtag

jdavidberger:

JTag fails to work properly on some hardware based on startup conditions. 

I tested the exact same verilog with and without this change, and the change fixed my jtag issue. 

I think this is the result of overzealous optimization passes (specifically Lattice based Synplify) since the state here does fully cover its bitset. 

# Impact on code generation

Minimal. It adds a default statement to a fully covered switch statement.


Dolu1990:
Thanks ^^

PR #1903, opened 2026-04-20
Add JohnsonCounter (Twisted-Ring/Möbius counter) to Utils

korbin:
This PR adds `JohnsonCounter`, a twisted-ring (a.k.a. Möbius) counter: a shift register whose inverted MSB feeds back into the LSB, producing a `2*width`-state sequence with only one bit transition per cycle. This is useful for:

- multi-phase clock generation (e.g. quadrature from `width=2`)
- 50%-duty-cycle clock division by `2*width` for any width
- glitch-free / low-switching state sequencing with half the flops and transitions of a `CounterOH`

The counter is self-recovering from illegal (unreachable-from-reset) states: any state with top bits `10` snaps back to 0 on the next increment, and every illegal cycle visits such a state within at most `2*width - 1` increments.

The implementation follows the conventions of `Counter` / `CounterOH`:

- `apply(width)` / `apply(width, inc)` factories
- `willIncrement` / `willClear`, `increment()`, `clear()`
- `willOverflowIfInc` / `willOverflow` (marking the wrap back to 0)
- `freeRun()`
- `clkDiv` helper tapping a middle bit for phase-symmetric post-reset behavior

PR #1904, opened 2026-04-24
Refactor `Counter`

korbin:
A common theme in my SpinalHDL code over the years is that I love to use the abstract `Counter` where possible, but then symmetry and functionality breaks any time I need a down-only `Counter`. `CounterUpDown` didn't have the same minimal RTL elaboration profile/behavior/fast paths or helper methods.

This PR builds on https://github.com/SpinalHDL/SpinalHDL/pull/1903 (which can be closed or merged first) to refactor, unify, and test the various `Counter` implementations that exist in `SpinalHDL`.

**This PR should not break any user code or harm performance.** Changes are mostly organization and consistency cleanup. Existing APIs are maintained, just unified and extended to all `Counter`s: no functional changes occur in user code. Emitted RTL is functionally identical and should be mostly-literally identical. I added simulation and/or formal verification for a large number of cases: let me know if I've missed any cases and I can add them. 

I am happy to discuss, break out, or rework these changes.

All `Counter`-related code has moved into `lib/src/main/scala/spinal/lib/Counter.scala` from `Utils.scala`.

### Available Counters

- `Counter`: Binary sequential (adder-based) up/down/bidirectional counters
  - Implements `CounterLike`, `CounterAddressable` and `BoundedCounter`
  - `CounterUpDown` is absorbed by `Counter(Both)`
  - `DownCounter` helper is similarly added for `Counter(Down)`
  - Gained directional/saturation capability, more range/width/flexibility and API unification
- `GrayCounter`: Gray-coded up/down/bidirectional counters
  - Gained full class with directional capabilities, API uniformity
- `OneHotCounter` (formerly `CounterOH`): Binary one-hot up/down/bidirectional ring counters
- `JohnsonCounter`: Twisted-ring (Möbius) counter with 1 bit transition per clock
- `CounterMultiRequest`:  Function-style helper that composes a list of function transforms into a single register update. Exists outside the `Counter` trait/class hierarchy, unmodified by this PR.

### New Traits
- `CounterLike[T <: BitVector]`: the minimum common surface for anything that steps through a discrete state sequence  
  -  Provides `value/valueNext`, a canonical control-signal set (`willClear`, `willLoad`, `willAdvance`, `willComplete`), `stateCount`, and default implementations of `clear()`, `load(raw)`, `frozen`, `toFlow()`, and `implicitValue.`
- `CounterAddressable[T <: BitVector]`: mixes on top of `CounterLike[T]` to declare the cycle is accessible-by-index  (you can jump to the k-th state for any k ∈ [0, stateCount).
  - Provides `loadOrdinal(index: UInt)` with `Int` and `BigInt` overloads for free.
- `BoundedCounter[T <: BitVector]`: abstract class layering direction and per-side boundary policies on top of `CounterLike` + `CounterAddressable`. 
  - Takes `(direction, upper, lower)` and provides direction-gated `increment()/decrement()`, `freeRun()/freeRunDown()`, `saturatedHigh/saturatedLow`, a shared `Freeze`-latch, and `willAdvance/willComplete`.
  - `JohnsonCounter` doesn't implement this, and neither will future LSFR, random, or other primitive counter implementations.

### New Configuration
- `CounterDirection`: allows capable counters to support `Up`-counting, `Down`-counting, or `Both` (bidirectional) counting.
- `BoundaryPolicy`: allows capable counters to wrap (default behavior), saturate, or freeze (latching) on either side.

### Core instance API (`CounterLike[T]`)

| API | Before | After |
|---|---|---|
| `value`, `valueNext` | Legacy `GrayCounter` had no class at all | Present on every counter |
| `clear()` | `Counter`, `CounterOH` | Every counter (`CounterUpDown` gains it) |
| `willClear: Bool` | `Counter`, `CounterOH` | Every counter |
| `load(raw: T)` | `Counter.load(UInt)`; `CounterOH.load(Bits\|Int\|UInt)` | Every counter |
| `willLoad: Bool` | — | Every counter (complements `willClear`) |
| `loadOrdinal(UInt\|Int\|BigInt)` | — | Every `CounterAddressable`; encoding-native (`Counter`: `start+k`, `OneHotCounter`: `1<<k`, `GrayCounter`: `k ^ (k>>1)` with parity fixup) |
| `stateCount: BigInt` | `CounterOH`, `CounterUpDown` | Every counter |
| `willAdvance: Bool` | — | Every counter; encoding-agnostic "stepped this cycle" |
| `willComplete: Bool` | — | Every counter; encoding-agnostic "boundary event this cycle" (overflow or underflow depending on direction) |
| `toFlow(): Flow[T]` | `Counter`, `CounterOH` only (and `Counter.toFlow` was hardwired to `willIncrement`, so nonsensical for Down/Both) | Every counter; gated by `willAdvance` so it works in any direction |
| `frozen: Bool` | — | Every counter (defaults to `False`; overridden by the shared Freeze-latch) |
| `implicitValue: T` | All class-based counters; legacy `GrayCounter` was a raw `UInt` (no class) | Every counter; all are `ImplicitArea[T]` |

### Direction + boundary API (`BoundedCounter[T]`)

| API | Before | After |
|---|---|---|
| `direction: CounterDirection` (`Up`/`Down`/`Both`) | Implicit via type: `Counter` = Up, `CounterUpDown` = Both, `CounterOH` ≈ Both, `GrayCounter`/`JohnsonCounter` = Up-only | Explicit constructor param; `Counter`, `OneHotCounter`, `GrayCounter` all accept `Up`/`Down`/`Both` |
| `upper: BoundaryPolicy`, `lower: BoundaryPolicy` (`Wrap`/`Saturate`/`Freeze`) | `Counter` hardcoded Wrap; `CounterUpDown` had a binary `handleOverflow: Boolean` flag (Wrap-at-stateCount if pow2, else `assert(false)`) | Orthogonal per-side params on every `BoundedCounter`. `JohnsonCounter` accepts `upper` only (Wrap or Freeze; Saturate rejected — no sensible pinned state) |
| `increment()`, `decrement()` | `Counter` had `increment` only; `CounterOH`/`CounterUpDown` had both; no elaboration-time guards | Uniform; `require(hasUp)`/`require(hasDown)` so calling the wrong one is an elaboration-time error instead of a silent no-op |
| `willIncrement`, `willDecrement` | `CounterUpDown` called them `incrementIt` / `decrementIt`; others used `willIncrement` / `willDecrement` | Uniform names everywhere |
| `willOverflowIfInc`, `willUnderflowIfDec` | `Counter` had `willOverflowIfInc` only; `CounterOH` had both but baked direction-cancellation in (`value.msb && !willDecrement`); `CounterUpDown` called them `mayOverflow`/`mayUnderflow` and also baked cancellation in | Uniform names; pure positional predicates (cancellation handled once, centrally, in the derived `willOverflow`/`willUnderflow` via `Counter.guardedComplete`) |
| `willOverflow`, `willUnderflow` | Ad-hoc per type, mixed cancellation semantics | Uniform derivation across all `BoundedCounter`s; `JohnsonCounter` has `willOverflow` only |
| `freeRun()` | `Counter`, `CounterOH`; **missing** on `CounterUpDown`; legacy `GrayCounter` had no class | Every `BoundedCounter` + `JohnsonCounter` |
| `freeRunDown()` | — | Every `BoundedCounter` with `Down` or `Both` |
| `saturatedHigh`, `saturatedLow` | — | Every `BoundedCounter`; returns `False` when that side isn't `Saturate` |

### Other Changes
- Helper methods added to relevant Counter constructor objects for up/down/both construction
- All Counters now completely and correctly implement non-power-of-two paths where possible (before, assertions were raised in some counters in some cases due to unimplemented behavior), fast paths are maintained where previously existing or otherwise added where possible.

### (Potentially) Breaking Changes
- I renamed `CounterOH` to `OneHotCounter` to have the same naming scheme as `GrayCounter` and `JohnsonCounter`
  - This was from my previous PR and hasn't landed in a released version yet, so should be OK
- `Counter`'s class signature changes (adding direction/saturation policies/etc with default) which could potentially break code utilizing reflection - all static code should be OK.


mrcmry:
This is nice.

I also find the `Counter` component very useful and was in need in the past for a downcounter.

I think we should keep `CounterOH` as an alias to `OneHotCounter` with a mention in the scaladoc to use the new naming convention. I don't think it should be deprecated, because there is nothing wrong in the name. But we can replace it in the lib and examples so people looking for inspiration use the new name.

It would be nice to add scaladoc in the code, my impression as a newcomer in SpinalHDL is that I tend to not use documented interface and doc is as important as code, at least at the helper level (it's probably different for large IPs).

If there is scaladoc, I can help by doing PR to add or modify SpinalHDL-RTD and the Spinal-bootcamp.

korbin:
> I think we should keep `CounterOH` as an alias to `OneHotCounter` with a mention in the scaladoc to use the new naming convention. I don't think it should be deprecated, because there is nothing wrong in the name. But we can replace it in the lib and examples so people looking for inspiration use the new name.

I renamed it because it hasn't been in a released version yet (and I am not sure people know about it, I added it), but can put the alias back.

> If there is scaladoc, I can help by doing PR to add or modify SpinalHDL-RTD and the Spinal-bootcamp.

I will add some.

Dolu1990:
Thanks :D

PR #1905, opened 2026-05-05
Have StreamFork2/3 helper functions generate the full component for tracability

jdavidberger:

This changes how streamFork2 and streamFork3 work so that the create a full component in those helper classes. 

The motivation here is that the StreamFork area contains state, and there is no way to see that state with any kind of interospection with it just as the area. As a component, it's a registered child it participates in the component tree. 

This is probably useful for any tooling, but my specific goal here was to enable a formal testing of components that use [PipelinedMemoryBusArbiter](https://github.com/jdavidberger/SpinalHDLExtras/blob/main/hw/spinal/spinalextras/lib/formal/fillins/PipelinedMemoryBusArbiterFormal.scala#L12). The formal proof does not work though without: [this constraint](https://github.com/jdavidberger/SpinalHDLExtras/blob/main/hw/spinal/spinalextras/lib/formal/fillins/StreamForkFormal.scala#L10). 

# Impact on code generation

Minimal. It generates slightly more code, but the effective net list is the same.

### DIFF
--- a/lib/src/main/scala/spinal/lib/Stream.scala
+++ b/lib/src/main/scala/spinal/lib/Stream.scala
@@ -1297,10 +1297,11 @@ object StreamFork {
 }
 
 object StreamFork2 {
-  def apply[T <: Data](input: Stream[T], synchronous: Boolean = false): (Stream[T], Stream[T]) = new Composite(input, "fork2"){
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks ^^

PR #1906, opened 2026-05-09
flatten nested concatnation operator to make RTL more friendly both for human and EDA tools

facebreeze:

Currently, spinal emit concatenated bits in a nested style, which is unfriendly when there are lots of bit to be concatenated. 
```scala
case class VecAsBitsDemo() extends Component {
  val io = new Bundle {
    val vec_in = in (Vec(UInt(8 bits), 4))
    val bits_out = out (Bits(32 bits))
  }
  io.bits_out := io.vec_in.asBits
}
```
corresponding verilog:
```verilog
module VecAsBitsDemo (
  input  wire [7:0]    io_vec_in_0,
  input  wire [7:0]    io_vec_in_1,
  input  wire [7:0]    io_vec_in_2,
  input  wire [7:0]    io_vec_in_3,
  output wire [31:0]   io_bits_out
);


  assign io_bits_out = {io_vec_in_3,{io_vec_in_2,{io_vec_in_1,io_vec_in_0}}};

endmodule
```
The example show above looks workable, but for a Vec containing 32/64 elements or more, this will leads to deeply nested concatenation operation, which becomes a challenge for EDA tools to parse, a pain for human to read. 

So it would be nice to flatten the concatenation operator.

# Impact on code generation

The below code look more concise while functions are identical.
```verilog
module VecAsBitsDemo (
  input  wire [7:0]    io_vec_in_0,
  input  wire [7:0]    io_vec_in_1,
  input  wire [7:0]    io_vec_in_2,
  input  wire [7:0]    io_vec_in_3,
  output wire [31:0]   io_bits_out
);


  assign io_bits_out = {io_vec_in_3,io_vec_in_2,io_vec_in_1,io_vec_in_0};

endmodule

```

# Checklist

- [ ] Unit tests were added


facebreeze:
@Dolu1990 The failed tests are unrelated with this commit，I have no idea what have happened，could you please restart the test and help figure out the root cause? Thanks a lot.

typingArtist:
I see your motivation and I second your effort but the `'&'` operator is a small little beast. It can stitch together base types and arrays of that base type into a wider array of that base type, like in `'0' & "10101" = "010101"`, it can also stitch together multiple arrays of the base type into a wider array, like in `"1001" & "0110" = "10010110"`, but it can also put together multiple arrays into an array of arrays, like in `"1001" & "0110" = (0 => "1001", 1 => "0110")`. From the 2nd and 3rd example you see that the result depends on the context and visible types, and sometimes even the elaboration fails if it cannot decide which of them to use. `pkg_cat()` was introduced to ensure only one possible outcome, like in the 2nd case. 

And while the use of `pkg_cat()` seems to fix the problem, it would be even better to implement `pkg_cat()` itself without the use of `'&'`:

```vhdl
function pkg_cat(a: std_logic_vector; b: std_logic_vector) return std_logic_vector is
  variable cat: std_logic_vector(a'length + b'length - 1 downto 0);
begin
  cat(ret'high downto b'length) := a;
  cat(b'length - 1 downto 0) := b;
  return cat;
end function pkg_cat;
```

Look at the following example code:

```vhdl
entity amptest is end entity;

library ieee; use ieee.numeric_bit.all;

architecture rtl of amptest is
  subtype T_MYVEC is bit_vector(3 downto 0);
--  type T_MYVEC_ARRAY is array(natural range <>) of T_MYVEC;
  signal a : unsigned(7 downto 0);
  constant C : T_MYVEC := "1001";
begin
  a <= unsigned(C & C);
end architecture rtl;
```

do a `ghdl -a amptest.vhd; ghdl -e amptest; ghdl -r amptest`. This should just finish without a hitch. Now uncomment the type declaration and retry.

For improving the generated VHDL code, we could define multiple versions of `pkg_cat()` with different numbers of arguments to replace `pkg_cat(a, pkg_cat(b, pkg_cat(c, d)))` with a more readable `pkg_cat(a, b, c, d)`. I think that variants up to 5 or 6 parameters make sense. If concatenations with more elements are needed, the VHDL code generator could back up to a binary tree of calls to `pkg_cat()`.

typingArtist:
Two additions: 

1. To test one of the failing tests inside your environment, like `spinal.core.MultiClockTesterCocotbBoot`, you can use the following command: `sbt ++2.11.12  "tester/testOnly spinal.core.MultiClockTesterCocotbBoot"`.
2. It should be simpler to just create one new implementation of `pkg_cat()` that accepts between 2 and 5 or 6 parameters, where every parameter beyond the second is optional, like in the following code:

```vhdl
function pkt_cat(
  a: std_logic_vector;
  b: std_logic_vector;
  c: std_logic_vector := "";
  d: std_logic_vector := "";
  e: std_logic_vector := "";
  f: std_logic_vector := "") return std_logic_vector is
  variable cat : std_logic_vector(
    a'length + b'length + c'length + d'length + e'length + f'length - 1 downto 0);
begin
  cat(cat'high downto b'length + c'length + d'length + e'length + f'length) := a;
  cat(b'length + c'length + d'length + e'length + f'length - 1 downto c'length + d'length +e'length + f'length) := b;
  cat(c'length + d'length + e'length + f'length - 1 downto d'length + e'length + f'length) := c;
  cat(d'length + e'length + f'length - 1 downto e'length + f'length) := d;
  cat(e'length + f'length - 1 downto f'length) := e;
  cat(f'length - 1 downto 0) := f;
  return cat;
end function pkg_cat;
```

facebreeze:
@typingArtist Thanks a lot. I'm not familiar with VHDL and missed the underlying problem, and looks that the problem comes with VHDL only. 

In my opinion, even if a binary-tree was introduced when there are more elements than function argument count, the concatenation operation can not get fully flattened for VHDL. Therefore, i think it's unnecessary to do this for VHDL.

Dolu1990:
Thanks ^^

PR #1914, opened 2026-05-15
Add SRAM based IMSIC file container

inochisa:

Current IMSIC implementation in SpinalHDL is very area intensive as it uses register for interrupt generation. This make it hard to use this in RISC-V CPU with multiple IMSIC register files.

There is a implementation result of 4 cores VexiiRiscv on XCKU5P, with 1 M-mode file, 1 S-mode file and 3 VS-mode file. Each file has 256 interrupts. This could be a minimum request for running Qemu.

Screenshot From 2026-05-13 09-36-24

It is very area intensive and left no room for other features.

With this PR, a new SRAM based IMSIC register file is introduced, it can simply reduce the area at the cost of some read/write delay.

The performance of current implementation for each operation:
read: 2 cycles
write: 4 cycles
query highest interrupt: immediately, it has a cache for it.

The implementation result with this PR (Each imsic file is marked with different color):

Screenshot From 2026-05-15 10-53-27

It can also be built with 8 cores:

Screenshot From 2026-05-15 16-54-48

The VexiiRiscv fork of this PR can be found at https://github.com/project-inochi/VexiiRiscv/tree/h-aia

# Impact on code generation

No change on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
+++ b/lib/src/main/scala/spinal/lib/bus/misc/BusSlaveFactory.scala
@@ -436,6 +436,19 @@ trait BusSlaveFactory extends Area{
   }
 
   /**
+   * Emit on that a transaction when a write happen at address, by using data placed at bitOffset in the word.
+   * Block the write transaction until the transaction succeeds (stream becomes ready).
+   */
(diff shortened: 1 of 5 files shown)


Dolu1990:
Thanks ^^

PR #1918, opened 2026-05-23
Fix Counter lazy val naming inconsistency on Scala 2.11

facebreeze:

The test `spinal.core.RepeatabilityTester` fails on scala 2.11.12 while it pass on scala 2.13.12. 

The root cause is that  the valCallbackRec method in Area  is the mechanism that captures Scala variable names via the iDSL compiler plugin. lazy val fields in Scala 2.11 likely don't trigger the valCallbackRec callback, causing them to lose their names. As a result, the `willOverflow` and `willUnderflow` lazy vals in BoundedCounter were not getting proper partial names on Scala 2.11, which leads to signal count mismatches in RepeatabilityTester (97 vs 98). 

Explicitly calling setCompositeName after enableStandardPruning() ensures consistent naming across all Scala versions will fix this.

# Impact on code generation

No impact on code generation. 
# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/Counter.scala
+++ b/lib/src/main/scala/spinal/lib/Counter.scala
@@ -283,6 +283,8 @@ class Counter(
   when(willClear) { valueNext := U(initVal, w bits) }
 
   enableStandardPruning()
+  willOverflow.setCompositeName(this, "willOverflow", true)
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks :D

PR #1925, opened 2026-05-28
com: spi: xdr: Fix missing logic and byte alignments

briansune:
com: spi: xdr: Add Pipeline Bus Logic

Current SPI XDR Master is missing necessary logic in
"fromPipelinedMemoryBus" with ??? marked. In order to function properly
on QSPI NOR flash or SPI NOR flash each quad bytes fast read must pack
into a word 32bits. This allow the instruction bus execute properly from
a big-endian SPI NOR flash read. With the formation completes,
necessary valid signal is given out in sync with the word instruction.

Signed-off-by: Brian Sune <briansune@gmail.com>

com: spi: xdr: Fix byte readout alignment

Current SpinalHDL SPI controller byte read out content is misaligned.
This is caused by shift register and concat logic error. By properly
using a full 8 bit shift register result the read out now properly
aligned.

Signed-off-by: Brian Sune <briansune@gmail.com>


briansune:
@Tectu 

Hi Sir,

Do i need to fix the test build?
I am not sure that section need some guidance on that

I do a verification on the current testbench which is a bit faulty:

Meantime one of the fail case:
image

Which if follow datasheet:
https://www.winbond.com/resource-files/w25q32jv%20dtr%20revf%2002242017.pdf?__locale=zh_TW

8.9.2 DTR Fast Read (0Dh)
Once the dummy edge is given the data should immediately ready while the sim shows X and the clock in of the original SPI master also wrongly latched in which this is not a simple tick shift or early.

I don't know but this patch only fixes the shift error under normal SPI with "cpol=true cpha=true" and verified under real hardware.

Thanks,

briansune:
@Tectu 

Sorry for the noise.

This patch is not a proper way to correct the faulty SPI.

The issue here or from first place is the alignment between two classes.

When you want a earliest data latch from the SPI Bus while expecting latches on inputPhy.
This completely breaks the data alignment.
Meantime the sampling of the shift is so late that completely missed the expected timing.
image

```
val readFill = sync(fsm.readFill, False)
val readDone = sync(fsm.readDone, False)
```
Either replying on these signals or the io.cmd.valid also breaks the data readout from the NOR FLASH.

I will resubmit the patch.

Thanks,

Tectu:
Sure, feel free to propose a new PR.
Thank you for your efforts!

briansune:
> Sure, feel free to propose a new PR. Thank you for your efforts!

https://github.com/SpinalHDL/SpinalHDL/pull/1928

Done

So far the investigation have done and either lowest clock prescaler nor DTR under the new PR still introduce assert in the current SPI testbench which the TB itself is really having a lot of design fault.

I don't plan to fix it this moment, where DTR itself is not really lot of NOR FLASH devices on the market.
While STR and QUAD DUAL data lane itself cannot see any serious fault in the first place.

Thanks

PR #1927, opened 2026-06-02
Feature/okay spelling

### LINKED ISSUE #932: Rename OKEY -> OKAY
This is a typo in the API XD

What to choose? `OK`  or long version `OKAY`?

- [x] Chose `OKAY`
- [ ] Find all occurrences of `/OKEY/` (as pattern, not as a full word because eg. `setOKEY` must be a match too)
- [ ] Let the previous items as `@deprecated("Typo, use XXX instead")`, do not remove them
- [ ] Add new items with correct spelling

Tectu:
This changes the spelling of "OKEY" to "OKAY".
This change is motivated by the fact that "OKEY" is an incorrect spelling as well as that both the AMBA and the Avalon spec spell it as "OKAY".

Fixes #932

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba3/ahblite/AhbLite3.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba3/ahblite/AhbLite3.scala
@@ -123,7 +123,7 @@ case class AhbLite3(config: AhbLite3Config) extends Bundle with IMasterSlave {
   val HREADYOUT = Bool()
   val HRESP     = Bool()
 
-  def setOKEY  = HRESP := False
+  def setOKAY  = HRESP := False
   def setERROR = HRESP := True
 
   override def asMaster(): Unit = {
@@ -131,7 +131,7 @@ case class AhbLite3(config: AhbLite3Config) extends Bundle with IMasterSlave {
     in(HREADYOUT, HRESP, HRDATA)
   }
 
-  def OKEY   = !HRESP
+  def OKAY   = !HRESP
   def ERROR  = HRESP
   def isIdle = HTRANS === AhbLite3.IDLE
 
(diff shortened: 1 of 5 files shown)


mrcmry:
A PR with some spelling fixes was already merged here : https://github.com/SpinalHDL/SpinalHDL/pull/1885

Instead of breaking the API I think it would be better to mark the function as @deprecated.

When we choose the activate the deprecation in an open question, I have chosen in my PR to comment the deprecation so we can activate it in the next minor. But Perhaps it's more practical the activate this right now with `since = ">1.14.2"` and later update it to since="1.14.3" or since="1.15.0". The "since" field in scala is purely informational and has no runtime effect. Using this pattern with deprecation  with ">x.y.z" remove one extra thing to think about when doing a release for @Dolu1990 and also do not break the API for user (and is also easier to debug "method do no exist" vs "deprecated, use OKAY").

Perhaps you can speak with @Dolu1990, decide how to do it and I can add a chapter "release management" (or whatever) in developer area in SpinalDoc-RTD so we dont forget what we decided.

By the way, thanks for the issue cleanup, it's nice for clarity.

Tectu:
> Instead of breaking the API I think it would be better to mark the function as @deprecated.

I agree with this. However, as per my current knowledge I think that won't work for the signals in the bundle.
By the looks of it this would only affect the `io.ahb.setOKEY` signal in `AhbLite3OnChipRam` and `AhbLite3OnChipRom`. Everything else I believe can be marked as deprecated and replaced.

Thoughts?

> Perhaps you can speak with @Dolu1990, decide how to do it and I can add a chapter "release management" (or whatever) in developer area in SpinalDoc-RTD so we dont forget what we decided.

Indeed. I'll keep you posted.
My current understanding is that @Dolu1990 's preference is to never remove deprecated items but realistically (or rather practically) they have to go at some point.
I understand that this is different from the software world tho. Maybe we can work towards a mayor release?

> By the way, thanks for the issue cleanup, it's nice for clarity.

Happy to help where I can. This was just the initial run. It will take quite some time to work through all the open issues and PRs and figuring out what to do with them. I don't want to be too aggressive at the beginning but I certainly want to clean up the mess as best as I can.

Issue #1922, opened 2026-05-27
SPI library bug?


briansune:
Hi all,

<img width="1146" height="412" alt="Image" src="https://github.com/user-attachments/assets/2f8d0318-900a-4994-8a98-c66ec2b3cd6d" />

The sample data return shows very wrong thing is happening.

I am not sure but I think the expected data should be 37h and the shift action also shows very puzzling behavior.

Thanks,
Brian

briansune:
@Dolu1990 

Hi all,

I am not sure this actually fix the bug but real hardware show a proper read out.

<img width="1220" height="403" alt="Image" src="https://github.com/user-attachments/assets/4338d9c1-6d71-4265-8dd3-75553a86557a" />

[v1-0001-Fix-SPI-byte-readout-alignment.patch](https://github.com/user-attachments/files/28325637/v1-0001-Fix-SPI-byte-readout-alignment.patch)

PR #1928, opened 2026-06-06
COM: SPI: XDR: Improve coherency between levels

briansune:
com: spi: xdr: Fix byte readout alignment
    
    Current SpinalHDL SPI controller payload is aligned with the cmd valid
    while the rsp.valid is registered by additional latches.
    This introduce data corruption which only 7bits register are use to
    latch the SPI readout and the last bit is simply combine via
    combinational logic.
    Such design is fast but timing slack is highly reduced.
    In order to fix this misalignment removing latches and simply gated by
    the necessary top logic level.
    
    Signed-off-by: Brian Sune <briansune@gmail.com>

    com: spi: xdr: Add Pipeline Bus Logic
    
    Current SPI XDR Master is missing necessary logic in
    "fromPipelinedMemoryBus" with ??? marked. In order to function properly
    on QSPI NOR flash or SPI NOR flash each quad bytes fast read must pack
    into a word 32bits. This allow the instruction bus execute properly from
    a big-endian SPI NOR flash read. With the formation completes,
    necessary valid signal is given out in sync with the word instruction.
    
    Signed-off-by: Brian Sune <briansune@gmail.com>

Real VEX RISC-V hardware test NOR FLASH W25Q64

image

For testbench I am not too convinced by the functionality, where it is somehow glitchy.

### DIFF
--- a/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
+++ b/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
@@ -408,13 +408,33 @@ object SpiXdrMasterCtrl {
 
     def fromPipelinedMemoryBus() = {
       val accessBus = new PipelinedMemoryBus(PipelinedMemoryBusConfig(24,32))
-      ???
-//      cmd.valid <> (accessBus.cmd.valid && !accessBus.cmd.write)
-//      cmd.ready <> accessBus.cmd.ready
-//      cmd.payload <> accessBus.cmd.address
-//
-//      rsp.valid <> accessBus.rsp.valid
-//      rsp.payload <> accessBus.rsp.data
+      cmd.valid <> (accessBus.cmd.valid && !accessBus.cmd.write)
+      cmd.address <> accessBus.cmd.address
+      accessBus.cmd.ready := cmd.ready
+
+      cmd.length := 3
+
+      val buffer  = Reg(Bits(32 bits)) init(0)
+      val counter = Reg(UInt(2 bits)) init(0)
+
+      accessBus.rsp.valid := False
+      accessBus.rsp.data := rsp.fragment ## buffer(23 downto 0)
+
+      rsp.ready := True
+
+      when(rsp.fire) {
+        counter := counter + 1
+
+        switch(counter) {
+          is(0) { buffer(7 downto 0)   := rsp.fragment }
+          is(1) { buffer(15 downto 8)  := rsp.fragment }
+          is(2) { buffer(23 downto 16) := rsp.fragment }
+          is(3) { buffer(31 downto 24) := rsp.fragment
+            accessBus.rsp.valid := True
+          }
+        }
+      }
+
       accessBus
     }
 
--- a/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
+++ b/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
@@ -896,8 +896,8 @@ object SpiXdrMasterCtrl {
     val inputPhy = new Area{
       def sync[T <: Data](that : T, init : T = null) = Delay(that,2,init=init)
       val mod = sync(io.config.mod)
-      // val readFill = sync(fsm.readFill, False)
-      // val readDone = sync(fsm.readDone, False)
+      val readFill = sync(fsm.readFill, False)
+      val readDone = sync(fsm.readDone, False)
       val buffer = Reg(Bits(p.dataWidth - p.mods.map(_.bitrate).min bits))
       val bufferNext = Bits(p.dataWidth bits).assignDontCare().allowOverride
       val widthSel = mod.muxListDc(p.mods.map(m => m.id -> U(widths.indexOf(m.bitrate), log2Up(widthMax + 1) bits)))
@@ -926,12 +926,12 @@ object SpiXdrMasterCtrl {
         for ((width,widthId) <- widths.zipWithIndex) {
           is(widthId) {
             bufferNext := (buffer ## dataRead(0, width bits)).resized
-            when(fsm.readFill) { buffer := bufferNext.resized }
+            when(readFill) { buffer := bufferNext.resized }
           }
         }
       }
 
-      io.rsp.valid := fsm.readDone
+      io.rsp.valid := readDone
       io.rsp.data := bufferNext
 
       switch(mod){
--- a/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
+++ b/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
@@ -895,8 +895,7 @@ object SpiXdrMasterCtrl {
 
     val inputPhy = new Area{
       def sync[T <: Data](that : T, init : T = null) = Delay(that,2,init=init)
-      // val mod = sync(io.config.mod)
-      val mod = io.config.mod
+      val mod = sync(io.config.mod)
       // val readFill = sync(fsm.readFill, False)
       // val readDone = sync(fsm.readDone, False)
       val buffer = Reg(Bits(p.dataWidth - p.mods.map(_.bitrate).min bits))
--- a/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
+++ b/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
@@ -896,8 +896,8 @@ object SpiXdrMasterCtrl {
     val inputPhy = new Area{
       def sync[T <: Data](that : T, init : T = null) = Delay(that,2,init=init)
       val mod = sync(io.config.mod)
-      val readFill = sync(fsm.readFill, False)
-      val readDone = sync(fsm.readDone, False)
+      // val readFill = sync(fsm.readFill, False)
+      // val readDone = sync(fsm.readDone, False)
       val buffer = Reg(Bits(p.dataWidth - p.mods.map(_.bitrate).min bits))
       val bufferNext = Bits(p.dataWidth bits).assignDontCare().allowOverride
       val widthSel = mod.muxListDc(p.mods.map(m => m.id -> U(widths.indexOf(m.bitrate), log2Up(widthMax + 1) bits)))
@@ -926,12 +926,12 @@ object SpiXdrMasterCtrl {
         for ((width,widthId) <- widths.zipWithIndex) {
           is(widthId) {
             bufferNext := (buffer ## dataRead(0, width bits)).resized
-            when(readFill) { buffer := bufferNext.resized }
+            when(fsm.readFill) { buffer := bufferNext.resized }
           }
         }
       }
 
-      io.rsp.valid := readDone
+      io.rsp.valid := fsm.readDone
       io.rsp.data := bufferNext
 
       switch(mod){
--- a/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
+++ b/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
@@ -895,7 +895,8 @@ object SpiXdrMasterCtrl {
 
     val inputPhy = new Area{
       def sync[T <: Data](that : T, init : T = null) = Delay(that,2,init=init)
-      val mod = sync(io.config.mod)
+      // val mod = sync(io.config.mod)
+      val mod = io.config.mod
       // val readFill = sync(fsm.readFill, False)
       // val readDone = sync(fsm.readDone, False)
       val buffer = Reg(Bits(p.dataWidth - p.mods.map(_.bitrate).min bits))
--- a/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
+++ b/lib/src/main/scala/spinal/lib/com/spi/xdr/SpiXdrMasterCtrl.scala
@@ -60,7 +60,9 @@ case class XdrPin(rate : Int) extends Bundle with IMasterSlave{
 
 case class SpiXdrParameter(dataWidth : Int,
                            ioRate : Int,
-                           ssWidth : Int)
+                           ssWidth : Int,
+                           inLatency : Int = 1,
+                           outLatency : Int = 1)
 
 case class SpiXdrMaster(val p : SpiXdrParameter) extends Bundle with IMasterSlave{
   import p._
@@ -894,7 +896,12 @@ object SpiXdrMasterCtrl {
 
 
     val inputPhy = new Area{
-      def sync[T <: Data](that : T, init : T = null) = Delay(that,2,init=init)
+      def sync[T <: Data](that : T, init : T = null) : T = {
+        if( (p.spi.inLatency + p.spi.outLatency) > 0)
+          Delay(that,(p.spi.inLatency + p.spi.outLatency),init=init)
+        else
+          that
+      }
       val mod = sync(io.config.mod)
       val readFill = sync(fsm.readFill, False)
       val readDone = sync(fsm.readDone, False)
--- a/tester/src/test/scala/spinal/lib/com/spi/xdr/SpinalSimSpiXdr.scala
+++ b/tester/src/test/scala/spinal/lib/com/spi/xdr/SpinalSimSpiXdr.scala
@@ -106,24 +106,9 @@ class SpinalSimSpiXdrMaster extends SpinalAnyFunSuite {
             dut.io.config.sclkToggle #= sclkToggle
 
             if (read) rspScoreboard.pushRef(readData)
-
-            for ((pin, tasks) <- mod.readMapping.groupBy(_.pin)) {
-              val initialDriveVal = tasks.filter { m =>
-                ((readData >> (m.target + mod.dataWidth - mod.bitrate)) & 1) != 0
-              }.map(1 << _.phase).fold(0)(_ | _)
-              dut.io.spi.data(pin).read #= initialDriveVal
-            }
-
             val spiThread = fork {
               for (beat <- 0 until mod.dataWidth / mod.bitrate) {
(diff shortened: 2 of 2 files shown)


briansune:
Sim log:

```
[Done] at 1.328
[Info] Workspace 'TopLevel' was reallocated as 'TopLevel_2' to avoid collision
[Progress] Simulation workspace in /home/intel/Desktop/SpinalHDL/./simWorkspace/TopLevel_2
[info] Found cached verilator binaries
[Progress] Verilator compilation done in 324.480 ms
[info] - verilator_compile
[Progress] Start TopLevel verilator_ test cpol=false cpha=false 0 simulation with seed 32
[Done] Simulation done in 105.222 ms
[info] - verilator_ test cpol=false cpha=false 0
[Progress] Start TopLevel verilator_ test cpol=false cpha=true 0 simulation with seed 32
[Done] Simulation done in 81.359 ms
[info] - verilator_ test cpol=false cpha=true 0
[Progress] Start TopLevel verilator_ test cpol=true cpha=false 0 simulation with seed 32
[Done] Simulation done in 86.836 ms
[info] - verilator_ test cpol=true cpha=false 0
[Progress] Start TopLevel verilator_ test cpol=true cpha=true 0 simulation with seed 32
[Done] Simulation done in 67.762 ms
[info] - verilator_ test cpol=true cpha=true 0
[info] Run completed in 15 seconds, 58 milliseconds.
[info] Total number of tests run: 15
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 15, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[success] Total time: 19 s, completed Jun 7, 2026,
```

briansune:
@Tectu
 
Please do a workflows.

Thanks,

Dolu1990:
Hi,

A few questions : 

1) Did you plugged the SPI from the controller directly on the FPGA io ? Or you used the placed in between the "phy" layer (which add a register on all pins)

> This introduce data corruption which only 7bits register are use to
latch the SPI readout and the last bit is simply combine via
combinational logic.
Such design is fast but timing slack is highly reduced.
In order to fix this misalignment removing latches and simply gated by
the necessary top logic level.

Do you mean that the design is buggy ("corruption") and slow ? or that the design is slow ?

Thanks ^^

briansune:
@Dolu1990 

> Hi,
> 
> A few questions :
> 
> 1. Did you plugged the SPI from the controller directly on the FPGA io ? Or you used the placed in between the "phy" layer (which add a register on all pins)
> 
> > This introduce data corruption which only 7bits register are use to
> > latch the SPI readout and the last bit is simply combine via
> > combinational logic.
> > Such design is fast but timing slack is highly reduced.
> > In order to fix this misalignment removing latches and simply gated by
> > the necessary top logic level.
> 
> Do you mean that the design is buggy ("corruption") and slow ? or that the design is slow ?
> 
> Thanks ^^

For 1)
I used the original Murax XIP SIP to verify.
No matter it is standalone or Murax itself also shows a wrong timing between class b.c. the valid in phy class added two ticks.
I had verified this patches under real HW with Murax demo and successfully run on NOR FLASH and SD card.

If you study the section "inputPhy" class you will clearly see that no matter what the idea of "Or you used the placed in between the "phy" layer (which add a register on all pins)" the data of the io.rsp.data will never be correct due to the passing of non latched SPI capture from the fsm sequencing.

Long story short: I did not introduce any additional register nor wrap reg/logic and simply map the generated IOs with bidirectional IO = (en) ? out : 1'z1; assign in = IO; so there are no additional concept of top level misalignment.

All misalignment are introduced by the SPI SpinalHDL controller design itself.

For 2)
Meantime with no fix the original simulation also did not pass all tasks which is clearly observed that the two ticks of sync introduce lag to the control which the early signal will simply be masked by the reset and two ticks.

Also with the current testbench sequence the first two bits or bit will be late and not properly sent to the DUT.

So I fixed by the push requested patches and also included descriptions of each patch does.

Thanks

Dolu1990:
> Murax XIP SIP

SIP ?

> class b.c

class b.c ?

> Real VEX RISC-V hardware test NOR FLASH W25Q64

This is with or without the patch ?

I'm trying to get a wave were i can see the issue. Mostly because the current design is already used by many, and so far, everything worked seemingly fine

briansune:
> > Murax XIP SIP
> 
> SIP ?

Typo SPI

> 
> > class b.c
> 
> class b.c ?

b.c <- because

> 
> > Real VEX RISC-V hardware test NOR FLASH W25Q64
> 
> This is with or without the patch ?

With my patches

> 
> I'm trying to get a wave were i can see the issue. Mostly because the current design is already used by many, and so far, everything worked seemingly fine

I don't think it is used many at least for XIP the code don't even complete for years.

Meantime it is highly doubted when the inputPhy is highly relied by the FSM counter. How could the valid signal correct when the data already gone while the LSb is combinational logic and the next level is rely on valid to pass input a correct byte.

I highly suggest you really check the signal under GTK-Wave not blindly trust the testbench pass result.

Please also read:
https://github.com/SpinalHDL/SpinalHDL/issues/1922

If you trust your instinct I am happy to close this PR which at lease I find a way to work on this side.
As for others Well I don't know.

I am sorry that I am more a HDL guy so Scala is really not my cup of tea but I think the ticks and timing should be the same idea.

Dolu1990:
> Please also read:
#1922

Ahh nice, i missed that one. i'm looking.

> If you trust your instinct I am happy to close this PR which at lease I find a way to work on this side.
As for others Well I don't know.

Mostly, i'm trying to find out if this maybe a bad configuration or something like it.

> I am sorry that I am more a HDL guy so Scala is really not my cup of tea but I think the ticks and timing should be the same idea.

No worries ^^

> phy

When i meant phy , i meant something like : 
https://github.com/SpinalHDL/VexRiscv/blob/680756065e9e6fc50d8c3d6c58191a16e867d822/src/main/scala/vexriscv/demo/Murax.scala#L468
(not the controller internal inputPhy)
That phy isn't contained in the Murax toplevel, but  added on the board specific toplevel.

It is expected for the user to added 1 cycle latency on outputs, and 1 cycle latency on inputs.
Your hadware tests included something like it ?

briansune:
> > Please also read:
> > #1922
> 
> Ahh nice, i missed that one. i'm looking.
> 
> > If you trust your instinct I am happy to close this PR which at lease I find a way to work on this side.
> > As for others Well I don't know.
> 
> Mostly, i'm trying to find out if this maybe a bad configuration or something like it.
> 
> > I am sorry that I am more a HDL guy so Scala is really not my cup of tea but I think the ticks and timing should be the same idea.
> 
> No worries ^^
> 
> > phy
> 
> When i meant phy , i meant something like : https://github.com/SpinalHDL/VexRiscv/blob/680756065e9e6fc50d8c3d6c58191a16e867d822/src/main/scala/vexriscv/demo/Murax.scala#L468 (not the controller internal inputPhy) That phy isn't contained in the Murax toplevel, but added on the board specific toplevel.

We are not in the same page can you really study the internal inputPhy and use GTK-Wave to review what my patch does?
There is nothing to do with input latency and output latency.

It is very hard for me to explain when you are not able to read the FSM <-> to inputPhy action and how the input coming bit from the SPI bus capture.

> 
> It is expected for the user to added 1 cycle latency on outputs, and 1 cycle latency on inputs. Your hadware tests included something like it ?

Even you add ticks on the input data it is also not going to work when inputPhy already missed the timing and also adding flops on input and output also have nothing to do with what my patch trying to fix.

Sorry Sir, form what the inputPhy if you really understand the coherence of two classes you will know what I am trying to lay down.

Just apply:
https://github.com/SpinalHDL/SpinalHDL/pull/1928/changes/10d8b4df6bf3dc9ddb1db9d63237ffe6416b5f51

is 100% not going to work.

Meantime, I see very questionable valid signal in the default SPI testbench which is 100% not convincing.

@Dolu1990 

Anyway If you feel this PR is useful feel free to use it, I am tired on this SPI controller which at least working properly with out issue now.

I am afraid you misunderstand this line, what i meant i will this PR the SPI is function properly.

Also I am not convinced by what input output 1 tick delay, if both are +1 clock so it simply cancel out the effect of external slave device to pass out the bit as respect to the two flops. So either of these ideas are not hold

Dolu1990:
> We are not in the same page can you really study the internal inputPhy and use GTK-Wave to review what my patch does?
There is nothing to do with input latency and output latency.

So, what i can see, is that the removal of the 2 cycle delay via this PR bypass of : 
```scala
      val mod = sync(io.config.mod)
      val readFill = sync(fsm.readFill, False)
      val readDone = sync(fsm.readDone, False)
```

will assume a zero cycle latency between the controller and the external physical pins, which will break all the currently implemented phy, aswell as make it impractical for high speed SPI interface.

Why removing those delay ?

> Also I am not convinced by what input output 1 tick delay, if both are +1 clock so it simply cancel out the effect of external slave device to pass out the bit as respect to the two flops. So either of these ideas are not hold

The idea is to have  :
```scala
      val mod = sync(io.config.mod)
      val readFill = sync(fsm.readFill, False)
      val readDone = sync(fsm.readDone, False)
```

That way it cancel out the SPI phy latency (1 output ff + 1 input ff), This is a wanted thing which allow to get high speed DDR / gearbox phy layers outside the spi controller, to connect to the physical pins.

I don't understand what is wrong with the current design.

briansune:
> > We are not in the same page can you really study the internal inputPhy and use GTK-Wave to review what my patch does?
> > There is nothing to do with input latency and output latency.
> 
> So, what i can see, is that the removal of the 2 cycle delay via this PR bypass of :
> 
> ```scala
>       val mod = sync(io.config.mod)
>       val readFill = sync(fsm.readFill, False)
>       val readDone = sync(fsm.readDone, False)
> ```
> 
> will assume a zero cycle latency between the controller and the external physical pins, which will break all the currently implemented phy, aswell as make it impractical for high speed SPI interface.

No Actual hardware had shown 25MHz or above normal function.
This is why you did not understand what coherent between the FSM and inputphy means.

Meantime if user really need addition metastable on the SPI MOSI MISO all it need just add REG / IOB / DRRIOB on top level.

> 
> Why removing those delay ?
> 
> > Also I am not convinced by what input output 1 tick delay, if both are +1 clock so it simply cancel out the effect of external slave device to pass out the bit as respect to the two flops. So either of these ideas are not hold
> 
> The idea is to have :
> 
> ```scala
>       val mod = sync(io.config.mod)
>       val readFill = sync(fsm.readFill, False)
>       val readDone = sync(fsm.readDone, False)
> ```
> 
> That way it cancel out the SPI phy latency (1 output ff + 1 input ff), This is a wanted thing which allow to get high speed DDR / gearbox phy layers outside the spi controller, to connect to the physical pins.

No this will not make a proper FSM coherency I expect whom design this should know very well.
For DTR it is also work in the same idea on the above.
You add input and output flop to gate the signal which still kept aligned from first place not because the bottom level is not align so I need top to make it algin.

i.e. MOSI this is aligned by the SCLK which both should tick in pace
While the incoming MISO also tick in pace with the SCLK with Tsu and Thld so a fast clock to capture and this controller done by a capture pulse under FSM and join lsb with combinational logic.
Due to the SCLK is controlled by the FSM while the capture signal as well so the entry signal must be aligned in first place.
When you need to add additional flop on the MOSI this also breaks the alignment of block output hances the SCLK <-> MOSI
SPI is a full-duplex protocol, so All three SCLK <-> MOSI <-> MISO must be aligned properly from device <-> FPGA/VLSI IO Buffer

If you delay MOSI +1 and MISO +1 on top of SCLK then it is very sure you cannot do (system clock / 2) or (system clock) speed.
Meantime it also very sure that the FSM with not able to capture the LSb proper under the capture pulse.

It is correct that you should have block box like analog stage, BIO, DDRIO buffer to reform the proper signal which again should be also aligned on three SCLK <-> MOSI <-> MISO then the SPI controller can in-sync with the FSM.

Also if this idea hold then the testbench also did not follow what you had descripted.

> 
> I don't understand what is wrong with the current design.

I cannot explain more when the coherency is not understandable on your side.
Which I am terribly sorry on that

For VLSI or any HDL design you will not expect top or bottom level to add additional tick to fulfill the signal sanity in the first place, which whatever idea or design that the current state can make to work is very improper.

Sorry again that I cannot explain more!

If this is not what your project expected or it is design in purpose of unalignment controller that I am terribly sorry I spend your time.

*REMARKS
One flop cannot fix metastable you at least need two or three and using flops as capture reformation is very wrong in first place.

There are proper IOB or Buffer on FPGAs so there are no concept of register the signal for fast design.

Meantime the latency might not be 1 tick it can be vary depends on design of the black-box. Then the SPI controller will eventually out of sync again.

So if you really understand what I am laying down I guess you will start understand.

Thanks,
Brian

briansune:
@Dolu1990 

image

This is the corrected patch waveform, how many tick you want in order the readfill pulse to capture a bit with in the SCLK period.

Either system/2 or system/4 have enough ticks for top level to undergo a latency of 1 ticks unless top level need to then you do start need to shift the readfill and valid.

Also if you want system/1 then 
image

This is the corrected patch waveform, so now if you want top black box to have latency you need bottom level to process this.
Then the bottom should dynamically sets to a proper latency settings i.e. FSM or register taps (the original FIXED 2 tick delay).

Meanwhile I am really confused?

This is the original design wave
image

You had mentioned many times 1 tick on top level input & output include SCLK?
Why you need additional tick on output?

This clearly shows no any reason from first place to be two, again unless system_clock = SPI_clock which you have no choice to delay 1 tick for flopping / black-box latency.

Again you assumed the latency of the top level on total of two ticks which breaks the idea of coherency in bottom block input and output.

briansune:
@Dolu1990 

I changed something that should make both side agree.
I had tested on real hardware with the same setup.
Which I simply set to:

```
xipConfig = ifGen(withXip) (SpiXdrMasterCtrl.MemoryMappingParameters(
      SpiXdrMasterCtrl.Parameters(8, 20, SpiXdrParameter(2, 1, 1, 0, 0)).addFullDuplex(0,1,false),
      cmdFifoDepth = 32,
      rspFifoDepth = 32,
      xip = SpiXdrMasterCtrl.XipBusParameters(addressWidth = 24, lengthWidth = 2)
    )),
```

Success boot via nor flash on Murax modified demo.

As for testbench I revert it. Which I am not convinced by the two tick delay inherent requirement.

Thanks

briansune:
```
COM: SPI: XDR: Improve coherency between levels
Current SPI controller assume top level bounding by 1 tick between input
and output signals. This is bad for user with different latency setup on
top level.

Improve latency support with parametric setup.

 - Add inLatency to Parameter with default value 1 tick
 - Add outLatency to Parameter with default value 1 tick
 - Introduce flops or wire by conditional check of total latencies

Signed-off-by: Brian Sune <briansune@gmail.com>
```

Dolu1990:
> For VLSI or any HDL design you will not expect top or bottom level to add additional tick to fulfill the signal sanity in the first place, which whatever idea or design that the current state can make to work is very improper.

The idea is to have a FPGA / ASIC agnostic SPI controller. On some plateforms , you can't infer IO flip-flop using regular VHDL / Verilog, you have to use vendor specific IP. This is why, this isn't something the SPI controller itself should deal with, but instead, this has to be externaly handled.

This design is kinda a bit designed like a DDR SDRAM, you have a controller, which is mostly hardware agnostic, and a phy layer which is very very hardware specific. Here, the SpiXdrMasterCtrl is that hardware agnostic thing, and the user is expected to introduce a phy layer in between the SpiXdrMasterCtrl and the FPGA pins.

> One flop cannot fix metastable you at least need two or three and using flops as capture reformation is very wrong in first place.

That is right, for the spi controller, this doesn't happen because it only sample the MISO data when it is assumed to be stable. But yes i agree, it isn't realy nice to assume that.

> You had mentioned many times 1 tick on top level input & output include SCLK?
Why you need additional tick on output?
Why using the sampled SPI MISO 2 cycle after the sclk_write falling edge ? 

cycle -1 =>  SCLK = 1
cycle 0 =>   SCLK = 0 MISO is being sampled at the end of the cycle
cycle 1 =>   SCLK physicaly on the FPGA = 0, Sampled miso is being used by the controller

This is a way to sample the MISO as late as possible to get the best timings.


I'm good with the PR, thanks :D

briansune:
> > For VLSI or any HDL design you will not expect top or bottom level to add additional tick to fulfill the signal sanity in the first place, which whatever idea or design that the current state can make to work is very improper.
> 
> The idea is to have a FPGA / ASIC agnostic SPI controller. On some plateforms , you can't infer IO flip-flop using regular VHDL / Verilog, you have to use vendor specific IP. This is why, this isn't something the SPI controller itself should deal with, but instead, this has to be externaly handled.
> 
> This design is kinda a bit designed like a DDR SDRAM, you have a controller, which is mostly hardware agnostic, and a phy layer which is very very hardware specific. Here, the SpiXdrMasterCtrl is that hardware agnostic thing, and the user is expected to introduce a phy layer in between the SpiXdrMasterCtrl and the FPGA pins.
> 
> > One flop cannot fix metastable you at least need two or three and using flops as capture reformation is very wrong in first place.
> 
> That is right, for the spi controller, this doesn't happen because it only sample the MISO data when it is assumed to be stable. But yes i agree, it isn't realy nice to assume that.
> 
> > You had mentioned many times 1 tick on top level input & output include SCLK?
> > Why you need additional tick on output?
> > Why using the sampled SPI MISO 2 cycle after the sclk_write falling edge ?
> 
> cycle -1 => SCLK = 1 cycle 0 => SCLK = 0 MISO is being sampled at the end of the cycle cycle 1 => SCLK physicaly on the FPGA = 0, Sampled miso is being used by the controller
> 
> This is a way to sample the MISO as late as possible to get the best timings.

No this is not correct you never need the extra register on SDR SPI.
And what you are trying to say is call "input output timing constraints", which this is what it is used for.

"POINT A": BTW if I design these in the first place I won't even allow the FSM to do bit deserialization.

Controller should always do ready packed data not PHY job, let vendor DESER or IP or foundry standard PHY to handle all fast timing and SpinalHDL to handle the data moving and sequencing.

I will say the for less than 150M adding register make no different on fitter but just extra power. The combination of the fill pulse will then trigger a valid to the APB bus that will again registered no matter how. This path basically is very short in logic flow as it is just a MUX -> REG no LUT -> MUX -> REG.

Long story short: I will highly suggests if anyone having energy to redo these controller, kept in mind of:
"POINT A"

> 
> I'm good with the PR, thanks :D

Welcome =]

PR #1932, opened 2026-06-10
Fix polarity of tristate input for Xilinx IOBUF blackbox

### LINKED ISSUE #1931: Xilinx IOBUF blackbox tristate polarity is reversed
It appears to me that the T input of blackbox for the Xilinx 7 Series IOBUF is reversed from the real device behavior. The Xilinx docs say "A logic High on the T pin disables the output buffer." ([UG471](https://docs.amd.com/v/u/en-US/ug471_7Series_SelectIO), page 39).

https://github.com/SpinalHDL/SpinalHDL/blob/749382b75f594f7ba112b914910dcae2d39f0df5/lib/src/main/scala/spinal/lib/blackbox/xilinx/s7/IO.scala#L139-L141

Just out of curiosity, when are the blackbox implementations used? It's my understanding that when a component is derived from BlackBox, its implementation is omitted when generating sources. Is the implementation here meant for simulation/formal verification?

obriensp:
Closes #1931

Reverses the polarity of the T input on the IOBUF blackbox primitive to match the Xilinx docs. When T is high, the output buffer is disabled.

# Impact on code generation
Since it's a blackbox component, this should only affect code generation for simulations. In this case, the generated code will now match the Xilinx docs and unisim library.

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/blackbox/xilinx/s7/IO.scala
+++ b/lib/src/main/scala/spinal/lib/blackbox/xilinx/s7/IO.scala
@@ -136,7 +136,7 @@ case class IOBUF() extends BlackBox{
   val O = out Bool()
   val IO = inout(Analog(Bool()))
 
-  when(T){
+  when(!T){
     IO := I
   }
   O := IO


Dolu1990:
Thanks ^^

PR #1933, opened 2026-06-11
feat: Add readAsyncWrite port inference for Mem

Aqwyh123:
Closes #

Currently, there is a lack of a model for single-port asynchronous read and synchronous write memory. This makes it inconvenient during the black-boxing process.

I added the relevant code for readAsyncWrite by following the same approach as for other types of ports. If there are any mistakes, please let me know.

# Impact on code generation

Only new additions without any modifications will not affect the existing code. The pattern matching branches have all been covered as well.

# Checklist

- [ ] Unit tests were added


Aqwyh123:
AI automatic review seems to found some problems in the original code. I think the AI's consideration is incorrect.

Dolu1990:
Thanks ^^

PR #1942, opened 2026-06-28
Refactor the TileLink peripherals with new pipeline API

inochisa:

As noticed, the current implementation of RAM, Cache and Hub use the old pipeline API, porting them to the new pipeline API so it can benefit from the improvement of the new design.

Adding some utility functions to the CtrlLink and the pipeline class.

Tested with the VexiiRiscv and litex.

# Impact on code generation

No impact on the public interface, but the internal interface is changed to the new design.

# Checklist

- [ ] Unit tests were added


Dolu1990:
Thanks ^^

PR #1947, opened 2026-07-26
Allow to change defaultKey of up/down of the CtrlLink

### LINKED ISSUE #1943: The defaultKey behavior in CtrlLink
I am trying to add the ability to set the default subkey for the node when initializing (For supporting a more complex pipeline). And found an interesting thing in the `CtrlLink`.

The `CtrlLink` defines the following:
```scala
def defaultKey : Any = null
```
And it use the default key as the following:
```scala
  /** Same as `Link.down(Payload)` */
  def apply[T <: Data](that: Payload[T]): T = down(that, defaultKey)

  /** Allows to conditionally override a [[Payload]] value between `link.up` -> `link.down`.
   *
   * This can be used to fix data hazard in CPU pipelines for instance.
   */
  def bypass[T <: Data](that: Payload[T]): T =  bypass(that, defaultKey)
```

It seems like the `defaultKey` defined in the `CtrlLink` is more like a override. But for the meaning, I think it does not respect the `defaultKey` of the node up and down. So I wonder what's the real meaning of the `defaultKey` in `CtrlLink`? 
At least I think we should adding a fix for the these two function to make sure it respect the `defaultKey` of the node, it does not match the comment (For example function `apply` above).

inochisa:
Closes #1943

Currently, the `defaultKey` of the `CtrlLink` does not respect the `defaultKey` of the node `up`/`down`, instead, it use a internal `defaultKey` defined in the `CtrlApi`. This means the `defaultKey` in the node does not take effect, it makes some necessary code for complex pipeline.

Introduce a flexible control for the `CtrlLink` and Node so it can control the `defaultKey` in a more easier way. Also, add the `defaultKey` to the `StagePipeline` and `StageCtrlPipeline`, so it can be used as internal component of the complex pipeline.

# Impact on code generation

Note: the current `defaultKey` API is marked as deprecated since the newly introduced `upDefaultKey`/`downDefaultKey` can provide fine-grant control

**The `CtrlLinkMirror.defaultKey` is meaningless as it always uses `upDefaultKey`/`downDefaultKey`. This may introduce a break to some code that overrides the `defaultKey` of `CtrlLinkMirror`.** Since this Area is more like a internal thing and the `defaultKey` should not be overrided, so I think it is fine to change this.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/misc/pipeline/Builder.scala
+++ b/lib/src/main/scala/spinal/lib/misc/pipeline/Builder.scala
@@ -74,12 +74,12 @@ class NodesBuilder() extends Area {
   }
 }
 
-class StagePipeline() extends Area {
+class StagePipeline(defaultKey: Any = null) extends Area {
   val nodes = mutable.LinkedHashMap[Int, Node]()
   val links = mutable.ArrayBuffer[StageLink]()
 
   def apply(i : Int) = node(i)
-  def node(i : Int) = nodes.getOrElseUpdate(i, new Node().setCompositeName(this, s"node_${i.toString}"))
+  def node(i : Int) = nodes.getOrElseUpdate(i, new Node(defaultKey).setCompositeName(this, s"node_${i.toString}"))
   class Area(i : Int) extends NodeMirror(node(i)) with spinal.core.Area
 
   def first = node(nodes.keys.min)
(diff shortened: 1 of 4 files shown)


Dolu1990:
Hi,

Is this PR ok with the current VexiiRiscv implementation ? (VexiiRiscv is the reference usage of the pipeline API)

inochisa:
> Hi,
> 
> Is this PR ok with the current VexiiRiscv implementation ? (VexiiRiscv is the reference usage of the pipeline API)

@Dolu1990 I think it is OK, at least it passed the regression test of vexiiriscv (100 tests).

inochisa:
Although the ci error is not related to this PR. I guess it is because the CPU of ci is overcommited too  much? If so, I guess we can relax the time check?

Dolu1990:
> Although the ci error is not related to this PR. I guess it is because the CPU of ci is overcommited too much? If so, I guess we can relax the time check?

Meaning increasing the github action timeout ?

inochisa:
> > Although the ci error is not related to this PR. I guess it is because the CPU of ci is overcommited too much? If so, I guess we can relax the time check?
> 
> Meaning increasing the github action timeout ?

I mean increasing timeout of the test. I think 60s is kind of limited.

Dolu1990:
Sure ^^

PR #1948, opened 2026-07-28
Add tilelink channel D corrupt line check

inochisa:

According to the latest tilelink spec 1.9.3. The corrupt line of the channel D should not be hard-wired to False. Make its behavior following the spec.

# Impact on code generation

The tilelink channel D now provides right corrupt line.

# Checklist

- [x] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/tilelink/sim/Checker.scala
+++ b/lib/src/main/scala/spinal/lib/bus/tilelink/sim/Checker.scala
@@ -199,6 +199,9 @@ class Checker(p : BusParameter, mappings : Seq[Endpoint], checkMapping : Boolean
   }
 
   override def onD(d: TransactionD) = {
+    if(d.withData) assert(!d.denied || d.corrupt, "Denied is also corrupt")
+    if(!d.withData) assert(!d.corrupt)
+
     d.opcode match{
       case Opcode.D.ACCESS_ACK | Opcode.D.ACCESS_ACK_DATA | Opcode.D.GRANT | Opcode.D.GRANT_DATA  =>  {
         val ctx = inflightA(d.source)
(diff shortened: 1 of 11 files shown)


lib/src/main/scala/spinal/lib/bus/tilelink/sim/Checker.scala
```diff
@@ -54,7 +54,7 @@ class Checker(p : BusParameter, mappings : Seq[Endpoint], checkMapping : Boolean
 
   val inflightA = Array.fill[InflightA](1 << p.sourceWidth)(null)
   val inflightB = mutable.LinkedHashMap[(Int, BigInt), TransactionB]()
-  val inflightC = mutable.LinkedHashMap[(Int, BigInt), TransactionC]()
```
Dolu1990, line 57: The idea of this is to follow the tilelink 1.9.3 chapter "6.4. Source and Sink Identiers"

```scala
Because Channel C responses to Channel B requests are routed to a single slave and uniquely
identied to an ongoing operation based on c_address, we can further relax the uniqueness
restriction on Channel B requests, requiring only that the combination of b_source and b_address be unique. This relaxation is necessary in order to simultaneously probe multiple masters on the same address, while also probing the same master on multiple addresses. Channel C responses may use any b_source associated with the sender; this signal is ignored for those message types.
```

"requiring only that the combination of b_source and b_address
be unique" :)

Dolu1990: (that is why this specific change and related things should be reverted)

inochisa: ~Good reminder, I only force channel D and forgot channel B. So I will try to split check for channel D/B. Thanks.~

Please see my next comment.

inochisa: Emmm, I have a small question, you mentioned channel C -> B use (source, address). But this is the channel C -> D for `RELEASE_DATA` and `RELEASE_DATA`. So I guess this should following this in sec 6.4?

```
Channels A and C route their requests based solely on address, but both provide source signals for use by their responses on Channel D. Because their Channel D responses can be differentiated based on d_opcode as well as d_source, we allow them to create their source identi ers from separate namespaces.
```

Dolu1990: Ahhhhhhhhhhhhhh, got it XD
different namespace <3

Hmm
So, to make things "perfect" we would need different hash maps for each kind of namespace on the C channel

inochisa: In fact I think the current checker does do different things on this. IIRC current implement just use (source, address) to check `inflightB` in `onC` for opcode `PROBE_ACK` and `PROBE_ACK_DATA`. Is there something I am missing?

But if we want a more prefect checker, (I mean the check follows the spec) a hardware improvement may be needed.
Actually the `c_source` is not the copy of `b_source`, but a hardware identifier (the spec allows to probe multiple master on different address simultaneously), so I think some change on hardware/sim agent is needed to adopt this behavior.

Dolu1990: Ahhh, all good ^^
Thanks :D

PR #1949, opened 2026-08-02
Add scaladoc for advanced features

mrcmry:
This PR add scaladoc to various advanced features of SpinalHDL.

The goal is to first write in-depth doc of individual method behaviors and then, based on these precise descriptions, add more accessible documentation in [SpinalDoc-RTD](https://github.com/SpinalHDL/SpinalDoc-RTD/). As the code is the "single source of truth" and contributors and advanced users navigate the lib much more that they look at SpinalDoc, I think it make more sense.

It is based on SpinalDoc-RTD, my experience of the behavior when using SpinalHDL, currently mainly with verilator and yosys+nextpnr. I also did some analysis of the code with and without LLM (mistral). I entirely wrote the text myself though.

It would be great if people with more experience could review it before merging, for example:

-  @briansune for verilator stuff in b831140
-  @Dolu1990 for data model stuff in 97987c6951, a15cf16516026 and 4b0203258
-  @inochisa for StreamDriver behavior in 3dacc00930

### DIFF
--- a/core/src/main/scala/spinal/core/Bool.scala
+++ b/core/src/main/scala/spinal/core/Bool.scala
@@ -322,11 +322,16 @@ class Bool extends BaseType with DataPrimitives[Bool]  with BaseTypePrimitives[B
   def ?[T <: SpinalEnum](whenTrue: SpinalEnumCraft[T]) = MuxBuilderEnum(whenTrue)
 
   /** Explicitly mark that this hardware signal can take any value in the current context.
-    * 
+    *
     * This is analogous to Verilog `'x`  or VHDL `'-'` assignment or implicit
-    * non-assignment, but  unlike those it is explicit. If a signal is unassigned 
+    * non-assignment, but  unlike those it is explicit. If a signal is unassigned
     * (e.g., no "don't care" or value) in any control path, an error will be raised.
-    * 
+    *
+    * At generation, `'x` in Verilog and `'-'` in VHDL will be generated, unless
+    * SpinalConfig.dontCareGenAsZero is explicitly set as `true`. Note that some
+    * simulators like Verilator use only 0 and 1 for performance reason and will
+    * use random or fixed 0 and 1 values depending on their config.
+    *
     * @see [[https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Data%20types/index.html#data-types Data type documentation]] 
     * @see [[https://en.wikipedia.org/wiki/Don%27t-care_term#X_value "Don't care term" wikipedia article]]
     */
--- a/core/src/main/scala/spinal/core/Bool.scala
+++ b/core/src/main/scala/spinal/core/Bool.scala
@@ -117,26 +117,22 @@ class Bool extends BaseType with DataPrimitives[Bool]  with BaseTypePrimitives[B
   /** this is assigned to `False` */
   def clear(): Unit = this := False
 
-  /** Set to True.
-    *
-    * Direct call to `set()` to reduce the number of nodes in the netlist.
-    */
+  /** Hardware assignment of all bits to `True` */
   override def setAll(): this.type = {
+    // Direct call to `set()` to reduce the number of nodes in the netlist.
     this.set()
     this
   }
 
-  /** Set to False.
-    *
-    * Direct call to `clear()` to reduce the number of nodes in the netlist.
-    */
+  /** Hardware assignment of all bits to `False` */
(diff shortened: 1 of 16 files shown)


mrcmry:
For some reason, the test `spinal.tester.scalatest.SpinalSimUsbHostTester` fail here:

https://github.com/SpinalHDL/SpinalHDL/actions/runs/30741645019/job/91480356821?pr=1949#step:5:4283

This point to this part of the test code:

https://github.com/mrcmry/SpinalHDL/blob/4b0203258d3c9470da616fd432394b322c494fd8/tester/src/test/scala/spinal/tester/scalatest/SpinalSimUsbHostTesterUtils.scala#L206

I did run a ` git log/diff/show upstream/dev..` session with an LLM and it confirmed that there was not behavioral change (only code change is `periodicaly(1000) → periodically(1000) in the RamTester`

I think the problem is due to the SpinalSimUsbHostTester not being deterministic due to the use of the global Random generator here:
- init : https://github.com/SpinalHDL/SpinalHDL/blob/bb298f1ce6d455d75e60af37ac0c055b44f228df/tester/src/test/scala/spinal/tester/scalatest/SpinalSimUsbHostTester.scala#L23
- generation (to point to a few): https://github.com/SpinalHDL/SpinalHDL/blob/bb298f1ce6d455d75e60af37ac0c055b44f228df/lib/src/main/scala/spinal/lib/bus/bmb/sim/BmbMemoryTester.scala#L17
https://github.com/SpinalHDL/SpinalHDL/blob/bb298f1ce6d455d75e60af37ac0c055b44f228df/lib/src/main/scala/spinal/lib/com/usb/sim/Agent.scala#L141

This is not really deterministic with multi-threaded simulation.

Also, I'm not sure if the `SimThread.fork {}` used in the test should not have they own Random object initialized at the start of the test so it's also deterministic.

I can do the changes for Random -> simRandom and single Random per fork, so it's more deterministic if needed.

Dolu1990:
Right, there is something broken with the usb host test.

> Also, I'm not sure if the SimThread.fork {} used in the test should not have they own Random object initialized at the start of the test so it's also deterministic.

Random should never be used. Instead, only the simRandom should be used, which will be the same for all threads in a simulation.

> I can do the changes for Random -> simRandom and single Random per fork, so it's more deterministic if needed.

Random should only be used if all the following criteria are meet : 
- Dedicated Random instance
- seed being set in a deterministic manner
- Not used by multiple threads (but can be used by multiple simulation threads in the same tests)

To debug underterministic behaviour, there is a spinal.sim.killRandom() function which destroy the scala default random, and will make things crash if it is ever used.

I just pushed a fix for the [SpinalSimUsbHostTester.scala, hopefully it will fix it.

core/src/main/scala/spinal/core/sim/package.scala
```diff
@@ -45,6 +45,17 @@ package object sim {
     field.setAccessible(true)
     field.set(r, null)
   }
+
+  /** Return a dedicated instance scala.util.Random for the current simulation.
+    *
+    * If only thread-less simulation API is used (`clockDomain.waitActiveEdge()`,
```
Dolu1990: Overall, the threaded simulation API is deterministic aswell. So, using simRandom across simulation thread is deterministic aswell.

core/src/main/scala/spinal/core/Bool.scala
```diff
@@ -325,6 +321,15 @@ class Bool extends BaseType with DataPrimitives[Bool]  with BaseTypePrimitives[B
   /** Conditional operation for Enumeration value */
   def ?[T <: SpinalEnum](whenTrue: SpinalEnumCraft[T]) = MuxBuilderEnum(whenTrue)
 
+  /** Explicitly mark that this hardware signal can take any value in the current context.
+    * 
+    * This is analogous to Verilog `'x`  or VHDL `'-'` assignment or implicit
+    * non-assignment, but  unlike those it is explicit. If a signal is unassigned 
+    * (e.g., no "don't care" or value) in any control path, an error will be raised.
+    *
```
inochisa: It could be better to stress the signal will could be assigned with 'x' and '-', and no guarantee that the value will be RO0 (it will be a random number in the most cases). This suggestion for all the `assignDontCare()` function

I am an victim of this XD.

mrcmry:
Thanks for your feedbacks. I updated the doc (after checking a bit that it's effectively the case in the code).

Dolu1990:
Thanks ^^

Issue #1294, opened 2024-01-23
Axi4Stream user signal width should not be multiplied by data width


KireinaHoro:
In the current implementation the `user` signal in `Axi4StreamBundle` is defined to be `dataWidth * userWidth`:

https://github.com/SpinalHDL/SpinalHDL/blob/ef0063da38424830a5b29f188e3887e2b043cb75/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala#L39

This directly contradicts the scaladoc comment defining `userWidth`:

https://github.com/SpinalHDL/SpinalHDL/blob/ef0063da38424830a5b29f188e3887e2b043cb75/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala#L11

The behaviour of enforcing the number of bits in `user` to be a multiple of number of bytes in `data` is, in the AXI Stream standard, a recommendation, but not a requirement:

> It is recommended, but does not require, that the number of TUSER bits is an integer multiple of the width of the
interface in bytes. The User signals for each byte must be packed together in adjacent bits within TUSER

Furthermore, there is a use-case of *transfer-based information*, which is what I'm trying to do:

> TUSER can be used to convey information that is relevant to an entire transfer rather than to individual bytes. For
example, when the same information applies to every byte in a transfer, it is more efficient to indicate the additional
information once only for the entire transfer rather than replicating it for each byte within the transfer.

In this case it does not make sense to enforce `user` to be a multiple of number of bytes in `data`.

oletf:
related to #1252 if I've understood correctly, 
there's an attempt here ce8d0bac06678946df7fff39a32ceb08eaf680cd in progress

xueweiwujxw:
I opened PR #1954  with a prototype implementation for independently sized TUSER fields.

It preserves the existing byte-associated `userWidth` behavior and adds the mutually exclusive `beatUserWidth` configuration.

The main point requiring feedback is width conversion. The current implementation treats `beatUserWidth` as transaction metadata:

- downsizing propagates it to the generated output beats;
- upsizing retains the first value of the assembled transaction;
- the value is expected to remain stable for that transaction.

I would appreciate feedback on this commit.

Issue #1252, opened 2023-11-27
Axi4Stream: userWidth tied to dataWidth


oletf:
I'm using a Vivado IP with AXI4Stream interface I have set up like this :
![image](https://github.com/SpinalHDL/SpinalHDL/assets/109208726/ea4c0311-378e-4304-b25f-af52afc472e6)

In my case my io is defined as :
```scala
val m_axis  = master(Axi4Stream(Axi4StreamConfig(dataWidth = 3, useLast = true)))
```
I can currently ignore the user bits and keep them unconnected, so it's not a problem, but if I were to need to use them,
I have no way to build a config which matches the IP data and user width, 
as [Axi4StreamBundle defines here](https://github.com/SpinalHDL/SpinalHDL/blob/a0f6837fae4557cfbb22af0c8d239bb22e954e39/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala#L39)  : 
```scala
user = Bits(config.userWidth*config.dataWidth bit)
```
while the docstring mentions : 
```
dataWidth Width of the bus in BYTES
userWidth Width of the User field in bits
```
but in the end `user`'s actual width ends up being `dataWidth*userWidth bits`
the userbit adaptation also [uses `dataWidth` here](https://github.com/SpinalHDL/SpinalHDL/blob/a0f6837fae4557cfbb22af0c8d239bb22e954e39/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala#L68-L69),
it makes sense together if the `dataWidth*userWidth bits` size is actually needed,
but I don't get why it's the case.

Is there an explicit `dataWidth` to `userWidth` ratio requirement in the Axi4 specification, 
or it's some kind of refactor typo ?

I have made a minimal change here : https://github.com/oletf/SpinalHDL/commit/6313695711e25ffaf4d9c0d6da2ff122728ae3cf, which made sense to me, 
I can PR if this change makes sense and I didn't miss any important part related to AXI4Stream specification.

likewise:
The width of the TUSER (user) field can be any number of bits, according to the AMBA AXI 4 Stream specification. 

Thus is no implicit correlation between the data width and the user width.

Quoting the specification: `TUSER width. Recommended number of bits is an integer multiple of the width of the interface in bytes.`(https://documentation-service.arm.com/static/642583d7314e245d086bc8c9?token=)

[...]

PR #1954, opened 2026-08-26
axis: support independently sized TUSER fields

xueweiwujxw:
Closes #

Related to #1294 and #1252.

`Axi4StreamConfig.userWidth` currently defines TUSER bits per byte lane, so the generated width is `dataWidth * userWidth`. This does not support AXI4-Stream interfaces whose TUSER width is independent of `dataWidth`.

This PR adds the mutually exclusive `beatUserWidth` parameter:

```scala
Axi4StreamConfig(
  dataWidth = 8,
  useUser = true,
  beatUserWidth = 4
)
```

This generates a 4-bit TUSER signal. Existing `userWidth` configurations keep their current byte-associated behavior.

The implementation distinguishes the two modes:

- `userWidth`: TUSER is sliced and adapted with its corresponding byte lanes.
- `beatUserWidth`: TUSER is handled as one complete field.
- Direct connections require matching TUSER association modes.
- `Axi4StreamSparseCompactor` propagates beat-associated TUSER without lane reordering.
- Both width adapters support beat-associated TUSER.
- Downsizing propagates the value to the generated output beats.
- Upsizing retains the first value of the assembled transaction.
- `Axi4StreamWidthAdapter` stores it independently from the byte buffer, similarly to `TID` and `TDEST`.

When several input beats are merged, `beatUserWidth` is treated as transaction metadata and is expected to remain stable for that transaction.

Tests cover width calculation, configuration validation, direct assignment, sparse compaction, adapter downsizing, upsizing, compact mode, association mismatches, and regression elaboration for the existing byte-associated path.

# Impact on code generation

Existing `userWidth` configurations retain the same generated TUSER width and byte-lane adaptation logic.

When `beatUserWidth` is used:

- TUSER width is independent of `dataWidth`.
- Direct connections and sparse compaction propagate TUSER as one field.
- `Axi4StreamSimpleWidthAdapter` directly propagates or registers TUSER.
- `Axi4StreamWidthAdapter` generates an additional `beatUserWidth`-wide register for TUSER.

No RTL changes are expected when TUSER is disabled or the existing byte-associated mode is used.

# Checklist

- [x] Unit tests were added
  - on [RTD](https://github.com/SpinalHDL/SpinalDoc-RTD):  not required
  - thanks to a [new tracking issue on RTD](https://github.com/SpinalHDL/SpinalDoc-RTD/issues/new?title=Document%20XXX&body=Do%20not%20merge%20until%20SpinalHDL/SpinalHDL%23XXX%20has%20not%20been%20merged): not required

### DIFF
--- a/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala
+++ b/lib/src/main/scala/spinal/lib/bus/amba4/axis/Axi4Stream.scala
@@ -8,13 +8,14 @@ import spinal.lib._
  * @param dataWidth Width of the bus in BYTES
  * @param idWidth Width of the ID field in bits
  * @param destWidth Width of the Destination field in bits
- * @param userWidth Width of the User field in bits
+ * @param userWidth Width of the User field in bits per byte lane
  * @param useStrb Use byte strobe bits
  * @param useKeep Use byte keep bits
  * @param useLast Use last bit
  * @param useId Use ID field, must specify idWidth
  * @param useDest Use Destination field, must specify destWidth
- * @param useUser Use User field, must specify userWidth
+ * @param useUser Use User field, must specify userWidth or beatUserWidth
+ * @param beatUserWidth Total width of the User field in bits for each beat, independent of the number of byte lanes
  */
 case class Axi4StreamConfig(dataWidth: Int,
                             idWidth:   Int = -1,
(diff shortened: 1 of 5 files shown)


xueweiwujxw:
Hi @Dolu1990 , could you take a look when you have time? I’d like to hear your thoughts on the `beatUserWidth` API and the width-adapter behavior.

PR #1956, opened 2026-08-31
Add SWD debug transport for RISC-V debug module

disdi:
Introduce DebugTransportModuleSwd (SW-DP PHY, DP register file, DMI gateway) and DebugModuleFiber.withSwdTransport() as the SWD counterpart of the existing JTAG DTM.

Closes #

# Impact on code generation

# Checklist

- [ ] Unit tests were added

### DIFF
--- a/lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
+++ b/lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
@@ -1,0 +1,489 @@
+package spinal.lib.cpu.riscv.debug
+
+import spinal.core._
+import spinal.lib._
+
+/**
+ * SWD transport for the RISC-V debug module.
+ */
+
+case class SwdDpCmd() extends Bundle {
+  val apNdp = Bool()          // request APnDP bit
+  val rnw   = Bool()          // request RnW bit
+  val addr  = Bits(2 bits)    // A[3:2] (addr(0) = A[2])
+}
+
+case class SwdDpRsp() extends Bundle {
+  val ack   = Bits(3 bits)    // OK=001, WAIT=010, FAULT=100 — sent LSB first
+  val rdata = Bits(32 bits)   // used when cmd.rnw and ack == OK
+}
+
+case class SwdDpWrite() extends Bundle {
+  val data     = Bits(32 bits)
+  val parityOk = Bool()       // false => WDATAERR material for the DP (Phase 2B)
+}
+
+object SwdAck {
+  def OK    = B"3'b001"
+  def WAIT  = B"3'b010"
+  def FAULT = B"3'b100"
+}
+
+case class SwdPhy() extends Component {
+  val io = new Bundle {
+    val swdio = new Bundle {
+      val i  = in  Bool()
+      val o  = out Bool()
+      val oe = out Bool()
+    }
+    val dp = new Bundle {
+      val cmd = master Flow(SwdDpCmd())
+      val rsp = slave  Flow(SwdDpRsp())
+      val wr  = master Flow(SwdDpWrite())
+    }
+  }
+
+  val dio = io.swdio.i
+
+  val oData  = Reg(Bool()) init(False)
+  val oDrive = Reg(Bool()) init(False)
+  io.swdio.o  := oData
+  io.swdio.oe := oDrive
+
+  // Response capture: rsp may be presented combinationally off cmd (same cycle the
+  // turnaround starts) or held valid for one cycle; latch it for the rest of the frame.
+  val rspHold = Reg(SwdDpRsp())
+  when(io.dp.rsp.valid){ rspHold := io.dp.rsp.payload }
+  val ackNow   = Mux(io.dp.rsp.valid, io.dp.rsp.payload.ack,   rspHold.ack)
+
+  val cmdValid   = Reg(Bool()) init(False)
+  val cmdPayload = Reg(SwdDpCmd())
+  cmdValid := False
+  io.dp.cmd.valid   := cmdValid
+  io.dp.cmd.payload := cmdPayload
+
+  val wrValid   = Reg(Bool()) init(False)
+  val wrPayload = Reg(SwdDpWrite())
+  wrValid := False
+  io.dp.wr.valid   := wrValid
+  io.dp.wr.payload := wrPayload
+
+  // Line reset: 50+ SWCLK cycles with SWDIO high while the host owns the line
+  // (ADIv6.0 B4.3.3). Recovers from any state, including protocol error.
+  val lineReset = new Area {
+    val counter = Reg(UInt(6 bits)) init(0)
+    val hit = counter === 50
+    when(oDrive || !dio) {
+      counter := 0
+    } elsewhen(!hit) {
+      counter := counter + 1
+    }
+  }
+
+  object EState extends SpinalEnum {
+    val IDLE, HEADER, ACK, READ_DATA, WR_TRN, WRITE_DATA, RELEASE, ERROR, RESET_WAIT = newElement()
+  }
+  val state = Reg(EState()) init(EState.RESET_WAIT)
+  val cnt   = Reg(UInt(6 bits)) init(0)
+  val hdr   = Reg(Bits(6 bits))          // apNdp, rnw, a2, a3, parity, stop (LSB first)
+  val wShift = Reg(Bits(32 bits))
+
+  switch(state) {
+    is(EState.IDLE) {
+      oDrive := False
+      when(dio) {                        // start bit
+        state := EState.HEADER
+        cnt := 0
+      }
+    }
+    is(EState.HEADER) {
+      hdr := dio ## hdr(5 downto 1)      // LSB-first shift-in
+      cnt := cnt + 1
+      when(cnt === 6) {                  // dio is the park bit; hdr holds apNdp..stop
+        val apNdp = hdr(0)
+        val rnw   = hdr(1)
+        val a2    = hdr(2)
+        val a3    = hdr(3)
+        val parityOk = (apNdp ^ rnw ^ a2 ^ a3) === hdr(4)
+        val stopOk   = !hdr(5)
+        when(parityOk && stopOk && dio) {
+          cmdValid := True
+          cmdPayload.apNdp := apNdp
+          cmdPayload.rnw   := rnw
+          cmdPayload.addr  := a3 ## a2
+          state := EState.ACK            // next cycle is the turnaround
+          cnt := 0
+        } otherwise {
+          state := EState.ERROR          // silent until line reset (B4.2.5)
+        }
+      }
+    }
+    is(EState.ACK) {                     // drives ACK[0..2], first bit lands after the turnaround
+      oDrive := True
+      oData  := ackNow(cnt(1 downto 0))
+      cnt := cnt + 1
+      when(cnt === 2) {
+        cnt := 0
+        when(ackNow === SwdAck.OK) {
+          when(cmdPayload.rnw) {
+            state := EState.READ_DATA    // target keeps the line: RDATA follows ACK directly
+          } otherwise {
+            state := EState.WR_TRN       // turnaround back to host before WDATA
+          }
+        } otherwise {
+          state := EState.RELEASE        // WAIT/FAULT: no data phase (no ORUNDETECT)
+        }
+      }
+    }
+    is(EState.READ_DATA) {               // 32 data bits LSB first + even parity
+      oDrive := True
+      oData  := Mux(cnt === 32, rspHold.rdata.xorR, rspHold.rdata(cnt(4 downto 0)))
+      cnt := cnt + 1
+      when(cnt === 32) {
+        state := EState.RELEASE
+        cnt := 0
+      }
+    }
+    is(EState.WR_TRN) {                  // release the line, then one turnaround bit period
+      oDrive := False                    // (a write has a 2nd turnaround after ACK; the host
+      cnt := cnt + 1                     //  drives its first WDATA bit only after it)
+      when(cnt === 1) {
+        state := EState.WRITE_DATA
+        cnt := 0
+      }
+    }
+    is(EState.WRITE_DATA) {              // sample 32 data bits LSB first, then parity
+      wShift := dio ## wShift(31 downto 1)
+      cnt := cnt + 1
+      when(cnt === 32) {                 // dio is the parity bit; wShift holds the data
+        wrValid := True
+        wrPayload.data     := wShift
+        wrPayload.parityOk := wShift.xorR === dio
+        state := EState.IDLE
+      }
+    }
+    is(EState.RELEASE) {                 // turnaround back to host after a target-driven phase
+      oDrive := False
+      cnt := cnt + 1
+      when(cnt === 1) {
+        state := EState.IDLE
+        cnt := 0
+      }
+    }
+    is(EState.ERROR) {                   // protocol error: line released, headers ignored
+      oDrive := False
+    }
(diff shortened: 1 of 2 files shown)


inochisa:
I guess the current RISC-V spec does not support this? Is there something I have missed?
https://github.com/riscv/riscv-debug-spec/issues/785

disdi:
> I guess the current RISC-V spec does not support this? Is there something I have missed? [riscv/riscv-debug-spec#785](https://github.com/riscv/riscv-debug-spec/issues/785)

Ratified Debug Spec 1.0 still only specifies JTAG (Ch. 6.1). 
[riscv-debug-spec#785](https://github.com/riscv/riscv-debug-spec/issues/785) is the request to add an SWD chapter; that hasn’t happened.

What Ch. 6 does say is that a JTAG DTM is not required:
`
This specification defines a JTAG DTM in Section 6.1. Additional DTMs may be added in future versions of this specification.
`

An implementation can be compatible with this specification without implementing any of JTAG DTM . In that case it must be advertised as conforming to “RISC-V Debug Specification, with custom DTM.”

That’s the path here. 

DebugTransportModuleSwd is a custom DTM: SW-DP on the pins, the existing DMI/DebugBus on the other side, same DebugModule as the JTAG DTM. 

The companion VexRiscv#483 (https://github.com/SpinalHDL/VexRiscv/pull/483) exposes it as an alternative to --jtag-tap, not alongside it (Ch. 6 leaves simultaneous DTMs unsupported).

Same idea as RP2350 (SW-DP + Mem-AP as the DTM). 

Conformance claim: “RISC-V Debug Specification, with custom DTM.”

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
```diff
@@ -0,0 +1,489 @@
...
+  val oData  = Reg(Bool()) init(False)
+  val oDrive = Reg(Bool()) init(False)
+  io.swdio.o  := oData
+  io.swdio.oe := oDrive
+
+  // Response capture: rsp may be presented combinationally off cmd (same cycle the
+  // turnaround starts) or held valid for one cycle; latch it for the rest of the frame.
+  val rspHold = Reg(SwdDpRsp())
+  when(io.dp.rsp.valid){ rspHold := io.dp.rsp.payload }
+  val ackNow   = Mux(io.dp.rsp.valid, io.dp.rsp.payload.ack,   rspHold.ack)
```
inochisa: `val rspHold = io.dp.rsp.stage()`?

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
```diff
@@ -0,0 +1,489 @@
...
+                                   dpidr : BigInt = BigInt("0BA11AAB", 16),
+                                   apIdr : BigInt = BigInt("74726976", 16)) extends Component {
+  val io = new Bundle {
+    val swd = new Bundle {
+      val swclk = in Bool()
+      val swdio = new Bundle {
+        val i  = in  Bool()
+        val o  = out Bool()
+        val oe = out Bool()
+      }
```
inochisa, line 474: Why defined a `Swd` bundle and initialize it with something `slave(Swd())`. And for bidirectional signal, why not using `ReadableOpenDrain` with InOutWrapper for verilog generation?

For signal `oe`, I guess it is SWO? If so, please make it optional. Thanks.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
```diff
@@ -0,0 +1,489 @@
...
+  // turnaround starts) or held valid for one cycle; latch it for the rest of the frame.
+  val rspHold = Reg(SwdDpRsp())
+  when(io.dp.rsp.valid){ rspHold := io.dp.rsp.payload }
+  val ackNow   = Mux(io.dp.rsp.valid, io.dp.rsp.payload.ack,   rspHold.ack)
+
+  val cmdValid   = Reg(Bool()) init(False)
+  val cmdPayload = Reg(SwdDpCmd())
+  cmdValid := False
+  io.dp.cmd.valid   := cmdValid
+  io.dp.cmd.payload := cmdPayload
```
inochisa, line 62: If understand this correctly. You can use the following
```scala
val cmd = Flow(SwdDpCmd())
cmd.valid := False
io.dp.cmd <-< cmd
```

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
```diff
@@ -0,0 +1,489 @@
...
+
+  val wrValid   = Reg(Bool()) init(False)
+  val wrPayload = Reg(SwdDpWrite())
+  wrValid := False
+  io.dp.wr.valid   := wrValid
+  io.dp.wr.payload := wrPayload
```
inochisa, line 68: The same as above

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
```diff
@@ -0,0 +1,489 @@
...
+  // (ADIv6.0 B4.3.3). Recovers from any state, including protocol error.
+  val lineReset = new Area {
+    val counter = Reg(UInt(6 bits)) init(0)
+    val hit = counter === 50
+    when(oDrive || !dio) {
+      counter := 0
+    } elsewhen(!hit) {
+      counter := counter + 1
+    }
+  }
```
inochisa, line 80: I think you can use the `Counter` utility for this logic.

inochisa:
That's good for for this clarification. It is important to let users know it is not a standard but a custom design supported by the SpinalHDL, and may be changed if the spec has SWD support.

inochisa:
As this is a good implementation, I think it can be translated as a generic library like jtag in `spinal.lib.com.jtag` so other can benefit from this design?

disdi:
> That's good for for this clarification. It is important to let users know it is not a standard but a custom design supported by the SpinalHDL, and may be changed if the spec has SWD support.

Thanks — I’ll make the custom-DTM status explicit by adding a short note  on DebugTransportModuleSwd block : 
"this is a Ch. 6 custom DTM, not a ratified SWD chapter, and the pin/AP map may change if the spec later defines one."

disdi:
> As this is a good implementation, I think it can be translated as a generic library like jtag in `spinal.lib.com.jtag` so other can benefit from this design?

On the library split: agreed that the wire-side pieces are not RISC-V-specific. The JTAG split is the right model:

• spinal.lib.com.jtag — pins + TAP
• spinal.lib.cpu.riscv.debug — DTM that consumes slave(Jtag())

I can do the same for SWD:

• spinal.lib.com.swd — Swd pin bundle, SwdPhy, SwdDp (ADI SW-DP)
• spinal.lib.cpu.riscv.debug — SwdDmiGateway + DebugTransportModuleSwd (RISC-V DMI)

PHY/DP stay reusable; the designer-AP / DMI map stays with the debug module.

As of now I prefer to land the DTM as-is and follow up with the split in a later PR.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
disdi: I think stage() is the wrong primitive for a same-cycle combinational ACK. It would delay ACK by a cycle.

The PHY starts driving ACK on the cycle after the header (the turnaround). The DP answers combinationally off the registered cmd pulse (io.dp.rsp.valid := cmd.valid), so the ACK bits have to be visible that cycle.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
disdi: Same for wr: it fires after 32 data bits + parity. The host has already sent the data. The PHY cannot stall, retry, or stretch the frame if the DP is not ready.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
disdi: I think it makes sense to use Counter instead of an open-coded Reg(UInt(6 bits)).

```
val lineReset = new Area {
  val counter = Counter(51) // [0, 50]
  when(oDrive || !dio) {
    counter.clear()
  } elsewhen(!counter.willOverflowIfInc) {
    counter.increment()
  }
  val hit = counter.willOverflowIfInc // value === 50
}
```

I think above code will not result in any FSM change.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
disdi: > For signal `oe`, I guess it is SWO? If so, please make it optional. Thanks.

oe is not SWO. SWO is ARM Serial Wire Output (the extra trace pin). oe is SWDIO output enable for the bidirectional data pin. It isn’t optional: the target has to release the line after ACK / RDATA.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
inochisa: > I think stage() is the wrong primitive for a same-cycle combinational ACK. It would delay ACK by a cycle.
> 
> The PHY starts driving ACK on the cycle after the header (the turnaround). The DP answers combinationally off the registered cmd pulse (io.dp.rsp.valid := cmd.valid), so the ACK bits have to be visible that cycle.

I think you Mux does already do this for some-cycle ACK.

I mean using stage() to replace 55-56, but in fact, stage does have problem as it does not hold the payload. So I think it should be `m2sPipe(holdPayload = true)`

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
inochisa: Sorry, I mistook the `io.dp.wr` is a Flow. so why not replacing the `Stream` with `Flow` in my example? It should be the same.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
inochisa: This confuse me, I think the physical interface is only two pins (`SWCLK` and `SWDIO`), means we only have at most 3 pin defined, one for `SWCLK` and two (a read/ a write) for `SWDIO`. So the thing you added does not describe the physical interface, but something else?

disdi: > Why defined a `Swd` bundle and initialize it with something `slave(Swd())`.

DebugTransportModuleJtagTap uses the library bundle:
```
  val io = new Bundle {
    val jtag = slave(Jtag())
    val bus = master(DebugBus(p.addressWidth))
```

asMaster() is the probe. slave(Jtag()) on the TAP flips it: tck/tms/tdi in, tdo out. That works because every JTAG pin is unidirectional.

SWD is an anonymous inner bundle with directions written by hand:
```
  val io = new Bundle {
    val swd = new Bundle {
      val swclk = in Bool()
      val swdio = new Bundle {
        val i  = in  Bool()
        val o  = out Bool()
        val oe = out Bool()
      }
    }
```
That is electrically complete (i/o/oe is an IOBUF). It is not a reusable type, and <> / slave() / toIo do not know it is “an SWD port”.

inochisa: OK. I am fine with this. Thanks.

disdi: >  And for bidirectional signal, why not using `ReadableOpenDrain` with InOutWrapper for verilog generation?

SWD (ADI) is push-pull tristate:

• Host drives 0 and 1 during header / WDATA / line reset (50 cycles of driven high).
• Target drives 0 and 1 during ACK / RDATA (OK = 001 starts with a driven 1).
• Both release for the turnaround (oe = 0). A pull-up is only for idle / high-Z, not for transmitting 1s.

InOutWrapper already has the matching case for that — TriState, not ReadableOpenDrain

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
disdi: For something like below :

```
val wr = Flow(SwdDpWrite())
wr.valid := False
io.dp.wr <-< wr
```

Type is right (Flow). 

<-< is still m2sPipe: a register slice on top of a combinational pulse. That can replace the explicit wrValid/cmdValid FFs, but only if payload is assigned every cycle (or held). Today those FFs are that pipe. Adding <-< on top of the existing Regs would delay the beat a cycle and break turnaround / WDATA commit.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
disdi: same answer as wr.

io.dp.cmd is a Flow — a one-cycle pulse, no ready. Stream would add a handshake the wire cannot use: after the park bit the PHY has to enter ACK on the next SWCLK and drive the ACK the DP already computed.

lib/src/main/scala/spinal/lib/cpu/riscv/debug/DebugTransportModuleSwd.scala
inochisa: Yes, I understand, You need to return immediately. If this I think you should also need a fix as you code does delay a cycle.

```scala
val wrValid   = Reg(Bool()) init(False)
wrValid := False
io.dp.wr.valid   := wrValid
```
This does not mean you set the `io.dp.wr.valid` as the `wrValid` is assigned a new value. Instead, it will be set after the `wrValid` is changed to True. I think you need something like this:
```scala
val wrValid = Bool()
val _wrValid = RegNext(wrValid, init = False)
wrValid := _wrValid
// some condition after
when (cond) {
  wrValid := xxx
}
```

disdi: wrValid is a Reg, so io.dp.wr.valid rises the cycle after `wrValid := True`. That’s intended, not a missed same-cycle path.

cmd is the one that has to be visible on a specific SWCLK: the park-bit assignment lands on the DP the next cycle, which is the turnaround, which is when combinational ACK is sampled into oData. That extra FF is the alignment.

wr has no such deadline. ACK already went out before WDATA. After the parity bit the host is finished; the DP only commits the write (WDATAERR / AP). A one-cycle delay into IDLE does not shift any wire bit.

A combinational pulse (`wr.valid := True` on the parity cycle, `io.dp.wr << wr` with no extra` <-<`) would commit one SWCLK earlier. I’d rather keep the current FFs — they are not a protocol stall.

inochisa: OK. I have no question on this. You can continue as it is.

Dolu1990:
While the SWD could be moved, i'm already merging this PR to keep things moving ^^
(can be done in another PR)

PR #1963, opened 2026-09-12
Adding support for Atomic op-codes.

jtredux:
Closes #

This patch adds support for the TileLink Atomic operations to the channelA definitions so that they can be used with custom TileLink slave components. This should be transparent to existing SpinalHDL tilelink components.

# Impact on code generation

This should have minimal effect on code generation as it is mostly extending enums.

# Checklist
 

No new unit tests have been added, however this code has been used with a new component which supports these new operations and integrates the original tilelink.Cache as an LLC and several other tilelink/AXI bridge components.


Dolu1990:
Thanks ^^

PR #1966, opened 2026-09-19
Move the SWD wire side into spinal.lib.com.swd

disdi:
Split the SWD transport the same way JTAG is split, so the ADI part can be reused outside the RISC-V debug module.

spinal.lib.com.swd (new):
- Swd: named pin bundle, swclk + swdio as TriState(Bool()). SWDIO is push-pull tristate, so InOutWrapper turns it into a real inout. asSlave() is overridden: both ends own a tristate driver, only SWCLK flips. clockDomain gives the BOOT-reset SWCLK domain.
- SwdPhy, SwdDp, SwdPhyDp and their Flow payloads, moved verbatim.
- sim.SwdHostDriver: bit-banged host model, like jtag.sim.JtagDriver.

spinal.lib.cpu.riscv.debug keeps what is RISC-V specific: SwdDmiGateway (designer-AP -> DMI) and DebugTransportModuleSwd, whose io.swd is now slave(Swd()). DebugModuleFiber.withSwdTransport() is unchanged; its io.swd.toIo now works on a named bundle.

The PHY, DP and DTM suites move here from VexRiscv, plus SwdBundleTest for the directions and the InOutWrapper case.

No functional change: the VexRiscv SMP cluster netlist is identical up to the i/o/oe -> read/write/writeEnable port renames. Re-verified on an Arty A7-35T over CMSIS-DAP (OpenOCD raw AP, riscv target, GDB).

Follow-up to #1956

# Impact on code generation

# Checklist

- [x] Unit tests were added


Dolu1990:
Thanks ^^

